PR #112, opened 2014-04-16
added block `formUserOptions` around user options in slideshow editor

novascreen:
i think it should be possible to change the user options in a widget that's based on the slideshow without copying the whole `formAll` block


boutell:

PR #139, opened 2014-06-27
Adding the ability to override the modal top offset

fotisp:
Sometimes depending on our design we may need to change several visual aspects of the the apos ui, which is set using javascript, and not css. 

Therefore it would be nice to have a way to override these values. This pull request deals only with the top offset of the modal dialogs.

A user extending apostrophe-snippet can add a block as such to the widgetEditor.html
```
{%block dataAttrs}
data-top-offset="150"
{%endblock}
```

And this way they are able to control that aspect. 

This method may be useful to control other aspects of the ui in the future.


boutell:

PR #160, opened 2014-09-09
Limit tags

pixeleet:
I think this is correct now. 

Care to test?

Cheers,
David


boutell:

PR #390, opened 2015-11-25
Update package clean-css, because we had problems with shorthand

matthaias:
We had a problem with the `clean-css` with version `1.x`, so i updated the package. Please review and check if you find some problems. Thx!

Example:
```    
font: 500 2em/1.1 'DIN Condensed',Arial,sans-serif;
```

was this after the minify-process:

```
font: 400 500 2em/1.1 'DIN Condensed',Arial,sans-serif;
```

### DIFF
--- a/lib/assets.js
+++ b/lib/assets.js
@@ -7,7 +7,7 @@ var fs = require('fs');
 // JS minifier and optimizer
 var uglifyJs = require('uglify-js');
 // CSS minifier https://github.com/GoalSmashers/clean-css
-var cleanCss = require('clean-css');
+var CleanCSS = require('clean-css');
 // LESS CSS compiler
 var less = require('less');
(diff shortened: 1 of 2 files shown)


Vispercept:
+1 :smile:

boutell:
@stuartromanek @jsumnersmith @kyjoya @suhmantha1 would y'all please take a look at this pull request and make sure it doesn't cause any CSS problems (mostly looking for flat-out errors here) with a few of our fancier websites when Apostrophe is using this newer version of the cleancss package? I can see that the older version is missing support for a valid CSS construct, and I'd like to accept this pull request, but we need some due diligence here first. Thanks!

boutell:
Closing because 0.5.x is in maintenance mode. Please use 2.x for new development. It uses a much newer release of clean-css.

PR #411, opened 2016-01-27
Refactored the contents of formMacros.html into sub-templates

agilbert:
Refactored the contents of formMacros.html into sub-templates under aposMacros. This allows your project to override individual macros easily.


jth-:
This change makes sense, makes it much easier to override individual macros. :+1: Merging.

Issue #252, opened 2015-05-17
Harmonize naming of custom jquery events in apostrophe


boutell:
If we're going to have them, they should be allInterCap, notSome-intercap-and-SomeHyphenated.

jth-:
I'm going to assume we're just looking at this in 0.6, let me know if you had something else in mind @boutell

PR #413, opened 2016-01-28
#252 custom jquery events triggered directly by apostrophe modules ha…

boutell:
…ve an apos prefix and are interCapNames. Fixed the only exceptions I could find


jth-:
Looks good, I couldn't find any other instances of this either. Merging and closing issue.

PR #427, opened 2016-02-10
JSON output in HTML markup must not crash on illegal unicode characte…

### LINKED ISSUE #397: Uncaught SyntaxError: Unexpected token ILLEGAL in $.extend(true, apos.data, { „mediaLibrary“: CONTENT })
Hi apostrophe-team,

we have some trouble with the script extending the apos.data object on the client side from https://github.com/punkave/apostrophe/blob/master/lib/push.js#L242.

The generated line in the browser extending the apos.data with the mediaLibrary information ($.extend(true, apos.data, { „mediaLibrary“: CONTENT })) throws an „Uncaught SyntaxError: Unexpected token ILLEGAL“ error in the browser. We can’t exactly specify what causes the error, as the CONTENT object is about 1.400.000 characters long (the object contains the blog page with all blogPost pages from the blog plugin).

CONTENT is similar to this:

```
{
  "mediaLibrary": {
    "aposPrefix": "",
    "doNotEdit": false,
    "offline": false,
    "demo": false,
    "loginButton": false,
    "siteTitle": "siteTitle",
    "shortName": "shortName",
    "hostName": "localhost:3000",
    "page": {
      "_id": "ID",
      "slug": "/",
      "path": "home",
      "title": "Home",
      "level": 0,
      "type": "blog",
      "published": true,
      "seoDescription": "",
      "tags": [],
      "orphan": false,
      "pagePermissions": ["edit-ID"],
      "andFromPagesIds": [],
      "andFromPagesRelationships": {},
      "sortTitle": "home",
      "highSearchText": "home",
      "highSearchWords": ["home"],
      "lowSearchText": "",
      "searchSummary": "",
      "blogWidget": {"slug": "/:blogWidget"...}  
      "_edit": true,
      "_publish": true,
      "url": "/",
      "ancestors": [],
      "peers": [
        {"_id": ID}
      ],
      "children": [...],                       
      "tabs": [...]
    }
  }
}
```

If we add an empty object to the mediaLibrary property the error won’t occur and everything works fine (it seems that other CMS functionalities won’t be affected). Is it a normal thing that the mediaLibrary property has so many data? Can we leave it blank or will this lead to another negative impact? Is it possible, that the "mediaLibrary" object contains unescaped characters?

Thanks in advance for your response!

boutell:
…rs (the line separator and paragraph separator unicode characters) fixes #397

### DIFF
--- a/lib/templates.js
+++ b/lib/templates.js
@@ -139,6 +139,9 @@ module.exports = function(self) {
     data = JSON.stringify(data); // , null, '  ');
     data = data.replace(/<\!\-\-/g, '<\\!--');
     data = data.replace(/<\/script\>/gi, '<\\/script>');
+    // unicode line separator and paragraph separator break JavaScript parsing
+    data = data.replace("\u2028", "\\u2028");
+    data = data.replace("\u2029", "\\u2029");
     return data;
   };
 


boutell:
cc @jth- this is a great stability fix which I'll forward-port to 0.6 once it's merged

jth-:
@boutell #428 is the corresponding 0.6 issue.

Issue #420, opened 2016-02-02
Slideshow: Implement default grid-view when choosing images.


jth-:
The grid view should be the default, should exist and should actually display images. Lots of work done toward this but there's no custom template to finish the job yet. Should be done with attention to the existing provisions that were created to help with it, talk to @boutell.

mcoppola:
closed with #429

boutell:
A++++

On Wed, Feb 10, 2016 at 4:07 PM, Matthew Coppola <notifications@github.com>
wrote:

> Closed #420 <https://github.com/punkave/apostrophe/issues/420>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/punkave/apostrophe/issues/420#event-546378213>.
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

PR #429, opened 2016-02-10
#420

mcoppola:


boutell:
This is good, you should render the attachment at size one-third rather than two-thirds for performance.

I would suggest one-sixth but that's actually too small even for your grid.

mcoppola:
closes #420

PR #434, opened 2016-02-11
Fixed virtual singleton editing in schemas.

### LINKED ISSUE #432: Virtual Singleton Editing not saving changes properly.
Could be something with singletons helper, could be something with singleton converter.

jth-:
Resolve #432

### DIFF
--- a/lib/modules/apostrophe-areas/lib/routes.js
+++ b/lib/modules/apostrophe-areas/lib/routes.js
@@ -55,36 +55,6 @@ module.exports = function(self, options) {
     return res.send(self.render(req, 'virtualArea', { options: options }));
   });
 
-  // // Render an editor for a virtual area with the content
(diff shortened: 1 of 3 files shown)


boutell:
Are both virtual areas and virtual singletons now calling edit-virtual-area in the same way?

jth-:
Just pushed a change that makes enableSingleton simply add a `type` to the options object and call `enableArea`, which in turn makes the actual request to `edit-virtual-area`. So now they are using common code to make that particular request.

mcoppola:
In testing I'm able to call /edit-virtual-area for a apostrophe-images singleton, so this is working.  @boutell  i'd like to merge this to fix another bug in the slideable modals stack.  just leaving a note here to take a look at jimmy's implementation if you'd like

PR #458, opened 2016-03-15
Views fallback

jth-:
@boutell @colpanik 

- Added an option to pass an optional path at project level to set a fallback folder for views for all modules, essentially allowing a developer to declare a global views folder.
- Added an option to set a different template for widgets (though it still falls back to `widget.html` if not set)


jth-:
Merging for now, since I've got a bunch of apos namespacing fixes in here now, and the change is pretty simple anyhow.

PR #459, opened 2016-03-15
when a media item is replaced, hide the old version in uploadfs. Crea…

boutell:
…te an array of uploadfs paths to the old versions and be sure to remove them in apostrophe:empty-trash


jth-:
Traced through the logic here, looks good to me. It appears `hideInUploadfs` should behave the same way as it did before for trashing a file, which is correct. Merging.

PR #481, opened 2016-04-04
Refactor Areas

jth-:
Several sizable chunks here:

- Refactor the rest of the rich text widget out of `apostrophe-areas`. `apostrophe-rich-text-widgets` now subclasses and fully obeys the contract specified by `apostrophe-widgets`.
- Add in contextual 'Add Content' insertion menus, so now you can add content between two existing widgets instead of perpetually at the top of an area. Some work around styling this appropriately as well.
- Attempt at namespacing all `data-area`, `data-widgets`, and `data-widget` to `data-apos-area`, `data-apos-widgets`, and `data-apos-widget` respectively.
- Remove all lockup logic. Eventually, we'll reimplement lockups as a nested area within rich text widgets, which is a much more elegant solution than in 0.5.

@boutell cc @mcoppola @colpanik

### DIFF
--- a/lib/modules/apostrophe-areas/public/css/user.less
+++ b/lib/modules/apostrophe-areas/public/css/user.less
@@ -9,7 +9,7 @@
   &.apos-limit-reached
   {
     >.apos-ui .apos-area-controls,
-    >[data-widgets]>[data-apos-widget-wrapper]>.apos-ui .apos-area-controls { display: none; }
+    >[data-apos-widgets]>[data-apos-widget-wrapper]>.apos-ui .apos-area-controls { display: none; }
   }
   // When we are dragging a widget, we want it to be at the forefront.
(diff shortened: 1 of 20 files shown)


boutell:
Looks like isText was retained but all usages of it were removed. Finish the job, kill it with fire; fits with your goal of areas not seeing text as special in any way.

This event handler in removeSeparators is going to hang around forever and get fired multiple times for the same things as various area editors are born:

```
$('[data-apos-refreshable] [data-apos-drag-item-separator]').on('transitionend webkitTransitionEnd oTransitionEnd', function() {
  $(this).remove();
})
```

Recommend you move it to the areas/public/js/user.js singleton where there are already some global event handlers, seen `enableShift`.

Shouldn't the `richText` flag and associated logic in `getDroppableAreas` die? That's to support "richText: false", is that even still a thing? I think not.

Otherwise good to go!

PR #482, opened 2016-04-05
Add more robust logic to composeFilters and composeColumns methods in…

jth-:
… pieces. Use this logic to remove the published filter from users and groups by default, since we remove that field.

### DIFF
--- a/lib/modules/apostrophe-docs/index.js
+++ b/lib/modules/apostrophe-docs/index.js
@@ -9,8 +9,7 @@ module.exports = {
     return async.series([ self.enableCollection, self.ensureIndexes ], callback);
   },
 
-  construct: function(self, options) {
-
+  beforeConstruct: function(self, options) {
     options.addFields = [
       {
         type: 'string',
@@ -48,6 +47,9 @@ module.exports = {
         def: false
       }
     ].concat(options.addFields || []);
+  },
+
+  construct: function(self, options) {
 
     self.schema = self.apos.schemas.compose(options);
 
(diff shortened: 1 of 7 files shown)


boutell:
Use beforeConstruct to manipulate options in a way that plays nicely with subclassing. That way the project-level code gets to go first and the options you're concatenating to your own are fully baked as far as the project level, or other subclass, is concerned. Your logic is otherwise correct, it just needs to be in beforeConstruct so that it's concatenating what the subclass has already messed with and so on up the line to apostrophe-pieces.

In your composeColumns and composeFilters methods, your "for" loop with "splice" is correct, but consider using _.filter for ease of understanding.

self.filters should be a clone of options.filters. This is just a startup cost, so basically free, and it means that if that options object gets modified and passed to something else, nothing breaks.

jth-:
@boutell I made changes according to feedback. Does what I'm doing with beforeConstruct now look good? I want to make sure I get that pattern correct, since I don't think we have that right everywhere just yet.

boutell:
Jimmy, this looks good.

Also, I overstated things when I was talking about not modifying the
options object, because we do that always and everywhere. I don't think you
need to clone that object. It's such a predominant practice for module
initialization that I think it makes more sense to say that the caller
needs to take care if they pass an object they intend to reuse.

On Wed, Apr 6, 2016 at 3:16 PM, Jimmy Harrison <notifications@github.com>
wrote:

> @boutell <https://github.com/boutell> I made changes according to
> feedback. Does what I'm doing with beforeConstruct now look good? I want to
> make sure I get that pattern correct, since I don't think we have that
> right everywhere just yet.
>
> —
> <https://github.com/punkave/apostrophe/pull/482#issuecomment-206520042>
>



-- 


*THOMAS BOUTELL, *DEV & OPS

jth-:
Sounds good, removed the cloning. I'm going to go ahead and merge.

PR #531, opened 2016-04-26
implemented attachment helpers such as apos.images.first, apos.attach…

boutell:
…ments.all. Also implemented migrations as we needed a type: "attachment" property on attachments to do this efficiently. All existing projects need to add apostrophe-migrations:migrate back into deployment/dependencies and deployment/migrate (in dependencies it needs the --safe flag).

### DIFF
--- a/defaults.js
+++ b/defaults.js
@@ -6,6 +6,7 @@ module.exports = {
     'apostrophe-i18n': {},
     'apostrophe-db': {},
     'apostrophe-caches': {},
+    'apostrophe-migrations': {},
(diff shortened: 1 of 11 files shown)


jth-:
Code looks very good, excited about the new API. :+1: I pulled down and ran the migration locally and verified that the attachment type was added to attachments across my project. The task caches properly. I took the liberty of removing one loosie variable declaration that wasn't used anywhere, but I'm going to go ahead and merge now.

Thanks @boutell !

PR #532, opened 2016-04-26
singleton limits working

mcoppola:
subclassing decorateManager for images-chooser, allows us to add the proper hook to visually denote existing choices in image-manager


boutell:
(Doesn't seem to have much to do with its title though?)

mcoppola:
@boutell looks like this branch was merged this morning, before I added the last commit.  @jth- referenced the related issue that this commit solves.

Issue #526, opened 2016-04-25
[versions] When on a piece show page, the "piece versions" button in the page menu should produce the instance version editor.


mcoppola:
Currently shows the parent class doc's versions.

jth-:
Merged in #541

PR #541, opened 2016-05-03
Page versions for pieces

mcoppola:
page menu for a piece now properly triggering versions for the instance of that piece, rather than the parent.  slight improvements to the version styling, giving the diff a little more room


mcoppola:
resolves #526

Issue #515, opened 2016-04-19
"info" tab should not show up by default if it has no items


colpanik:

mcoppola:
fixed in #542

PR #542, opened 2016-05-03
Resolves "info" tab appearing in editors without fields (#515)

mcoppola:
fixes bug where contextual fields such as trash were causing info schema tab to display without any fields


mcoppola:
resolves #515

Issue #514, opened 2016-04-18
[apostrophe-tags] Implement tag manager


jth-:

jth-:
Tags created in the manager should be a piece with a type of `tag` (piece with a tag with a limit of 1).

jth-:
Good progress on this, hopefully going to wrap up early in this next sprint.

jth-:
Merged in #544

PR #544, opened 2016-05-04
Implement tag editor

jth-:
This could likely still use some styling and UX consideration, but the tag editor is implemented and functional with add, rename, and delete all working. Further revisions to this functionality can be broken into smaller, more granular issues.

@mcoppola @boutell


jth-:
#514

mcoppola:
💯  All the core features here are working for me.  

One thing - when a tag is updated or removed via the manager, the search terms for items with that tag are not updated.  In example, if I update "TagA" to "TagX", a search for "TagX" returns no results, while a search for "TagA" returns the old results.  This is in spite of the fact that the tags themselves have been properly updated.

jth-:
That is a good point, I'm probably not updating the search text properly. I'll look into that.

jth-:
This is actually going to be less trivial than I imagined...does this work in 0.5? If it does I'm not sure how.

boutell:
Good point, it doesn't. It would be necessary to trigger a rebuild of the
search text. This is over and above what 0.5 does in this situation, but
it's still reasonable to call it a bug not to have it, so I suggest opening
a separate issue for it.

The issue is that lowSearchText and highSearchText are what gets directly
indexed by full text search, while tags currently aren't. I think. But
should be, and weighted right so they can be booted out of the special
search text properties. Then this would Just Work. However I'm not
referring to the code right now so I'm not 100% certain I'm right.
On May 13, 2016 1:39 PM, "Jimmy Harrison" <notifications@github.com> wrote:

This is actually going to be less trivial than I imagined...does this work
in 0.5? If it does I'm not sure how.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/punkave/apostrophe/pull/544#issuecomment-219110389>

jth-:
Alright, then I'll merge this and create a separate issue for rebuilding search text after a tag is renamed or deleted.

PR #548, opened 2016-05-09
If there is a defaultElement set in the options of apostrophe-rich-te…

### LINKED ISSUE #251: Rich Text: CKEditor should default to the first 'style' when an Area is created
Right now CKEditor starts you off with an unclassed `<div>` tag (or a forced `<p>` tag in the case of `InitialContent`). The editor should default to markup from the `style` array passed into it. Perhaps the first item (Usually Normal Text, a `<p>` tag)

CKEditor has an API for setting this explicitly 
http://docs.cksource.com/ckeditor_api/symbols/CKEDITOR.config.html#.font_style

`InitialContent` should probably also take this into account?

The win here being that if you have an Area / Text Singleton that can only be a single style (a heading, for example):
1. You don't have to write unsemantic CSS that forces an element to be something it's not
```css
.something-special .apos-area div {
font-size: 3em; /** something very headline'y **/
}
```
And 2. The correct markup and style is applied upon rendering so the user doesn't have to select it from the style dropdown (also potentially confusing because you're switching from an element that you can't switch back)

jth-:
…xt-widgets, use that instead of an empty div. If that's not set, use the first of the styles array instead of an empty div. If that's also not set, then go ahead and use the empty div. Trying to resolve #251 

@stuartromanek Assigning this to you since it is originally your issue. The `font_style` option in ckeditor didn't appear to work for this, but changing the markup we inject appears to make ckeditor automatically select a style.

### DIFF
--- a/lib/modules/apostrophe-rich-text-widgets/public/js/editor.js
+++ b/lib/modules/apostrophe-rich-text-widgets/public/js/editor.js
@@ -36,6 +36,12 @@ apos.define('apostrophe-rich-text-widgets-editor', {
 
       self.toolbar = self.options.toolbar || [];
 
+      if (self.options.defaultElement) {
+        self.defaultElement = self.options.defaultElement;
+      } else if (self.styles.length) {
+        self.defaultElement = self.styles[0].element;
+      }
+
       // This will allow loading of extra plugins for each editor
       var extraPlugins = [ 'split' ];
       // Additional standard plugins can be configured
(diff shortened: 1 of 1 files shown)


jth-:
I guess the one thing I could see as an issue here is I'm not allowing for other attributes to be passed through, which may limit flexibility..

jth-:
Stu & I talked this through, think it makes sense, merging for now.

cc @mcoppola

PR #553, opened 2016-05-17
Fix an issue where, in certain cases, moving a page ahead of another …

### LINKED ISSUE #552: [apostrophe-pages] Page order is not preserved when syncing a database.
The ordering of pages in the reorganize menu is subject to changing when syncing, for instance, from local to prod. My initial thoughts would be that, since Mongo doesn't seem to preserve natural order on sync, that something about are page ordering is falling back to natural order. I would imagine this shouldn't be the case, since it's nondeterministic, and that this should be entirely based upon page rank or something of that ilk.

jth-:
…page in the reorganize menu wouldn't nudge all the pages to a unique rank. Resolves #552

Just changed a `$gt` to a `$gte` here, because otherwise, if you move a page to a new rank that collides with the one immediately succeeding it, it doesn't nudge that page that it is colliding with.

I'm not too familiar with this logic though, so let me know if you see any issue with this @boutell

### DIFF
--- a/lib/modules/apostrophe-pages/lib/api.js
+++ b/lib/modules/apostrophe-pages/lib/api.js
@@ -206,7 +206,7 @@ module.exports = function(self, options) {
     function nudgeNewPeers(callback) {
       // Nudge down the pages that should now follow us
       // Always remember multi: true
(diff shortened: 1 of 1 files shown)


boutell:
This looks good but I am not happy that we are changing this code while our
unit tests are still in "oh well" territory.

On Tue, May 17, 2016 at 10:04 AM, Jimmy Harrison <notifications@github.com>
wrote:

> Assigned #553 <https://github.com/punkave/apostrophe/pull/553> to @boutell
> <https://github.com/boutell>.
>
> —
> <https://github.com/punkave/apostrophe/pull/553#event-663017102>
>



-- 


*THOMAS BOUTELL, *DEV & OPS

Issue #500, opened 2016-04-13
[apostrophe-images] Chooser template needs search ability/filters


jth-:

boutell:
I think you mean the manager template? But doesn't it have those things
already?

The chooser is only displaying what has already been chosen so pagination
is not an issue?

On Wed, Apr 13, 2016 at 4:23 PM, Jimmy Harrison <notifications@github.com>
wrote:

> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/punkave/apostrophe/issues/500>
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

jth-:
Specifically talking about the images chooser here, and there is currently an issue with filters displaying on managers which are slidable (like choosers). The idea is to eventually have bottomless hooked up on the images manager. Maybe these are separate issues though.

boutell:
Okay, that makes sense. Just making sure the chooser wasn't escaping its
intended scope (:

On Thu, Apr 14, 2016 at 10:59 AM, Jimmy Harrison <notifications@github.com>
wrote:

> Specifically talking about the images chooser here, and there is currently
> an issue with filters displaying on managers which are slidable (like
> choosers). The idea is to eventually have bottomless hooked up on the
> images manager. Maybe these are separate issues though.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/punkave/apostrophe/issues/500#issuecomment-209987066>
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

jth-:
@mcoppola I've currently implemented a stopgap here by bringing traditional paging back in the grid view of images. Unless you particularly need filtering for your training, I think I'll probably this one back a sprint in order to prioritize other issues, is that cool?

PR #554, opened 2016-05-20
Added filters and search to chooser modal view, which needed to be in…

jth-:
… the manage-view body because of it being a slideable modal. Made style tweaks accordingly. Also added a `allowedInChooser` option to filters, which, if explicitly set to false, doesn't display that filter in choosers (for filters like trashed and published, which shouldn't be used when joining)

Ref #500 

@boutell


boutell:
Looks good, one question: does it really make sense to forbid joining with unpublished things in all cases? For one thing, you could use it to set relationships up in advance of publication.

boutell:
Merging in the meantime but we should hash that out.

jth-:
Yeah, I'm not sure what we ultimately want to do in those particular cases, though trash would almost certainly make sense to hide. It's configuration now though, so we can adjust this depending on our decision.

Issue #535, opened 2016-04-27
[apostrophe-modals] Add confirm dialog before leaving page if unsavedChanges is true.


jth-:

jth-:
Resolved in #555

PR #555, opened 2016-05-20
Add onbeforeunload handler added on modal creation, removed on modal …

jth-:
…cancel. If there are unsaved changes, proc a confirm cancel on page unload.

We talked about this one, @boutell, let me know if it looks good. I verified that I get rid of the event when the modal is cancelled.

### DIFF
--- a/lib/modules/apostrophe-modal/public/js/modal.js
+++ b/lib/modules/apostrophe-modal/public/js/modal.js
@@ -192,6 +192,11 @@ apos.define('apostrophe-modal', {
       self.$el.on('aposModalHide', function() {
         self.getLastSlide().hide();
       });
+      window.onbeforeunload = function() {
+        if (self.unsavedChanges) {
+          return self.getBeforeUnloadText();
+        }
+      };
     };
 
     // Return the last slide or the modal itself if it has no nested slides.
@@ -437,6 +442,11 @@ apos.define('apostrophe-modal', {
       return callback();
     };
 
+    self.getBeforeUnloadText = function() {
+      return "You are about to discard unsaved changes to this "
+        + (options.label ? options.label.toLowerCase() : 'item') + '.';
+    };
+
     self.getConfirmCancelText = function() {
       return 'Are you sure you want to discard unsaved changes to this '
         + (options.label ? options.label.toLowerCase() : 'item') + '?';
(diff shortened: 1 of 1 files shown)


jth-:
Will resolve #535

boutell:
Close, but you need to push and pop these. If I open modal A and then it
opens modal B, and then I close modal B, I'll want the handler for modal A
restored.

On Fri, May 20, 2016 at 4:14 PM, Jimmy Harrison <notifications@github.com>
wrote:

> Will resolve #535 <https://github.com/punkave/apostrophe/issues/535>
>
> —
> You are receiving this because you were assigned.
> <https://github.com/punkave/apostrophe/pull/555#issuecomment-220707652>
>



-- 


*THOMAS BOUTELL, *DEV & OPS

jth-:
Good call, adjusted to store parent's `onbeforeunload` handler accordingly, and re-apply it on `hide` (which should cover save and cancel).

mcoppola:
It seems that when a hash is appended to the URL path by a javascript event, that our ability to read the history is disrupted.  This happens, for example, when the Array editor is opened.  

Still, the functionality of this branch is working.  The array editor appending a "#" could be seen as a bug with the array editor.

boutell:
Disrupts whose ability to read from the history? Have we thought through why we are touching the hash at all in the array editor? As a general rule we leave that for the project level developer, since Apostrophe's admin states can't be revisited by copying and pasting such a URL anyway, and the project level may have a hash thing goin' already when our modal opens.

jth-:
Is this merge-able then?

boutell:
I would like to understand this hash business better! Why are we using the
hash at all in Apostrophe's admin? Apostrophe admin modals are not
bookmarkable, and probably should not be unless we did it with a ton of
care to not screwing up people who are using hash URLs on the frontend...

On Tue, Jun 28, 2016 at 9:34 AM, Jimmy Harrison <notifications@github.com>
wrote:

> Is this merge-able then?
>
> —
> You are receiving this because you were assigned.
> <https://github.com/punkave/apostrophe/pull/555#issuecomment-229049999>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB9fUGM7XN1yTTm6Mm1nZgFgw0dcHPPks5qQSLhgaJpZM4Ijh27>
> .
>



-- 


jth-:
@mcoppola ?

boutell:
I am merging this branch. I think the "#" thing is probably because somebody forgot to return false, which would be a different bug not really related to this.

PR #557, opened 2016-05-23
Remove deprecated files module, which has been replaced by attachment…

jth-:
…s. Used that same module name (apostrophe-files) to create a piece and pieces-widgets subclass that functions as a home for 'office' type files.

Note that currently this will automatically pop up in the admin bar unless the admin bar has been overridden. Do we want this module to be automatically present in every project?

cc @boutell @mcoppola


jth-:
Resolves #488

boutell:
I think we do want this to be present in every project by default, because
PDFs are so common.

How hard is it, though, to shut it off? Do you have to override a template
or can you just disable the button via a flag of some sort? I forget.


On Mon, May 23, 2016 at 3:06 PM, Jimmy Harrison <notifications@github.com>
wrote:

> Resolves #488 <https://github.com/punkave/apostrophe/issues/488>
>
> —
> <https://github.com/punkave/apostrophe/pull/557#issuecomment-221060990>
>



-- 


*THOMAS BOUTELL, *DEV & OPS

boutell:
Shouldn't "subtype" be in use here to ensure people don't just get confused
thinking this is the place for *all* types of files? The "images" module is
locked down to that subtype. This one could be locked down to the office
subtype.


On Mon, May 23, 2016 at 3:20 PM, Tom Boutell <tom@punkave.com> wrote:

> I think we do want this to be present in every project by default, because
> PDFs are so common.
>
> How hard is it, though, to shut it off? Do you have to override a template
> or can you just disable the button via a flag of some sort? I forget.
>
>
> On Mon, May 23, 2016 at 3:06 PM, Jimmy Harrison <notifications@github.com>
> wrote:
>
>> Resolves #488 <https://github.com/punkave/apostrophe/issues/488>
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly or view it on GitHub
>> <https://github.com/punkave/apostrophe/pull/557#issuecomment-221060990>
>>
>
>
>
> --
>
>
> *THOMAS BOUTELL, *DEV & OPS
> P'UNK AVENUE | (215) 755-1330  |  punkave.com
>



-- 


*THOMAS BOUTELL, *DEV & OPS

jth-:
We should probably make this an option instead of having to override a template. It should be something easy to shut off.

Regarding subtype, I see that images sets a "subtype" option for their attachments field, but the string "subtype" is not used anywhere else in the codebase. Is that value actually associated with any behavior at the moment?

boutell:
Ha! We messed that up pretty good.

subtype doesn't exist anywhere else.

"group" is a real thing, attachments get a "group" property based on the
file extension and the fileGroups configuration of the attachments module.

You should, by rights, be able to configure a "group" (not subtype)
property for an attachment field, to limit that field to that group. But
there's missing plumbing at the moment.

On Mon, May 23, 2016 at 3:34 PM, Jimmy Harrison <notifications@github.com>
wrote:

> We should probably make this an option instead of having to override a
> template. It should be something easy to shut off.
>
> Regarding subtype, I see that images sets a "subtype" option for their
> attachments field, but the string "subtype" is not used anywhere else in
> the codebase. Is that value actually associated with any behavior at the
> moment?
>
> —
> <https://github.com/punkave/apostrophe/pull/557#issuecomment-221071549>
>



-- 


*THOMAS BOUTELL, *DEV & OPS

boutell:
Okay, so actually there is a "groups" property of attachment fields that is
implemented. It's an array property, so you can accept more than one group
if you want to.

This works through server side validation at the time the field is
converted.

However this is lousy UX, you don't discover it was rejected until after
you save the beast. We should be rejecting unacceptable uploads up front.

attachments.accept currently takes no options. It should take an options
object and that ought to have an options.groups property which gets
compared to the group of the attached file; if it's inappropriate the file
should be discarded and not accepted.

This brings us to how the options make it from browser to server. The
"populate" method of the attachments user.js is the obvious place, but how
exactly?

It is not hard to programmatically send extra fields with a jquery
fileupload:

```

$('#fileupload').fileupload({
    formData: {example: 'test'}
});


Then in routes.js of attachments, req.body can be passed as the options
object and we're there.

If there is no options.groups it should accept any file with an extension
that appears for any configured group (the current behavior).

(It would be even sassier if we got the list of valid extensions for the
allowed groups into browserland and, where supported, told users before
they waited for the upload. But we haven't had that before, and we still
need the above solution because there are still supported browsers that
wouldn't allow that.)

On Mon, May 23, 2016 at 3:45 PM, Tom Boutell <tom@punkave.com> wrote:

> Ha! We messed that up pretty good.
>
> subtype doesn't exist anywhere else.
>
> "group" is a real thing, attachments get a "group" property based on the
> file extension and the fileGroups configuration of the attachments module.
>
> You should, by rights, be able to configure a "group" (not subtype)
> property for an attachment field, to limit that field to that group. But
> there's missing plumbing at the moment.
>
> On Mon, May 23, 2016 at 3:34 PM, Jimmy Harrison <notifications@github.com>
> wrote:
>
>> We should probably make this an option instead of having to override a
>> template. It should be something easy to shut off.
>>
>> Regarding subtype, I see that images sets a "subtype" option for their
>> attachments field, but the string "subtype" is not used anywhere else in
>> the codebase. Is that value actually associated with any behavior at the
>> moment?
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly or view it on GitHub
>> <https://github.com/punkave/apostrophe/pull/557#issuecomment-221071549>
>>
>
>
>
> --
>
>
> *THOMAS BOUTELL, *DEV & OPS
> P'UNK AVENUE | (215) 755-1330  |  punkave.com
>



-- 


*THOMAS BOUTELL, *DEV & OPS

boutell:
Do you want me to take this one?

On Mon, May 23, 2016 at 3:54 PM, Tom Boutell <tom@punkave.com> wrote:

> Okay, so actually there is a "groups" property of attachment fields that
> is implemented. It's an array property, so you can accept more than one
> group if you want to.
>
> This works through server side validation at the time the field is
> converted.
>
> However this is lousy UX, you don't discover it was rejected until after
> you save the beast. We should be rejecting unacceptable uploads up front.
>
> attachments.accept currently takes no options. It should take an options
> object and that ought to have an options.groups property which gets
> compared to the group of the attached file; if it's inappropriate the file
> should be discarded and not accepted.
>
> This brings us to how the options make it from browser to server. The
> "populate" method of the attachments user.js is the obvious place, but how
> exactly?
>
> It is not hard to programmatically send extra fields with a jquery
> fileupload:
>
> ```
>
> $('#fileupload').fileupload({
>     formData: {example: 'test'}
> });
>
>
> Then in routes.js of attachments, req.body can be passed as the options
> object and we're there.
>
> If there is no options.groups it should accept any file with an extension
> that appears for any configured group (the current behavior).
>
> (It would be even sassier if we got the list of valid extensions for the
> allowed groups into browserland and, where supported, told users before
> they waited for the upload. But we haven't had that before, and we still
> need the above solution because there are still supported browsers that
> wouldn't allow that.)
>
> On Mon, May 23, 2016 at 3:45 PM, Tom Boutell <tom@punkave.com> wrote:
>
>> Ha! We messed that up pretty good.
>>
>> subtype doesn't exist anywhere else.
>>
>> "group" is a real thing, attachments get a "group" property based on the
>> file extension and the fileGroups configuration of the attachments module.
>>
>> You should, by rights, be able to configure a "group" (not subtype)
>> property for an attachment field, to limit that field to that group. But
>> there's missing plumbing at the moment.
>>
>> On Mon, May 23, 2016 at 3:34 PM, Jimmy Harrison <notifications@github.com
>> > wrote:
>>
>>> We should probably make this an option instead of having to override a
>>> template. It should be something easy to shut off.
>>>
>>> Regarding subtype, I see that images sets a "subtype" option for their
>>> attachments field, but the string "subtype" is not used anywhere else in
>>> the codebase. Is that value actually associated with any behavior at the
>>> moment?
>>>
>>> —
>>> You are receiving this because you were mentioned.
>>> Reply to this email directly or view it on GitHub
>>> <https://github.com/punkave/apostrophe/pull/557#issuecomment-221071549>
>>>
>>
>>
>>
>> --
>>
>>
>> *THOMAS BOUTELL, *DEV & OPS
>> P'UNK AVENUE | (215) 755-1330  |  punkave.com
>>
>
>
>
> --
>
>
> *THOMAS BOUTELL, *DEV & OPS
> P'UNK AVENUE | (215) 755-1330  |  punkave.com
>



-- 


*THOMAS BOUTELL, *DEV & OPS

jth-:
Go ahead, it looks like you have this pretty well thought through.

boutell:
I think this should be accepted and I'll open a new issue about the stuff I was raising.

jth-:
Thanks!

PR #564, opened 2016-05-25
fix for replacing files in attachment fields (the first upload worked…

boutell:
…, later uploads did not)


jth-:
Verified this is working, merging.

PR #568, opened 2016-05-26
Swap out jcrop for cropper, given that cropper has a richer featurese…

jth-:
…t, a more active community, and is more recently updated.

Also fixed several bugs:

- Fixed bug where top, left, width, or height being set to 0 would cause the crop to not show up in the cropper view.
- Fixed bug where image attachment was rendering below the rest of the form in the edit view of a piece accessed from the chooser.

I believe we now have functional parity with what we had with Jcrop. I've also refactored the crop-editor front-end module to be less opinionated in terms of cropper, so theoretically it'd be pretty easy to override with another cropping utility at project level if desired.

@boutell @mcoppola


boutell:
I don't think Apostrophe willalways respond well to cropping coordinates with fractional parts. It'll definitely make crazy URLs. You should Math.round() the coordinates.

jth-:
We are getting rounded data.

```
return self.$cropImage.cropper("getData", true);
```

The true parameter there is an option for whether the result comes back rounded:

https://github.com/fengyuanchen/cropper#getdatarounded

boutell:
Oh, baller.

On Thu, May 26, 2016 at 2:59 PM, Jimmy Harrison <notifications@github.com>
wrote:

> We are getting rounded data.
>
> return self.$cropImage.cropper("getData", true);
>
> The true parameter there is an option for whether the result comes back
> rounded:
>
> https://github.com/fengyuanchen/cropper#getdatarounded
>
> —
> <https://github.com/punkave/apostrophe/pull/568#issuecomment-221963116>
>



-- 


*THOMAS BOUTELL, *DEV & OPS

PR #570, opened 2016-05-31
Prefix support for fast login

benirose:
Adding prefix support to the login/logout windows that appear when you use the "fast login" with shift+escape.

### DIFF
--- a/views/loginWindow.html
+++ b/views/loginWindow.html
@@ -7,7 +7,7 @@
       {% if user %}
         logout bro
       {% else %}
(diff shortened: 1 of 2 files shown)


benirose:
This has been a hassle on the Way to Health dashboard for a while, finally broke down and made the change. Realized I don't have commit access anymore, but that's good because I should be using pull requests anyways :)

boutell:
Published.

benirose:
👍

PR #573, opened 2016-06-01
cursors are now fully moog-powered. Rather than overriding find() you…

### LINKED ISSUE #571: [apostrophe-docs] Cursor clone doesn't always clone everything
Docs cursor has a `.clone()` method, which is notably used to run `toCount` queries using an equivalent cursor ahead of running a `toArray` query. However, since it creates a new cursor through: 

```javascript
var cursor = self.apos.create(self.__meta.name, options);
```

It doesn't get piece-specific filters added to it that are added by overriding the pieces .`find()` method and adding filters to the retrieved cursor. For example, running a `toCount` on the image piece type doesn't get the `minSize` filter added, causing the `toCount` results to potentially differ from the results of the `toArray`. (The implementation of the `minSize` filter is in images cursor.js, the override of the pieces find for images is in images api.js).

@boutell Not sure where the fix belongs on this one. Should we change how we are adding filters for specific piece types, or change how we construct an additional cursor when cloning?

boutell:
… should provide a "things-cursor" type corresponding to your "things" piece type. You can call filters in construct() to implement different defaults which is the most common use case for find() other than adding filters. You can also call addFilters() in construct of course. Fixes #571

### DIFF
--- a/lib/modules/apostrophe-docs/lib/api.js
+++ b/lib/modules/apostrophe-docs/lib/api.js
@@ -77,11 +77,7 @@ module.exports = function(self, options) {
     if ((!req) || (!req.res)) {
       throw new Error("I think you forgot to pass req as the first argument to find()!");
     }
-    var cursor = self.apos.create('apostrophe-cursor', { docs: self });
-    cursor.set('req', req);
-    cursor.set('criteria', criteria);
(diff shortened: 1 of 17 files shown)


jth-:
Verified that this is now working in Lansdowne and WRT. Merging.

PR #575, opened 2016-06-01
Find for editing

### LINKED ISSUE #574: Subclasses of pieces that add more default filters for the frontend are frustrating to edit because you can't find future-dated blog posts and the like
I'm addressing this with a new findForEditing method which is used consistently when we want to see stuff we can edit without regard to when it was published, etc. Basically inclusive of everything except trash as long as we have the permissions for it.

boutell:
findForEditing method provides a logical extension point for changing the filters used when fetching things for editing purposes, fixes #574

### DIFF
--- a/lib/modules/apostrophe-pieces-pages/index.js
+++ b/lib/modules/apostrophe-pieces-pages/index.js
@@ -123,13 +123,18 @@ module.exports = {
     };
 
     self.showPage = function(req, callback) {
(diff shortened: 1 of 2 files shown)


boutell:
Oops, this one makes sense to merge after the moog-powered cursors one since I apparently forgot to go back to unstable first

boutell:
BOOM, I interactively rebased it, no need to wait for moog cursors to land.

PR #576, opened 2016-06-01
Html widget

jth-:
Add `apostrophe-html-widget`, which supports not rendering the widget if the option safeMode is present (as a get or a post param) or if the rendering is occuring due to an AJAX request (because in that context, a document.write could break our world).

@boutell I've made the changes we discussed.


boutell:
Beauty! @mcoppola, the html widget has landed, thank Jimmy, I just did a quick consult.

PR #581, opened 2016-06-03
Undefined `req.url` would break render pipeline

jth-:


boutell:
We decided not to do this, right?

jth-:
Correct, we're going to make our test `req`s have a url property instead.

PR #584, opened 2016-06-06
Factored out all 'req' generation into a 'testUtils.js' file. Also ad…

jth-:
…ded a url property to our test req.

This fixes the broken template tests, along with making our test suite more maintainable.


boutell:
This looks great. Assuming the tests are passing (:

Issue #567, opened 2016-05-25
[apostrophe-videos] Implement widget and piece type to embed videos from third-party sources


jth-:

boutell:
Long since built

PR #589, opened 2016-06-08
Video

boutell:
We now have an apostrophe-video widget, a video schema field type that does much of the work, and an apostrophe-oembed module that factors out that functionality. Implements #567 

An "embed" widget is still to come, however this implementation was done with an eye to making that straightforward. A little refactoring of the video field type and widget will probably be needed to make them useful as base classes for the embed widget, but the storage of the field is correctly an object rather than just a URL, so there's room to accommodate details like "alwaysIframe" and "iframeHeight".

The title and thumbnail URL of the embedded content are stored along with the url in the video field:

video.url (ready for oembed)
video.title (captured from oembed when initially saved)
video.thumbnail (URL; captured from oembed when initially saved)

The widget displays the thumbnail initially, with the title and a progress animation superimposed, and that gets swapped out for the player as soon as a fresh oembed response is available. This also improves both accessibility and SEO, so it should be maintained in some way even if we don't like it as-is. The initial appearance is surprisingly unterrible, but some attention to my title font choice and spinner choice is probably in order.


mcoppola:
@boutell Nearly there and looking good!  I've found an error that crashes the server, when a bad YouTube URL is provided.  The `openGraph` method of the oembed api fails on `self.truncatePlaintext` [(here)](https://github.com/punkave/apostrophe/blob/8e6ea2714d44d6ca0922d0f0eb02ff55ff2569ec/lib/modules/apostrophe-oembed/lib/api.js#L151), which isn't available.

boutell:
Oops! I will dig into that.

On Thu, Jun 9, 2016 at 10:39 AM, Matthew Coppola <notifications@github.com>
wrote:

> @boutell <https://github.com/boutell> Nearly there and looking good! I've
> found an error that crashes the server, when a bad YouTube URL is provided.
> The openGraph method of the oembed api fails on self.truncatePlaintext
> (here)
> <https://github.com/punkave/apostrophe/blob/8e6ea2714d44d6ca0922d0f0eb02ff55ff2569ec/lib/modules/apostrophe-oembed/lib/api.js#L151>,
> which isn't available.
>
> —
> <https://github.com/punkave/apostrophe/pull/589#issuecomment-224915464>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB9fd-M9CHd7gJyQ-wNhy6LND2XkVHgks5qKCWOgaJpZM4IxcCL>
> .
>



-- 


*THOMAS BOUTELL, *DEV & OPS

boutell:
This has been addressed, please take another peek. Thanks!

I also handled a *very* basic error/inappropriate content indicator.

I'm noticing that when you paste a URL in the widget editor your enhanced progress meter looks wacky because there is no thumbnail to superimpose it on yet.

mcoppola:
@boutell I fixed fixed a couple cases where an `undefined` url would break the front-end video/user.js, and also crash the server in the `query` method of the oembed module.  I think this is good to go if you can review those adds.  

A+ for that bad video/url gif in the preview.

boutell:
I made one more round of changes and merged this. My change was to consolidate responsibility for falsy URLs in queryAndPlay, and to make a change to ensure new video widgets work with the latest logic. I had neglected to make sure "video" existed before looking at "video.html".

PR #590, opened 2016-06-10
fixed bug in image cropper that occured when original src image was l…

mcoppola:
…arger than 'full' size.  fixed by requesting 'original' size attachment for cropper, and by reimplementing scaleUp/scaleDown serialize methods

### DIFF
--- a/lib/modules/apostrophe-attachments/public/js/cropEditor.js
+++ b/lib/modules/apostrophe-attachments/public/js/cropEditor.js
@@ -11,7 +11,7 @@ apos.define('apostrophe-attachments-crop-editor', {
       self.cropWidth = self.$cropImage.width();
       self.$saveButton = self.$el.find('[data-apos-save]');
       self.busy(true);
-      self.$cropImage.attr('src', apos.attachments.url(self.attachment));
+      self.$cropImage.attr('src', apos.attachments.url(self.attachment, {size: 'original'} ));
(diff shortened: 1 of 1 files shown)


mcoppola:
@colpanik @boutell could you check this PR?  especially Kerry if you could test in your projects that the cropper still works 👌.

I was seeing a bug with images larger than the "full" size.  The cropper coordinates wouldn't be relative to the original file.

boutell:
Makes sense. It could be a big download, but cropping a tiny piece of a gigantic image is a legitimate use of our cropper, so it needs be there.

mcoppola:
cool! thanks tom

boutell:
Actually Matt, it looks like you added in some missing scaleUp/scaleDown
calls, which ought to make it work even if the image is not the original...

However, I think we should keep it this way because of the use case I
pointed out. (We have zooming enabled in the cropper right?)

On Fri, Jun 10, 2016 at 1:22 PM, Matthew Coppola <notifications@github.com>
wrote:

> cool! thanks tom
>
> —
> <https://github.com/punkave/apostrophe/pull/590#issuecomment-225243301>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB9fQocwcYIU1Q2QatSoZrLawtlSYdRks5qKZ1MgaJpZM4IzJwD>
> .
>



-- 


*THOMAS BOUTELL, *DEV & OPS

PR #592, opened 2016-06-10
Minify is fixed. The jqtree.css code was crashing when treated as LES…

boutell:
…S, I removed an IE6/7/8 workaround we don't need and now it's fine. There were other bugs that prevented minify from completing and are now fixed. Also Apostrophe was failing to print initialization errors and gracefully exit. md5File is now in utils because it is not really specific to attachments.


boutell:
Closes #591

PR #602, opened 2016-06-30
Fix issue where a 'group' select field in the user editor would always reset …

jth-:
…to 'guest' (or whatever the first group option was), regardless of the actual value. This was happening because that 'group' field doesn't actually correspond to a real field, but rather is just an easier to work with frontend for a `joinByArray`. The conversion (saving) pipeline took this into account, but when retrieving the record from the database we never set the value of the field in order for it to properly populate on the frontend.

### DIFF
--- a/lib/modules/apostrophe-users/index.js
+++ b/lib/modules/apostrophe-users/index.js
@@ -326,6 +326,16 @@ module.exports = {
       });
     };
 
+    var superRequirePiece = self.requirePiece;
+    self.requirePiece = function(req, res, next) {
+      return superRequirePiece(req, res, function() {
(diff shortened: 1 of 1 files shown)


boutell:
Good catch.

On Thu, Jun 30, 2016 at 4:42 PM, Austin Starin <notifications@github.com>
wrote:

> Merged #602 <https://github.com/punkave/apostrophe/pull/602>.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/punkave/apostrophe/pull/602#event-709722894>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAB9fVrDKX7jZ7xF-8NtKiQk-FvSSdipks5qRCpCgaJpZM4JCkbF>
> .
>



-- 

PR #603, opened 2016-06-30
the pieces pages default index template should contain the pager as w…

boutell:
…e are already starting to see projects missing the pager


jth-:
Looks good to me. This template won't often be used, given that most projects will extend a `layout.html`, but this should serve as a good reminder to implement paging.

PR #606, opened 2016-07-07
Fallback area and singletons to schema field options

jth-:
If no options are specified for a given area or singleton, fallback to use the options specified for the schema field declaring the area or singleton, if they exist.

Not sure if I have this quite right, let me know what you think @boutell

### DIFF
--- a/lib/modules/apostrophe-areas/lib/api.js
+++ b/lib/modules/apostrophe-areas/lib/api.js
@@ -152,6 +152,15 @@ module.exports = function(self, options) {
     });
   };
 
+  self.getSchemaOptions = function(doc, name) {
+    var schema = self.apos.docs.getManager(doc.type).schema;
+    var field = _.find(schema, 'name', name);
+    if (!(field && field.options)) {
+      return {};
+    }
+    return field.options;
+  };
+
   // merge new methods with all apostrophe-cursors
   self.apos.define('apostrophe-cursor', require('./cursor.js'));
 };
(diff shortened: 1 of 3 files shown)


boutell:
Feels good, merged. I went over your other changes, which seem to only affect legibility of some peripherally related code. I assume you've test-fired this?

jth-:
I tested this both with areas and singletons, seems to work well for both! 

Yeah, the other changes should be functionally-equivalent with one exception: we now always add the `.apos-area-widgets` wrapper around the widgets of an area, where prior we only added it if the user had edit permissions. This seems safer, because otherwise you could theoretically style `.apos-area-widgets` and have unexpectedly different styles when logged out.

boutell:
Are you talking about adding an entire div that wouldn't otherwise be there?

We made that choice because historically frontend developers have asked for
less markup for logged out users whenever we don't need it.

I recommend you chat with Kerry about it, basically somebody other than me
(since I'm the backend guy who got told not to do it the first time in A1.5
(: )

On Thu, Jul 7, 2016 at 2:23 PM, Jimmy Harrison <notifications@github.com>
wrote:

> I tested this both with areas and singletons, seems to work well for both!
>
> Yeah, the other changes should be functionally-equivalent with one
> exception: we now always add the .apos-area-widgets wrapper around the
> widgets of an area, where prior we only added it if the user had edit
> permissions. This seems safer, because otherwise you could theoretically
> style .apos-area-widgets and have unexpectedly different styles when
> logged out.
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/punkave/apostrophe/pull/606#issuecomment-231164625>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB9fXVofnCyEbLFgDJcT8S3W9IfGi7jks5qTUQHgaJpZM4JHP4W>
> .
>



-- 


jth-:
Fair enough, I went ahead and removed it: 41adc1576ce07569355f29b131901a243ce7e9b4

PR #607, opened 2016-07-08
fixed contextual sanitization so it does not blindly trust contextual…

boutell:
… fields but we can still accept them when present


jth-:
Verified to be working with nested widgets in WRT

PR #628, opened 2016-08-13
Permissions editing

boutell:
NOTE: I assigned everybody as PR reviewers because I didn't want to leave anyone out of what I think is a very important read, but I'jm assuming I'll go through this with Greg or Kylee on Monday to actually land it. CoP needs that to happen ASAP.

Lots of new things here. I needed to improve a lot of infrastructure in order to land granular permissions in 0.6.s

* PIECES AND PAGES BOTH HAVE PERMISSIONS FEATURES EQUAL TO 0.5 (MINUS WORKFLOW)

Both pieces and pages can now have nuanced "view" and "edit" permissions locked down by user and group. (You must be an admin to change edit permissions for anything.)

This time the permissions editing interface was built using normal A2 0.6 schema forms rather than as a one-off world unto itself. This required a lot of legwork but I think it was ultimately worth it.

* ANY SCHEMA FIELD CAN REQUIRE A PERMISSION

If you add `permission: 'admin'` to a schema field definition, you have to be an admin to see that field at all or change its value. Widely requested, and necessary for the permissions editing.

* APOSTROPHE ADMIN BAR
Implemented apostrophe-admin-bar, which provides the admin bar as a well-thought-out service and understands permissions properly. Here is what the pieces module does to register the button for each pieces subclass:

```
    self.apos.adminBar.add(self.__meta.name, self.pluralLabel, self.isAdminOnly() ? 'admin' : ('edit-' + self.name));
```

First argument: name of item (for reference when grouping them). Second: label. Third: the permission the user must have, which can be null if no permission is required other than being logged in.

Fourth: an optional `options` object, which can use an `after` property to put this item after a specific item.

This produces a data attribute on the browser side:

`data-apos-admin-bar-item="apostrophe-events"` (for example)

On the browser side a convenience method is provided to link this to a function. Example:

```
apos.adminBar.link(self.__meta.name, function() {
  self.manage();
});
```

A module can also group items together into a dropdown:

`self.apos.group([ 'apostrophe-users', 'apostrophe-groups'])`

Or give it a special label:

`self.apos.group({ label: 'People', items: [ 'apostrophe-users', 'apostrophe-groups'] })`

But most of the time, the developer will make grouping choices:

```
modules: {
  'apostrophe-admin-bar': {
    addGroups: [
      { label: 'People', items: [ 'apostrophe-users', 'apostrophe-groups'] }
    ]
  }
}
```

(You can use `groups` instead if you don't want to keep any groups that modules suggested.)

I just wrote an epic one for Science Center, replacing the hand-written menu code that was not permissions-ready.

If the user lacks all of the permissions or is not logged in, they automatically get no admin bar at all. (Generally they will always be allowed the logout action though.)

`outerLayoutBase` now calls `apos.adminBar.output` to render this.

`logo.html` has moved to the `apostrophe-admin-bar` module.

Please DON'T custom-template the admin bar anymore; use this functionality so that your project is permissions-aware. Talk to me if you have concerns about the feature set.

* JOINS NOW WORK FOR CUSTOM PAGES WHEN VIEWING THOSE PAGES

From now on, when you fetch docs and/or pages, joins always run unless the joins() filter is set to false, even if you didn't call the `type` filter. This fixes the issue that you weren't able to see any joins on a page that introduced them via its `apostrophe-custom-pages` schema. TODO: it can also simplify search but I haven't taken full advantage yet.

* EASY CREATION OF BROWSER-SIDE SINGLETONS FOR ADMIN FEATURES OF MODULES

Added a `pushCreateSingleton` method to all modules. Tphis method takes care of instantiating the browser-side singleton with the same moog type name as the module, and does it safely so you don't wind up with duplicate event handlers. It calls back to `getCreateSingletonOptions` to get the object to pass to the new singleton as `options`, which by default just contains the `self.action` property. So just override `getCreateSingletonOptions` to pass more.

This happens only if the user is logged in.

You can wait until `pageBeforeSend` and pass `req` to `pushCreateSingleton` if your options depend on exactly who the user is.

You have to actually call `pushCreateSingleton` if you want it. Not all modules need browser-side admin code. Pushing the `user.js` that defines the singleton's moog type is also up to you.

`apos.modules['full-module-name']` and `apos.moduleAlias` are automatically made available.

Many modules have their own `pushCreateSingleton` already, overriding this. It would be nice to migrate them to it. I migrated those I needed for permissions to work well.

* EASY REGISTRATION OF NEW PERMISSIONS TO BE OFFERED FOR GROUPS

If your module has a custom permission (not just the standard pieces stuff), you can call `self.apos.permissions.add({ value: 'edit-thingies', label: 'Edit the Thingies' })` and it will be available as a permission that can be assigned to groups. If you start the name with `edit-`, anyone with the `editor` permission can also automatically do it. If you don't, they must have that specific permission (or be an admin, of course).

* "EDIT ACCOUNT" FEATURE TEMPORARILY REMOVED

This was just a button to edit the user's own piece — it was clever but it is not permissions-safe at all, they could add themselves to new groups, etc. We need to replace it with a real "reset your password, change fields you are allowed to change, etc." stack for 0.6.


boutell:
As far as I'm concerned, landing this PR makes unstable a release candidate for 2.0.0 — fix deprecation warnings, write the more obviously missing tutorials and HOWTOs, publish that thing.

PR #631, opened 2016-08-18
refactored doc type managers to moog by introducing the apostrophe-do…

boutell:
…c-type-manager base class for modules that manage any type of doc. No more hacky mixins. All tests passing, also tested with UE, Lansdowne and Sci Center.


boutell:
A “doc type manager” is the object responsible for providing a custom “find” method for a given type in aposDocs, among other things. The docs module provides a “getManager” method to retrieve these. They are used to carry out joins and other cool things.

A module that subclasses pieces becomes the doc type manager for that piece type. A module that subclasses custom-pages becomes the doc type manager for that page type. And so on.

However, the “base class” — the shared schema fields and shared methods of all doc type managers — was a hot steaming mess in a spooky file, docs/lib/manager.js.  Anything that wanted to inherit its features had to call getManager (which created a generic manager based on that file), merge its methods into itself with _.defaults (bypassing moog), and then call setManager to become the official manager. It was like nothing else in A2, it was a shitty way of implementing what some languages call mixins or multiple inheritance or traits.

As a result there was no good clean obvious place to extend or change the methods provided by *all* managers, because they had no base class in moog. Want to change the default behavior? Good luck with that.

With this PR, we get “apostrophe-doc-type-manager”, a module that is the base class for all modules that plan to manage a doc type (pieces, custom-pages, global, a few oddballs).

In addition, we get plumbing that:

* Makes a dynamic subclass of custom-pages on the fly for every “ordinary” page type.
* Makes a dynamic subclass of doc-type-manager for any “orphan” types that exist in your database, but also complains loudly about it, so you know you have crap that’s not relevant anymore in your database.

PR #637, opened 2016-08-28
autoload cursor definitions

boutell:
## TL;DR SUMMARY

With this change, any subclass of pieces with a "lib/cursor.js" will automatically load it. This includes project level.

It must export a moog definition (an object with `construct` or perhaps `afterConstruct` functions). The cursor is "self" in these functions. In other words it looks just like a module "index.js" file. 

No "extend" property is required, unless you're actually looking to avoid extending the usual thing.

Boom!

No overriding "self.find", no calling "apos.define". Nuttin. It just works. 

This addresses the pain point of remembering how to `apos.define` and require cursors, the one thing we frequently have to manually define on the server side, every time we want to add filters. I think this has been one of the reasons why folks have been slow to stop adding filters in `find`, which is wrong (it breaks pagination).

## BC BREAK

This impacts all the projects that are defining cursors the hard way, or requiring a cursor.js file from find() calls. But the former is unnecessary work and the latter is wrong (it breaks pagination). So let's just fix them. I have fixes standing by for the blog, events, and places modules, plus sci center and urban engineering. I will walk everybody through anything that comes up with this; I really want to land it for 2.x.

## IMPORTANT TO KNOW

*"So what is cool to do in an override of a `find` method?*

It is OK to CALL filters there, but you should NEVER ADD filters there. Do that in cursor.js.

You know what's better though? Call those filters in `afterConstruct` in your `cursor.js`. Don't bother overriding `find`.

"What if I need access to my module in my filters?"

You can get at the module from the cursor:

`self.options.module`

## HOW MUCH DO YOU TRUST THE CODE?

I tested the code pretty thoroughly. I'm more interested in buy-in for making this change. I feel strongly enough about it that I think we should do it before publishing 2.0. It would definitely be a "3.0" BC break if we didn't.

## MORE ANTICIPATED QUESTIONS

*"What about other doc type managers?"* Oh cool, you've been paying attention. Correct, this is actually for all doc type managers, but most folks are generally familiar only with pieces.

*"What if your parent class also has a cursor.js?"* They both work, they extend like you hope they do.

*"So if I put cursor.js in pieces..."* Yes, you can have a lib/modules/apostrophe-pieces/lib/cursor.js and that affects *all* pieces.

*"Do I have to have an index.js for the module?"* No, you don't have to have an index.js for the module *at all*, it still just works.

*"Is this a special snowflake one-off pattern to clutter up our world?"* Nope! If you want to create a cool new "related type" like cursors for your module, check out the new "defineRelatedType" and "createRelatedType" methods, now available on all modules. These kill the weirdness of defining and creating cursors and other things that are objects, but are not modules. You get snazzy autoloading of definitions, just like for cursors.

*"What about pages?"* Pages break out of this pattern, for reasons (they are loaded by apostrophe-pages-cursor, they are of many different types, it's a bad assumption to think their own cursor will get used).

*"What about adding methods to ALL cursors? Even the generic `apostrophe-cursor` that doesn't care about types at all?"* Wow, you really have been paying attention. Yes, we do that sometimes, but it's a big deal and I think explicitly calling `apos.define('apostrophe-cursor', require(...))` makes sense for such a momentous occasion.

### DIFF
--- a/lib/modules/apostrophe-custom-pages/index.js
+++ b/lib/modules/apostrophe-custom-pages/index.js
@@ -61,16 +61,17 @@ module.exports = {
 
     // Return a cursor for finding pages of this type only. The cursor is an
     // `apostrophe-pages-cursor`, so it has access to filters like
-    // `ancestors` and `children`. Subclasses will often override this
-    // to create a cursor of a more specific type that adds more filters
+    // `ancestors` and `children`.
+    //
+    // Because pages are normally displayed by Apostrophe's page loading mechanism,
+    // which uses an `apostrophe-pages-cursor`, it doesn't really make sense to return
+    // a custom cursor subclass here. It would not be used when actually viewing the
+    // page anyway. If you must have extra filters for specific page types, implicitly
+    // subclass apostrophe-pages-cursor and add filters that are mindful of the
+    // type of each page.
 
     self.find = function(req, criteria, projection) {
-      return self.apos.create('apostrophe-pages-cursor', {
-        apos: self.apos,
-        req: req,
-        criteria: criteria,
-        projection: projection
-      }).type(self.name);
+      return self.apos.pages.find(req, criteria, projection).type(self.name);
     };
 
   }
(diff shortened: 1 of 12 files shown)


boutell:
Here is the corresponding PR for science center, so you can see how simple it is to be compatible with the change:

https://github.com/punkave/science-center/pull/186/commits/e64ad3c83cf97f6bf049e341cf656436cc0cfade

colpanik:
👍

PR #653, opened 2016-09-09
files widget should display browse button in the same position that t…

boutell:
…he images widget does


boutell:
This is a copy and paste of the editorModal template for the images widget. Looks much better and keeps the actual upload button from getting buried in the Info tab, which I think qualified as a bug in UX terms at the least.

Nice schemas hack, whoever did this for the images widget.

PR #659, opened 2016-09-15
apos.areas.richText helper and method

boutell:


boutell:
This helper replaces the old aposAreaRichText helper, which returns the rich text found in an area. Please see "files changed" for complete inline docs on how it will behave. Note that the default behavior is to join the rich texts from various widgets with a newline. Does that make sense? What is the best default behavior? Do we like my "delimiter" and "wrapper" options?

Speak now or forever hold your peace; I have an ES client waiting on this.

Thanks!

PR #670, opened 2016-09-20
100% height for login takeover page

grdunn:
fix for boilerplate `login.html` styles. feels a little strange pushing styles for `html` and `body` just within this module.


grdunn:
Refactored this a bit not to include global html / body styles.

PR #677, opened 2016-09-22
showFields fixes: function properly if a field is shown by two choice…

boutell:
…s. Function properly when `choices` with `showFields` have been set for boolean fields. Ignore `required` for fields that are hidden by `showFields`, so that it can be used as intended to say that these things are required only in the presence of a particular select or boolean choice.


boutell:
@austinstarin I've assigned this PR to you since you opened a ticket about it.

PR #691, opened 2016-10-04
added slugs to autocomplete titles for all page types and for the "an…

boutell:
…y page" case, in a way that extends whatever projection is being returned for all doctypes. In addition, fixed a bug in the apostrophe-global module: it would crash on startup if an existing "global" doc contained widgets referencing doc types not yet initialized, such as "plain old page types" which get initialized at the last minute. Also, this module now correctly respects its "slug" option.


boutell:
Oops, this was supposed to go out last night but only an earlier version of it did, so I published it a 2.2.2 as it includes an important bug fix.

Issue #705, opened 2016-10-17
minSize cropping constraints don't work


boutell:
Our minSize cropping constraints flat-out don't work in 2.0.

We are using cropper:

https://github.com/fengyuanchen/cropper

It doesn't support this. It has `minCropBoxWidth` and `minCropBoxHeight` options but they are "relative to the page," whatever that means, and don't take into account scaling of the image. So our current implementation is a nonstarter.

Here's a hack that apparently works without modifying the plugin:

http://jsfiddle.net/dekkard/9r6czo65/12/

That's not complete but it does suggest the appropriate points to add the code.

Throwing out the plugin would be a bc break. We could fork it in a backwards compatible way, or find a way to do it without plugin modifications.

boutell:
Also: https://github.com/fengyuanchen/cropper/issues/339

boutell:
This may be the right workaround:

https://github.com/fengyuanchen/cropper/issues/576

PR #708, opened 2016-10-19
fix crop minSize

boutell:


boutell:
This is the fix for #705

PR #709, opened 2016-10-19
Defines and requires fs

### LINKED ISSUE #707: fs is never defined in apostrophe-express
[apostrophe-express/index.js#L394](https://github.com/punkave/apostrophe/blob/dfc813d856495d768a526fcc5372a013e83094b3/lib/modules/apostrophe-express/index.js#L394) and [apostrophe-express/index.js#L404](https://github.com/punkave/apostrophe/blob/dfc813d856495d768a526fcc5372a013e83094b3/lib/modules/apostrophe-express/index.js#L404) fails every time because fs is never defined. 

Error: `[ReferenceError: fs is not defined]`

poacher2k:
Closes #707

### DIFF
--- a/lib/modules/apostrophe-express/index.js
+++ b/lib/modules/apostrophe-express/index.js
@@ -141,6 +141,7 @@
 // [connect-multiparty](https://npmjs.org/package/connect-multiparty) npm module.
 // This middleware is used by [apostrophe-attachments](../apostrophe-attachments/index.html).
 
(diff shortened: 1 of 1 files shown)


boutell:
Ha! The dangers of try/catch...

Since setting the PORT and ADDRESS environment variables works and is a much more common technique, we're not going to rush this one out, but it will go out in the next routine patchlevel.

Thanks!

PR #714, opened 2016-10-22
Corrected link to tutorial page

fredrikekelund:
Was missing a ".html" extension, and the hash wasn't quite right either


boutell:
Thanks!

FYI you should sign up for our developer directory if you're building sites with Apostrophe:

https://tomboutell1.typeform.com/to/NeqckD

Issue #647, opened 2016-09-08
Admin buttons block content areas


alexgleason:
This is annoying if you're trying to click a particular element and the buttons get in the way, or if the button itself is taller than the content element and you get buttons overlaid on top of buttons... etc.

![screenshot from 2016-09-08 00 12 05](https://cloud.githubusercontent.com/assets/3639540/18336891/556478e4-7559-11e6-9cb3-5ac0f09e7820.png)

What's worse, the markup/styles don't make it simple to just hover the buttons to the right side. I realize that since this is happening on the front-end it will require some customization on each site to make things look perfect, but there has to be a better way. Could the buttons be outside of the hovered region?

![screenshot from 2016-09-08 00 12 44](https://cloud.githubusercontent.com/assets/3639540/18336904/7dc11928-7559-11e6-8603-e64d937deab2.png)

Here's an extreme example where I've added two empty fields. Note that it also happens on single line text fields though, which isn't such a fringe case.

stuartromanek:
addressed in #720

PR #699, opened 2016-10-09
Align richtext contextual hover controls above text

alexgleason:
Sample area:
screenshot from 2016-10-08 21 15 04

The problem (hovering over the text covers it when it's only a single line):
screenshot from 2016-10-08 21 15 18

Fixed:
screenshot from 2016-10-08 22 12 04

EDIT: This still needs work, actually. I just discovered that this will interrupt the hover states of adjacent objects.

PR #720, opened 2016-10-24
[Area UX] Changes to Areas controls

stuartromanek:
Addresses #647 & #699 

A couple subtle changes:
## Rich Text Controls

Rich text widget controls now float completely above the widget, leaving the content of the widget unobscured while interacting with it. To achieve this without interfering with the widget above I'm toggling the `pointer-events` property to only be active when moving from the widget to the control set.

image
## Areas inside Areas (content widgets inside of layout widgets)

The classic Areas inside Areas pattern where the top and bottom controls of the outer widget are on the same horizontal as the first widgets of the inner areas

image

I've added a `role` option to the `apos.area` configuration that can be set to `layout`

```
{# ########################### #}
{# THE LAYOUT WIDGETS #}
{% macro layout(context, name) %}
  {{ apos.area(context, name, {
    widgets: {
      'pl-marquee': {},
      'pl-block-title-column': {},
      'pl-block-one-column': {},
      'pl-block-narrow-one-column': {},
      'pl-block-one-two-column': {},
      'pl-block-two-one-column': {}
    },
    role: 'layout'
  } )}}
{% endmacro %}
```

This gives some special consideration to the outer area. It gives its UI the set `apos-secondary` color and it also sets its controls to be to be transparent but _in document flow_ (@boutell don't fall out of your chair). Ultimately attempts to put the outer area controls in special offset places felt worse than the occasional 20px of white space between larger widgets.

The interface then works like this

image

Area interactions have lots of edge cases so I'd like to have some people test this branch in real life projects, hopefully ones with complex layout configurations. Also looking for feedback on the `role: layout` separation. This is a pattern both my 2.x projects have used, hoping it will be useful to people besides me.

Would be great if you could bash around on this in a complex page layout and see how it feels.

@agilbert @austinstarin @grdunn @jonlong @colpanik @boutell @kyjoya @bgantick

### DIFF
--- a/lib/modules/apostrophe-areas/public/css/user.less
+++ b/lib/modules/apostrophe-areas/public/css/user.less
@@ -113,22 +113,27 @@
   padding: 0;
 }
 
+// [data-apos-area-controls] .apos-area-controls { opacity: 1 !important;}
 .apos-ui .apos-dropdown.apos-dropdown--area-controls
 {
+  height: 5px;
   display: inherit;
   .apos-drop-shadow(0, 0, 0, 0);
   >.apos-button
   {
+    position: relative;
+    top: -7px;
     border: 2px solid @apos-white;
     .apos-rounded(@apos-padding-3);
     .apos-drop-shadow();
+
   }
   >.apos-dropdown-items
   {
     text-align: left;
     left: 0;
     right: 0;
-    top: 12px;
+    top: 6px;
     margin-left: auto;
     margin-right: auto;
     max-width: 200px;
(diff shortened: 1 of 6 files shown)


boutell:
Codewise I'm good with it.

stuartromanek:
U P D A T E
The syntax of this changed to the much less coded `blockLevelControls: true` as an option in your `apos.area` call. Also gave area controls a peek when hovering the widget.

I'm done hacking on this branch, has anyone tested it out?

stuartromanek:
this merge closes #647 and #699

PR #723, opened 2016-10-26
optional inline schema forms for editing relationships; use them for …

boutell:
…permissions editing by default

### DIFF
--- a/lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
+++ b/lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
@@ -14,6 +14,8 @@ apos.define('apostrophe-doc-type-manager-chooser', {
   },
   afterConstruct: function(self, callback) {
     self.removed = self.options.removed;
+    self.enableInlineSchema();
+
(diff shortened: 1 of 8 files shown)


boutell:
The immediate problem: nobody can find the "apply to subpages" option for individual user and group permissions. It's buried behind the "edit relationship" button for each person or group you've chosen.

So users assume there isn't one, or they misuse the "apply to subpages" option intended for the overall view permissions (I've seen both scenarios go down).

More generally: while join relationship editing is better in theory in 2.x because it supports real schemas, it's worse because it's hard to find.

My solution here: you can set "inline: true" for any individual field in a joinByArray `relationship` schema. And, the permissions relationship fields do this by default.

So you wind up with this:

https://dl.dropboxusercontent.com/spa/jfyjetktfp495r1/q1oipf28.png

Which feels like a significant improvement over hunting down the little list icon to open a separate modal to pick one thing.

We have always discouraged complex relationship forms, so I would expect most people who use them to opt for inline: true on all of the fields. For bc I don't want to force it though.

PR #725, opened 2016-10-28
Apply action button styles to view file link

austinstarin:


boutell:
Merged, thanks. I think we should also have a "View URL" button to make it painfully obvious there's a way to get the URL. It comes up super often.

PR #726, opened 2016-10-28
Select input type style enhancements

austinstarin:
Center arrow in input wrapper and add cursor pointer for hover state


boutell:
Thanks! btw why the hotfix/ prefix? I would think of a "hotfix" as something done off a release branch to rush it into production without waiting to merge with newer work on master.

PR #729, opened 2016-11-02
if you create a piece when using the manager to choose pieces, that p…

boutell:
…iece is automatically chosen. A limit reached message has been added to the chooser, but only when limit is 2 or more. When limit is 1, a new selection automatically replaces the old one, like in every program ever. When choosing pieces the manager modal shows correct singular language if limit is 1. A new pieceInserted apos event is emitted when the piece editor modal is saved for a new piece.


stuartromanek:
made nice progress on this

progress

passing back to @boutell because I removed some js that was triggering hide/show and moved its responsibility to CSS, want to make sure i'm not stepping on anything important

Issue #736, opened 2016-11-08
2.6.0 breaks array field JS


jonlong:
I'm having an issue where the only solution I've found is rolling back to 2.5.1.

Widgets that feature `array` fields appear to have unresponsive chooser buttons in the Admin UI. Here's a quick look at what I'm seeing — this is a slideshow widget where each slide is added as an array item:

![image](http://g.recordit.co/iHsrCRoSGF.gif)

stuartromanek:
also seeing this ... @boutell possible side effect of #723 ?

boutell:
Taking a look.

boutell:
Pull request for review in #737

jonlong:
Confirmed that #737 fixes the issue for me. Thanks for the quick fix @boutell!

PR #737, opened 2016-11-08
fixed the array edit button. When apos.schemas.findSafe started actua…

boutell:
…lly working properly, buggy code in the array editor was revealed. Introduced a findSafeInFieldset method, an affordance to developers who might otherwise make the relevant mistake.


jonlong:
Confirmed that this fixes #736 for me

PR #743, opened 2016-11-10
declutter the versions database by robustly checking whether saving a…

boutell:
…n area is really necessary after server-side sanitizers have run. We were getting burned by differences that do not matter, like `&nbsp;` versus the unicode character for it.


bgantick:
Looks sane to me 👍

PR #753, opened 2016-11-22
apostrophe now has concurrency locks. apos.locks.lock("lockName", cal…

boutell:
…lback); apos.locks.unlock("lockName", callback);


boutell:
Reviewed with Austin.

PR #755, opened 2016-11-28
support for generating asset bundles to be handed off to prod environ…

boutell:
…ments where it is awkward to run apostrophe:generate or we want every server to get the same results. Also support for various environment variables to set common cloud requirements like the mongo URI and the S3 settings. Lightly refactored the assets module, breaking it up into methods called in afterConstruct so it is easier to understand, in hopes this will encourage contributions to its pipeline such as autoprefixer and source maps.


austinstarin:
👍

PR #757, opened 2016-11-28
Lodash’s indexBy (keyBy) as helper

austinstarin:


boutell:
Idea is fine, but please fix the comment documenting it and the name of the second parameter. Sure, in lodash "iteratee" can be a worker function *or* a property name, but in lodash it can only be a property name, so use that less confusing term.

PR #762, opened 2016-11-28
Allow body-parser customization at the app.js level

fotisp:
Apostorphe uses body-parser however this cannot be customized at the application level. Not even by overriding the middlewares using the "middelware" array.

body-parser also provides a way to set the "limit" of the request body, which is a common solution for the *request entity too large* nodejs error.

The proposed change allows the developer to customize the body-parser "limit" at the app.js level, or data/local.js level

### DIFF
--- a/lib/modules/apostrophe-express/index.js
+++ b/lib/modules/apostrophe-express/index.js
@@ -277,12 +277,21 @@ module.exports = {
         self.apos.app.use(self.csrf);
       }
 
+      
+      // Allow the options for bodyParser to be customized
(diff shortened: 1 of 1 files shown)


boutell:
Thanks, Fotis.

FYI I will be stripping out those leading underscores. We only use that to mean one thing anymore: properties of a doc that shouldn't be persisted to the database.

(Well, OK: and you'll occasionally see _foo in a callback when I need to receive it and then assign it to the real foo. But honest, that's it.)

PR #783, opened 2016-12-27
#771 made static texts translatable

timaebi:
I added a few __(...) for translations.
The i18n library used for translation renders an error if the message key is undefined. Unfortunately in some cases this has happened and therefore I added the `| d('')` to avoid errors.


boutell:
Thanks! I changed the approach for the login error message a little and in the process made them actually work (they didn't before, whoops). Otherwise accepted it as-is.

PR #784, opened 2016-12-27
sourcemaps in dev

boutell:
This is just basic sourcemap support for less files in dev, where the less-middleware is in use. Just needed to be configured properly.

However there is a known issue in both Chrome and Firefox that hampers the utility of this:

https://github.com/less/less.js/issues/2033


boutell:
Yooo so does anybody want these sourcemaps I built? ( see this though: https://github.com/less/less.js/issues/2033 )

austinstarin:
This is great, let's merge. Regardless of the Chrome issue I think this is still valuable or better than no sourcemaps.

boutell:
Great, merge away

PR #785, opened 2017-01-04
wondering if it makes sense to give the area widget controls more z-index

agilbert:
The issue I'm trying to solve for is when you have a widget that with editing controls that would always be obscured by the Apostrophe menu button or the page menu button. In my specific case, it was the footer:

image

Without this change, it's basically impossible to edit my footer content.

The change does mean that as you scroll by widget controls will hover over the Apostrophe menu button. This might be something we don't like:

image

Not sure if this is the best solution, but I think we should figure out a way to make this work that doesn't require doing area/widget specific overrides in the project level

### DIFF
--- a/lib/modules/apostrophe-areas/public/css/user.less
+++ b/lib/modules/apostrophe-areas/public/css/user.less
@@ -71,9 +71,9 @@
 .apos-area-widget-controls
 {
   position: absolute;
-  z-index: @apos-z-index-3;
+  z-index: @apos-z-index-5;
   .apos-transition;
-  &:hover { z-index: @apos-z-index-4; }
+  &:hover { z-index: @apos-z-index-6; }
 }
 .apos-area-widget-controls--context
 {


austinstarin:
I'm all for a solution that gets away from project level overrides or positioning of controls.

This might be a separate concern, but sometimes i think it would be valuable to remove trash/drag and drop controls as well for globally scoped widgets that are mission critical like the header or footer (i.e. widgets you would never want to delete and in some cases may influence the page tree). I see this as potentially a configuration option in `index.js`.

agilbert:
Thanks @austinstarin, I would agree that your other concern makes a lot of sense to consider as a separate issue.

stuartromanek:
I think that this change makes sense based on the other control hovering behavior but imaging project-level changes for things like marquees will always be a concern.

To @austinstarin's other point, a big ++ on that. I often want to remove drag controls from very specific widget configurations where it wouldn't make sense to drag something like a specialized headline or image out a card and into a general area widget.

PR #808, opened 2017-01-27
Make explicit that apostrophe-users config affects groups admin bar button

austinstarin:


boutell:
Looks like you opened bold (two asterisks) but didn't close it?

austinstarin:
Thanks for catching that Tom. See 6d66f95.

boutell:
One more tweak (:

Configuration `apostrophe-users` at project level? A-OK! Lots of reasons you might do that. Adding fields, etc.

Setting the `groups` option is the one and only no-no. It means "I want to provide a short list of simple groups so there's no groups UI."

PR #815, opened 2017-02-05
Pass root option to clean-css when minifying CSS

### LINKED ISSUE #809: Error when minifying non-LESS CSS files
It looks like Apostrophe has some trouble minifying non-LESS CSS files. My main CSS files lives in lib/modules/apostrophe-assets/public/css/site.css, and I reference it through my app.js in the standard way. After enabling the `minify` option and running the apostrophe:generation task, I get the following error:

```
[ 'Broken @import declaration of "/modules/my-apostrophe-assets/css/site.css"' ]
```

It took some debugging, but after figuring out that changing the file extension of site.css to .less made the minification work, I think I was able to track down the problem. What seems to happen is that the `self.compileStylesheet` callback in `self.minifyStylesheet` ([source](https://github.com/punkave/apostrophe/blob/master/lib/modules/apostrophe-assets/index.js#L819)) is receiving a code string where the `@import` directive for the CSS file has been left in place by `less` **but the path is now absolute**. So instead of `@import '../modules/my-apostrophe-assets/css/site.css';` we now have `@import '/modules/my-apostrophe-assets/css/site.css';`. The `clean-css` module doesn't know where to find that file, so it returns an error.

I think there are two possible solutions to this:

1. if I understand it correctly, Apostrophe will generate a main LESS file that contains one `@import` directive per stylesheet that has been pushed. If we force all of those `@import` directives to inline the referenced files, that should hopefully work. So `@import '../modules/my-apostrophe-assets/css/site.css';` would become `@import (less) '../modules/my-apostrophe-assets/css/site.css';`
2. Pass the option `root: 'public'` to the `clean-css` constructor

Perhaps the first option is more in line with the intended behavior anyway?

fredrikekelund:
This PR fixes #809. LESS will preserve @import statements for regular CSS files when compiling files, which means that when minifying, clean-css is the module responsible for the final concatenation of all CSS files. Before this change, the @import paths would be absolute, and clean-css wouldn't know where to start looking.

Looks like my editor removed a bit of trailing whitespace here as well, so the diff is a bit larger than the actual change.

### DIFF
--- a/lib/modules/apostrophe-assets/index.js
+++ b/lib/modules/apostrophe-assets/index.js
@@ -66,7 +66,7 @@ module.exports = {
       // at this point we should just finish
       return callback(null);
     });
-    
+
     self.extractBundleIfAppropriate();
     self.setAssetTypes();
     self.setTypeMap();
@@ -78,7 +78,7 @@ module.exports = {
     self.pushDefaults();
 
   },
-  
+
   beforeConstruct: function(self, options) {
     if (options.minify === undefined) {
       options.minify = options.apos.launder.boolean(process.env.APOS_MINIFY);
@@ -215,7 +215,7 @@ module.exports = {
         }
       };
     };
-    
+
     self.setTypeMap = function() {
       // Name of both folder and extension in
       // public/ for this type of asset
@@ -269,7 +269,7 @@ module.exports = {
 
       self.generation = generation;
     };
-    
+
     // Extract an asset bundle if appropriate. The default implementation
     // looks for an APOS_BUNDLE=XYZ environment variable and, if present, extracts
     // a bundle with the name XYZ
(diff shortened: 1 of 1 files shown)


fredrikekelund:
Ping @boutell @austinstarin, any chance we can get this merged? Without it, I can't create a minified stylesheet that references normal, static CSS files. Right now I'm converting SCSS > CSS > LESS, with a few hacks to get around the difference in how SCSS and LESS treat `calc()` statements

austinstarin:
Thanks for nudging this @fredrikekelund. I will take a closer look today. Stay tuned.

austinstarin:
Looks good @fredrikekelund. Thank you!

cc/ @boutell

fredrikekelund:
Thanks for merging! Happy to see that this only needed a fix this small in the end

PR #817, opened 2017-02-08
max width causing modal controls to float off page

grdunn:


austinstarin:
Thanks Greg! Looks good to me.

PR #818, opened 2017-02-08
limit for arrays, field defaults for arrays, refactoring

boutell:
Also:

* Refactored crazy redundant event handlers
* Documented methods
* Reduced redundant refreshes

### DIFF
--- a/lib/modules/apostrophe-schemas/public/css/components/array-editor.less
+++ b/lib/modules/apostrophe-schemas/public/css/components/array-editor.less
@@ -44,10 +44,8 @@
   .apos-array-limit-reached
   {
     display: none;
-    background-color: @apos-red;
-
     &:hover {
-      background-color: @apos-red;
+      background-color: @apos-base;
     }
   }
 }
--- a/lib/modules/apostrophe-schemas/public/css/components/array-editor.less
+++ b/lib/modules/apostrophe-schemas/public/css/components/array-editor.less
@@ -44,8 +44,10 @@
   .apos-array-limit-reached
   {
     display: none;
+    background-color: @apos-red;
+
     &:hover {
-      background-color: @apos-base;
+      background-color: @apos-red;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-schemas/public/js/array-modal.js
```diff
@@ -174,51 +254,86 @@ apos.define('apostrophe-array-editor-modal', {
...
+    
+    // Implements the `limit` option by showing and hiding
+    // the limit message and the add button, respectively.
+
+    self.limit = function() {
+      if (!self.field.limit) {
+        return;
+      }
+      if (self.arrayItems.length >= self.field.limit) {
+        self.$el.find('[data-apos-add-array-item]').hide();
```
austinstarin: Would prefer to manage visual state by toggling a selector here. Not the end of the world and probably breaks patterns already being used.

austinstarin:
Looks good to me Tom. Made a fix here 0164028 and an adjustment here 08b3469.

lib/modules/apostrophe-schemas/public/css/components/array-editor.less
```diff
@@ -28,16 +28,26 @@
...
+    left: 0;
+    width: 15%;
+    @media @apos-breakpoint-desktop-xl { width: 25%; }
+    text-align: center;
+    padding: @apos-padding-2*.75 0;
+  }
+  .apos-array-limit-reached
+  {
+    display: none;
+    background-color: @apos-red;
```
agilbert: Do we want this to feel like an error? I think a red error state might be too strong in this case. Thoughts?

boutell:
I agree, being at the max is not an error. It's a normal thing in your life. Calling some attention to it seems appropriate but red will make people think they Did It Wrong.

lib/modules/apostrophe-schemas/public/css/components/array-editor.less
austinstarin: See c0364053b271841cad82bf917107d5b9eebfe4ab.

PR #832, opened 2017-03-01
Fix test on adminBar subItem a link

ludovicbret:
Previously not closing the a markup


boutell:
Thanks! We'll be pushing a release by week's end which will include this.

PR #842, opened 2017-03-08
Performance improvement when viewing home page

boutell:


boutell:
Normally Apostrophe populates `req.data.home` from `req.page._ancestors[0]`. However there is fallback code to do this when `req.page._ancestors[0]` is not available, for instance on the home page itself, or a route which is calling `renderPage` but is separate from our standard page route (there is no `req.data.page`).

The fallback code applies the same options passed in for the use of the `ancestors()` filter, however that filter also calls `.areas(false).joins(false)` up front.

I factored that code out to a method which is now called from both the `ancestors()` filter and the fallback logic for always populating `req.data.home`, so they behave consistently.

This results in a significant performance win when looking at the home page: we don't wind up fetching areas and joins twice for what is often a very complex page. This eliminates many unnecessary queries.

boutell:
There is also a bug fix for a duplicate parameter name in a method of the schemas module. Babel caught it and refused to compile our code.

boutell:
Code reviewed with Austin.

Issue #843, opened 2017-03-11
We keep reinventing global preferences


boutell:
We keep reinventing global preferences, and hurting performance by loading them as a separate doc, not as part of the `global` doc.

These two things should get married, and `apostrophe-global` should extend `apostrophe-pieces` to do it with a minimum of code.

By default there are no global schema fields. If there are some, thanks to the project level, then the "global content" admin button should  launch the editor modal for that, which should also offer a versions button.

Whereas if there is no schema for it, the "global content" admin button should go straight to the versions modal, so you can at least roll back changes to shared headers and footers.

PR #844, opened 2017-03-11
Implemented #843, global preferences editing via the apostrophe-global module

boutell:
See #843

Implemented this, and also fixed a deep, hilarious bug: the pieces module has been clobbering the pushAsset call for user.js and certain others all this time. We missed it because people kept re-pushing these for their subclasses rather than kicking me hard enough. A simple application of the super pattern resolves this forever.

I think people probably just figured pushAsset didn't really cascade and push the same files in subclasses (it absolutely does, but if you override pushAsset and never call superPushAsset...)


austinstarin:
This is great!

PR #848, opened 2017-03-19
implemented batch uploads per @stuartromanek's sketch

### LINKED ISSUE #363: Attachments: Implement batch upload and annotation

boutell:
Will close #363.

`createViaUpload` is a new boolean flag for pieces subclasses. It is on by default for images and files, and when it is on, the "new/add" button triggers the browser's file upload dialog directly, plus you can drag and drop to the manager modal.

You can enable it for other pieces too but they must directly possess a schema field of type attachment — not to be confused with a widget that refers to images or files.

If you do a big fat test, like dragging in a big folder of images, it will upload relatively quickly but be prepared for a large delay at save time. That's autocropping. The spinner is working properly during this time, although it would be a nice enhancement later to say something specific.

I also fixed some other UX issues:

* Non-chooser managers had space set aside to display the current selection, which we've discussed won't be a thing because the selection will potentially be very large (too large, even, to display all at once) when working with bulk operations — it's not the same as choosing things for a slideshow.
* The logic for adjusting the heights of the elements in a manager modal assumed the footer was not a child of the body, but it has been for a long time, which is why there's been non-reachable space at the bottom when scrolling a tall page of images etc. I fixed the math both for the body and for the chooser.
* `minSize` constraints are shown to the user, and if you batch upload images that are too small you'll be told so. We could do it better, but this is a good iterative improvement.
* The image chooser had two search fields... still! I'm not sure if this was a regression or a special case where it was never fixed. The latter seems likely because it has a custom template.


stuartromanek:
Hey @boutell , looking and feeling very awesome, 🍾 

In testing around in apostrophe-boilerplate, noticed that regular files get a front-end error when uploading a **single** file, both via the button and drag-and-drop. Doesn't happen if you go the multi route or if you do it through Images.

Starts with
```
Uncaught TypeError: Cannot read property 'edit' of undefined at feedback (http://localhost:3000/modules/apostrophe-pieces/js/manager-modal.js:299:27)
```

boutell:
Good catch! Fixed that bug.

lib/modules/apostrophe-pieces/public/js/manager-modal.js
```diff
@@ -223,6 +224,93 @@ apos.define('apostrophe-pieces-manager-modal', {
...
+      if (!self.options.createViaUpload) {
+        return;
+      }
+      // Additional parameters to be passed to the route. Used to ensure
+      // we get things like `minSize` via the override the chooser puts in place
+      var query = {};
+      self.beforeList(query);
+      // The actual input element, hidden, which gets clicked by the button
+      // (see below).
+      console.log(getUploader().length);
```
stuartromanek: left a log in

boutell:
Quieted that log. Stuart did you review the source code or should I set something up with Alex?

stuartromanek:
I did review it, nothing jumped out at me but by all means get more eyeballs in there

lib/modules/apostrophe-pieces/lib/routes.js
```diff
@@ -19,6 +19,7 @@ module.exports = function(self, options) {
     self.route('post', 'create-modal', self.requireEditor, self.routes.createModal);
     self.route('post', 'trash', self.routes.trash);
     self.route('post', 'rescue', self.routes.rescue);
+    self.route('post', 'create-via-upload', self.apos.attachments.middleware.canUpload, self.apos.middleware.files, self.routes.createViaUpload);
```
boutell: insert-via-upload? Check what you called the other one.

lib/modules/apostrophe-pieces/public/js/manager-modal.js
```diff
@@ -223,6 +224,92 @@ apos.define('apostrophe-pieces-manager-modal', {
...
+      }
+      // Additional parameters to be passed to the route. Used to ensure
+      // we get things like `minSize` via the override the chooser puts in place
+      var query = {};
+      self.beforeList(query);
+      // The actual input element, hidden, which gets clicked by the button
+      // (see below).
+      getUploader().fileupload({
+        dataType: 'json',
+        dropZone: self.$el,
```
boutell: Should this be the outer modal container and not just the current slide?

lib/modules/apostrophe-pieces/views/chooserModalBase.html
```diff
@@ -30,7 +30,7 @@
 {%- block body -%}
   <div class="apos-chooser" data-chooser></div>
   <div class="apos-chooser-manage-view-wrapper">
-    <div class="apos-clearfix apos-manage-view-filters apos-modal-filters" data-filters></div>
+{#    <div class="apos-clearfix apos-manage-view-filters apos-modal-filters" data-filters></div> #}
```
boutell: Make this die.

lib/modules/apostrophe-pieces/public/js/manager-modal.js
boutell: Works as-is for dragging to the button etc.

PR #849, opened 2017-03-24
each widget now has one piece of admin chrome that combines Edit butt…

stuartromanek:
…ons with position controls and trash buttons, hopefully less overlaps, also some general style refinements to buttons and their hover states

image


boutell:
I like it, do you think this is something we can spring on existing users without a heads-up?

boutell:
I think we can just do this.

PR #865, opened 2017-04-13
Update api.js

fotisp:
If we don't parse options to the self.oembetter.fetch then these do not propagate to the filters


boutell:
Good call, thanks Fotis

PR #868, opened 2017-04-18
in dev, we leave relative URLs in LESS files strictly alone. Producti…

boutell:
…on should match the expectation this creates. The old behavior worked when assets stayed put but broke relative URLs in asset bundles uploaded to S3. 

(This does require that the assets being fetched by relative URLs be part of the bundle, but this is already an appropriate assumption because S3 is a separate origin server.)

### DIFF
--- a/lib/modules/apostrophe-assets/index.js
+++ b/lib/modules/apostrophe-assets/index.js
@@ -1045,12 +1045,8 @@ module.exports = {
       return less.render(fs.readFileSync(src, 'utf8'),
       {
         filename: src,
-        rootpath: path.dirname(stylesheet.web) + '/',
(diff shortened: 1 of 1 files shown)


boutell:
(The new code is based on a review of exactly what the less-middleware passes to successfully get this end result, and sure enough gets the same result for production minified and bundled assets.)

agilbert:
Any possible scenarios where this could lead to a BC break for somebody's Apostrophe 2.x site?

boutell:
In any such situation, the site would be breaking now in dev... and working in production! I think "it should work like it did in dev" is the reasonable expectation when we're forced into a choice like that.

PR #872, opened 2017-04-28
Add field type object

falkodev:
Add a new field type "object" to handle nested objects in mongo data.

It can be used as an "array" type, that is to say with a schema using other field types, even objects (so objects in objects):

```js
module.exports = {
  extend: 'apostrophe-pieces',
  name: 'module-test',
  label: 'Test',
  addFields: [
    {
        name: 'objectFirstLevel',
        label: 'Object First Level',
        type: 'object',
        schema: [
          {
            name: 'randomField1',
            label: 'Random Field 1',
            type: 'string',
            required: true
          }, {
            name: 'randomField2',
            label: 'Random Field 2',
            type: 'boolean'
          }, {
            name: 'objectSecondLevel',
            label: 'Object Second Level',
            type: 'object',
            schema: [{
              name: 'randomField3',
              label: 'Random Field 3',
              type: 'string',
              required: true
            }, {
              name: 'randomField4',
              label: 'Random Field 4',
              type: 'integer'
            }]
          }
        ]
      }]
    }
  ]
}
```

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -810,7 +810,7 @@ module.exports = {
         return partial(field);
       }
     });
-    
+
     // Given a schema and a cursor, add filter methods to the cursor
     // for each of the fields in the schema, based on their field type,
     // if supported by the field type. If a field name exists in `options.override`
@@ -1527,6 +1527,16 @@ module.exports = {
     });
 
     self.addFieldType({
+      name: 'object',
+      converters: {
+        form: function(req, data, name, object, field, callback) {
+          object[name] = data[name];
+          return callback(null);
+        }
+      }
+    });
+
+    self.addFieldType({
       name: 'joinByOne',
       converters: {
         string: function(req, data, name, object, field, callback) {
@@ -1589,7 +1599,7 @@ module.exports = {
           safeFor: 'manage',
           launder: joinFilterLaunder
         });
-        
+
         self.addJoinSlugFilter(field, cursor, '');
       }
     });
@@ -1709,7 +1719,7 @@ module.exports = {
         }
       }
     });
-    
+
     function joinFilterLaunder(v) {
       if (Array.isArray(v)) {
         return self.apos.launder.ids(v);
@@ -1731,7 +1741,7 @@ module.exports = {
       }
       return undefined;
     };
-    
+
     // You don't need to call this. It is called for you when you invoke `toChoices` for any
     // cursor filter based on a join. It delivers an array of choice objects to its callback.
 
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-schemas/index.js
```diff
@@ -1527,6 +1527,16 @@ module.exports = {
     });
 
     self.addFieldType({
+      name: 'object',
+      converters: {
+        form: function(req, data, name, object, field, callback) {
+          object[name] = data[name];
```
boutell: This needs to sanitize. See the implementation of the `array` field type which is very similar to what you need to do, minus the outer loop.

lib/modules/apostrophe-schemas/index.js
```diff
@@ -1589,7 +1599,7 @@ module.exports = {
           safeFor: 'manage',
           launder: joinFilterLaunder
         });
-        
+
```
boutell: Please submit without your whitespace-only changes in future as they make diffs confusing.

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -572,6 +572,46 @@ apos.define('apostrophe-schemas', {
...
+
+          if (subField.required === true) {
+            $subFieldSet.addClass('apos-required');
+          }
+
+          if (subField.type === 'object') {
+            // recursive call : no need for callback as the real callback will be called at the end of this loop
+            populate(data[name], subField.name, $subFieldSet, $fieldset, subField, function() { return; })
+          } else {
+            $input.val(data[name][subField.name]);
```
boutell: Please revisit this block in general: you should be using the populate method of schemas to do this work, so that all field types work automatically, and in fact you don't have to write your own _.each loop at all.

This will eliminate my concern about your not invoking callbacks and it will also work with all field types — arrays, etc. — that are not as simple as calling val().

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -572,6 +572,46 @@ apos.define('apostrophe-schemas', {
...
+      },
+      convert: function convert(data, name, $field, $el, field, callback) {
+        var $fieldset = self.findFieldset($el, name);
+        _.each(field.schema, function (subField) {
+          var $input = self.findSafeInFieldset($fieldset, subField.name, '[name=' + subField.name + ']');
+
+          if (subField.type === 'object') {
+            var $subFieldSet = self.findFieldset($fieldset, subField.name);
+            // recursive call : no need for callback as the real callback will be called at the end of this loop
+            convert(data[name], subField.name, $subFieldSet, $fieldset, subField, function() { return; })
```
boutell: This code needs to invoke apos.schemas.convert() with the schema of the object field, so all field types work, all callbacks are invoked, proper sanitization occurs, etc.

PR #873, opened 2017-05-01
* The `expressMiddleware` property of a module can already be a funct…

boutell:
…ion or array of functions, which are executed as middleware in the order the module is activated. In addition, it may now be an object. The object must have a `middleware` property, which is a function or array of functions as before. It may also have a `before` property, which must be the name of another module. This changes the execution order: the middleware for the current module will run prior to that of the module specified with `before`.
* Use `expressMiddleware` more consistently within Apostrophe's core modules.
* Provide `apostropheContextMenuWrapper`, `beforeApostropheContextMenu` and `afterApostropheContextMenu` nunjucks blocks to allow extension of the context menu in the outer layout.

### DIFF
--- a/lib/modules/apostrophe-templates/views/outerLayoutBase.html
+++ b/lib/modules/apostrophe-templates/views/outerLayoutBase.html
@@ -15,12 +15,20 @@
     {% endblock %}
 
     {% if data.user %}
-    <div class="apos-ui">
-      <div class="apos-context-menu-container">
-        {{ apos.pages.publishMenu({ publishMenu: data.publishMenu, page: data.page, piece: data.piece, bottom: true }) }}
-        {{ apos.pages.menu({ contextMenu: data.contextMenu, page: data.page, bottom: true })}}
-      </div>
-    </div>
+      {% block apostropheContextMenuWrapper %}
+        <div class="apos-ui">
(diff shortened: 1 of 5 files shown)


lib/modules/apostrophe-templates/views/outerLayoutBase.html
```diff
@@ -15,12 +15,20 @@
...
+              {% block beforeApostropheContextMenu %}
+              {% endblock %}
+              {{ apos.pages.publishMenu({ publishMenu: data.publishMenu, page: data.page, piece: data.piece, bottom: true }) }}
+              {{ apos.pages.menu({ contextMenu: data.contextMenu, page: data.page, bottom: true })}}
+              {% block afterApostropheContextMenu %}
+              {% endblock %}
+            </div>
+          {% endblock %}
+        </div>
+      {% endblock %}
```
agilbert: Is this change being made to support new workflow UI?

boutell: Yes. I think I did it right to make this space more flexible in general, but it is reasonable to ask if this is even where @stuartromanek will want markup to go. Also we could approach it by registering helpers to be called or something, so it doesn't require overriding outerLayout for every project.

PR #876, opened 2017-05-04
* Ability to disable contextual editing for a specific request with `…

boutell:
…req.disableEditing = true`, utilized by workflow module; but the data attributes necessary to gather information about the area are still output, which is also important for workflow.

* areas module exported to `apos.areas` browser side as well, making its options accessible to the workflow module.
* docs and pages modules don't call `ensureIndexes` until `modulesReady`, emit new events allowing other modules to alter the index properties (workflow)
* `self.apos.pages.addAfterContextMenu(helper)` accepts a helper function to be called by `outerLayout` immediately after the context menu is emitted; this allows extension of the UI by new modules without forcing developers to manually alter `outerLayoutBase`. The helper function receives `req` for convenience, it is always `apos.templates.contextReq` since we are already rendering the page at this point
* `self.apos.templates.addBodyClass(req, bodyClass)` adds a body CSS class to be output on this request. Multiple classes may be added as a single string separated by spaces. They are combined with other calls to this method and the `bodyClass` nunjucks block in `outerLayoutBase`. This allows extension of the UI by new modules without forcing developers to manually alter `outerLayoutBase`.
* Reverted some unnecessary new nunjucks blocks from previous workflow accommodations commit now that we have a cleaner way.
* Fixed some stray references to "req.page" in documentation (should be `req.data.page`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -139,7 +139,7 @@ Thanks to Michelin for their support of this work.
 
 All tests passing.
 
-* Performance enhancement: when fetching `req.data.home` directly in the absence of `req.page._ancestors[0]`, such as on the home page itself or a non-page route like `/login`, we must apply the same default filters before applying the filter options, namely `.areas(false).joins(false)`, otherwise duplicate queries are made.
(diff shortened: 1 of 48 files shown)


boutell:
There's a lot more in this branch now. Someone will have to sit down with me to review this.

PR #882, opened 2017-05-18
Feature proposal : "readonly" property for schema fields

falkodev:
Here is my proposal to have readOnly fields in schemas in Apostrophe.

I needed to import data in Mongo and then display them in the admin interface without being able to edit them (even the admin user). So, I wanted to specify some fields to be readonly.

It can be used for every field type. 
- area and singleton :
```
    {
      name: 'test',
      label: 'Test Area',
      type: 'area',
      options: {
        widgets: {
          'apostrophe-rich-text': {
            controls: [ 'Bold', 'Italic', 'Link', 'Unlink' ],
            readOnly: true
          },
          'apostrophe-images': {
            readOnly: true
          }
        }
      }
    }, {
      name: 'test2',
      label: 'Test Singleton',
      type: 'singleton',
      widgetType: 'apostrophe-rich-text',
      options: {
        controls: [ 'Bold', 'Italic', 'Link', 'Unlink' ],
        readOnly: true
      }
    }
```

- choices-like : 
```
    {
      name: 'test2',
      label: 'Test',
      type: 'tags',
      readOnly: true
    }, {
      name: 'test3',
      label: 'Test',
      type: 'boolean',
      readOnly: true
    },{
      type: 'checkboxes',
      name: 'test4',
      label: 'Test',
      choices: [
        {
          label: 'Big',
          value: 'big'
        },
        {
          label: 'Friendly',
          value: 'friendly'
        },
        {
          label: 'Furry',
          value: 'furry'
        }
      ],
      readOnly: true
    }
```

- string-like : 
```
    {
      name: 'test5',
      label: 'Test',
      type: 'integer',
      readOnly: true
    }, {
      name: 'test6',
      label: 'Test',
      type: 'float',
      readOnly: true
    }, {
      name: 'test7',
      label: 'Test',
      type: 'url',
      readOnly: true
    }, {
      name: 'test8',
      label: 'Test',
      type: 'date',
      readOnly: true
    }, {
      name: 'test9',
      label: 'Test',
      type: 'time',
      readOnly: true
    }, {
      name: 'test10',
      label: 'Test',
      type: 'password',
      readOnly: true
    }
```

- arrays and objects : the array or object will be readonly, and items in it will also appear disabled. Therefore, no need to add "readOnly" to child items and even a child object in a readOnly parent object will be disabled (and its sub-items too) : 
```
    {
      name: 'test11',
      label: 'Test',
      type: 'array',
      schema: [
        {
          type: 'string',
          name: 'item1',
          label: 'Item 1'
        },
        {
          type: 'integer',
          name: 'item2',
          label: 'Item 2'
        },
      ],
      readOnly: true
    },
    {
      name: 'test11bis',
      label: 'Test',
      type: 'object',
      schema: [
        {
          type: 'string',
          name: 'item1',
          label: 'Item 1'
        },
        {
          type: 'integer',
          name: 'item2',
          label: 'Item 2'
        },
      ],
      readOnly: true
    }
```

- attachments : 
```
    {
      name: 'test12',
      label: 'Test',
      type: 'attachment',
      readOnly: true
    }, {
      name: 'test13',
      label: 'Test',
      type: 'video',
      readOnly: true
    }
```

- joins : 
```
   {
      name: '_test14',
      label: 'Test',
      withType: 'dgad-article',
      type: 'joinByArray',
      idsField: 'test14Id',
      readOnly: true
    }
```


boutell:
This is well done so far, thanks for contributing it. However right now, the convert() method of schemas on the server side still reads all of the data, which is a security risk. Fields with the `readonly` property should be skipped by convert() to complete the picture.

Skipping that in the browser-side convert() method is also a good idea, but it's critical on the server side.

Also, one slightly annoying request: we would prefer `readOnly` rater than `readonly`.

Thanks!

falkodev:
Hi Tom,

Thanks for the advice.

I changed "readonly" to "readOnly" as asked, and added a condition to skip readOnly properties in apostrophe-schemas/index.js line 451, and apostrophe-schemas/public/js/user.js line 128.

I guess this is correct, at least on the manual tests I did. Let me know if I missed something.

boutell:
Thanks for that change.

I am trying to understand this part:

"arrays and objects : the array or object will be readonly, so items in it can't be changed but each item can be also be readonly (only for visible effect of being disabled because items in it won't be updated anyway)"

I think you are saying that unless you explicitly make an individual field in an array or object schema readOnly, it's going to appear to be editable, even though it isn't really...?

If so, I think that should be changed. There's no reason for a field to appear editable when it is not.

Very close!

falkodev:
Yes, you are right. There was no reason for a field in readOnly array or object to appear enabled. I changed that.

lib/modules/apostrophe-schemas/index.js
```diff
@@ -802,7 +802,8 @@ module.exports = {
       toGroups: function(fields) {
         return self.toGroups(fields);
       },
-      field: function(field) {
+      field: function(field, readOnly = false) {
```
boutell: Default values are an ES6-ism, please remove. Apostrophe 2.x core is compatible back to ES5.

lib/modules/apostrophe-schemas/index.js
```diff
@@ -802,7 +802,8 @@ module.exports = {
...
+        if (readOnly) field.readOnly = true;
```
boutell: Coding style: please always use { ... } for if statements.

falkodev:
Corrected ES6 code and coding style as requested

PR #894, opened 2017-06-01
notifications!

agilbert:
Built out a reasonable starting place for the front-end for these. Note: @boutell, we still need to add support for passing parameters into the message that don't need to be considered for the localization in the template.

```
apos.notify('Your changes have been saved.');
apos.notify('Your changes have been saved.', { type: 'success' });
apos.notify('Your changes have been saved. But something is weird.', { type: 'warn' });
apos.notify('There was an error!', { type: 'error' });
apos.notify('There was an error! This message will self destruct.', { type: 'error', dismiss: true });
apos.notify('There was an error! This message will self destruct in 1 second.', { type: 'error', dismiss: 1 });
```

screencast

Not shown here is that you can simply click on a notification to dismiss it.

And yes, I am on fire. 🔥 🔥 🔥

### DIFF
--- a/defaults.js
+++ b/defaults.js
@@ -43,6 +43,7 @@ module.exports = {
     'apostrophe-images-widgets': {},
     'apostrophe-files': {},
     'apostrophe-files-widgets': {},
-    'apostrophe-service-bridge': {}
+    'apostrophe-service-bridge': {},
+    'apostrophe-notifications': {}
   }
 };
--- a/defaults.js
+++ b/defaults.js
@@ -15,6 +15,7 @@ module.exports = {
     'apostrophe-permissions': {},
     'apostrophe-assets': {},
     'apostrophe-admin-bar': {},
+    'apostrophe-notifications': {},
     'apostrophe-login': {},
     'apostrophe-browser-utils': {},
     'apostrophe-ui': {},
@@ -43,7 +44,6 @@ module.exports = {
     'apostrophe-images-widgets': {},
     'apostrophe-files': {},
     'apostrophe-files-widgets': {},
-    'apostrophe-service-bridge': {},
-    'apostrophe-notifications': {}
+    'apostrophe-service-bridge': {}
   }
 };
--- a/lib/modules/apostrophe-notifications/index.js
+++ b/lib/modules/apostrophe-notifications/index.js
@@ -13,7 +13,7 @@ module.exports = {
      * Render a notification partial with the passed message.
      */
     self.route('post', 'notification', function (req, res) {
-      res.send(self.render(req, 'notification', { message: req.body.message }));
+      res.send(self.render(req, 'notification', { message: self.apos.launder.string(req.body.message) }));
     });
   },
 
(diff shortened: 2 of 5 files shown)


defaults.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
     'apostrophe-images-widgets': {},
     'apostrophe-files': {},
     'apostrophe-files-widgets': {},
-    'apostrophe-service-bridge': {}
+    'apostrophe-service-bridge': {},
+    'apostrophe-notifications': {}
```
boutell: apostrophe-service-bridge is intended to be last. It lets all the core modules know that all the core modules are ready.

lib/modules/apostrophe-notifications/index.js
```diff
@@ -0,0 +1,24 @@
...
+    self.pushAssets = function() {
+      self.pushAsset('script', 'user', { when: 'user' });
+      self.pushAsset('stylesheet', 'user', { when: 'user' });
+    };
+
+    /**
+     * Render a notification partial with the passed message.
+     */
+    self.route('post', 'notification', function (req, res) {
+      res.send(self.render(req, 'notification', { message: req.body.message }));
```
boutell: Run `message` through `self.apos.launder.string`.

lib/modules/apostrophe-notifications/public/js/user.js
```diff
@@ -0,0 +1,63 @@
...
+      setTimeout(function() {
+        $notification.removeClass('apos-notification--hidden');
+      }, 100);
+    };
+    
+    self.dismiss = function($notification) {
+      $notification.addClass('apos-notification--hidden');
+      
+      setTimeout(function() {
+        $notification.css('display', 'none');
```
boutell: I think it should be removed, otherwise they will pile up in the DOM, consuming resources.

agilbert:
@boutell ready for review / merge

lib/modules/apostrophe-notifications/views/notification.html
```diff
@@ -0,0 +1,3 @@
+<div class="apos-notification apos-notification--hidden" data-apos-notification>
+  <p class="apos-notification-message" data-apos-notification-message>{{ data.message }}</p>
```
agilbert: @boutell doesn't this template need to include the strings in here?

boutell: In index.js I already interpolated them, and I pass the prepared string as data.message. I do it this way because I need to use .apply to pass a prepared array of arguments.

lib/modules/apostrophe-notifications/index.js
```diff
@@ -0,0 +1,28 @@
...
+      self.pushAsset('stylesheet', 'user', { when: 'user' });
+    };
+
+    /**
+     * Render a notification partial with the passed message.
+     */
+    self.route('post', 'notification', function (req, res) {
+      var message = self.apos.launder.string(req.body.message);
+      var strings = self.apos.launder.strings(req.body.strings);
+      var args = [ message ].concat(strings);
```
boutell: Building the arguments for the __ function...

lib/modules/apostrophe-notifications/index.js
```diff
@@ -0,0 +1,28 @@
...
+      message = self.apos.i18n.__.apply(self.apos.i18n, args);
```
boutell: ... Invoking __ with our arguments array, via JavaScript's `apply` keyword

PR #906, opened 2017-06-09
complete implementation of explicitOrder, replacing implementation th…

boutell:
…at was never actually reached and incomplete


jevillard:
Hi,
I have been test and it's good for us. Thanks !
Wainting merge ;)

PR #908, opened 2017-06-12
Do not crash on missing widget types, print good warning messages

boutell:


lib/modules/apostrophe-areas/lib/api.js
```diff
@@ -24,7 +24,18 @@ module.exports = function(self, options) {
 
   self.getWidgetManager = function(name) {
     return self.widgetManagers[name];
-  }
+  };
+
+  // Print warning message about a bogus widget type — only once per run per type.
+  self.warnBogusWidgetType = function(name) {
```
agilbert: Maybe another word other than `Bogus` would be less subjective here?

agilbert: `Missing`? `Invalid`?

PR #910, opened 2017-06-14
Add fromHtml to areas api

austinstarin:


boutell:
I'd prefer you rename it "fromRichText," the reason being that "fromHtml" suggests it'll be able to hack img tags and youtube and such, which it can't do (we'll factor that out of one of our import tools sometime I'm sure).

boutell:
(When we do get that feature it won't be with this API, it'll be a heavier, async method.)

austinstarin:
@boutell See b521a47

PR #923, opened 2017-06-28
Prepend protocol to dev server URL

alexgleason:
Rationale for this:

1. It will always be http for a local dev server.
2. In some terminal programs, you can click a valid URL to have it open in the browser. Prepending the protocol allows this.


boutell:
Thanks, makes sense.

boutell:
(Even on a prod server, it is technically http, because apostrophe does not implement https directly — nginx is much better at answering the front door so we always use it as a reverse proxy. Some folks use Apache but the principle is the same.)

PR #932, opened 2017-07-14
resolves #900. pieces are now contextual by default.

### LINKED ISSUE #900: Functionality for contextual true for pieces should be default behavior

colpanik:
Closes #900.

I have switched the logic up ever so slightly so that contextual is now true by default. I have tested this by creating a new piece type and checking to see how the option affects its behavior when creating a new piece when a page module is/isn't present and whether or not an index page exists.

The `contextual` flag is already smart enough to know NOT to try to go to a show page if a `-pages` module does not exist for the piece type in question or if an index page does not yet exist in the database.

Two iterations ago, we discussed the potential negative implications of default `contextual: true` and it was noted that many developers create piece types that are intended to appear in a list view on an index page, but NEVER have a show page (ex. news posts that just link out to external articles).

It was proposed that we introduce a `showPage` flag to `pieces-pages` that would address this use case. My thinking is that `showPage: false` will automatically set its corresponding piece type to `contextual: false` and override a `contextual: true` flag if explicitly set.

I created a separate issue for the `showPage` flag and will close that separately: #931

✌️

### DIFF
--- a/lib/modules/apostrophe-pieces/index.js
+++ b/lib/modules/apostrophe-pieces/index.js
@@ -26,7 +26,7 @@ module.exports = {
   },
 
   beforeConstruct: function(self, options) {
-    self.contextual = options.contextual;
+    self.contextual = true ? (options.contextual !== false) : false;
 
     if (self.contextual) {
       // If the piece is edited contextually, default the published state to false
@@ -115,7 +115,7 @@ module.exports = {
         style: 'pill'
       }
     ].concat(options.addFilters || []);
-    
+
     options.batchOperations = [
       {
         name: 'trash',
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-pieces/index.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
   },
 
   beforeConstruct: function(self, options) {
-    self.contextual = options.contextual;
+    self.contextual = true ? (options.contextual !== false) : false;
```
boutell: this doesn't make sense, the else clause can never happen. The thing you are testing is the constant: `true`

stuartromanek:
We agreed at the last iteration meeting to scrap this change and make `contextual:true` a part of our normal, recommended workflow. Correct? @colpanik @boutell

boutell:
Correct!

PR #933, opened 2017-07-14
widgets can have template options "moveable" and "trashable" set to f…

stuartromanek:
…alse to remove apostrophe move/trash UI

```
    {{ apos.singleton(data.page, 'imageSingleton', 'apostrophe-images', {
      trashable: false,
      moveable: false
    }) }}
```


stuartromanek:
@boutell you had concerns about this because of something Michelin is requesting, yes?

All I'm doing is passing options through singletons that skip rendering the Move and Trash controls

stuartromanek:
Wait on Tom's workflow button config dealy

stuartromanek:
closing this as @boutell implements a more elegant, configurable solution

PR #934, opened 2017-07-14
Bring Lansdowne image bug fix to Apos images core

bobclewell:
The below code causes an exception that blows up the site if `piece.attachment` doesn't exist.

``

The logic in this PR prevents that exception.

See Zendesk issue 650.


bobclewell:
@boutell once you get settled back in after your time off, can we take a couple minutes to pair up on this, and perhaps figure out a root cause.

PR #935, opened 2017-07-17
Add an insert method to apostrophe-attachments to match pieces behaviour

HollyPony:
Allow backend modules to insert attachments. Basically it's an override of the `accept` method with a `permissions: false` option.

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -27,6 +27,11 @@ module.exports = function(self, options) {
   // permissions for backend tasks purposes.
 
   self.insert = function(req, file, options, callback) {
+    if (arguments.length === 3) {
+      callback = arguments[2];
+      options = {};
+    }
+
     var extension = path.extname(file.name);
     if (extension && extension.length) {
       extension = extension.substr(1);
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -20,6 +20,13 @@ module.exports = function(self, options) {
   // configure jquery fileupload to submit one per request.)
 
   self.accept = function(req, file, callback) {
+    self.insert(req, file, {permissions: true}, callback);
```
boutell: Please use spaces around `permissions: true`

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -20,6 +20,13 @@ module.exports = function(self, options) {
...
+  };
+
+  // Extending the functionnality of `accept`. Provide a way to bypass
+  // permissions for backend tasks purposes.
+
+  self.insert = function(req, file, options, callback) {
     var extension = path.extname(file.name);
```
boutell: Since your goal here is to be parallel with pieces.insert, please support calls with only three arguments, implying an empty options object. See the pieces.insert implementation for the preferred way to structure this.

Once you do that, accept can just be a bc wrapper that invokes this method with three arguments.

HollyPony:
@boutell Mission complete

PR #950, opened 2017-07-26
Add a callOne method beside the callAll

HollyPony:
This allow us to keep a unique function used by callAll but for specific purposes.


HollyPony:
cc @ludovicbret

PR #952, opened 2017-07-27
pass a class in a widgets options to be applied to the outer wrapper …

stuartromanek:
…of the singleton

### DIFF
--- a/lib/modules/apostrophe-areas/views/widget.html
+++ b/lib/modules/apostrophe-areas/views/widget.html
@@ -2,7 +2,7 @@
 {# however ginormous joined objects are elided. Use APIs to #}
 {# get them if you need them (TODO: consider ways of #}
 {# whitelisting them for the inline JSON attributes) #}
-<div class="apos-area-widget-wrapper" data-apos-widget-wrapper="{{ data.widget.type }}">
+<div class="apos-area-widget-wrapper{% if data.options.class %} {{ data.options.class }}{% endif %}" data-apos-widget-wrapper="{{ data.widget.type }}">
   {# This wrapper exists for editor.js to inject contextual widget insertion controls into,
     since those are area level controls rather than widget level #}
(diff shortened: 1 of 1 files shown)


stuartromanek:
hey @abea what are your thoughts on this?

abea:
The implementation looks great. I think I was unclear by using "wrapper element". I meant to refer to the area or singleton wrapper (the `data-apos-area` element, I think in area.html) since it's the top-most wrapper that we don't have direct template access to.

The purpose would be to have clearer access to lay out the areas without excess wrapper divs around each one.

stuartromanek:
Wouldn't you just do that in your template, wrapping the area? Is there a case where it would _not_ be specific to a template?

stuartromanek:
it makes sense for widgets in an area because then you could selectively wrap different parts of your areas with classes.

abea:
Applying a class to each widget within and area does seem useful, though thus far layout doesn't seem to be an issue within an area as much as for the area within the region. I also don't think I know how this would be applied--whether uniquely to different widgets or in the area's option object and the same for all widgets.

My original goal for the issue was to reduce excess markup in templates. The same layout wins can be gained by wrapping the area in a `div` with a layout class for sure. My thought was that doing so is unnecessary if we can apply that layout class to the area itself.

austinstarin:
I think this opens up an issue where you're writing project level code to affect Apostrophe's UI, which strongly couples the project level code with the CMS itself.

stuartromanek:
I don't necessarily see it as affecting apostrophe UI, just leveraging markup that will be there anyway.

That said, now that I understand the original ask, I would say passing an area a class to be dealt with by apostrophe rather than just wrapping your own markup would make the code less readable, less straight forward, and I'm less of a fan of the idea.

I COULD see passing widgets in an area special options because it puts the specialness on the context, not the vanilla widget.  In the past this has been done by passing the widgets options that the developer deals with in the template, at the project level.

I would still do it the old way, we could trash this branch

abea:
Sounds good. Fine if we want to close the issue as well.

Issue #73, opened 2014-01-13
What's the recommended way to deal with tables within Apostrophe?


etodanik:
What's the best way to implement user editable tables?

PR #955, opened 2017-07-28
Moving the apostrophe buttons and menu above the WYSWYG editor.

bobclewell:
Specifically fixes a Michelin bug #73, found in their workflow testing.

### DIFF
--- a/lib/modules/apostrophe-ui/public/css/components/admin-bar.less
+++ b/lib/modules/apostrophe-ui/public/css/components/admin-bar.less
@@ -4,7 +4,7 @@
   height: 60px;
   top: @apos-padding-2;
   left: @apos-padding-2;
-  z-index: @apos-z-index-5;
+  z-index: @apos-z-index-7;
   background-color: @apos-base;
   color: @apos-white;
   .apos-button-border(@apos-white);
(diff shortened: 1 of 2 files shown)


bobclewell:
@stuartromanek this should not have the odd behavior that you found with the pages menu and the modals.

boutell:
So what happens if the dropdown style menu of the rich text editor is open and it overlaps  the workflow controls? In that situation the editor should win...

bobclewell:
I think this looks right. Right? 
apostrophe_sandbox__home

bobclewell:
This too...
apostrophe_sandbox__home

boutell:
Any weird interactions when a modal is opened on top of the rich text editing controls?

How about notifications? Try apos.notify('hello') a few times in the console.

bobclewell:
Yep. @boutell notifications look OK. 
apostrophe_sandbox__home

bobclewell:
@boutell would you rather this is against the master branch? Even though it was found in the context of the workflow module, there's no code that depends on it. I can change the PR if you like.

boutell:
Accepted — my Big Damn Project this afternoon is to merge this branch anyway so after that it won't matter, hooray

PR #960, opened 2017-07-28
907 next previous

boutell:


lib/modules/apostrophe-docs/lib/cursor.js
```diff
@@ -1250,5 +1186,171 @@ module.exports = {
...
+      // Finalize the sort if it hasn't been already,
+      // we need to know what it is so we can build
+      // $gte/$lte queries around it
+      if (!self.get('sortMongo')) {
+        self.finalizeSort();
+      }
+      var sort = self.get('sortMongo');
+      var direction = (verb === 'next') ? 1 : -1;
+      if (!sort) {
+        return callback(null);
```
boutell: Bug: should just return, there is no callback

lib/modules/apostrophe-docs/lib/cursor.js
```diff
@@ -1250,5 +1186,171 @@ module.exports = {
...
+      }
+      var clauses = [];
+      var criteria = {
+        _id: {
+          $ne: doc._id
```
boutell: This is wrong. Instead, push one more clause at the end, in which everything is equal, and the id must be less (not equal).

PR #962, opened 2017-07-31
Changed page tree trashcan from utf8 icon to a simple text label

bobclewell:
I didn't make it a font awesome icon, since it was passed as a label param, and I didn't see any other precedent in the code for passing HTML as a label in the code. Plus, it was a quick fix, and we already use the text string "Home" to label the root of the tree.


boutell:
That dog will hunt! Hadn't realized it was this way already until workflow.

Issue #919, opened 2017-06-16
Be able to position a widgets controls with a flag


stuartromanek:

austinstarin:
This has been merged.

boutell:
This exists now, but requires documentation before we close it.

mtthwmnc:
@boutell PR for docs: https://github.com/punkave/apostrophe-documentation/pull/70

PR #969, opened 2017-08-09
TKT-919 / Position widget controls via options object in widget

mtthwmnc:
https://github.com/punkave/apostrophe/issues/919

Allows developers to position widget controls by passing in a 'position' property into a widget. The following values can be passed in:

- bottomLeft
- bottom RIght
- topRIght
- topLeft (default position, but developer may use it for granularity)

`{{ apos.singleton(data.global, 'footer', 'my-footer', { position: "bottomLeft" }) }}`

### DIFF
--- a/lib/modules/apostrophe-areas/public/css/user.less
+++ b/lib/modules/apostrophe-areas/public/css/user.less
@@ -75,7 +75,7 @@
   .apos-transition;
   &:hover { z-index: @apos-z-index-6; }
 }
-.apos-area-widget-controls--context, .apos-area-widget-controls--topLeft,
+.apos-area-widget-controls--context, .apos-area-widget-controls--top-left,
 {
   top: @apos-padding-1;
   left: @apos-padding-1;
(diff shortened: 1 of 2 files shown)


abea:
On a purely code format point, the CSS formatting doesn't match the current stylelint config, though it looks to match the existing code. It might be an open question whether to incorporate those new formatting styles to Apostrophe.

mtthwmnc:
@abea I was talking about this with @austinstarin – formatting in piecemeal as we we see it vs. maintaining the format of a project that doesn't use our current formatting practices is something we should formally discuss with the whole dev team

austinstarin:
My vote would be to follow best practices and formatting within Apostrophe currently.

Maybe we can use our new house styles in the next version.

abea:
That makes sense. Maybe we should create an issue to add a stylelint and eslint config to the project? I could create an issue and plan to do that.

austinstarin:
@tom @stuartromanek This looks good to me?

stuartromanek:
camelcase selectors? do we do that elsewhere? otherwise fine

austinstarin:
Good point. This felt right though as a property you would pass in `{ position: topRight }`, for example.

stuartromanek:
pass in JS' y camelcase config and convert it to hyphenated classes in the template with `apos.utils.cssName(string)`

austinstarin:
👍

boutell:
I think what you're proposing with your latest commit is that the values passed by the developer would be `bottom-left`, `top-right`, etc. to start with, rather than camelCased.

I'm fine with that if that is the intent.

mtthwmnc:
I jumped the gun here - we're keeping the JS param camelCase and using a the nunjucks pipe filter {{ property | css }}

stuartromanek:
Going to merge this for now

PR #972, opened 2017-08-10
Updates to move notifications within modals to the lower right corner…

bobclewell:
… of the modal so that it doesn't block the buttons


stuartromanek:
@bobclewell quick path to see this in action?

boutell:
Oops, screenshot: https://www.dropbox.com/s/u5or2i8l5dw2j5e/87v-kcvx.png?raw=1

bobclewell:
Fix the issue with the overlapping notifications and resubmit. 

To reproduce use `apos.notify('foo')` in the Console.

bobclewell:
@boutell The styles have been updated and I confirm that the notification now stack correctly.

PR #977, opened 2017-08-18
Removing slash from @fa-font-path variable

ValentinJeudy:
There is a conflict between this variable **@fa-font-path** which finish with a slash, and urls inside **paths.less** which are beginning with a slash.
Exemple -> 
`@fa-font-path: '/modules/apostrophe-ui/fonts/';`
`url('@{fa-font-path}/fontawesome-webfont.eot?v=@{fa-version}');`


bobclewell:
@boutell this makes sense to me. See `apostrophe-ui/public/css/vendor/font-awesome/path.less`. If you agree, let me know and I'll merge.

PR #978, opened 2017-08-18
Check if an attachment is still existing to avoid template errors

dirago:
Hello punkave,

This change is related to a problem we are currently working on (DGAD), to check if a joigned object (regardless of its type) in a template is still existing to avoid rendering errors, this one is just for images and it works well.

What do you think?

Have a nice day :)


boutell:
I hear what you're saying, and while I've resisted making this change because we feel developers should be planning for what to do if the attachment is missing, it would make sense from a production resiliency standpoint to just gracefully display nothing.

I do not think returning `/` is a good idea though. The default URL should be a valid file from the static assets of the apostrophe-attachments module. Perhaps it would be better if a configuration option to the apostrophe-attachments module determined whether you got a 1x1 transparent GIF URL, or an animated GIF calling your attention to the problem. The default would be the animated GIF, while in production you could choose the invisible GIF.

Either way a message should be written to `console.error` to call attention to the problem.

From a code style perspective, if we were to use an "else" clause there we would drop to the next line right after the opening bracket. However, we would not need an else clause, because we would also make a check like this that eliminates most of the code right up at the top to avoid extra nesting, like:

```
if (!attachment) {
  // Eliminate this strange case right up front
  return self.getMissingAttachmentUrl(req);
}
// Normal case follows here without the need for nesting
```

PR #985, opened 2017-08-25
fixed aposPrefix to apos.prefix

thecb4:
Pull request to fix apostrophe-login/loginBase.html to use apos.prefix instead of aposPrefix.


boutell:
Thank you!

Issue #916, opened 2017-06-16
Create official apostrophe confirm window


stuartromanek:
@stuart find the designs for this

boutell:
`apos.notifications.trigger` is aliased as `apos.notify` to encourage adoption. So this should be aliased as `apos.confirm`.

I think the API should look like:

```
return apos.confirm('Are you sure you want to delete %s?', doc.title, function() { 
  // ... Do the thing ...
}, function() {
  // ... OPTIONAL function if they click cancel, usually omitted ...
});
```

This needs to be implemented with careful attention to the existing pattern for the `apostrophe-notifications` module. Like notify, confirm gets rendered on the server. Note the support for %s placeholders, which permits internationalization to happen on the server side when rendering the confirm modal. This won't be hard to achieve because we already did it for `apos.notify`, but it would be reasonable to implement just a string argument at first and ask me to port the %s stuff from notify.

This is not quite the usual function(err) { ... } callback interface because:

1. Errors are unexpected things. Clicking cancel is not an unexpected thing.

2. There really aren't any other possible interesting outcomes beyond "yes" and "no" (there are no errors, apart from a network error rendering the confirm modal, which is unlikely unless your whole Apostrophe experience is already pretty much failville).

3. It facilitates the easy refactoring of existing `confirm()` code, which looks like this:

```
if (confirm('Really?')) {
  // DO THE THING
}
```

There is almost never an "else" clause here.

That code just becomes:

```
// I LIKE THIS
return apos.confirm('Really?', function() {
  // DO THE THING
});
```

I think the convenience of this is a win over something like "clicking 'cancel' is an error":

```
// I DON'T LIKE THIS

return apos.confirm('Really?', function(err) {
  if (err) {
    return;
  }
  // DO THE THING
});
```

Should this code live in the `apostrophe-notifications` module? I am leaning toward yes. They will hopefully share some implementation details, and in the end both are aliased onto `apos` anyway.

stuartromanek:
Attaching some designs. Not sure the green/red button pattern is always applicable
![apos-confirm-alt](https://user-images.githubusercontent.com/1889830/27746817-aec9d110-5d96-11e7-8540-6ff26639beae.png)
![apos-confirm](https://user-images.githubusercontent.com/1889830/27746816-aec7a2b4-5d96-11e7-8267-5e56d37f00f7.png)

boutell:
[...]

PR #987, opened 2017-08-27
840 conflict resolution

boutell:
Conflict resolution has been implemented:

* If you start editing what someone else is editing, you get stopped with a prompt to either take control, or leave it alone
* If control is taken from you, you get an alert and your page reloads
* If the person who has control is you, but in another tab/window/whatever, this is called to your attention, although the same rules still apply for the same reasons (no loss of work, no weird competing save callback scenarios)
* All of this applies to both pieces and pages
* Apostrophe doesn't start trying to autosave an area until you engage with it for the first time, which eliminates major causes of excess page versions and conflicts that shouldn't even be conflicts

Right now there are a fair number of confirm() and alert() calls. We really need #916 for that to be nice. But our clients are in serious need of this feature right away so there's no reason to hold it...

Except for good testing & code review by someone who isn't me! This is significant.


boutell:
Must reproduce and fix the bug where after taking control of a piece in the editor and saving it (?), I am unable to close the manage modal. And/or verify it is not our bug.

lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -514,26 +527,62 @@ apos.define('apostrophe-areas-editor', {
...
-          dotPath: self.$el.attr('data-dot-path'),
-          options: apos.areas.getAreaOptions(self.$el),
-          items: items
-        },
-        function() {
-          self.previousData = items;
+        request,
+        function(result) {
+          if (result.status !== 'ok') {
+            if (result.status === 'locked') {
```
boutell: What about locked-by-me?

lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -612,5 +661,62 @@ apos.define('apostrophe-areas-editor', {
...
+        return succeed();
+      });
+      function succeed() {
+        self.registerAutosave();
+        return callback(null);
+      }
+    };
+    
+    // Similar to `startAutosaving`, this method
+    // obtains a session lock and starts autosaving of
```
boutell: session lock -> context lock

lib/modules/apostrophe-areas/public/js/user.js
```diff
@@ -230,12 +231,67 @@ apos.define('apostrophe-areas', {
...
+          requests.push(request);
+        }
+      });
+      return $.jsonCall(self.options.action + '/save-areas-and-unlock',
+        {
+          async: false
+        },
+        {
+          areas: requests,
+          htmlPageId: apos.htmlPageId
```
boutell: dead code - this is implicit now

lib/modules/apostrophe-areas/public/js/user.js
```diff
@@ -230,12 +231,67 @@ apos.define('apostrophe-areas', {
...
+    //
+    // Both the sync flag and the callback may be
+    // omitted entirely. The default is asynchronous.
+    //
+    // This method is ideal in situations where
+    // you wish to be sure everything has been saved
+    // before transitioning to a UI such as a commit dialog box
+    // that displays what has changed, etc.
+
+    self.saveAllIfNeeded = function(sync, callback) {
```
boutell: Do I still need the option of a sync flag here?

lib/modules/apostrophe-assets/public/js/always.js
```diff
@@ -172,6 +172,14 @@ _.extend(apos, {
       });
     }
   },
+  
+  enableHtmlPageId: function() {
```
boutell: Move this to user.js, it should not be forced on browserland for logged out users

lib/modules/apostrophe-assets/index.js
```diff
@@ -1397,9 +1404,18 @@ module.exports = {
...
-          'window.apos = ' + JSON.stringify(_.pick(self.apos, 'prefix', 'csrfCookieName')) +
-            '\n' +
-          '</script>\n';
+          'window.apos = ' + JSON.stringify(
+            _.assign(
+              _.pick(self.apos, 'prefix', 'csrfCookieName'),
+              { 
+                // A unique identifier for the lifetime of this
+                // HTML page in the browser
+                htmlPageId: self.apos.utils.generateId()
```
boutell: Only if there is a user

lib/modules/apostrophe-areas/public/js/user.js
boutell: Yeah it's good for parallelism with things we have to keep for bc.

PR #994, opened 2017-09-01
#886: Applies box-sizing border-box to .apos-ui

arti5m:


lib/modules/apostrophe-ui/public/css/global/generic.less
```diff
@@ -0,0 +1,9 @@
+// ===============================================================
+// Generic Styles for Apostrophe UI
+// ===============================================================
+
+* {
```
boutell: Is this imported inside something that will scope it appropriately?

Is @stuartromanek among the reviewers?

boutell:
Austin pointed out it's scoped to apos-ui. I have no remaining concerns about this.

PR #997, opened 2017-09-05
up and running and unit tested rather heavily

boutell:


stuartromanek:
In testing around:

- The tests don't get refreshed, so when you take action on an area you are in the incorrect state for a refresh. Weird but maybe no way around it.
- This should work on singletons, yes? I am getting that to work intermittently.. When i put a singleton on a page and give the singleton a brand new `name`, it seems to work (test shows empty). If I populate it and refresh, test shows not empty. If i delete the singleton the test continues to show not empty
- If I put the tests inside the template of a widget and test singleton properties inside the widget, the test always comes back Not Empty even when they are
-  For actual areas I'm getting the same behavior as singletons (point 2) so I am really getting stuck.

Let's make sure we're on the same page about usage IRL

boutell:
We ascertained that Stuart was passing the area directly to the helper — it doesn't work that way, you pass the doc or widget and the area name, or an object like `{ area: foo }`. This is consistent with the syntax of `apos.area` and `apos.singleton`.

With that clarified it works as expected for areas in widgets.

Re: refreshing people generally use this inside `{% if data.user %}` or `{% if data.page.body._edit %}` anyway to avoid a chicken and egg problem.

Merging, thanks for the tests!

PR #1001, opened 2017-09-07
fixed by blessing all widget schemas on pageServe as we already do fo…

boutell:
…r page/doc type schemas


boutell:
The package.json update is a consequence of typing npm update but if you look closely none of those changes change the semver ranges. npm just pretties it up to reflect the latest it installed but keeps the ranges the same. Any other concerns?

I agree it would be better to leave that file out of my next PR, it's just confusing.

PR #1004, opened 2017-09-08
Feature/contextual true widget

stuartromanek:
add optional `skipInitialModal` flag to widget configuration that will skip the initial editor modal from firing but still hook up the edit functionality for later, keeps the edit button on the widget.

Current use case is adding 'layout' widgets to page and having optional, secondary configuration be editable later


lib/modules/apostrophe-widgets/public/js/user.js
```diff
@@ -8,17 +8,33 @@ apos.define('apostrophe-widgets', {
...
+    // Widget can opitonally be set to skipInitialModal which skips the first
+    // edit modal but binds future editing interactions
+
     self.edit = function(data, options, save) {
-      // Only create a modal editor if the schema is not contextual only
-      if(self.options.contextualOnly){
-        // Check for an id or assign a generated one
+      
+      // Check for an id or assign a generated one
+      var checkOrGenerateId = function(data) {
```
boutell: Please make this an ordinary named function (not an assignment statement) and move it to the end of the function so it doesn't interrupt easy "top down" reading of what the method does.

PR #1017, opened 2017-09-18
in the reorganize modal attempts to drag a page above or below home, …

boutell:
… or below the conventional trashcan, no longer produce crashes or confusing errors, they are simply rejected. Also, if trashInSchema is active (for workflow) so that every page that is a parent has its own "virtual trashcan," the original "conventional trashcan" appears below any contextual trash, and it is given a special "legacy trash" label to clarify its status.

### DIFF
--- a/lib/modules/apostrophe-pages/public/js/reorganize.js
+++ b/lib/modules/apostrophe-pages/public/js/reorganize.js
@@ -237,7 +237,22 @@ apos.define('apostrophe-pages-reorganize', {
         targetId: e.move_info.target_node.id,
         position: e.move_info.position
       };
-
+      
+      // Refuse requests to move something before the
+      // home page, or after it (as a peer). Inside it is fine
+      var target = e.move_info.target_node;
+      if (!target.parent.parent) {
+        if (e.move_info.position !== 'inside') {
+          return;
+        }
+      }
+      // You also can't move something after the conventional trashcan
+      if ((target.type === 'trash') && (!target.virtualTrashcan)) {
+        if (e.move_info.position === 'after') {
+          return;
+        }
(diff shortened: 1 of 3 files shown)


lib/modules/apostrophe-pages/public/js/reorganize.js
```diff
@@ -237,7 +237,22 @@ apos.define('apostrophe-pages-reorganize', {
         targetId: e.move_info.target_node.id,
         position: e.move_info.position
       };
-
+      
+      // Refuse requests to move something before the
+      // home page, or after it (as a peer). Inside it is fine
+      var target = e.move_info.target_node;
+      if (!target.parent.parent) {
+        if (e.move_info.position !== 'inside') {
```
bobclewell: Instead of nested ifs, what about..
```
if (!target.parent.parent) &&
   (e.move_info.position !== 'inside') {
     return;
}
```

boutell: How's this?

lib/modules/apostrophe-pages/public/js/reorganize.js
```diff
@@ -237,7 +237,20 @@ apos.define('apostrophe-pages-reorganize', {
         targetId: e.move_info.target_node.id,
         position: e.move_info.position
       };
-
+      
+      // Refuse requests to move something before the
+      // home page, or after it (as a peer). Inside it is fine
+      var target = e.move_info.target_node;
+      if ((!target.parent.parent) && (e.move_info.position !== 'inside')) {
```
bobclewell: Thanks! 👍

PR #1022, opened 2017-09-20
for video widgets, display a clear error message and do not crash if …

boutell:
…there is a 401 response from the upstream oembed server. Also, do not waste time on open graph for a video widget (although it would be allowed for an embed widget, if we get around to one).


lib/modules/apostrophe-video-fields/views/video.html
```diff
@@ -5,6 +5,7 @@
 {% macro video(field) %}
   {% if not field.readOnly %}<input type="text" class="apos-field-input apos-field-input-text" name="{{ field.name }}" />{% endif %}
   <div class="apos-video-player" data-apos-video-player></div>
+  <div class="apos-video-error" data-apos-video-error="401">Embedding is not allowed for this video.</div>
```
boutell: Need to localize this.

PR #1026, opened 2017-09-23
support for the .njk file extension, which is preferred if both exist…

boutell:
… for the same template. To allow the use of njk as a substitute extension for existing templates, it does not matter whether the developer requests .html or .njk when calling render().

Implements #685

### DIFF
--- a/lib/modules/apostrophe-templates/lib/nunjucksLoader.js
+++ b/lib/modules/apostrophe-templates/lib/nunjucksLoader.js
@@ -7,20 +7,18 @@
 // Note that if apostrophe-templates has a project-level override
 // of outerLayout.html, that will be loaded instead. This is
 // intentional.
-//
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-templates/lib/nunjucksLoader.js
```diff
@@ -7,20 +7,18 @@
...
 
-module.exports = function(moduleName, searchPaths, noWatch, templates) {
+module.exports = function(moduleName, searchPaths, noWatch, templates, options) {
 
   var self = this;
+  options = options || {};
+  var extensions = options.extensions || [ 'njk', 'html' ];
   self.moduleName = moduleName;
-  self.templates = templates;
+  self.templates = templates;
```
bobclewell: whitespace.

boutell:
Thank you, I will clean up the whitespace in the merge.

PR #1027, opened 2017-09-25
drag and drop of widgets not respected due to oversight in conflict r…

boutell:
…esolution, also drag and drop targets were missing for subwidgets


boutell:
NOT finished.

PR #1048, opened 2017-10-09
Feature/add custom lang tag

HYannick:
I wanted to custom the ``outerLayout.html`` but came accross an issue. I was not able to custom the ``lang`` tag so here is a pull request for this.


stuartromanek:
The blocks should default to `en` since that is what we were using before.

boutell:
Our documents are HTML5, not XHTML, so I think the xml:lang attribute should be dropped per:

https://stackoverflow.com/questions/4199832/can-someone-explain-what-the-xmllang-attribute-does-in-html5

PR #1059, opened 2017-10-20
1055 get option

boutell:


lib/modules/apostrophe-doc-type-manager/lib/api.js
```diff
@@ -109,9 +109,31 @@ module.exports = function(self, options) {
...
     var schema = _.filter(self.schema, function(field) {
       return (!field.permission) || self.apos.permissions.can(req, field.permission);
     });
+    var typeIndex = _.findIndex(schema, { name: 'type' });
+    if (typeIndex !== -1) {
+      // This option exists so that the
+      // apostrophe-option-overrides and apostrophe-workflow modules,
+      // if present, can be used together to disable various types based
+      // on locale settings
+      var disabled = self.apos.pages.getOption(req, 'disabledTypes');
```
bobclewell: You declared `disabled` in line 112, so I don't think you need `var` here.

lib/modules/apostrophe-module/index.js
```diff
@@ -34,7 +35,7 @@ module.exports = {
       self.apos[self.options.alias] = self;
     }
   },
-
+
```
bobclewell: whitespace

test/base-module.js
```diff
@@ -38,4 +40,14 @@ describe('Base Module', function(){
...
     return done();
   });
+  
+  it('should produce correct responses via the getOption method', function() {
+    var mod = apos.modules['test-get-option'];
+    var req = apos.tasks.getReq();
+    assert.equal(mod.getOption(req, 'flavors.grape.sweetness'), 20);
+    assert.equal(mod.getOption(req, 'flavors.cheese.swarthiness'), undefined);
+    assert.equal(mod.getOption(req, 'flavors.grape.ingredients.0'), 'chemicals');
+    var markup = mod.render(req, 'test.html');
```
bobclewell: I appreciate that you check the markup here too. 👍

PR #1064, opened 2017-10-24
fix(login): use the right callback method. Remove misleading comment

HollyPony:
I've got an error while logged with a dab user and I seen that. I think it's a but urgent and it needs additionnal automated tests. but for now it's enough.

Sorry for the linting updates it was blinded by my env 🙈 

As I can see, this error was introduced recently with : https://github.com/punkave/apostrophe/commit/ddaa53fbe614f42da73b79b98b43b11131c298f6#diff-297b2881cc29c9a42671c4bc64f95986R182


boutell:
I added this fix to master directly, thank you.

However that comment you removed is correct (: You can indeed add a callback argument to your method, that's how `callAll` works, callbacks are not mandatory for every module, only those that need to do something asynchronous for that method.

PR #1068, opened 2017-10-26
feat: adding block for alt image attribute

dirago:
Hello P'unk Av',

In the `apostrophe-images-widgets/views/widgetBase.html` file, the  alt attribute in the img tag is as follows: `<img alt="{{ image.title }} ...>`

I thought it would be nice to have an `imgAlt` block for that: `<img alt="{% block imgAlt %}{{ image.title }}{% endblock %} ...>`

SYS 🖐


boutell:
I see where you're going with this idea. However, it's inside a "for" loop. Perhaps that will work — perhaps you can override the block with code that needs to execute separately for each invocation of the for loop that surrounds the block, and it will do the right thing — but I have a hunch it won't. Have you tested this with at least two images in the widget to verify the results can be separate for each one?

dirago:
I just finished testing with multiple images, it works well, here is how I override afterwards:

```njk
{% extends 'widgetBase.html' %}

{% block imgAlt %}{{ image.alt or image.title }}{% endblock %}
```

**with the `alt` field defined in the BO**:

avec


**without**:

sans

stuartromanek:
I think this is a good addition .. any reason we don't want this @boutell ?

PR #1069, opened 2017-10-27
Adding a page by task

ehibes:
I extended the apostrophe-pages module to propose a task to add a page. 
I think is a "developer-friendly" feature to include in apostrophe-pages, but if you prefer I can create a new module on npm for that.

### DIFF
--- a/lib/modules/apostrophe-pages/index.js
+++ b/lib/modules/apostrophe-pages/index.js
@@ -291,6 +291,76 @@ module.exports = {
       ], callback);
     };
 
+    // Task to add a page from apostrophe
+    // command line
+
+    self.apos.tasks.add('apostrophe-pages', 'add',
+      'Usage: node app apostrophe-pages:add title /slug type /parentslug\n\n' +
+      'This adds a new page .\n',
+      function(apos, argv, callback) {
+        return self.addFromTask(callback);
+      }
+    );
+
+    self.addFromTask = function(callback) {
+      var argv = self.apos.argv;
+      if(argv._.length !== 5 ){
+        return callback('Incorrect number of arguments.');
+      }
+
+      var title = argv._[1];
+      var slug = argv._[2];
+      var type = argv._[3];
+      var parentslug = argv._[4];
+      var req = self.apos.tasks.getReq();
+
+      var page = {
+          editUsersIds: [],
+          loginRequired: '',
+          orphan: '0',
+          published: '1',
+          slug: slug,
+          tags: [],
+          title: title,
+          type: type
+        };
+
+      
+      var parentPage;
+      var safePage;
+      return async.series({
+        findParent: function(cb) {
+          return self.find(req, { slug: parentslug }).permission('publish-apostrophe-page').toObject(function(err, _parentPage) {
+            if (err) {
+              return cb(err);
+            }
+            if (!_parentPage) {
+              return cb('Parent page not found');
+            }
+            parentPage = _parentPage;
+            safePage = self.newChild(parentPage);
+            return cb(null);
+          });
+        },
+        convert: function(cb) {
+          var manager = self.apos.docs.getManager(self.apos.launder.string(page.type));
+          if (!manager) {
+            return cb('Template not found');
+          }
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-pages/index.js
```diff
@@ -291,6 +291,76 @@ module.exports = {
...
+          editUsersIds: [],
+          loginRequired: '',
+          orphan: '0',
+          published: '1',
+          slug: slug,
+          tags: [],
+          title: title,
+          type: type
+        };
+
```
boutell: It would make more sense to populate this by looping over the schema and populating the fields that exist as command line arguments (properties of argv).

In particular it's not a good idea to force `published=1`. If it's not the default then it's not the default (on that particular site), but you would be able to pass `published=1`.

lib/modules/apostrophe-pages/index.js
```diff
@@ -291,6 +291,76 @@ module.exports = {
...
+      
+      var parentPage;
+      var safePage;
+      return async.series({
+        findParent: function(cb) {
```
boutell: Our convention is `callback` rather than `cb`.

lib/modules/apostrophe-pages/index.js
```diff
@@ -291,6 +291,76 @@ module.exports = {
...
+            return cb(null);
+          });
+        },
+        convert: function(cb) {
+          var manager = self.apos.docs.getManager(self.apos.launder.string(page.type));
+          if (!manager) {
+            return cb('Template not found');
+          }
+          var schema = manager.allowedSchema(req);
+          return self.apos.schemas.convert(req, schema, 'form', page, safePage, cb);
```
boutell: This would be a good place to loop over the schema and look for matching arguments in argv and push those into `page`.

lib/modules/apostrophe-pages/index.js
```diff
@@ -291,6 +291,76 @@ module.exports = {
...
+        },
+        insert: function(cb) {
+          return self.insert(req, parentPage, safePage, cb);
+        }
+      }, function(err) {
```
boutell: This is correct but you could just pass the callback as... well, the callback (:

lib/modules/apostrophe-pages/index.js
```diff
@@ -291,6 +291,76 @@ module.exports = {
       ], callback);
     };
 
+    // Task to add a page from apostrophe
+    // command line
+
+    self.apos.tasks.add('apostrophe-pages', 'add',
+      'Usage: node app apostrophe-pages:add title /slug type /parentslug\n\n' +
+      'This adds a new page .\n',
```
boutell: Remove extra space before .

lib/modules/apostrophe-pages/index.js
```diff
@@ -291,6 +291,76 @@ module.exports = {
       ], callback);
     };
 
+    // Task to add a page from apostrophe
+    // command line
+
+    self.apos.tasks.add('apostrophe-pages', 'add',
+      'Usage: node app apostrophe-pages:add title /slug type /parentslug\n\n' +
```
boutell: A good idea to show an example here using quotation marks in the shell as will almost always be necessary with multiword titles

lib/modules/apostrophe-pages/index.js
boutell: The slug might not match what they asked for due to the automatic deduplication of slugs. It would make sense for this task to output JSON of the page object to the console before exiting.

boutell: A good idea. Some changes we think make sense before merging.

stuartromanek:
Hi @ehibes , thanks for the work on this.

Any chance you have interest in making the requested changes?

ehibes:
Hi @stuartromanek,
I miss the time but on 2018, yes!

boutell:
Hi @ehibes, any chance of bringing this PR back to life?

localghost443:
We have a rest API now via apostrophe-headless

PR #1070, opened 2017-10-29
Fix error in apostrophe-assets documentation for self.pushAsset method

fredrikekelund:
It seems that previously, the docs have instructed to use the `options.scene` property rather than `options.when`. I believe this was just a plain error, let me know if not.

Some trailing whitespace was also removed in the diff, didn't think it would be much of an issue.


boutell:
Thanks!

Issue #115, opened 2014-05-04
renderPager doesn't really works with only 5 links


dannyblv:
I've noticed that on top of the file view/global/pagerMacros.html writed 
`{# ACHTUNG: the math was done pretty carefully to accommodate 5 links in the pager. #}`  and in my first page(?page=1) its showing 8 links and in the last page its showing 5 and in some random page (between the first and last one) its still showing 8 links (see attached images, and ignore the arrows I want them in addition to the 5 links.)

this is my view/global/pagerMacros.html file
http://pastebin.com/ni05FgKW

any idea why its not functioning good ?
thanks
![2](https://cloud.githubusercontent.com/assets/6795014/2872890/7e888a86-d37f-11e3-94a8-a2a39e49004e.jpg)
![3](https://cloud.githubusercontent.com/assets/6795014/2872892/7e88b06a-d37f-11e3-93d3-59cf58da69b5.jpg)
![1](https://cloud.githubusercontent.com/assets/6795014/2872891/7e88aae8-d37f-11e3-94d9-0d25898fa1d9.jpg)

PR #1091, opened 2017-10-31
Fix for workflow not working for globals #115

waqarini:


boutell:
Hmm, I would think the "Edit" controls and language would always be appropriate since there is only one global doc and it always exists.

stuartromanek:
@boutell Is this something we feel is appropriate?

boutell:
It's appropriate but it's in the wrong place, it should be on the edit controls rather than the "new" controls because you're never making a "new" global doc.

stuartromanek:
@waqarini can you make the above edit?

PR #1092, opened 2017-10-31
factored out a new getBaseUrl method of pages allowing this to be dyn…

boutell:
…amic based on req, an accommodation for apostrophe-workflow


index.js
```diff
@@ -182,7 +182,7 @@ module.exports = function(options) {
   self.instanceOf = function(object, name) {
     return self.synth.instanceOf(object, name);
   };
-  
+
```
bobclewell: whitespace

Issue #1097, opened 2017-11-07
Select All (not just all on page) should be available in manage view of pieces


boutell:
Both for images (grid view) and list view (everything else). Make sure it interacts sensibly with the expanded chooser (or decide it's not a thing in the chooser).

boutell:
Reassigning to Stuart since the core implementation has been code reviewed and we're just looking for styles at this point (I clicked merge, which was premature, I'll try to keep an eye out and not publish it to npm prematurely too of course).

PR #1098, opened 2017-11-07
Support for "select everything" (#1097), requires styling

boutell:
Assigning to Bob for code review and Stuart for styling.

Stuart, note that in a grid view there is one checkbox which is currently hanging out north of the filters it's meant to be next to. In a list view, you get that, and you also get the checkbox you always got before. Both work sensibly, but feel free to hide the new one with CSS if you don't think that is a good idea.


boutell:
There's a ticket about this, #1097. Also dgad 262.

PR #1100, opened 2017-11-07
Use overflow: auto; rather than overflow: scroll; in UI

fredrikekelund:
`overflow: scroll` will force browsers to always render scrollbars, while `overflow: auto` will tell them to only do it when necessary. On mac OS, this is less of an issue, since users will often have overlay scrollbars, but a lot of other platforms don't, in which case it gives a more elegant impression to use `overflow: auto`.

Reference screenshots for comparison:

`overflow: scroll`
screenshot-2017-11-7 projects 1

`overflow: auto`
screenshot-2017-11-7 projects


fredrikekelund:
Fixes #810

fredrikekelund:
Cool, great to see it merged! :+1:

PR #1101, opened 2017-11-07
Miscellaneous UI improvements

fredrikekelund:
I attempted to make a few adjustments to the overlay UI for improved consistency in where borders are placed, the colors of those borders and the spacing of the breadcrumb items.

I realize that these things are always subjective, but I still felt that these changes had some pretty clear aesthetic gains.

If there is something you are unsatisfied with, then what I feel most strongly about is the bottom border of .apos-modal-breadcrumb as well as the lighter border color of the .apos-schema-tab's

I attached two before and after screenshots for reference.

Before:
localhost_3000_ 3

After:
localhost_3000_ 2


fredrikekelund:
Great! :+1:

PR #1115, opened 2017-11-15
fixed bug where grid size wasnt adjusted in a certain select case, al…

stuartromanek:
…so allow user to toggle Select Everything without affecting Select All, created a few common methods
Bonus, if a user manually selects all pieces on the grid, activate Select All state and show 'em the Select Everything checkbox

### DIFF
--- a/lib/modules/apostrophe-pieces/public/js/manager-modal.js
+++ b/lib/modules/apostrophe-pieces/public/js/manager-modal.js
@@ -526,11 +526,9 @@ apos.define('apostrophe-pieces-manager-modal', {
         // temp hack while we refactor modals to use flexbox
         if (self.$el.find('.apos-modal-slide-current .apos-manage-grid-page').length) {
           if (self.$el.find('.apos-modal-footer').css('bottom') === '0px') {
-            self.$el.find('.apos-modal-footer').css('bottom', '72px');
-            self.$el.find('.apos-modal-body').css('height', 'calc(100% - 144px)');
+            self.shrinkGrid();
           } else {
-            self.$el.find('.apos-modal-footer').css('bottom', '0px');
-            self.$el.find('.apos-modal-body').css('height', 'calc(100% - 72px)');
+            self.growGrid();
           }
         }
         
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-pieces/public/js/manager-modal.js
```diff
@@ -567,8 +565,28 @@ apos.define('apostrophe-pieces-manager-modal', {
         } else {
           self.removeChoice(id);
         }
+
+        var gridIds = self.$el.find('[data-piece]').map(function() {
+          return $(this).attr('data-piece');
+        }).get();
+
+        if (self.choices.length === gridIds.length) {
```
boutell: Choices might not have been made on the same page. You need to make sure they are the same ids.

lib/modules/apostrophe-pieces/public/js/manager-modal.js
```diff
@@ -619,6 +638,25 @@ apos.define('apostrophe-pieces-manager-modal', {
       self.hideSelectEverything();
       self.reflectChoiceCount();
     };
+
+    self.clearEverythingChoices = function() {
+      var $pieces = self.$el.find('[data-piece]');
+      
+      self.choices = [];
```
boutell: Has this been tested with the chooser too? I think this needs to be self.clearChoices.

lib/modules/apostrophe-pieces/public/js/manager-modal.js
stuartromanek: This is only checking for a normal select piece checkbox that happens to complete all selections on a page .. reaching beyond the page still triggers UI changes in the usual way.

boutell: Right, but you're testing that by checking to see if the number of choices matches the number of ids on the page, and they could've made those choices on page two and then navigated to page three (you can do that).

lib/modules/apostrophe-pieces/public/js/manager-modal.js
stuartromanek: DIdn't realize the chooser was at play here , will test.

One thing i'm coming across immediately playing with the chooser is that the chooser bar glitches off the page when scrolling and using select all .. different issue..

boutell: To be "clear" I mean that `self.choices = []` needs to be a call to `self.clearChoices()` instead. The chooser overrides some of this stuff, if we stick to calling the methods that won't trip us up.

stuartromanek:
Made the changes and got rid of the detection around normal checking and activating the Select'd All checkbox .. @boutell turning this over to you to play nicely in the chooser

PR #1118, opened 2017-11-16
parking a page should only push permanent properties

boutell:


boutell:
Otherwise you wind up with `_children` as an actual property in the database, which is Not Supposed To Happen and causes problems.

PR #1120, opened 2017-11-20
fix: fill attachment length on insert

maninga:
Hello,

I was giving a try to your amazing framework, and I have seen that there was something missing in attachment insertion (upload or copy), from what I could test, I saw that the file length was present in data returned by uploadfs (as size field and not length), but in case of a copy, this info is not present, so I have added an utility function to get file size in bytes (in utils) and use it to fill the length property before inserting.

I also notice that there was a typo in schema for area and singleton field type in the isEmpty method.

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -48,7 +48,6 @@ module.exports = function(self, options) {
     var image = group.image;
     var info = {
       _id: self.apos.utils.generateId(),
-      length: file.length,
       group: group.name,
       createdAt: new Date(),
       name: self.apos.utils.slugify(path.basename(file.name, path.extname(file.name))),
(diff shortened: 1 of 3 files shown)


lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -153,6 +162,18 @@ module.exports = function(self, options) {
         function(callback) {
           self.uploadfs.copyImageIn(tempFile, croppedFile, { crop: crop }, callback);
         },
+        // Question: enable this is if we needed length for cropped images
```
boutell: This would not be an appropriate way to check as uploadfs is used for the purpose of ensuring it doesn't matter if files are local or in s3, etc. Assuming they live in the filesystem is against the rules (:

We don't use the length property for much (which must be evident since it wasn't working properly yet). It's a bit misleading really since we provide various scaled versions anyway.

I think for now we'll just accept your change for correctly recording the raw original size, which could be useful as something like a rough cut filter for PDFs that are bigger than the user likes... maybe someday (:

boutell:
Yes I think that's a fine use of the information since it will be close to
accurate if you are offering access to the original.

PR #1123, opened 2017-11-20
updates split plugin to work with 2.x

stuartromanek:


stuartromanek:
authored by tom and stuart

Issue #1126, opened 2017-11-21
Projections should accept common "mistakes" and turn them into features


boutell:
It is currently a mistake to write:

```
projection: {
  _url: 1,
  _project: 1
}
```

`_url` is a mistake because it's a "computed property," it's set dynamically based on other properties: `slug`, `tags` and `type` (could be other stuff if overrides are present). It's not in the database, trying to project it (or not project it) at the mongo level doesn't make sense.

`_project` is a computed property too — a join. It's set dynamically based on `projectId` which is used to fetch a project.

But, Apostrophe never stores properties starting with `_` in the database. Which means we are allowed to be clever and map these to the right things automagically:

`_url` can map to `slug`, `tags` and `type` by default (should be configurable, per piece type, in case you overrode the logic for setting `_url`).

`_project` can be looked up in the schema and turned into the right idField or idsField.

If `_project` turns out to be a **reverse** join, then the projection actually has no impact on it at all, because it's based on a property of **something else** that's pointing back to us. But we should print a warning to help the user understand that if they include it in the projection.

stuartromanek:
Love the convenience of this but is it worth dropping a warning in the console when this happens? Any worry about devs misunderstanding the structure of pieces and happily plowing along?

boutell:
Well, I would see it as a feature at this point, so a warning would seem unnecessary? It's a right thing you're allowed to do... a warning if you try to get an _ property we can't figure out would definitely make sense.

PR #1128, opened 2017-11-22
you can now "project" _url or any joinByOne or joinByArray field name…

boutell:
…. For another _ field you get a friendly warning that it does not help


boutell:
See #1126

boutell:
Reviewed in person with @austinstarin

PR #1134, opened 2017-11-28
Use reference tracking to make attachments inaccessible when none of …

boutell:
…the docs that directly contain them are outside the trash. Also, added promise support to pieces.trash, pieces.rescue and attachments.insert.

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -832,11 +832,12 @@ module.exports = function(self, options) {
       var method = trash ? self.uploadfs.disable : self.uploadfs.enable;
       return async.series([
         original,
-        scaled,
         crops,
         update
       ], callback);
 
+      // Handle the original image and its scaled versions
+      // here ("original" means "not cropped")
       function original(callback) {
         if ((!trash) && (attachment.trash === undefined)) {
           // Trash status not set at all yet means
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -605,4 +626,300 @@ module.exports = function(self, options) {
...
+        if ((!trash) && (attachment.trash === undefined)) {
+          // Trash status not set at all yet means
+          // it'll be a live file as of this point,
+          // skip extra API calls
+          return callback(null);
+        }
+        return async.eachSeries(attachment.crops, cropOne, callback);
+      }
+
+      function cropOne(crop, callback) {
```
boutell: You forgot the cropped original, you bounding cad

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -605,4 +626,300 @@ module.exports = function(self, options) {
...
+
+      function original(callback) {
+        if ((!trash) && (attachment.trash === undefined)) {
+          // Trash status not set at all yet means
+          // it'll be a live file as of this point,
+          // skip extra API calls
+          return callback(null);
+        }
+        var path = self.url(attachment, { uploadfsPath: true, size: 'original' });
+        return method(path, function(err) {
```
boutell: This might be redundant, you should be able to just treat `original` as a size, self.url knows what to do with that.

PR #1153, opened 2017-12-07
Bugfix/overlapping controls in empty areas

stuartromanek:


abea:
Maybe the add widget button could be invisibly full width, making it an easier target as well. Or otherwise bigger than it appears. Full w/h minus a few px?

That'd reduce the need to adjust the empty area size and cross fingers that some layout in the future wouldn't again raise this problem.

stuartromanek:
Added some fatter hit states for Add Content UI, outline attached.

@abea can you pull this branch into a project or two and review editing controls still work as intended?

image

stuartromanek:
thanks @abea !

PR #1155, opened 2017-12-10
parked properties should not be editable in page settings

boutell:
See #951 for the benefits of this change.


stuartromanek:
i put this comment in the wrong place..

"A little ++ on this enhancement would be disabling the fields with a little note / help text to be like 'this page is parked, these settings can't be changed'"

boutell:
@stuartromanek does that mean you looked the code over already or is this just a comment on the functionality?

I see your point. Two concerns:

1. Terminology: "parking" currently means something to developers but usually not to site users. 
2. Technical issue: sometimes the current, parked value is not a legal one, in particular if "type" is parked and we're taking advantage of that to take "type" off the dropdown, *which will finally be possible with this fix in place.* So the dropdown in read-only mode would have no way of looking up the proper label and Something Confusing would appear instead.

I'd like to push this one and see if that further enhancement makes sense down the line, if you don't mind.

stuartromanek:
I have reviewed it and tested it. By all means plow ahead.

boutell:
@stuartromanek can you mark this as an approved review? I'm trying to cut down on my nasty "as an administrator, override the review approval process" habit. Thanks!

PR #1157, opened 2017-12-12
Fixed broken less calculation

Acombes:
Added parenthesis forcing less to actually calculate the division instead of leaving it to CSS.


boutell:
@stuartromanek please review

PR #1169, opened 2017-12-17
allow any widget module to specify a wrapperTemplate module-level opt…

### LINKED ISSUE #1044: Allow any widget module to extend how its wrapper divs are output
It would be helpful to be able to pass a project level selector to `.apos-area-widget-wrapper` so that you could use CSS adjacent sibling selectors for area/widget layouts.

https://github.com/punkave/apostrophe/blob/cf79201825bd11d23adaa31d8278d520c73ab182/lib/modules/apostrophe-areas/views/widget.html#L5

boutell:
…ion, in which case it is rendered instead of apostrophe-areas:widget.html, and should extend that module and override the blocks that have been added for this purpose in order to add attributes, styles, etc.

See #1044. Closes #1044.


stuartromanek:
What does it look like to implement the wrapperTemplate ?

boutell:
Stuart, it looks like this:

```
// in lib/modules/my-widgets/index.js

module.exports = {
  wrapperTemplate: 'wrapper'
};
```

```
{# in lib/modules/my-widgets/views/wrapper.html #}
{% extends "apostrophe-areas:widget.html" %}
{% block extraWrapperAttributes %}
style="background-color: {{ data.widget.color }}"
{% endblock %}
```

My example here works with the original request in #1044, but other blocks are provided too.

This feature is documented by comments in apostrophe-areas/widget.html, which is where the OP wound up having to override fully.

boutell:
@stuartromanek bump... since you've reviewed what it will do, could you take a peek at the code as well and submit an approved review, assuming you're OK with the general strategy of course?

Issue #1174, opened 2017-12-23
Add an `email()` method to all modules


boutell:
This feature is required in order for any number of tickets to be implemented:

#282, #284, #288, #289, probably more tickets all require email confirmation messages be sent.

This has historically been challenging to get right from an API perspective. In 0.5 it was better than before but people still punted and used the nodemailer API directly a lot.

Here is the proposed API:

```
self.email(req, 'welcome',
  { name: 'Joe Smith' },
  { subject: 'Welcome', to: 'joe@whatever.com' }
).then(function() { ... oh good, my email was sent ... });
```

You can also use a callback of course.

The above code renders `welcome.html` in your module's views folder, exactly like it would for an ordinary page.

In addition, it feeds the result through the `html-to-text` module with an adjusted formatter I wrote to keep links from getting ruined by capitalization. This module is otherwise quite good at presenting links along with text in the plaintext version, so it is good enough for password confirmation emails. And that means... nobody hsa to write plaintext templates.

Except for the `templateName` and `data` arguments, which are used to render the message, and the notes above, this is very close to calling `nodemailer` directly. Everything else in your options object (after the data object for the template) is passed straight to nodemailer.

You are expected to configure the `apostrophe-email` module, giving it a `nodemailer` option, which is passed on to Nodemailer's `createTransport` method.

This time around, I gave up on a dedicated "subject" parameter to the method because the number of arguments seemed to just cause confusion in previous implementations. Also Nodemailer doesn't do it, so why would we?

PR #1175, opened 2017-12-23
Implemented email method of all modules

boutell:
See #1174 for discussion of the rationale and API for this feature.

I also took the opportunity to eliminate package-lock.json from the repo. Having a package-lock.json file in a module (as opposed to an app or website) is just bad news. It prevents updates when you install the module manually, i.e. in unit testing, but when it is installed as a dependency of an actual project the file is ignored, which is very misleading. We need to know ASAP if the latest versions of modules depended upon break the apostrophe module.


lib/modules/apostrophe-module/index.js
```diff
@@ -19,6 +19,8 @@ var async = require('async');
 var moment = require('moment');
 var fs = require('fs');
 var syntaxError = require('syntax-error');
+var htmlToText = require('html-to-text');
```
boutell: Remove this redundant line.

PR #1218, opened 2018-01-29
clonePermanent 

mtthwmnc:
`clonePermanent` was updated on server but not browser to treat only true arrays as arrays. Now this is consistent, fixing a problem with attachments in schemas seen on CP.


mtthwmnc:
@boutell I know you probably have a branching/release process for this, just opening this to keep it on your radar. Feel free to close and handle the PR/branching as you do.

boutell:
This is published to npm.

PR #1231, opened 2018-02-09
Enhancement/colorpicker and range fields

stuartromanek:
Having not added fieldtypes to apostrophe before I expect some review and changes will be necessary

**Range syntax**
```
        {
          name: 'cornerRadius',
          label: 'Corner Radius',
          type: 'range',
          min: 0, // int
          max: 30, // int
          step: 1 // int, optional
        },
```

**Colorpicker syntax**
```
        {
          name: 'backgroundColor',
          label: 'Background Color',
          type: 'colorpicker'
        }
```


lib/modules/apostrophe-schemas/index.js
```diff
@@ -1395,6 +1395,29 @@ module.exports = {
       }
     });
 
+    self.addFieldType({
+      name: 'colorpicker',
+      converters: {
+        string: function (req, data, name, object, field, callback) {
```
boutell: This must actually sanitize. Make sure it's a string (self.apos.launder.string), then make sure it's in the format you accept.

It should be possible for a color to be left null (a non-required field), so if the string is empty that should be treated as null, and on the populate/editing end of things there should be a way to go back to having no selection.

lib/modules/apostrophe-schemas/index.js
```diff
@@ -1395,6 +1395,29 @@ module.exports = {
...
+        },
+        form: 'string'
+      },
+    });
+
+    self.addFieldType({
+      name: 'range',
+      converters: {
+        string: function (req, data, name, object, field, callback) {
+          object[name] = data[name];
```
boutell: Same concern. Must sanitize.

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -1040,6 +1039,41 @@ apos.define('apostrophe-schemas', {
       convert: convertString
     });
 
+    self.addFieldType({
+      name: 'colorpicker',
```
boutell: Change this too if you go with my suggestion.

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -1040,6 +1039,41 @@ apos.define('apostrophe-schemas', {
...
+      populate: function (data, name, $field, $el, field, callback) {
+        $field.val(data[name]);
+        apos.ui.enhanceColorpicker($field);
```
boutell: This kinda makes me wish the whole field lived in its own module, like apostrophe-attachments, but I could go either way on that. We have a lot of modules and it's not as big a deal as apostrophe-attachments.

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -1040,6 +1039,41 @@ apos.define('apostrophe-schemas', {
...
+        $field.next().text(data[name]);
+        $field.on('change', function() {
+          $field.next().text($field.val());
+        });
+        return setImmediate(callback);
+      },
+      convert: function (data, name, $field, $el, field, callback) {
+        data[name] = $field.val();
```
boutell: Deal with required-ness. If that's not being picked up elsewhere somehow, you might test first.

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -1040,6 +1039,41 @@ apos.define('apostrophe-schemas', {
...
+        var $valueIndicator = $field.prev();
+        $valueIndicator.text($field.val());
+        $field.on('input', function() {
+          $valueIndicator.text($field.val());
+          $field.trigger('change');
+        });
+        return setImmediate(callback);
+      },
+      convert: function (data, name, $field, $el, field, callback) {
+        data[name] = $field.val();
```
boutell: You should reject bad input here too, so users don't have to wait until they save a doc with three levels of nested arrays to find out it's mysteriously discarded by the server-side converter (which does have to start doing that, as noted earlier).

boutell:
General direction is great, OK with these being part of the main apostrophe-schemas module with that one shim method in apostrophe-ui which is comparable to others. Need real converting/sanitizing/rejecting going on and attention & testing around required-ness (or the lack thereof).

stuartromanek:
Made all suggested improvements, thanks @boutell 

The only thing I was left unsure how to deal with is that when a `range` element's value is left unset it defaults to it's middle position (if min is 0 and max is 10, it sets itself to 5).

This goes against the range being optional.

Would a better thing be to detect whether the range is set or not and then set it to its minimum value?

stuartromanek:
@boutell can we get another look at this one?

lib/modules/apostrophe-ui/public/js/ui.js
```diff
@@ -264,6 +264,64 @@ apos.define('apostrophe-ui', {
...
+
+      // set up preview on load
+      if ($el.spectrum('get')) {
+        populateColorpicker();
+      } else {
+        emptyColorpicker();
+      }
+
+      function emptyColorpicker() {
+        $label.text('None Selected');
```
boutell: Nom this from a data attribute of the label or field that is set in the template, so it can go through localization with __()

PR #1232, opened 2018-02-09
Option to use jquery 3

### LINKED ISSUE #1202: jQuery 1.x is past end of life, not receiving fixes, how should Apostrophe 2.x respond?
An enterprise client pointed out this unpatched issue in jQuery 1.x:

https://github.com/jquery/jquery/issues/2432

We could simply say "don't do that"; we don't do that - we don't use $.get or $.post or $.getJSON or $.ajax to access other domains, only the local site.

But yeah, it's not good that the response from such a site gets executed without cross-domain origin protections. 

And more importantly... jQuery 1.x is past its end of life and not getting patches for anything, except very serious security holes:

https://github.com/jquery/jquery.com/issues/162

(This policy was under-publicized until recently as you can see.)

So what can we do?

Well, Apostrophe 3.x will not ship with jQuery at all. So that's not an issue.

As for Apostrophe 2.x today, we could work around this one issue by moving to one of the 1.12.x versions of jQuery where it's fixed - which will impact anyone depending on the bug. But this is a lame solution - there are later versions in the 1.12 series where they un-fixed it. 

Or we could move to 3.x. 3.x is only for modern browsers but it has a very generous definition of modern browser, going back to IE9. 

This would, of course, break things:

https://jquery.com/upgrade-guide/3.0/

It might not break enough things to be a big deal or impact Apostrophe's UI, but it would certainly affect some client project, somewhere.

So my recommendation is that we start shipping jQuery 3.0 with a flag to enable loading it, and use that flag in our various boilerplate/CLI projects, after first regression testing of course and, if necessary, making sure our UI code works with either 1.x or 3.x.

Thoughts?

@stuartromanek @ludovicbret @agilbert

boutell:
Resolves #1202 

Notes on the approach:

I check for a simple option, "jQuery," passed to apostrophe-assets. If it is set to 3, I push the latest version of jQuery 3 that has been validated for Apostrophe, along with a version of jQuery UI which is also known to be current, supported and compatible with jQuery 3 (note: it is not actually version 3 of jQuery UI, that's not a thing, but it's appropriate to tag it by the way we think of it: latest good version for jQuery 3).

I do it this way, rather than letting people specify a version or supply their own file or whatever, because (a) we can't tell them that will actually work because we haven't tested with it, and (b) we're not looking for a big shift in this functionality before Apostrophe 3.x. We just want the option of using a fully supported version of jQuery.

We don't force the use of jQuery 3 as a new default because backwards compatibility is an issue. I had to fix two archaic deprecated jQuery method calls in Apostrophe core that no longer exist in 3.x.

I have validated this against the apostrophe sandbox, including images, videos, and the manage/create/etc. modal experiences. It will get further validation when we regression test the release.

The use of this flag for NEW client projects internal to P'unk Avenue should be made standard. I think we should set it in apostrophe-boilerplate too. In that way new projects being born will not have security warnings about old versions of jQuery (although we were never using jQuery in an exploitable way in our own code).


boutell:
I have updated the CSS with the same tweaks that were made in the old file to accommodate paths and z-indexes in apostrophe.

PR #1241, opened 2018-02-16
all apostrophe logging flows through apos methods, and alternative lo…

boutell:
…ggers can be plugged in

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ module.exports = function(options) {
   self.rootDir = options.rootDir || path.dirname(self.root.filename);
 
   // Install logger extremely early, before we might log anything
-  self.logger = options.logger(self) || require('./lib/logger.js')(self);
+  self.logger = options.logger ? options.logger(self) : require('./lib/logger.js')(self);
   testModule();
 
   self.options = mergeConfiguration(options, defaults);
--- a/index.js
+++ b/index.js
@@ -13,6 +13,8 @@ module.exports = function(options) {
   self.root = options.root || getRoot();
   self.rootDir = options.rootDir || path.dirname(self.root.filename);
 
+  // Install logger extremely early, before we might log anything
+  self.logger = options.logger(self) || require('./lib/logger.js')(self);
   testModule();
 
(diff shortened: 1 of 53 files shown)


index.js
```diff
@@ -13,6 +13,8 @@ module.exports = function(options) {
   self.root = options.root || getRoot();
   self.rootDir = options.rootDir || path.dirname(self.root.filename);
 
+  // Install logger extremely early, before we might log anything
+  self.logger = options.logger(self) || require('./lib/logger.js')(self);
```
boutell: Hello, what if options.logger does not exist silly person?

lib/modules/apostrophe-express/index.js
```diff
@@ -497,40 +497,69 @@ module.exports = {
...
+        } else if (self.options.port) {
+          port = self.options.port;
         }
+
         if (!self.options.forcePort) {
           if (process.env.PORT) {
             port = process.env.PORT;
           } else {
-            try {
+            if (fs.existsSync(self.apos.rootDir + '/data/port')) {
```
boutell: Look at whether throwing an exception here will bubble reasonably or not, are we in a callback?

lib/modules/apostrophe-jobs/lib/implementation.js
```diff
@@ -25,15 +25,15 @@ module.exports = function(self, options) {
 
     context.checkingStop = true;
 
-    return self.db.findOne({ _id: context._id }, function(err, job) {
+    return self.db.self.apos.error({ _id: context._id }, function(err, job) {
```
boutell: WTF!!!

lib/modules/apostrophe-templates/index.js
```diff
@@ -31,8 +31,8 @@ module.exports = {
 
     self.templateApos = {
       modules: {},
-      log: function(o) {
-        console.log(o);
+      log: function(msg) {
+        self.apos.log.apply(self.apos, arguments);
       },
```
boutell: Explicitly return an empty string

PR #1249, opened 2018-02-26
moved autocollapse behavior of admin bar into its own method to be overridden at project level

stuartromanek:


stuartromanek:
some of the other methods could probably stand to be moved out as well but they're a mix of purposes and I wasn't sure what to name stuff .. future enhancement

PR #1253, opened 2018-02-28
Removed the email macro--doesn't seem to be in use

arti5m:


boutell:
Better not to change package.json yourself when submitting these. That will get done based on the sum of what is being released on a release day (might be patchlevel or minor that changes).

PR #1262, opened 2018-03-06
need a way to inject our own argv object so that in multisite we can …

boutell:
…run CLI tasks in one apos object without others seeing the arguments


boutell:
tnx

PR #1267, opened 2018-03-08
pre-add the virtual trashcans before we pass the data to jqtree, whic…

boutell:
…h moves nodes very slowly in volume

### DIFF
--- a/lib/modules/apostrophe-pages/public/js/reorganize.js
+++ b/lib/modules/apostrophe-pages/public/js/reorganize.js
@@ -443,7 +443,6 @@ apos.define('apostrophe-pages-reorganize', {
     };
 
     self.updateVirtualTrashcan = function(node) {
-      self.$tree.tree({ slide: false });
       if (!apos.docs.trashInSchema) {
         return;
       }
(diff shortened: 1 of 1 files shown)


boutell:
It turns out that while it is technically not broken, the jqtree widget has a performance issue when moving hundreds of nodes to the virtual trashcan (the home page has over 100 children in the trash). This has been resolved by pre-arranging the tree of nodes with appropriate virtual trashcans before the jqtree widget is asked to visualize them.

lib/modules/apostrophe-pages/public/js/reorganize.js
```diff
@@ -417,13 +418,32 @@ apos.define('apostrophe-pages-reorganize', {
...
+      }
+
     };
 
     self.updateVirtualTrashcans = function() {
       self.updateVirtualTrashcan(self.$tree.tree('getTree').children[0]);
     };
 
     self.updateVirtualTrashcan = function(node) {
+      self.$tree.tree({ slide: false });
```
mtthwmnc: Remove this.

PR #1268, opened 2018-03-08
Added class to h4 image title

gaco79:
Allows the title to be hidden using CSS when not required. 

It's not always pretty to show the image title when the widget is used, and especially when it's embedded in other widgets. Adding a class to the h4 image title allows it to be easily hidden using CSS only when it's not required.


boutell:
We usually resolve this by overriding that template block, but I have no objection to this approach, except that the class should be named `apos-image-widget-image-title` for clean namespacing.

gaco79:
Thanks for the correct class name. Amended the PR.

Hiding the element using CSS is better for me because I might not want to hide the image title in all widgets or on all pages. Also it's less maintenance for me in future if the original template is changed.

Thanks!

boutell:
Thank you!

Issue #1283, opened 2018-03-16
Default bodyParser size limit too small for many plausible Apostrophe scenarios


boutell:
The default limit is 100K, which would not accommodate a particularly verbose page with apostrophe-headless, and could perhaps even crop a long area while editing in the UI.

PR #1284, opened 2018-03-16
implemented 1283, also the code for permissions in widgets must not c…

boutell:
…rash if there is no manager for a widget

### DIFF
--- a/lib/modules/apostrophe-express/index.js
+++ b/lib/modules/apostrophe-express/index.js
@@ -353,8 +353,12 @@ module.exports = {
       // Allow the options for bodyParser to be customized
       // in app.js
       var bodyParserOptions = _.extend({
-        json: { },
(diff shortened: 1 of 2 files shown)


boutell:
See #1283

lib/modules/apostrophe-express/index.js
```diff
@@ -353,8 +353,12 @@ module.exports = {
       // Allow the options for bodyParser to be customized
       // in app.js
       var bodyParserOptions = _.extend({
-        json: { },
-        urlencoded: {extended: true}
+        json: {
```
austinstarin: Is this related?

boutell: Indeed, this is the piece that *is* related (to 1283, the issue in question). Plus per the title of the PR there's one other minor fix in the other spot.

PR #1290, opened 2018-03-23
651: accommodations for the apostrophe-optimizer module (which is a separate jawn)

boutell:


boutell:
Tests now pass, good catch, I broke sorting.

PR #1292, opened 2018-03-26
close up the Add Content button on next click no matter where it goes, also escape closes it up

stuartromanek:


boutell:
Doesn't killing it on any keyup break keyboard accessibility for the dropdown? (Is that even a thing that could potentially work?)

stuartromanek:
it only kills it on the esc key. It could, conceivably, still be set up to leverage tab through or arrow key

boutell:
LGTM

PR #1296, opened 2018-03-28
add check to see if controls.clone === false and skip the ui

stuartromanek:


stuartromanek:
Changed the setting name for consistency.

Noticing that clone is showing up in the singleton UI which doesn't make sense anyway ... and we are talking about this in IRL right now!

PR #1315, opened 2018-04-04
express.static middleware runs before apostrophe-global middleware; n…

boutell:
…ew "when" property can be used in "expressMiddleware" objects to run early too, by setting when: beforeRequired

### DIFF
--- a/lib/modules/apostrophe-assets/index.js
+++ b/lib/modules/apostrophe-assets/index.js
@@ -1227,15 +1227,14 @@ module.exports = {
         }
       };
       _.merge(lessMiddlewareOptions.render, self.options.less || {});
-
-      self.apos.app.use(lessMiddleware(self.apos.rootDir + '/public', lessMiddlewareOptions,
+      self.lessMiddleware = lessMiddleware(self.apos.rootDir + '/public', lessMiddlewareOptions,
         {
           // parser options
         },
         {
           compress: false
         }
-      ));
+      );
     };
 
     // Prefix all URLs in CSS with the global site prefix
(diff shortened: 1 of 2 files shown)


boutell:
DGAD-402.

You can already do this in any module to set middleware to run before that from another module:

```
self.expressMiddleware = {
  before: 'products',
  middleware: function(req, res, next) { ... }
}
```

However some middleware is added very early by the `apostrophe-express` module. For most things this makes sense: POST bodies should always be parsed, sessions should be honored, etc.

However for a few things, notably the `express.static` middleware for static files, this is a bad idea because it burns CPU for no good reason. 

The issue was doubly a concern for `express.static` because `apostrophe-assets` was still calling `self.apos.app.use` on its own for that, rather than using the `expressMiddleware` property. Which meant that it ran very late indeed, even after `data.global` was populated. 

Ludo wants to see this fixed. In addition he has other reasons why needs to be able to run middleware early, beyond fixing that for `express.static`.

So, you can now write this in any module:

```
self.expressMiddleware = {
  when: 'beforeRequired',
  middleware: function(req, res, next) { ... }
}
```


`when` can be set to `beforeRequired`, `afterRequired` (the default), or `afterConfigured` (if you want your middleware to run after project level `middleware` you passed as an option to the `apostrophe-express` module, which is probably pretty rare). These names are chosen with reference to how we already named the method of `apostrophe-express` that adds the "required" middleware (body parser, sessions, etc).

And, `apostrophe-assets` now leverages this feature.

There is some code marked "bc" here to make sure that if someone has overridden some of these methods at project level, they will still behave as they used to. That is why there is still a `modulesMiddleware` property containing *all* of the middleware, in case an old-school override expects that still.

Issue #1319, opened 2018-04-06
Should be easy to declare a separate npm root dir and locales root dir, versus the root dir used for "almost everything else" (uploads, assets, etc.)


boutell:
This is a necessity for multisite.

boutell:
With this change multisite can point most things to dashboard/ or sites/, resolving several conflicts over assets.

PR #1320, opened 2018-04-06
npmRootDir and localesDir options that override apos.rootDir, allowin…

boutell:
…g rootDir to be used to point all asset-ish, data-ish things painlessly to an alternate location


boutell:
See #1319

PR #1331, opened 2018-04-16
dont check existence, just populate, the way the real float field does

stuartromanek:


boutell:
Worth asking later if required: false means anything for this type of field. Or a float.

PR #1353, opened 2018-05-04
Set 301 HTTP status code when redirecting with apostrophe-soft-redirects

### LINKED ISSUE #1352: [apostrophe-soft-redirects] Sending HTTP 301 instead of 302 with changed slugs
Hi there,
although the name of the module `apostrophe-soft-redirect` might suggest otherwise, I believe the better way to redirect user agents with a HTTP 301 instead of the standard HTTP 302 as used by the [`res.redirect`](https://expressjs.com/en/4x/api.html#res.redirect) method in express.js. The change of the URL in Apostrophe is *permanent*. This is also recommended by Google, see https://support.google.com/webmasters/answer/93633?hl=en

> If you need to change the URL of a page as it is shown in search engine results, we recommend that you use a server-side 301 redirect. This is the best way to ensure that users and search engines are directed to the correct page. The 301 status code means that a page has permanently moved to a new location.

The redirect is happening here: https://github.com/apostrophecms/apostrophe/blob/master/lib/modules/apostrophe-soft-redirects/index.js#L45

I'll provide a PR. What do you think?

Thanks for making Apostrophe!
Leo

leomelzer:
Closes #1352

Happy to elaborate or adapt the changes!

### DIFF
--- a/lib/modules/apostrophe-soft-redirects/index.js
+++ b/lib/modules/apostrophe-soft-redirects/index.js
@@ -42,7 +42,7 @@ module.exports = {
           return callback(null);
         }
         if (self.local(doc._url) !== req.url) {
-          return req.res.redirect(self.local(doc._url));
+          return req.res.redirect(301, self.local(doc._url));
         }
         return callback(null);
       });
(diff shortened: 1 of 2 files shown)


boutell:
Absolutely, but this should be an option, not default behavior.

leomelzer:
@boutell Good idea! Just added an example.

boutell:
One very last thing I swear:

Let's just change it to `statusCode`. That's consistent with Apostrophe's general philosophy of not duplicating stuff that's implicit in the name of the module.

leomelzer:
@boutell Done!

boutell:
This will be in our next release. Thanks for contributing!

PR #1360, opened 2018-05-07
apostrophe-groups:add CLI task, plus tasks can return promises and al…

boutell:
…l modules have a convenient addTask method

### DIFF
--- a/lib/modules/apostrophe-groups/index.js
+++ b/lib/modules/apostrophe-groups/index.js
@@ -18,6 +18,7 @@
 // you may implicitly subclass it at project level to change its behavior.
 
 var _ = require('lodash');
+var Promise = require('bluebird');
 
 module.exports = {
 
@@ -57,6 +58,10 @@ module.exports = {
       .concat(options.removeFilters || []);
   },
 
+  afterConstruct: function(self) {
+    self.enableAddGroupTask();
+  },
+
   construct: function(self, options) {
 
     self.modulesReady = function() {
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-groups/index.js
```diff
@@ -86,6 +91,37 @@ module.exports = {
...
+        function(apos, argv) {
+          var req = self.apos.tasks.getReq();
+          var groupname = argv._[1];
+          if (!groupname) {
+            throw 'You must specify a group name.';
+          }
+          var permissions = argv._.slice(2);
+          return Promise.try(function() {
+            _.each(permissions, function(permission) {
+              if (!_.contains(_.pluck(self.apos.permissions.getChoices(), 'value'), permission)) {
```
bgantick: `_.pluck` removed in favor of `_.map` in lodash 4 - not a problem for us here, but we might want to stop using `_.pluck` in case we ever decide we want to upgrade lodash?

lib/modules/apostrophe-permissions/index.js
```diff
@@ -1,11 +1,26 @@
...
   alias: 'permissions',
 
   construct: function(self, options) {
     require('./lib/api')(self, options);
     require('./lib/strategiesApi')(self, options);
+    self.addListTask = function() {
+      return self.addTask('list', 'Usage: node app apostrophe-permissions:list', function() {
+        return Promise.try(function() {
+          // eslint-disable-next-line no-console
+          console.log(_.pluck(self.getChoices(), 'value').join('\n'));
```
bgantick: same `_.pluck` comment

boutell:
Map doesn't work in 3.x unfortunately. It's on our roadmap to fix the
plucks in apos 3.

PR #1361, opened 2018-05-07
Added block to the context menu

TheIceBreaker7:
Added a block surrounding the context menu code. This will allow developers to control access to this menu by simply overriding the block inside of their local "lib/modules/apostrophe-templates/views/outerLayout.html"


boutell:
Thanks! In next release.

PR #1363, opened 2018-05-09
Hotfix/forbidden type

abea:
I just ran into a project that tried to add a field named `type` to indicate the type of role the person serves at the organization. This was then being overwritten by Apostrophe (rightfully), but there was no indication of that happening.

This adds this as a forbidden field for docs and pages.

### DIFF
--- a/lib/modules/apostrophe-custom-pages/lib/api.js
+++ b/lib/modules/apostrophe-custom-pages/lib/api.js
@@ -52,7 +52,7 @@ module.exports = function(self, options) {
 
   self.composeSchema = function() {
     superComposeSchema();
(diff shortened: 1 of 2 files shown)


stuartromanek:
the console does print a warning about this, doesn't it (i just did this recently)

abea:
Ah, good point. Undone for pages.

@stuartromanek It already does for widgets, but not pieces (via doc-type-manager).

boutell:
doc-type-manager is everybody's parent class - pages and pieces.

abea:
Alright. This doesn't seem like the approach. I'll create an issue for now and close this.

PR #1369, opened 2018-05-11
Fix preview of crop image in Cross-Domain scenario

### LINKED ISSUE #1217: Image not visible during cropping when using S3 storage on different domain.
This issue doesn't appear when running locally in my dev environment, only on a hosted site. My images are hosted with a third-party S3 provider, and in the browser console I get errors saying that the request is lacking CORS headers.

It's possible to resolve this by using the AWS api to setup the S3 bucket's CORS configuration, but I feel like there should be a way around this since images appear everywhere else just fine.

This is the the crop modal that I'm talking about, and what I see.
screen shot 2018-01-29 at 11 09 35 am copy

leomelzer:
Closes #1217 

Disable Cross Origin Check of copper.js plugin by passing relevant option, as per discussion here https://github.com/apostrophecms/apostrophe/issues/1217

### DIFF
--- a/lib/modules/apostrophe-attachments/public/js/cropEditor.js
+++ b/lib/modules/apostrophe-attachments/public/js/cropEditor.js
@@ -21,7 +21,7 @@ apos.define('apostrophe-attachments-crop-editor', {
     };
 
     self.initCropper = function() {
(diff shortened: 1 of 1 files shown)


boutell:
Thanks!

tortilaman:
Huge thanks!!!

Issue #1370, opened 2018-05-11
Images in the trash are not correctly made inaccessible


boutell:
When an image is in the trash, the original should be inaccessible, and so should all of the resized versions except the designated `sizeAvailableInTrash`, if any, which by default is `one-sixth`.

Currently this is not happening due to an omitted callback and a sneaky bug in a unit test that should have caught it. Only one size is getting locked down, which isn't very helpful.

As security bugs go, this is not the worst, because it does not apply to other types of attachments other than images and expectations for really "locking down" images anyone could screen-capture on the web must always be low. But it is still a security bug and an urgent fix. A migration to fix the issue retroactively is undergoing code review.

PR #1371, opened 2018-05-11
fixed the issue (see #1370) and added a migration to address it retroactively

boutell:
#1370


boutell:
Austin approved this at the kitchen table.

PR #1375, opened 2018-05-15
Revert "Parses date before moment processing"

abea:
Reverts apostrophecms/apostrophe#1362

The ISO conversion removes consideration of the time zone. The issue with MomentJS requiring strict ISO or RFC formats in a future release does need to be dealt with, but there's not an obvious solution and this is causing problems now.

### DIFF
--- a/lib/modules/apostrophe-templates/index.js
+++ b/lib/modules/apostrophe-templates/index.js
@@ -408,9 +408,6 @@ module.exports = {
         if (!date) {
           return '';
         }
-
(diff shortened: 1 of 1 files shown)


boutell:
We should be accommodating this:

https://stackoverflow.com/questions/6549223/javascript-code-to-display-twitter-created-at-as-xxxx-ago

By providing a way to give a date *input format*, not just output format, to the date filter, or some helper method.

PR #1383, opened 2018-05-17
automagical migrations for sortified fields

boutell:


boutell:
The idea here is that all you have to do is run apostrophe-migrations:migrate after setting sortify: true. No additional code required on the developer's part to leverage the feature.

PR #1385, opened 2018-05-18
sortable choosers roughed in

stuartromanek:
Kinda guess at how you would 'validate' good choices vs hacked in choices, that could probably be more elegant, lmk.

but whoah Nelly
image
image

### DIFF
--- a/lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
+++ b/lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
@@ -331,42 +331,25 @@ apos.define('apostrophe-doc-type-manager-chooser', {
         handle: '[data-apos-sort-item]',
         update: function(event, ui) {
           var $choices = self.$el.find('[data-chooser-choice]');
-          var proposedChoices = [];
-          var intact = true;
 
-          self.$choices.find('[data-chooser-choice]').each(function () {
-            var $choice = $(this);
-            proposedChoices.push({ value: $choice.attr('data-chooser-choice') });
-          });
-
-          // account for all original items
-          _.each(proposedChoices, function(proposed) {
-            var matched = false;
-            _.each(self.choices, function(choice) {
-              if (proposed.value === choice.value) {
-                matched = true;
-              }
-            });
-            if (matched === false) {
-              intact = false;
-            }
-          });
-
-          if (intact === true) {
-            self.choices = proposedChoices;
-            if (!$choices.length) {
-              // Markup is too custom, just re-render
-              return self.refresh();
-            }
-          } else {
+          if ($choices.length !== self.choices.length) {
             apos.notify('Something went wrong, reverting changes', { type: 'error' });
-            self.$choices.sortable('cancel');
+            return self.$choices.sortable('cancel');
+          }
+
+          self.choices = [];
+
+          $choices.each(function () {
+            self.choices.push({ value: $(this).attr('data-chooser-choice') });
+          });
+
+          if (!$choices.length) {
+            // Markup is too custom, just re-render
+            return self.refresh();
           }
         }
       };
-
       self.$choices.sortable(sortableOptions);
-
     };
 
     self.getBrowserType = function() {
--- a/lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
+++ b/lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
@@ -24,6 +24,7 @@ apos.define('apostrophe-doc-type-manager-chooser', {
       }
       self.enableLinks();
       self.enableAutocomplete();
+      self.enableSortable();
       return callback(null);
     });
   },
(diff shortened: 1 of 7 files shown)


lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
```diff
@@ -324,6 +325,50 @@ apos.define('apostrophe-doc-type-manager-chooser', {
         position: { collision: 'flip' }
       });
     };
+
+    self.enableSortable = function () {
+      var sortableOptions = {
+        handle: '[data-apos-sort-item]',
```
boutell: How is this attribute actually getting into the markup somewhere?

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
```diff
@@ -324,6 +325,50 @@ apos.define('apostrophe-doc-type-manager-chooser', {
...
+          // account for all original items
+          _.each(proposedChoices, function(proposed) {
+            var matched = false;
+            _.each(self.choices, function(choice) {
+              if (proposed.value === choice.value) {
+                matched = true;
+              }
+            });
+            if (matched === false) {
+              intact = false;
```
boutell: Interesting. We don't sanity check regular up/down moves in this way. Do you feel it is necessary because you hit some cases where it didn't behave?

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
```diff
@@ -324,6 +325,50 @@ apos.define('apostrophe-doc-type-manager-chooser', {
...
+
+    self.enableSortable = function () {
+      var sortableOptions = {
+        handle: '[data-apos-sort-item]',
+        update: function(event, ui) {
+          var $choices = self.$el.find('[data-chooser-choice]');
+          var proposedChoices = [];
+          var intact = true;
+
+          self.$choices.find('[data-chooser-choice]').each(function () {
```
boutell: This seems to duplicate the query in line 333.

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
```diff
@@ -324,6 +325,50 @@ apos.define('apostrophe-doc-type-manager-chooser', {
         position: { collision: 'flip' }
       });
     };
+
+    self.enableSortable = function () {
+      var sortableOptions = {
+        handle: '[data-apos-sort-item]',
+        update: function(event, ui) {
+          var $choices = self.$el.find('[data-chooser-choice]');
+          var proposedChoices = [];
```
boutell: This would be a better place for a quick sanity check: if not $choices.length then we can't do this and should just cancel.

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
boutell: I think if there are any elements with your attribute (which is new, after all) we can trust the DOM.

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
```diff
@@ -324,6 +325,50 @@ apos.define('apostrophe-doc-type-manager-chooser', {
...
+              }
+            });
+            if (matched === false) {
+              intact = false;
+            }
+          });
+
+          if (intact === true) {
+            self.choices = proposedChoices;
+            if (!$choices.length) {
```
boutell: See above re: just giving up much earlier if we don't see choices in the world.

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
stuartromanek: https://github.com/apostrophecms/apostrophe/pull/1385/files#diff-adc89237501d9c40a3a33cf2c1c963f2R7 

and the like

lib/modules/apostrophe-doc-type-manager/public/js/chooser.js
stuartromanek: just checking the length of the DOM now

lib/modules/apostrophe-global/index.js
```diff
@@ -139,25 +140,55 @@ module.exports = {
...
+    // If called with two arguments, the first is `req` and the second is
+    // invoked as `callback`.
+    //
+    // If called with one argument, that argument is `req` and a promise
+    // is returned.
 
     self.addGlobalToData = function(req, res, next) {
-      if (self.options.deferWidgetLoading) {
-        req.deferWidgetLoading = true;
+      if (arguments.length >= 3) {
```
boutell: I think you must've based this branch off another that isn't merged yet? One of mine.

stuartromanek: i rebased from master? would that do this? Should i copy this out into a fresh from-master branch?

boutell:
I dunno. Master has this stuff already. It's weird. I'm going to merge and take the consequences.

boutell:
There were no consequences. Maybe rebase confused github's diff viewer? Who knows.

PR #1393, opened 2018-05-18
we need a clean way to add the global doc to req.data from a task etc…

boutell:
we need a clean way to add the global doc to req.data from a task etc. without doing extra work if it is already there, for apostrophe-pieces-orderings

### DIFF
--- a/lib/modules/apostrophe-global/index.js
+++ b/lib/modules/apostrophe-global/index.js
@@ -34,6 +34,7 @@
 // `_id`: the MongoDB ID of the global doc. Available after `modulesReady`.
 
 var async = require('async');
+var Promise = require('bluebird');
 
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-global/index.js
```diff
@@ -139,25 +140,55 @@ module.exports = {
       self.expressMiddleware = self.addGlobalToData;
     };
 
-    // Standard middleware. Fetch the global doc and add it to `req.data` as `req.data.global`.
+    // Fetch the global doc and add it to `req.data` as `req.data.global`, if it
+    // is not already present. If it is not already present, skip the
```
abea: Typo (I think): "If it is not already present" should be "If it is already present."

PR #1401, opened 2018-05-22
added options for adding an ID to a notification so we can find it la…

stuartromanek:
…ter, and a pulse style for ongoing work


stuartromanek:
I am fudging a cancel.. there is not a good way to programmatically dismiss

boutell:
Ain't hurtin' nothin', we can do more with it later

PR #1407, opened 2018-06-01
Tag creation: check if value before adding tag

falkodev:
Currently, in the tag management modal (in the admin bar category "Tags"), if the input is empty and the user clicks the "+" button to add a tag, there is an error message "An error occurred. Please try again.".

Several users complained it is not clear the problem is coming from an empty value. That is why I propose this pull request to have a clearer explanation. Also, this PR blocks the whole process of going through apostrophe-pieces and apostrophe-db `insert` method only to find out the there is no title in the new piece to add (because the title is filled with an empty tag value).


boutell:
Thanks!

PR #1411, opened 2018-06-05
warn developers if they use bad area names

boutell:


boutell:
It does not stop the process, but because this hasn't stopped the process in the past, I don't think we can quite get away with this update taking the site entirely down for one bad area name. It would be a good idea for it to be a process-stopper in 3.x.

PR #1421, opened 2018-06-11
Paul dedupe doc trash

starsinmypockets:
Please review implementation to ensure that refactoring dedupe code for backwards compat

### DIFF
--- a/lib/modules/apostrophe-docs/lib/api.js
+++ b/lib/modules/apostrophe-docs/lib/api.js
@@ -3,6 +3,7 @@ var _ = require('@sailshq/lodash');
 var Promise = require('bluebird');
 
 module.exports = function(self, options) {
+
   self.enableCollection = function(callback) {
     self.apos.db.collection('aposDocs', function(err, collection) {
       self.db = collection;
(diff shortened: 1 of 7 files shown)


lib/modules/apostrophe-docs/lib/api.js
```diff
@@ -3,6 +3,12 @@ var _ = require('@sailshq/lodash');
 var Promise = require('bluebird');
 
 module.exports = function(self, options) {
+  self.trashPrefixFields = [ 'slug' ];
+
+  // this alias is for bc
```
boutell: This can't be for bc if it's not in the module it used to be.

lib/modules/apostrophe-docs/lib/api.js
```diff
@@ -3,6 +3,12 @@ var _ = require('@sailshq/lodash');
 var Promise = require('bluebird');
 
 module.exports = function(self, options) {
+  self.trashPrefixFields = [ 'slug' ];
```
boutell: This should move to apostrophe-doc-type-manager so that each doc type can have its own array.

lib/modules/apostrophe-docs/lib/api.js
boutell: This actually belongs in apostrophe-doc-type-manger.

lib/modules/apostrophe-docs/lib/api.js
```diff
@@ -469,6 +475,84 @@ module.exports = function(self, options) {
...
   };
 
+  // After moving document to the trash, prefix any fields that have
+  // unique indexes so that other pieces are allowed to reuse
+  // those usernames, email addresses, etc.
+
+  self.deduplicateTrash = function(req, doc, callback) {
+    var prefix = 'deduplicate-' + doc._id + '-';
+    var $set = {};
+    _.each(self.trashPrefixFields, function(name) {
```
boutell: Rather than having only one array of trashPrefixFields, we need to flexibly get this based on the doc type. So we write self.getManager(doc.type).trashPrefixFields.

lib/modules/apostrophe-docs/lib/api.js
```diff
@@ -469,6 +475,84 @@ module.exports = function(self, options) {
...
+
+  // When rescuing docs from the trash, attempt to un-prefix any fields
+  // that have unique indexes. However, if we then find it's been taken
+  // in the meanwhile, leave the prefix in place and leave it up to
+  // the user to resolve the issue.
+
+  self.deduplicateRescue = function(req, doc, callback) {
+    var prefix = 'deduplicate-' + doc._id + '-';
+    var $set = {};
+    _.each(self.trashPrefixFields, function(name) {
```
boutell: Must use the relevant doc type manager's trashPrefixFields.

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -4,11 +4,9 @@ var Promise = require('bluebird');
 
 module.exports = function(self, options) {
 
-  self.trashPrefixFields = [ 'slug' ];
-
-  self.addTrashPrefixFields = function(fields) {
-    self.trashPrefixFields = self.trashPrefixFields.concat(fields);
-  };
+  // we need to alias this so that
```
boutell: I see how this worked, but as discussed this method will exist after all and live in apostrophe-doc-type-manager (factor it up to parent class).

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -395,7 +394,7 @@ module.exports = function(self, options) {
         if (err) {
           return callback(err);
         }
-        return self.apos.docs.deduplicateTrash(req, piece, callback);
+        return self.apos.docs.getManager(self.name).deduplicateTrash(req, piece, callback);
```
boutell: This is redundant, this module is the manager of its own type, so just say self.deduplicateTrash.

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -423,7 +422,7 @@ module.exports = function(self, options) {
         if (err) {
           return callback(err);
         }
-        return self.apos.docs.deduplicateRescue(req, piece, callback);
+        return self.apos.docs.getManager(self.name).deduplicateRescue(req, piece, callback);
```
boutell: This is redundant, this module is the manager of its own type, so just say self.deduplicateRescue.

lib/modules/apostrophe-pages/lib/api.js
```diff
@@ -701,28 +701,7 @@ module.exports = function(self, options) {
...
-  // with `_id` and `slug` properties.
-
-  self.rescueInTree = function(req, _id, callback) {
-    return self.trashInSchemaBody(req, _id, false, callback);
-  };
-
-  // Implementation detail of `trashInSchema` and `rescueInTree`. Accepts a boolean
-  // to do either.
-  self.trashInSchemaBody = function(req, _id, trash, callback) {
+  self.trashInSchema = function(req, _id, toTrash, callback) {
```
boutell: What's up with these changes to the pages module, I thought you were coming to that later...

PR #1426, opened 2018-06-14
we cannot retry on duplicate keys forever, limit it to 20 attempts wh…

boutell:
…ich makes it statistically impossible that one of the deduplicatable keys we actually know about is still at fault

### DIFF
--- a/lib/modules/apostrophe-docs/lib/api.js
+++ b/lib/modules/apostrophe-docs/lib/api.js
@@ -543,6 +543,9 @@ module.exports = function(self, options) {
 
   self.retryUntilUnique = function(req, doc, actor, callback) {
     var done = false;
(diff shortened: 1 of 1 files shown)


boutell:
This method is responsible for making the slug/path/etc. unique if there is a unique key error on an insert or update. It does so by adding random digits until it's unique, which produces a unique key very fast. callAll methods implementing `docFixUniqueError` can take care of making additional properties more unique, as needed; this is how it works for slug and path out of the box and other modules can address others. Great, except when it doesn't work because the problem is a unique index nobody wrote a docFixUniqueError method for. The solution is to try no more than 20 times; it is statistically impossible that the issue is still the slug/path after 20 random digits. Then the original unique key error should be reported.

Issue #1429, opened 2018-06-14
Widgets not always being 'unlocked' properly


stuartromanek:
If a widget has an area in it, a combination editing that area and opening/saving the widget's modal will lock the widget's areas until a refresh.

**To reproduce**

1. Clone https://github.com/apostrophecms/apostrophe-samples/tree/bugs/locked-area
2. Add a Two Column Widget to the page
3. Save the editor modal
4. Add a Rich Text widget to one of the areas, add some text
5. Re-enter the Two Column modal (Edit Two Column)
6. Save nothing
7. Interact with the previously created rich text widget

Interestingly, if the areas inside the two column widget are _not_ set to `contextual: true`, you get an unblessed error, opening another ticket for that.

Issue #1428, opened 2018-06-14
Unblessed Widget Errors


stuartromanek:
A rather simple widget config throws a nasty unblessed error on the server and locks the page

**To reproduce**

1. Clone https://github.com/apostrophecms/apostrophe-samples/tree/bugs/locked-area
2. Add a Three Column Widget to a page
3. Save the modal
4. Add a rich text widget to one of the areas. Add some text to it
5. Re-enter the Three Column editor modal
6. Should have an unblessed error in your console

PR #1430, opened 2018-06-14
we have to fix dotPaths for replaced widgets too (after the modal), n…

boutell:
…ot just inserted widgets. Also, killed noisy logging of a thing that can happen normally


stuartromanek:
#1429 looks solved by this but not #1428

boutell:
#1428 is addressed by a better explanatory message about what the developer has to do which has also been added to master.

Issue #1427, opened 2018-06-14
Apostrophe core needs a standard method for performance profiling


boutell:
Apostrophe needs a standardized method for performance profiling. It should be easy to track how long something took in a way that helps people debug slow sites.

The actual implementation of profiling can be an external module, i.e. the rapidly developing `apostrophe-debug` module (which should probably be renamed to `apostrophe-profiler`). But, it should be possible for any module to safely report profiling data without checking whether a particular profiler is installed. The default implementation would simply do nothing.

The proposed API is:

```
const p = self.apos.utils.profile(req, 'topic.subtopic.subsubtopic')
doTheThingWhichMightBeSlow(...)
p()
```

Note that the second call can be made in a callback of course. It doesn't actually matter if the thing being done is async or not because you call p() whenever you are ready.

However there is an alternate syntax that should also be accepted:
```
const start = self.apos.utils.now();
doTheThingWhichMightBeSlow(...)
self.apos.utils.profile(req, 'topic.subtopic.subsubtopic', self.apos.utils.now() - start)
```

Because sometimes you want to use timing information you are already gathering for other reasons.

As for naming things:

It would be nice to have a `profile` module in core, and for 3.x we probably should, but for 2.x we don't want to add more modules in core that have aliases for bc reasons, and not having a convenient name would make it unlikely anyone would use this.

So as with the new logging functions, `utils` is not a bad compromise. I don't see a better one.

See also DGAD-472.

boutell:
Another possible API:

```
const p = self.apos.utils.profile(req, 'topic.subtopic.subsubtopic')
doTheThing(...)
p.end()
```

I think I like this better. Nobody has to make their own `now()` calls to get high resolution times.

boutell:
Note that `req` may be null. Heck, let's say it's entirely optional.

boutell:
After a little bit of internal discussion and a lot of revision I've got a PR ready for this.

PR #1431, opened 2018-06-14
implemented profiling API (with do-nothing placeholder version, so it is always safe to call)

boutell:
See #1427 

Placeholder implementation, with tests verifying it does no harm. For now at least the version in the core intentionally does nothing, profiler modules then have an open door to implement it (and the overhead it might involve). The apostrophe-debug module now expects this API.

### DIFF
--- a/lib/modules/apostrophe-utils/lib/api.js
+++ b/lib/modules/apostrophe-utils/lib/api.js
@@ -5,6 +5,7 @@ var XRegExp = require('xregexp').XRegExp;
 var crypto = require('crypto');
 var cuid = require('cuid');
 var fs = require('fs');
+var now = require('performance-now');
 
 module.exports = function(self, options) {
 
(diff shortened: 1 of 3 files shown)


lib/modules/apostrophe-utils/lib/api.js
```diff
@@ -622,4 +623,52 @@ module.exports = function(self, options) {
...
+  // `callAll.pageBeforeSend.module-name` as categories. Note that the
+  // most general category should come first.
+  //
+  // To avoid overhead and bloat in the core, the default implementation
+  // does nothing. Also most core modules and methods do not invoke this method.
+  // However, the `apostrophe-profiler` module extends them to invoke it,
+  // for performance reasons: the profiler itself can have
+  // a performance overhead.
+
+  self.profile = function(req, key /* , duration */) {
```
abea: Was this supposed to be uncommented? Duration is referenced on L631, but then the use seems incomplete on L662.

lib/modules/apostrophe-utils/lib/api.js
```diff
@@ -622,4 +623,52 @@ module.exports = function(self, options) {
...
+        return self.profile(null, arguments[0], arguments[1]);
+      } else {
+        return self.profile(null, arguments[0]);
+      }
+    }
+    if (arguments.length === 2) {
+      // This is a do-nothing placeholder implementation
+      return function() {};
+    } else {
+      // All-in-one version with duration, there's nothing to fake
```
abea: I think this should be left out if it's not doing anything yet.

lib/modules/apostrophe-utils/lib/api.js
```diff
@@ -622,4 +623,52 @@ module.exports = function(self, options) {
...
+    }
+  };
+
+  // Returns time since the start of the process in milliseconds,
+  // with high-resolution accuracy. Used by apos.utils.profile.
```
abea: `Used by apos.utils.profile.` is confusing since it doesn't seem to actually be used.

boutell:
@abea please see the description of the PR:

"Placeholder implementation, with tests verifying it does no harm. For now at least the version in the core intentionally does nothing, profiler modules then have an open door to implement it (and the overhead it might involve). The apostrophe-debug module now expects this API."

None of this stuff gets used in the default implementation which is just a placeholder to ensure developers can call it safely without worrying about whether people have chosen to install the forthcoming profiler module or not. For performance reasons a do-nothing placeholder implementation is right for production environments.

PR #1436, opened 2018-06-21
Add hyphens to some keywords

crstnio:
NPM treats a "content management system" as three different keywords in search.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "apostrophe-cms",
     "apostrophecms",
     "cms",
-    "content-management",
-    "content-management-system",
+    "content management",
+    "content management system",
     "punkave"
   ],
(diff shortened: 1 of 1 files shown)


boutell:
I see your comment about npm treating that as three separate keywords. You're saying npm won't find a space separated keyword at all?

crstnio:
I tested `keywords:content management system` as search and NPM shows only modules with these three keywords.

But I now see that keywords:"content management system" shows exactly that ...

So, it doesn't seem to matter at all. You prefer with spaces, right? I will update all PRs if so.

boutell:
Prefer spaces yes. Thanks!

crstnio:
Sorry for the mess. I updated these PRs. Closing this now. 👍

PR #1438, opened 2018-06-22
Accommodations for reliable functional browser testing

boutell:
Includes fix for the context menu getting in the way of the add content menu. That fix is pretty blunt, it hides the context menu until the add content menu is dismissed. Without this tests don't pass (and usability on small screens is genuinely busted).

### DIFF
--- a/lib/modules/apostrophe-areas/public/js/editor.js
+++ b/lib/modules/apostrophe-areas/public/js/editor.js
@@ -207,7 +207,7 @@ apos.define('apostrophe-areas-editor', {
       self.$el.removeClass('apos-context-content-menu-active');
       // It is now OK to show this again without worrying about
       // menu interactions being suppressed due to stacking order issues. -Tom
-      $('.apos-context-menu-container').show();
+      $('.apos-context-menu-container').removeClass('apos-content-menu-active');
     };
 
     self.dismissContextContentMenuKeyEvent = function (e) {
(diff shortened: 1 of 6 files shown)


lib/modules/apostrophe-modal/public/js/modal.js
```diff
@@ -453,6 +456,16 @@ apos.define('apostrophe-modal', {
       // we need a gap of time to trigger the transition
       setImmediate(function() {
         $wrapper.find('[data-modal-content]').removeClass('apos-modal-slide-current');
+        var fired = false;
+        $content.one('webkitTransitionEnd otransitionend oTransitionEnd msTransitionEnd transitionend', function() {
+          if (fired) {
+            // Has been seen firing twice in nightwatch tests, in spite of `one`
```
stuartromanek: chrome runs both `webkitTransitionEnd` and `transitionend` events simultaneously

lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -205,6 +205,9 @@ apos.define('apostrophe-areas-editor', {
       self.$el.find('[data-apos-dropdown]').removeClass('apos-active');
       self.$el.find('.apos-area-controls').removeClass('apos-active');
       self.$el.removeClass('apos-context-content-menu-active');
+      // It is now OK to show this again without worrying about
+      // menu interactions being suppressed due to stacking order issues. -Tom
+      $('.apos-context-menu-container').show();
```
stuartromanek: can a class be toggled here? `.show()` is not transparent

PR #1440, opened 2018-06-22
Remove drag button in singleton ...

crstnio:
... for the same reason the move buttons are hidden.

Or is it possible to drag widgets into other areas?


boutell:
You can drag widgets into other areas. No need for this change. There are other ways to address it if you really want to. From the docs:

```
If `_options.controls.movable` is false, the widget may not be dragged out of the singleton.

options.controls.removable` is false, the widget may not be removed entirely.
```

crstnio:
Ah, found this part in the docs. Searched for the wrong query. Thanks for the reminder!

PR #1443, opened 2018-06-23
Vertically center dropdown arrows for select fields

crstnio:
This ensures that these arrows are vertically centered – no matter what height the field has in different browsers.

capto_capture 2018-06-23_02-22-21_pm

I could imagine a mix-in like

```less
.apos-center-vertical-abs() {
  top: 50%;
  transformY: (-50%);
}
```

to enable consistent styles across Apostrophe and project level changes.

And how about using `::` instead of `:` for pseudo-elements, since that is the current spec?

### DIFF
--- a/lib/modules/apostrophe-ui/public/css/components/fields.less
+++ b/lib/modules/apostrophe-ui/public/css/components/fields.less
@@ -42,7 +42,8 @@
   {
     content: "\f0d7";
     position: absolute;
-    top: @apos-padding-3;
+    top: 50%;
+    transform: translateY(-50%);
     right: @apos-padding-3;
     .fa;
     pointer-events: none;


boutell:
cc @stuartromanek

stuartromanek:
I think we'd rather be moving to flexbox centering that absolute hacking. I agree we should convert pseudo element colon syntax to `::`

@crstnio can you see if flex would work in this scenario?

crstnio:
I don't think `flexbox` would work in case of `select` field styling. This and other scenarios need `absolute` positioning for proper styling – where a mix-in could simplify things. I use this _hacking_ a lot and don't see disadvantages over `flexbox`. They are different things for me.

Or do I misinterpret something?

stuartromanek:
I think flexbox would work in this case because we're doing a bunch of pseudo styling of a select field with its wrapper elements, but I take your point that this is sort of an out-of-flow piece of UI.

crstnio:
Wait. I tested this a bit further and it works flawless. Never aligned an absolute element with `flexbox`.

```css
.apos-field-input-select-wrapper {
  display: flex;
  align-items: center;
}
```

I am a bit scared about IE bugs, though. Do you prefer this approach?

stuartromanek:
@crstnio yes, if you could make that update that'd be great. Thanks!

crstnio:
@stuartromanek I will take a look. Side question: Does Apostrophe handle prefixing? And what browsers are supported?

houmark:
@crstnio Would you mind creating a new PR replacing this approach with the flexbox technique on the wrapper? You cannot add to this PR after it has been merged. It has to be a new PR (GitHub limitation).

I'm doing some improvements on a part of the UI where the flexbox technique would be helpful.

crstnio:
Already done: #1457

PR #1444, opened 2018-06-25
implemented default value support for widget schema fields

boutell:


lib/modules/apostrophe-widgets/public/js/user.js
```diff
@@ -13,6 +13,13 @@ apos.define('apostrophe-widgets', {
 
     self.edit = function(data, options, save) {
 
+      if (!data) {
```
boutell: For legacy reasons, this method might be called with null or with an existing data object that could still be empty and have no _id, which still means it's effectively "brand new" and should get default values.

So we create the object if necessary and reuse it if not, and then we copy the properties of a new object created with `apos.schemas.newInstance`, which assigns default values as found in the schema for each field.

This gives us the same support for `def` properties in widgets that we've always had for doc schemas (pieces and pages).

PR #1446, opened 2018-06-26
Fix arrow position in filter dropdowns

houmark:
This PR fixes the position of the dropdown arrow in filters. Basically there was a hardcoded `top` in the `::after` which took precedence against the recent update to all select dropdowns. In a perfect world the `right` position should also be in the "core" dropdown styling but I don't want to break more than I fix at this time.


stuartromanek:
@houmark did this get addressed in the other dropdown UI fix?

houmark:
Not exactly because there’s absolute overwrites and left / right that needs to be removed. Let me recheck this one and see if I have to push another commit to fully correct this one.

houmark:
I updated this one to be compatible with the flex approach, so it should now again look right and be a better more stable approach. Ready to test/merge.

Issue #1392, opened 2018-05-18
Image Cropper: pad cropping interface into the modal


stuartromanek:
![image](https://user-images.githubusercontent.com/1889830/40253124-980faddc-5aac-11e8-90e1-2f98ca16b763.png)

When the crop handles are at the edge of the image they are being cut off by the viewport of the modal. Pad the entire interface into the modal more so that crop handles are easy to grab.

Consider disabling the zoom feature.

boutell:
@stuartromanek any concerns with this one?

stuartromanek:
I think this is a fine solution .. the main concern was when the cropper handles would be on the edge of the browser window and difficult to find, so this addresses that.

PR #1447, opened 2018-06-26
Fix image cropper on re-crop when using minSize

houmark:
Please see PR commit message. Relates to #1392


houmark:
I wanted to add to my **magical experience** here as it may not be obvious what's happening and how to re-test.

**With no padding and `autoCrop: true;` (current master):**

* Open editor for cropping an image that is very tall (taller than your screen maybe?)
* Crop and save
* Refresh just to make sure we have a non-polluted load of the editor next time
* Open the image, enter the crop editor
* Will show blank image, and no cropped area; Time to start over

**With `autoCrop: true;` but no padding:**

* Re-opening the crop editor after having saved the first crop, shows the crop on the wrong position

**Magic setup: `autoCrop: true;` and modal internally padded with 20px:**

* Re-opening shows the crop at the right position and everything seems to be.... well... magic...

Note: Only it seems that this is not 100% of the time, but almost right seems better than 100% wrong (it's easier to move the cropper to the right position). I hit a bad position once in my testing, so adding the cropper with a delay may help more, but for now I hope this improvement will at least get into the core.

houmark:
[This is the image](https://cl.ly/sXLU) I used to test this problem on my MacBook Pro 15" screen.

houmark:
I'm sorry. I just realized that this PR includes the commit with the arrow fix for dropdowns which has another PR. I can redo this PR unless you somehow can merge the other one first and then this one, which should then merge cleanly? Let me know...

I was merging all my PR's into my master fork when I should have merged them into another branch and I only noticed after I committed my latest PR. Too little sleep...

houmark:
I reverted the wrong arrow position commit in this PR to make this one merge clean against master no matter.

PR #1448, opened 2018-06-26
Make it possible to configure if admin bar loads on page load

houmark:
This PR introduces two new configuration options in app.js:

```bash
modules: {
  'apostrophe-admin-bar': {
    openOnLoad: true,
    stayOpenTimeout: 5000
  },
    ......
};
```

**Features:**

* It's possible to set `openOnLoad` to false, making the admin bar stay closed on page loads. If not set defaults to true, opening the admin bar on every page load.
* It's possible to configure the timeout until the admin bar closes after load if `openOnLoad` is `true` or `undefined` / not set. `stayOpenTimeout` configures the time in milliseconds before the admin bar closes. If **not set** or set to 0, it will default back to 3000.

**Thoughts:**

I thought about making it possible to set `stayOpenTimeout` to 0 and make it stay open infinite or until the user clicks the Apostrophe logo, but decided to not complicate the code further unless there's a want from core devs for this. In that case, let me know and I will update this PR.

This needs a bit of documentation, but I am not sure how that works, I hope the above text helps adding it for whoever has access/knowledge.

### DIFF
--- a/lib/modules/apostrophe-admin-bar/index.js
+++ b/lib/modules/apostrophe-admin-bar/index.js
@@ -207,6 +207,11 @@ module.exports = {
     self.pushAssets = function() {
       self.pushAsset('script', 'user', { when: 'user' });
       options.browser = options.browser || {};
+      var stayOpenTimeout = self.apos.launder.integer(self.options.stayOpenTimeout);
+      _.extend(options.browser, {
+        openOnLoad: (!!(typeof self.options.openOnLoad === 'undefined' || self.options.openOnLoad)),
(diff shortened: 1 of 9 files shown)


houmark:
I will create new PR based on master. I forgot I had merged a bunch of other recent things into my fork master. brb...

boutell:
Markdown documentation you provide in a `//` comment at the beginning of a module's `index.js` file appears on its page of the docs site. 

An example of this being done right for documenting options:

https://github.com/apostrophecms/apostrophe/blob/master/lib/modules/apostrophe-assets/index.js

(You don't need to go overboard here, methods also get documentation based on their preceding comments, just document the one new module-level option.)

PR #1455, opened 2018-06-28
slug and path should be deduplicated via suffixes, not prefixes, to avoid confusing the hierarchy. slug and path deduplication must be propagated, with tolerance for new pages that have taken slugs since.

boutell:


boutell:
This is building on Paul's earlier work to address deduplicating slugs and paths for pages when they move to the trash in workflow, where the trash is really just a flag, so they would "camp" on their old slug, preventing the creation of new pages with that same slug, if nothing were done to prevent that.

grdunn:
@tom conceptually this looks good and swell. Wouldn't mind potentially PR'ing this IRL to go through some of the workflow stuff, as I'm not well versed in that specific module.

boutell:
Sure. Would help my own clarity. I'm on the patio if you'd like to go through it.

PR #1456, opened 2018-06-28
Make it easy to prepend or append content at various critical points in the layout from any module without bc breaks

crstnio:
This adds a hook to `outerLayoutBase.html` for NPM modules to add code into the `startHead` block.

Basically a copy of `apos.pages.addAfterContextMenu`.

### DIFF
--- a/lib/modules/apostrophe-templates/views/outerLayoutBase.html
+++ b/lib/modules/apostrophe-templates/views/outerLayoutBase.html
@@ -2,6 +2,7 @@
 <html lang="{% block locale %}en{% endblock %}" {% block extraHtml %}{% endblock %}>
   <head>
     {% block startHead %}
+      {{ apos.pages.startHead() }}
     {% endblock %}
     <title>{% block title %}{% endblock %}</title>
     {{ apos.assets.stylesheets(data.when) }}


bobclewell:
@boutell would a similar `startBody()` hook also make sense to add? I think it would make the element I'm adding to the page in the review and deploy/visual diff updates cleaner. (This could also be `endBody()` or `appendBody()`)

crstnio:
I think it would be useful to add as many hooks as possible for module authors. But this may need a simpler method of adding hooks. Currently each hook needs a separate method, array variable and helper in two different files. Could this be streamlined into one method?

boutell:
You're not wrong. Hmm.

I feel like these insertion points, or the functions we're passing to insert stuff at them, need a name.

They are akin to nunjucks blocks, it's like a programmatic version of that.

How about:

```
self.apos.templates.atBlock('startHead',  function(req) { ... return some markup ... })
self.apos.templates.atBlock('extraHead',  function(req) { ... return some markup ... })
self.apos.templates.atBlock('startBody',  function(req) { ... return some markup ... })
self.apos.templates.atBlock('extraBody, function(req) { ... return some markup ... })
self.apos.templates.atBlock('afterContextMenu', function(req) { ... return some markup ... })

^^^ pages.addAfterContextMenu would become a wrapper for this last one ^^^
```

Now, in `outerLayoutBase`, we would need to mark the supported "at-blocks":

```
<head>
  {# Simple nunjucks block, for project level devs %}
  {% block startHead %}{% end startHead %}
  {# at-block, for us hardcore folk writing npm modules #}
  {{ apos.templates.atBlock('startHead') }}
  <title>blah</title>
  {% block extraHead %}{% end extraHead %}
  {{ apos.templates.atBlock('extraHead') }}
</head>
<body>
  {% block startBody %}{% end startBody %}
  {{ apos.templates.atBlock('startBody') }}
  <main>...</main>
  {{ apos.templates.atBlock('extraBody') }}
  {% block extraBody %}{% end extraBody %}
</body>
```  

(In theory it would be nice if the atBlock calls magically added the regular nunjucks blocks too, or something, but there won't be 100's of these, and that would probably be hideously difficult to implement.)

boutell:
You would be able to call `atBlock` from your javascript code as many times as you wanted, just like you can now with `addAfterContextMenu`.

bobclewell:
I don't have a strong opinion about `atBlock`, but it seems to me that rather than `startBody`/`extraBody` naming them `prependBody`/`appendBody` would be better.

bobclewell:
What about...
```
self.apos.templates.prepend('head',  function(req) { ... return some markup ... })
self.apos.templates.append('head',  function(req) { ... return some markup ... })
self.apos.templates.prepend('body',  function(req) { ... return some markup ... })
self.apos.templates.append('body', function(req) { ... return some markup ... })
self.apos.templates.append('contextMenu', function(req) { ... return some markup ...
```

crstnio:
I think the names of these hooks should match the existing block names to avoid confusion.

`atBlock` sounds good as a universal function, but I would place these inside existing blocks so that project level developers can choose whether to place their block-code before or after these `atBlock`s by utilizing `{{super}}`.

boutell:
Hmmm, I see the attraction of placing them inside existing blocks, but that would mean that if you're already overriding an existing block, you are dead in the water the first time you install a module that relies on this mechanism.

As for the functions, I like Bob's proposed syntax a lot. It's highly readable and it sidesteps the block naming issue. 

Bob's syntax needs a corresponding syntax for what gets inserted in `outerLayoutBase` to output what you prepended and appended. I'm thinking it's as simple as:

```
<head>
  {{ apos.templates.prepended('head') }}
  <title>blah</title>
  {{ apos.templates.appended('head') }}
  {% block extraHead %}{% end extraHead %}
</head>
```

Use of the past tense clarifies that we're outputting what we prepended or appended elsewhere.

We can continue to add convenience blocks for project level development, but I don't think they should contain the automatic stuff  because misuse of that just blows up apostrophe's interface in ways you don't expect or understand. If there needs to be a way to shut off a module's interface, if that might make sense in some projects, then it should have an option to do that at config time.

crstnio:
I think it is more important to establish a standard by using `{{ super }}` all the time rather than taking care of project level developers who don't use that nor read the documentation of modules that rely on it and wonder later on why things aren't working anymore.

If it's meant to be written in a certain way, they should do it. Otherwise the compatibility is lost.

As to the `prepend`/`append` syntax: It might simplify the naming, but could add more complexity to the code. Doesn't it?

boutell:
That would be a reasonable approach for 3.0 @crstnio, but for 2.x we can't break BC by suddenly putting potentially mission-critical stuff in "extraHead", which developers currently reasonably expect to override without any consequence to their use of apostrophe modules.

boutell:
prepend/append doesn't add any complexity, no. Clarity for developers is worth it.

boutell:
Having a quick internal convo to make sure there's consensus on the syntax. Thanks for your patience!

boutell:
Closing this PR, we shipped `apos.templates.append` and `apos.templates.prepend`, see the changelog for details. Thanks for the input that led in this direction!

PR #1459, opened 2018-06-29
Simple way to serve a folder of static assets in the cloud, and update it or remove it all later

boutell:
"Efficiently sync a local folder of static files to uploadfs,
with the ability to remove the folder and its contents later.
This ensures your application will still work if you move
it to s3, azure blob storage, etc. via uploadfs.

It is meant for delivering the output of tools like backstop,
sitemap generators, etc. that generate output as HTML files
and associated assets.

As long as you use this module, you won't have to worry about
making changes to those tools on the day you switch to S3."

`apos.cloudStatic.syncFolder('/local-folder', '/uploadfs/folder')` --> pushes it up, replacing folder’s contents efficiently if it exists
`apos.cloudStatic.removeFolder('/uploadfs-folder')` --> gets rid of one of these folders when you’re done with it
`var url = apos.cloudStatic.getUrl('/uploadfs-folder')`

I think I've limited the scope properly here, and keeping it inside Apostrophe reflects that this is a utility module for us that makes sense given what Apostrophe has on hand, not something more cosmic in scope.


boutell:
Oops, fixing a couple things.

boutell:
OK, we're really ready now (:

boutell:
We made this a standalone npm module instead.

PR #1463, opened 2018-06-29
Fix styling of autocomplete item active state

houmark:
This relates to PR #1432 which never made it in, because it was modifying the vendor file.

This overrules the basic styling of jQuery UI in a way that jQuery UI vendor can be updated without this breaking (unless there's changes to the class names or structure of the autocomplete of course).

I kept the styling pretty basic (bold font on active), but I could add more transitions if someone would point me in the direction of what you would like.

### DIFF
--- a/lib/modules/apostrophe-ui/public/css/components/autocomplete.less
+++ b/lib/modules/apostrophe-ui/public/css/components/autocomplete.less
@@ -6,16 +6,19 @@
 
   .ui-menu-item
   {
-    color: @apos-white;
-    background-color: @apos-primary;
+    color: @apos-primary;
+    background-color: @apos-light;
     border: none;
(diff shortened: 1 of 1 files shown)


houmark:
This one also improves the UI and gets it back to a more Apostrophe like look. Maybe the "active" state needs a different styling, but I have never seen the correct styling, so if you have any feedback on that @stuartromanek I will update the PR. Otherwise it should be ready to merge.

stuartromanek:
Thanks for this @houmark 

Can you update the styles to match something like this?

image

you can use `@apos-light` as the background gray color of un-hovered list items

houmark:
Yes will do later today.

houmark:
The requested changes has been done. Pretty happy with how it looks now.

PR #1465, opened 2018-07-02
Fix label for finished button in tags

houmark:
This PR corrects the label for the tags modal.

This modal is the only modal using the label "Finish", where all the other modals use "Finished".


boutell:
Thanks!

PR #1469, opened 2018-07-03
you can now run tasks programmatically

boutell:


lib/modules/apostrophe-tasks/index.js
```diff
@@ -27,9 +28,83 @@ module.exports = {
...
+    // without using the command line or using the `child_process` module.
     //
-    // The group name, by convention, should be the name of your module.
+    // Except for `name`, all arguments may be omitted.
+    //
+    // If you do not pass a callback, a promise is returned.
+    //
+    // Examples (assume `products` extends `apostrophe-pieces`):
+    //
+    // `self.apos.tasks.invoke('apostrophe-users:add', [ 'admin', 'admin' ]).then(function() { ... })`
```
stuartromanek: differentiate username from group in examples, can be confusing `['bob', 'admin']`

stuartromanek:
Looks good, I assume there is no difference between just running the task function vs spinning up a full node process and running it?

boutell:
Thanks. To answer your question: sure there's a difference, some tasks might not be written with the continued existence of the process in mind and might do stuff like calling process.exit() or whatever. But, if they do behave, you've saved a lot of time and RAM and avoided the pitfalls and hassles of escaping arguments for the shell. And the possible issues re: poorly behaved tasks are documented with the method.

PR #1470, opened 2018-07-03
Improve check for current page is piece being deleted

houmark:
This is an update to a prior already merged PR, attention @boutell.


boutell:
Thanks!

PR #1475, opened 2018-07-05
implement ability to display a count for each distinct choice in piec…

boutell:
…esFilters. The new distinctCounts filter alters the behavior of toDistinct to use an aggregation pipeline. toDistinct replies normally but you can then call cursor.get("distinctCounts") to get an object with the counts. This sidesteps a ton of bc breaks to existing "choices" methods for every type of filter that would otherwise be needed for them to gather more data, skipping ahead to the insight that everybody relies on toDistinct in the end.

In `piecesFilters` all you have to do is add `counts: true` to a filter to get this behavior.

Of course it is not as fast as plain distinct.

### DIFF
--- a/lib/modules/apostrophe-docs/lib/cursor.js
+++ b/lib/modules/apostrophe-docs/lib/cursor.js
@@ -1049,9 +1049,7 @@ module.exports = {
         return Promise.promisify(body)(property, options);
       }
       function body(property, options, callback) {
-        var countsWas;
         if (options.counts) {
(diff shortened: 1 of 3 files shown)


stuartromanek:
Looks good but Travis is hollering about

```
/home/travis/build/apostrophecms/apostrophe/lib/modules/apostrophe-docs/lib/cursor.js
  1052:13  error  'countsWas' is assigned a value but never used. Allowed unused vars must match /apos/  no-unused-vars
```

PR #1476, opened 2018-07-05
Improve batch operations UI

houmark:
This PR improves batch operations in the editor modal.

Due to missing disabled attributes on the button, the button is clickable even if no items has been selected for a batch operation. This is now corrected with a check in JavaScript.

At the same time the text for the confirmation has also been made "plural aware".

The UI is improved by rounding the dropdown, fixing the arrow position, by removing specific styling for the dropdown, and relying on the recent flexbox additions to arrows in general.

For tags, the field to add tags has been moved to the left, so it appears between the batch operation dropdown and the button to perform the action. It was appearing to the right of button before this PR, which could be confusing.

### DIFF
--- a/lib/modules/apostrophe-modal/public/css/components/modal-batch-operations.less
+++ b/lib/modules/apostrophe-modal/public/css/components/modal-batch-operations.less
@@ -8,9 +8,11 @@
 	.apos-manage-batch-operations-select {
 		display: inline-block;
 		width: auto;
+    min-width: 180px;
 		margin-right: @apos-margin-2;
 		font-size: 14px;
 		line-height: 21px;
+    .apos-rounded
 	}
 
   .apos-field-input,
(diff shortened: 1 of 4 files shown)


stuartromanek:
I am good with the front-end of this, @boutell do you see any issues with the length checks?

houmark:
@boutell Could you show me an example of other similar server side localization techniques currently used by Apostrophe? I have this one and other similar ones (especially form validation errors) on the PR list. I would like to try to get most user facing errors localized over time, but before starting, I would need to understand what's the best way to do it.

houmark:
I figured that out. An object for browserCall with messages in is the way to go, but I have texts with dynamic values that needs to have %s replaced to the value in browserland, which seems to become complicated because neither Apostrophe, Lodash, jQuery or vanilla JavaScript has a function to do that (PHP has lovely sprintf). Cannot do it server side (where it is possible already) because I can have situations like this:

* Field A in a schema has a min. length validation of 5
* Field B has a min. length validation of 10
* The schema will have error messages for all fields and just one for `min` but that one needs to be replaced client side with the field configuration value to output "min length of 5" for field A in the error message

On this one, I would love to get some feedback on what is the next step to be able to parse %s. I thought about a utility function, but not sure what you prefer?

PR #1484, opened 2018-07-10
Allow passing options to serve-static middleware

### LINKED ISSUE #1482: Allow setting options on `serve-static` middleware for e.g. client-side caching
Currently it is not possible to set options on the Express middleware [`serve-static`](https://github.com/expressjs/serve-static/) which serves static content from the `/public` directory. One effect is the resulting Cache-Control header `max-age: 0`, [which is the default](https://github.com/expressjs/serve-static/blob/master/index.js#L64). This de-facto disables client-side caching for generated assets.

I understand some setups don't require this as there will be e.g. an NGINX (which modifies headers etc) in front of Node but there might be other situations. Some CDNs also reflect the Cache-Control headers send through the application. As those are currently the defaults, no caching at all might occur. 

In general the assets can be cached for a long time because Apostrophe already hashes the generated file names.

What notices should be put in place to enable a fluent development flow? E.g. `Cache-Control` etc. should be left out during development. One option would be to **not** set the options if `self.options.minify === true` doesn't hold.

I'll provide a PR.

leomelzer:
The options can be set by providing the `serveStaticOptions` option in `apostrophe-assets`.

I have added a test scenario to `test/express.js` to cover this but unfortunately this is a bit weird to test: 

1. We would need to set `minify: true` which increases test time a lot (would need to increase timeout)
2. The existing file `test.less` (from `apostrophe-test-module-push`) can not be built during bundling because the required less file isn't present. We would need to introduce an extra static file then. 

Here is the test in any case: https://gist.github.com/leomelzer/e67ac6f86e87f3e29cb1b708c37ed7fa (last test case)

In the end this feature is only about passing options further down to the middleware.

Closes #1482.


lib/modules/apostrophe-assets/index.js
```diff
@@ -32,6 +32,23 @@
 // that you enable this option in production, and also in staging so you can see
 // any unexpected effects. If this option is undefined, the APOS_MINIFY
 // environment variable is consulted.
+//
+// ### `serveStaticOptions`
+//
+// Pass options to the [serve-static](https://github.com/expressjs/serve-static/)
+// Express middleware, such as `Cache-Control` and more. If no options are defined,
+// the default options from the middleware will be used. Please note these options
+// only take effect if you set `minifiy: true`. This is to ensure up to date assets
```
boutell: Typo

lib/modules/apostrophe-assets/index.js
```diff
@@ -1288,7 +1305,10 @@ module.exports = {
         // this property won't be set
         middleware.push(self.lessMiddleware);
       }
-      middleware.push(self.apos.express.static(self.apos.rootDir + '/public'));
+      middleware.push(self.apos.express.static(
+        self.apos.rootDir + '/public',
+        _.defaults({}, self.options.minify === true ? self.options.serveStaticOptions : {}))
```
boutell: I don't think we should be making assumptions about how others will use this option. It might be about caching — or something totally different.

Instead, developers should pass what they want to pass in each environment.

This can be done via the `data/local.js` mechanism, which merges the export of that file with the `app.js` apos configuration object, or via the use of environment variables etc. when writing configuration.

leomelzer:
@boutell Thanks for the suggestions. I've updated the PR.

lib/modules/apostrophe-assets/index.js
```diff
@@ -1288,7 +1306,10 @@ module.exports = {
         // this property won't be set
         middleware.push(self.lessMiddleware);
       }
-      middleware.push(self.apos.express.static(self.apos.rootDir + '/public'));
+      middleware.push(self.apos.express.static(
+        self.apos.rootDir + '/public',
+        _.defaults({}, self.options.serveStaticOptions)
```
boutell: Does `_.defaults` guarantee reasonable behavior if `self.options.serveStaticOptions` is null? I don't see it in the docs. I would usually use `|| {}` here.

Also, it's a given that things in `options` are options, so we should just call it `static` and in context it's clear that it contains options.

I suggest we just say `static` because the express documentation, which is what people are probably looking at, never explicitly says express.static is serve-static, although that is what is in the source. And brevity is nice if there's no lack of clarity.

lib/modules/apostrophe-assets/index.js
```diff
@@ -33,7 +33,7 @@
 // any unexpected effects. If this option is undefined, the APOS_MINIFY
 // environment variable is consulted.
 //
-// ### `serveStaticOptions`
+// ### `static`
 //
 // Pass options to the [serve-static](https://github.com/expressjs/serve-static/)
```
boutell: Just reference `express.static` since we don't control what express actually does with that. Sorry for the nitpicking this is the last one (:

leomelzer:
@boutell Thanks, good note. My idea with using `defaults` was to allow the addition of Apostrophe core defaults at some point but this might be too early. I changed it to use the `|| {}` pattern now.

Think the behavior would have been fine. For reference:

```js
> const _ = require("@sailshq/lodash")
undefined
> _.defaults({}, {maxAge: "foobar"})
{ maxAge: 'foobar' }
> _.defaults({}, null)
{}
> _.defaults({}, undefined)
{}
> _.defaults({}, "foobar")
{ '0': 'f', '1': 'o', '2': 'o', '3': 'b', '4': 'a', '5': 'r' }
> _.defaults({}, 123)
{}
```

lib/modules/apostrophe-assets/index.js
leomelzer: Good point, changed with reference to https://expressjs.com/en/4x/api.html#express.static

boutell:
🎉 thank you

PR #1485, opened 2018-07-11
Make it possible to set field classes and attributes on form fields

houmark:
This PR adds the possibity of setting a schema field with the following configuration:

```
name: 'name',
type: 'string',
label: 'Name',
...
fieldClasses: 'custom-class',
fieldAttributes: 'data-custom-attribute'
```

Currently Apostrophe supports adding both `classes` and `attributes` in the same way, but those are added to the `fieldset` which wraps the form field. Some CSS frameworks requires a class put on the field itself (this is related to a submit piece where schema fields are used for a frontend submit).

**Note:**
There's a slight annoynce here in naming, as `classes` and `attributes` are already in use for the`fieldset` but for bc compatibility, it's not really possible to use the existing ones for the field and create new ones for the `fieldset` so this is the only way, even if it's not 100% obvious that this is a bit opposite of what one would expect. Maybe this can be "swapped" For 3.0, as this seems to be an undocumented feature, it's maybe not much in use yet.

### DIFF
--- a/lib/modules/apostrophe-schemas/views/fieldset.html
+++ b/lib/modules/apostrophe-schemas/views/fieldset.html
@@ -1,7 +1,7 @@
 {% macro render(field, bodyMacro) %}
   {# WARNING: if you remove apos-field as a class you will have problems with nested schemas reusing #}
   {# field names. We use this class with our $.findSafe plugin. #}
-  {% set options = { 'id': apos.utils.generateId() } %}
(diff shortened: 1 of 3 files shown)


houmark:
Yes, I thought about that also, considering I did that kind of work recently for the body tag attributes. There's some inconsistency in the core here both on naming and also arguments to functions, but I guess a later cleanup could take of that. For 3.0 it could be interesting to do a deprecation approach where wrapper functions are set to accept the old behavior but warns about the correct use, and then in 4.0 for example, the code is removed to only accept the new way.

I will put the documentation on my list :)

PR #1510, opened 2018-07-29
Enhance apos.singleton - attempts to fix #1023

### LINKED ISSUE #1023: Address edit: false for singletons, area-level options in general for singletons
Currently it is not possible to pass edit: false to a singleton. But people try, and this results in rich text widgets being configured with no toolbars etc., because the presence of the options argument signals to Apostrophe that it doesn't have to go get the options from the schema of the piece.

What is needed is either (a) a clear and unambiguous way to pass *area level* options to a singleton, such as "edit", or (b) a list of options we're going to recognize as area-level and pull out and not consider part of the widget options.

I prefer option one, but I can see it being problematic. You would have to at least give the widget options argument as a placeholder. This isn't obvious:

`{{ apos.singleton(data.page, 'body', 'apostrophe-rich-text', null, { edit: false }) }}`

Option two would look like:

`{{ apos.singleton(data.page, 'body', 'apostrophe-rich-text', { edit: false }) }}`

But it means the widget itself can never have an "edit" option, and as we add more area level options that are deemed singleton-worthy, there could be more such conflicts.

In fairness, an "edit" option at widget level seems pretty unlikely.

manoj-krishnan:


manoj-krishnan:
Commit history is messy, I'll push a cleaner version.

PR #1512, opened 2018-07-30
Avoid adding UI drop zones to non-movable widgets

fredrikekelund:
This is a very minor fix, but I ran into it when looking into the drag and drop functionality. If you have multiple areas on the same page where the same type of widget can be added – the drag and drop functionality supports moving a particular widget between these different areas. 

It is however possible that another area has marked this widget as being "non-movable" – ie `options.controls.movable` is false. This PR makes it so that drop zones aren't added to those areas.

There is a second change in here that fixes another issue I ran into. The drag and drop functinality resets certain CSS styles after a widget has been dropped. It seems to assume very optimistically what these styles should be though (both `height` and `width` are simply set to `auto`). This 2nd change makes it so that the styles are actually removed from the DOM, and not just reset to the default browser value for each CSS rule.


boutell:
I think this makes sense, @stuartromanek can you confirm Fredrik's change to the CSS isn't going to do anything it shouldn't?

PR #1513, opened 2018-07-30
implemented 854: additional joins in the schema of apostrophe-pieces-…

boutell:
…widgets are now loaded as you would expect them to be


lib/modules/apostrophe-pieces-widgets/index.js
```diff
@@ -7,6 +7,14 @@
 // To learn more and see complete examples, see:
 //
 // [Reusable content with pieces](../../tutorials/getting-started/reusable-content-with-pieces.html)
+//
+// ## options
+//
+// ### `loadManyById`
```
boutell: This is just missing documentation, not a new feature.

PR #1517, opened 2018-07-31
Fix nested error messages

houmark:
This PR fixes nested error messages.

Consider a checkbox field with a fieldset containing a label for the main label, and checkboxes with a label for each checkbox.

Before this PR, all labels would get the "required" error and become styled accordingly.

This PR fixes that behavior and ensures the error and styling is only applied to the first label within the fieldset.


boutell:
Are you sure the CSS rules are necessary with the jquery fix present?

houmark:
Yeah, kind of, because it would otherwise style with colors etc. on the checkboxes for example. But I can give this another check if you like? I did have these changes stashed a few days.

boutell:
OK yes, I see that now, the fieldset is still what gets the class and so specificity is still important here. ✔️

PR #1520, opened 2018-07-31
Fix background color on notification on uploads when file extension is not accepted

houmark:
This PR sets the type error of the notification, when a file upload extension is not allowed by configuration.

Before this PR it would be white (no background) — after this PR it will be red.


boutell:
Thanks.

PR #1530, opened 2018-08-04
Fix late population of default values

houmark:
This PR fixes late population of default values which results in an overwrite of a user input value.

Before this PR a schema field value would **always** be set on populate. Depending on latency of the user JS and how early in a custom module the `newInstance` would be called, this could affect user input values. A user may have started applying values in a field, and once the populate finally ran, that value would be overwritten to "" (effectively emptying the field).

This fix makes populate only apply a value if not `undefined` meaning `null` values are still applied (date type field uses that).

### DIFF
--- a/lib/modules/apostrophe-schemas/public/js/user.js
+++ b/lib/modules/apostrophe-schemas/public/js/user.js
@@ -889,7 +889,9 @@ apos.define('apostrophe-schemas', {
     }
 
     function populateString(data, name, $field, $el, field, callback) {
-      $field.val(data[name]);
+      if (data[name] !== undefined) {
+        $field.val(data[name]);
+      }
       if (field.max) {
         if (field.textarea) {
           if (field.max) {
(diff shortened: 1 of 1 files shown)


boutell:
Intended behavior, a form should not be displayed to the user until
populate succeeds. That is the case for modals. Perhaps not for array
items...

houmark:
Hmmm, and not for a user module frontend form either. Seems like some of these edge cases shows up mainly because it's not running in the modal as it was intended. Does this fix have side effects? If not, then it does fix cases where the form may show before. I am currently not sure how to avoid rendering my "test" form until populate has happened but I will of course play around with it (my form is using so many different tricks that I am really pushing things to the limit some places, but I am still using widget.play and the basic same concept as "pieces submit").

boutell:
Populate needs to finish before any interaction is allowed but yes, we
don't yet offer a simple way to ensure that for forms already on the page.
Hmm.

Some field types might not react well at all to being interacted with
prematurely.

houmark:
Premature interaction may be a problem, but as an example the date field will need `null` to not set today as default date, but my PR maintains that behavior as `null !== undefined` so if you enter to interact with the date field too early (considering it's prerendered like my case), then the date picker simply doesn't open.

Issue #1415, opened 2018-06-06
callAll is not promise-friendly, awkward when you don't know if you need to use the super pattern; replace with promisified events in 3.0


boutell:
(Discussion moved to github because the forum has not proven to be an effective place to spark technical conversation.)

Now that we've shared our general goals for 3.0, we'd like to begin sharing a series of conversations about planned changes. Your input here in the forum on these issues is valued and we look forward to it.

Here is the first such proposal. It concerns Apostrophe's "callAll" and "super" patterns.

## What callAll does, and what's good and bad about it

We have been thinking about the challenges people encounter with the “super” pattern, mainly in simple cases where they only want to add behavior, not change or override existing behavior. This comes up a lot with `callAll` methods like `pageBeforeSend`.

Some background:

If any module invokes `apos.callAll('addYourNotes')`, Apostrophe looks for an `addYourNotes` method in every module. If it exists, it gets called. This is a powerful and useful feature. Many developers have implemented at least one `docBeforeSave` or `pageBeforeSend` method.

This is great. But when your module extends another module, it is very difficult to be sure if the parent module already has an implementation. And therefore very hard to tell if you need to use the `super` pattern.

Here is an example of the `super` pattern, which we regularly use when we want to extend a method but still call the original version:

```
var superAddYourNotes = self.addYourNotes;
self.addYourNotes = function(notes) {
  superAddYourNotes(notes);
  notes.push(['my note']);
};
```

This is great, but if you don't know if there is already an `addYourNotes` method, it becomes more verbose:

```
var superAddYourNotes = self.addYourNotes || function(notes) {};
self.addYourNotes = function(notes) {
  superAddYourNotes(notes);
  notes.push(['my note']);
};
```

And in some cases much more complex than this, for instance if `addYourNotes` takes a callback. 

This is confusing and developers make mistakes.

## What about events?

Other systems address this by using hooks and events instead. Apostrophe supports events too:

apos.emit('addYourNotes', notes);

However, here are some problems with events:

* It’s hard to track where they are being handled, and hard or impossible to override behavior of other event handlers if you *do* want to do that. This is a big deal for us; we extend and override `callAll` methods on purpose all the time.
* Another objection used to be no support for async. However a modern implementation of events could support returning promises from event handlers, and resolving them before the next handler is run.

Recently we worked out a strategy to solve the first issue. This is significant because we've been puzzling over whether we can do better for a long time now and the richer Apostrophe becomes, the more these conflicts occur.

[...]

PR #1531, opened 2018-08-07
1415 promise events

boutell:
Implements #1415 

Adds one feature: a shortcut to add a method while registering it as an event handler. 

See the tests for examples of use.


boutell:
Assigned to the enterprise team because this can't be one I just run by a single developer who doesn't spend time in the back end of Apostrophe.

lib/modules/apostrophe-module/index.js
```diff
@@ -681,5 +681,87 @@ module.exports = {
       });
     }
 
+    self.eventHandlers = {};
```
austinstarin: dead code?

lib/modules/apostrophe-module/index.js
```diff
@@ -681,5 +681,87 @@ module.exports = {
...
+    // "Why can't I use a methodName that is identical to
+    // the event name?" Doing so sets you up for an accidental
+    // collision with other event handlers in subclasses.
+    // Your method name should describe what your method does
+    // in response to the event.
+
+    self.on = function(name, methodName, fn) {
+      var moduleName, eventName, index;
+      index = name.indexOf(':');
+      if (_.camelCase(name) === _.camelCase(methodName)) {
```
austinstarin: Test needs to happen on name not module part.

lib/modules/apostrophe-module/index.js
```diff
@@ -681,5 +681,87 @@ module.exports = {
...
+        self[methodName] = fn;
+      }
+      eh[eventName] = eh[eventName] || [];
+      if (_.find(eh[eventName], function(item) {
+        return (item.moduleName === self.__meta.name) && (item.methodName === methodName);
+      })) {
+        // The "event name and method name must differ" rule helps
+        // prevent this situation, but it is still possible and a
+        // warning is in order
+        self.apos.utils.warn('The method ' + methodName + ' has been registered twice for the event\n' + eventName + ' in the module ' + self.__meta.name + '. It will only be called once.\n\nYou should rename the method in your code to differ from that in the module\nit extends so they both run.\n\nIf you are intentionally overriding it, assign a new function to\nself.' + methodName + ' instead.');
```
austinstarin: Would not be invoked once, but twice.

austinstarin:
Should we require innovations of this? You need to be able to do `pageBeforeSend` in this new style at the very least...

boutell:
@austinstarin please see the new commit in which I address the concerns raised in the review, including implementing actual emitting of the events for all of the current callAll situations and a unit test demonstrating that you can listen to those.

boutell:
Example from the tests:

```
      modules: {
        'test1': {
          alias: 'test1',
            self.on('apostrophe-docs:beforeInsert', 'beforeInsertReverseTitle', function(req, doc, options) {
              if (doc.type === 'default') {
                return Promise.delay(50).then(function() {
                  doc.title = doc.title.split('').reverse().join('');
                });
              }
            });
            self.on('apostrophe:modulesReady', 'modulesReadyCoreEventsWork', function() {
              coreEventsWork = true;
            });
            // Old school callAll works too (for now)
            self.docBeforeInsert = function(req, doc, options, callback) {
              doc.oldSchool = true;
              return setImmediate(callback);
            };
          }
        },
      }
```

Use of Promise.delay is a contrived way of proving you can do async stuff and it’ll wait for you before continuing.

Note the `apostrophe:modulesReady` example. A few callAlls were coming from the Apostrophe object itself. After staring at that for a bit, I concluded that the `apos` object simply needs to be capable of emitting promise events too, and its namespace is of course just `apostrophe`.

Also, a thought experiment: if I teach `apostrophe-pieces` at project level to emit an event, and I listen for that event, I won't "hear" it unless I listen like this...

```
// one of my piece types
apos.on('people:coolEvent', ...)
// another one
apos.on('trucks:coolEvent', ...)
```

Etc.

I think this is fine because sometimes people extend modules they also use directly, and that would lead to confusion and no way to tell the events apart if our handlers also fired for parent classes.

However, nothing stops us from introducing a syntax for that later if it's desired.

PR #1534, opened 2018-08-07
Feature/pieces placeholder

abea:
For #1528 (Add a class to apos-pieces-widgets when no matching content )

I tried to use mostly existing UI classes. If there are some that'd make sense for the background and border let me know. We could also move the styles out of the piece widget, but it doesn't seem like that's necessary yet.

screen shot 2018-08-07 at 2 14 18 pm
screen shot 2018-08-07 at 2 14 13 pm


boutell:
I like the general direction, I would suggest text like "mouse over and click the edit button to select some [label of piece type here]."

boutell:
Bump... any thoughts on my suggestion?

abea:
I think that's a good idea. I'll get an update back up here soon.

abea:
Finally got back to this. I noticed the same problem will persist with the blog widget template. Should I repeat this there?

boutell:
I would imagine this will have little impact on existing sites, because everyone overrides the template, but in future they will hopefully copy it with this nice placeholder as a starting point.

PR #1540, opened 2018-08-09
Enable showFields on checkboxes

falkodev:
In a schema, allow checkboxes to have a `showFields` property:
```
{
  name: 'awesomeCheckboxes',
  label: 'Awesome Checkboxes',
  type: 'checkboxes',
  choices: [{
    label: 'Choice 1',
    value: 'choice1',
  }, {
    label: 'Choice 2',
    value: 'choice2',
    showFields: ['choice2Options']
  }],
}, {
  name: 'choice2Options',
  label: 'Options for choice 2',
  type: 'string',
  required: true
}
```

The `choice2Options` field will be displayed only if `choice2` is checked in `awesomeCheckboxes`, and it will be required only if displayed (as in this example the field is required).

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -242,7 +242,7 @@ module.exports = {
 
         // Extra validation for select fields, TODO move this into the field type definition
 
-        if (field.type === 'select') {
+        if (field.type === 'select' || field.type === 'checkboxes') {
           _.each(field.choices, function(choice) {
             if (choice.showFields) {
               if (!_.isArray(choice.showFields)) {
--- a/lib/modules/apostrophe-schemas/public/js/user.js
+++ b/lib/modules/apostrophe-schemas/public/js/user.js
@@ -583,6 +583,12 @@ apos.define('apostrophe-schemas', {
                 show = true;
               }
             }
+          } else if (field.type === 'checkboxes') {
+            _.each($field || [], function(checkbox) {
+              if (checkbox.checked && checkbox.value === choice.value && choice.showFields) {
+                show = true;
+              }
+            });
           } else {
             // type select
             if (val === choice.value) {


boutell:
I believe the `isVisible` method of `apostrophe-schemas` will also need an update to be compatible. This does server-side checks of whether `required` is truly required or not when it's applied to a field that might be hidden. Otherwise you get stuck on a validation error you can't get past.

Thanks for working toward this!

houmark:
This is great stuff. On a related note; Should `boleean` also be able to show a field when true?

falkodev:
I can try for the boolean type. But what if the user wants to show some fields when it is true, but other fields when it is false? Shall we create a `showFieldsWhenTrue` and `showFieldsWhenFalse` methods as well?

I will take care of the `isVisible` method too, as requested by Tom.

________________________________

houmark:
I think `showFieldsWhenTrue` and `showFieldsWhenFalse` could work okay. 

Another direction is an object `showFields` on root level:

```javascript
{
  name: 'myField',
  type: 'boolean',
  showFields: {
    true: ['fieldOne', 'fieldTwo'],
    false: ['fieldThree']
  }
}
```

I personally like this one more because it uses the same name as for `choices` and true/false is self explanatory.

 Maybe @boutell has opinions about which way to go and naming?

falkodev:
I lke your idea too. Waiting for @boutell's reply.

About the `isVisible` method, I tried something but we only have the current field. How is it possible to know if this field is part of a hidden parent field?
Or maybe I did not understand correcty how to update this method.
________________________________
De : houmark <notifications@github.com>
Envoyé : samedi 11 août 2018 03:40
À : apostrophecms/apostrophe
Cc : Falkodev; Mention
Objet : Re: [apostrophecms/apostrophe] Enable showFields on checkboxes (#1540)


I think showFieldsWhenTrue and showFieldsWhenFalse could work okay.

Another direction is an object showFields on root level:

{
  name: 'myField',
  type: 'boolean',
  showFields: {
    true: ['fieldOne', 'fieldTwo'],
    false: ['fieldThree']
  }
}

I personally like this one more because it uses the same name as for choices and true/false is self explanatory.

Maybe @boutell<https://github.com/boutell> has opinions about which way to go and naming?

falkodev:
Ok for `showFieldsIfTrue` and `showFieldsIfFalse`.

About `isVisible`, yes, I have looked at the existing method. It seems to work ok, even with my modifications. But, I am not sure if I am missing/forgetting something (that was the meaning of my previous message).

________________________________
De : Tom Boutell <notifications@github.com>
Envoyé : lundi 13 août 2018 14:20
À : apostrophecms/apostrophe
Cc : Falkodev; Mention
Objet : Re: [apostrophecms/apostrophe] Enable showFields on checkboxes (#1540)


I like showFieldsIfTrue and showFieldsIfFalse best (note "if" rather than "then" for consistency with a few other things around Apostrophe).

Re: isVisible, have you looked at the existing method? It takes the entire schema and computes the visibility of everything. Just needs to be able to take into account your new stuff. apostrophe-schemas/index.js, line 511.

boutell:
I don't see how it could be working OK, given that the current implementation is only written for single select. See the tests of such cases at the end of `schemas.js`.

falkodev:
Hi,

I checked for having `showFieldsIfTrue` and `showFieldsIfFalse` but it added too much complexity to `self.enableShowFields` in `user.js`.

My proposal is to have this structure for `boolean` in a schema: 
```
{
  "name": "awesomeBoolean",
  "label": "Awesome Boolean",
  "type": "boolean",
  "choices": [
    {
      "value": true,
      "showFields": ["otherField1"]
    },
    {
      "value": false,
      "showFields": ["otherField2"]
    }
  ]
}
```

This way, it works with the existing code (there is already a condition for the `boolean` type in `self.enableShowFields` in `user.js`).

If you prefer to have `showFieldsIfTrue` and `showFieldsIfFalse` anyway, I will change the code. Let me know.

@boutell Your remark about tests pointed me to the right direction. I updated the `isVisible` method and added some tests for checkboxes and boolean.

falkodev:
I also removed one unused line in `user.js` (line 579).

falkodev:
Hi,

Is anything needed in this PR?

PR #1596, opened 2018-08-14
helper functions to prepend and append to the head, the body, the con…

boutell:
…text menu, etc. in a standardized way from modules that need to do that without making devs hack up their templates

### DIFF
--- a/test/templates.js
+++ b/test/templates.js
@@ -24,7 +24,16 @@ describe('Templates', function() {
         'express-test': {},
         'templates-test': {},
         'templates-subclass-test': {},
-        'templates-options-test': {}
+        'templates-options-test': {},
+        'apostrophe-pages': {
+          park: [
+            {
+              title: 'With Layout',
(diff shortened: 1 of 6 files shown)


test/templates.js
```diff
@@ -81,17 +90,46 @@ describe('Templates', function() {
...
+    var bodyIndex = result.indexOf('<body');
+    var afterContextMenu = result.indexOf('<h4>After the Context Menu</h4>');
+    assert(titleIndex !== -1);
+    assert(beforeTestIndex !== -1);
+    assert(afterTestIndex !== -1);
+    assert(bodyIndex !== -1);
+    assert(afterContextMenu !== -1);
+    assert(beforeTestIndex < titleIndex);
+    assert(afterTestIndex > titleIndex);
+    assert(afterTestIndex < bodyIndex);
```
bobclewell: A nifty way to test this. I like it!!

PR #1603, opened 2018-08-18
Add rich text styling for most common tags in Apostrophe modal rich text editor

### LINKED ISSUE #1599: Preview issue with rich-text-widgets in pieces modal
Expected behavior:
1. Applying styles (bold, italic, custom) to content in a rich-text-widgets editor show in preview

Actual behavior:
1. No matter what styles are applied, content appears plain text thereby making it hard for users to see what the end result will be (See screenshot below). Note that if I include the rich-text-widgets on a page, then bold and italic and link previews will work but custom styles and bulletedlist don't. 

Pieces screenshot:
image

Pages screenshot:
image

Relevant code:
In the piece definition (/lib/modules/articles/index.js), I have the following field definition:

`
{
      name: 'content',
      label: 'Content',
      type: 'singleton',
      widgetType: 'apostrophe-rich-text',
      options: {
        toolbar: [ 'Bold', 'Italic', 'Styles', 'BulletedList', 'Link', 'Unlink' ],
        styles: [
          { 
            name: 'Heading 1',
            element: 'h1'
          }, 
          { 
            name: 'Heading 2',
            element: 'h2'
          },
          { 
            name: 'Heading 3',
            element: 'h3'
          },
        ]
      }
    }
`

/lib/modules/apostrophe-rich-text-widgets/index.js

`
module.exports = {
  // The standard list copied from the module, plus sup and sub
  sanitizeHtml: {
    allowedTags: [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul', 'ol',
      'li', 'b', 'i', 'strong', 'em', 'strike', 'code', 'hr', 'br', 'div',
      'table', 'thead', 'caption', 'tbody', 'tr', 'th', 'td', 'pre',
      'sup', 'sub'
    ],
    allowedAttributes: {
      a: [ 'href', 'name', 'target' ],
      // We don't currently allow img itself by default, but this
      // would make sense if we did
      img: [ 'src' ]
    },
    // Lots of these won't come up by default because we don't allow them
    selfClosing: [ 'img', 'br', 'hr', 'area', 'base', 'basefont',
      'input', 'link', 'meta' ],
    // URL schemes we permit
    allowedSchemes: [ 'http', 'https', 'ftp', 'mailto' ],
    allowedSchemesByTag: {}
  }
};
`

/lib/modules/apostrophe-pages/views/pages/articlepage.html

`
{% extends "layout.html" %}

{% block main %}
  <div class="main-content">
    {{ apos.singleton(data.page, 'textBlock', 'apostrophe-rich-text', {
      toolbar: [ 'Bold', 'Italic', 'Styles', 'BulletedList', 'Link', 'Unlink' ],
        styles: [
          { 
            name: 'Heading 1',
            element: 'h1'
          }, 
          { 
            name: 'Heading 2',
            element: 'h2'
          },
          { 
            name: 'Heading 3',
            element: 'h3'
          }
        ]
    }) }}
  </div>
{% endblock %}
`

houmark:
This PR adds rich text editor styling for the most common tags in the Apostrophe modal rich text editor.

It seems these were forgotten, leading to random styling depending on how the project was styled. This PR adds basic support for the most common tags.

Tags like H1, H2 is not included as they are not considered by me to be basic and those tags should be added using the `styles` option of the rich text editor.

To test use the following configuration on a field:

```javascript
options: {
  toolbar: [ 'Styles', 'Bold', 'Italic', 'Link', 'Unlink', 'BulletedList', 'NumberedList' ]
}
```

I will post a screenshot to display the various styles based on the improvements introduced in this PR.

And yes, underline is not there, because the underline tag is no included html 5, and ckeditor does not even allow adding a underline button to the editor (can be done using a custom style).

Colors of `a` tags are inherited, but could be a fixed blue (or green) if someone thinks that would be better?

Fixes #1599.

### DIFF
--- a/lib/modules/apostrophe-rich-text-widgets/public/css/components/rich-text-widget.less
+++ b/lib/modules/apostrophe-rich-text-widgets/public/css/components/rich-text-widget.less
@@ -35,3 +35,28 @@
 }
 
 [data-apos-widget-wrapper="apostrophe-rich-text"]:hover { z-index: @apos-z-index-6; }
+
+.apos-modal-body [data-rich-text]
+{
+  p {
+    margin: 10px 0;
+  }
+  a {
+    color: inherit;
+    text-decoration: underline;
+  }
+  b, strong {
+    font-weight: bold;
+  }
+  em, i {
+    font-style: italic;
+  }
+  ul {
+    margin-left: 30px;
+    list-style-type: disc;
+  }
+  ol {
+    margin-left: 30px;
+    list-style-type: decimal;
+  }
+}


houmark:
This is the editor after applying the styles:

image 2018-08-18 at 4 33 09 pm

houmark:
I'd say that this for me is considered a bug. Using the rich text editor in a modal with buttons and functionality facilitated by Apostrophe core but with none of those buttons apparently doing any styling. If this were not to make it into the core, as simple as basic as it is, then the getting started guides should include information about adding this to the project site, but that I think is very repetitive.

Docs could explain that there's the most basic styles included, and if one wants to overwrite or even use the same font and general styling as the site, then it would be possible to overwrite.

As a fairly new user to Apostrophe, I myself thought this was a bug and weird behavior, but in the pace of getting other stuff done, I did a project based overwrite with a note to look into why no styling was added (maybe I was overwriting by mistake somewhere?) but when it was brought up by someone else I realized that this was "intended" and no styles are added by Apostrophe.

boutell:
I don't really disagree, it's not very newcomer friendly not to have this!

stuartromanek:
I struggle with this because applying these rules does not reflect project-level rules for these elements .. on one hand these widget toolbar buttons are broken in a modal without some boilerplate but on the other a developer has to make decisions about what a list looks like on the page, that does not come baked in (should it?)

This feels like a good first step.

An aside: How are new apostrophe developers styling rich text? There are a lot of entry points and we should try to understand what people  are doing out of the gate.

houmark:
OK, so if we are not disagreeing between us, LGTM? :)

@stuartromanek about your question. I can only answer for me and I think I did mostly in my prior comment but here's my experience getting started with Apostrophe in relation to styling:

* I had looked into Bulma before and decided to use Bulma as my first project together with Apostrophe as my first project.
* Bulma does most of the styling as any other frame work, but I did need to add a `.content` class to make it possible to style rich text for the client side part.
* I got confused when trying to use the WYSIWYG because *nothing* happened when clicking buttons in the editor. Was looking into configuration, did I miss something, did I do something wrong?
* After some looking around (and maybe I am less scared to look into the core than some other new users) I figured that Apostrophe core simple does not concern itself with styling modal rich text areas, or I had somehow overwritten that with a too wide CSS selector somewhere.
Frantic to move on and get things "working", I added most of the styles that became this PR, just in my site link colors and with a nicer underline as a border, because I generally like that more.

My proposal is to add a section around the rich text editor configuration in the documentation that explains that Apostrophe does basic styling in the modal rich text areas, and that this can be overwritten by doing something like `.apos-modal ..... full explantation of powerful overwrite CSS selectors` so editors who cares about that the link style for example, or UL's are the same style as the site style, can overwrite.

stuartromanek:
@boutell would this kind of overview belong in the How to config CKE (https://apostrophecms.org/docs/tutorials/howtos/ckeditor.html) or in its own how-to?

boutell:
`apostrophe-boilerplate` should ship with reasonable styles for elements that are enabled in the rich text editor, by default, so that new projects made with our CLI behave reasonably. This includes H3 and H4. But I don't see any reason it shouldn't.

The question then becomes whether we do the same in the core when the rich text editor is inside apos-ui. I suggest we do not. Instead, when we're telling people all about the CLI and the benefits of using the boilerplate project, we mention that this is one of the things they won't otherwise have "out of the box."

boutell:
Scoped to the modal body as this is, I agree it makes sense. Thanks!

PR #1604, opened 2018-08-19
better string concat, validates

stuartromanek:


stuartromanek:
addresses https://stackoverflow.com/questions/51739344/css-errors-in-apostrophe-ui-best-approach-to-solve-it

boutell:
LGTM

PR #1614, opened 2018-08-30
Added apos.global.whileBusy(fn) method

boutell:
This method allows you to run a function to completion while the entire site is paused. Obviously this should be super rare, used for things like deploying entirely new content to the site, and would have a hideous performance impact if used routinely. But it's very useful for a big operation like replacing all the content without a period of inconsistent results (as requested in DGAD-449).

HOW YOUR FUNCTION GETS RUN

A lock is obtained, the `global` doc is marked busy, and then requests in progress are waited out with a simple timer rather than trying to actually track them because the performance cost of tracking them across load-balanced cores and servers would be too high.

Next your function runs. If it returns a promise the promise is awaited. This is a new API and supporting both can be complicated, so callbacks are not an option.

After your function runs, the global doc is marked not busy, and the lock is released.

MEANWHILE

When any request comes in, we check to see if the global doc (which we always fetch anyway) is marked busy.

If it is, we wait until we can get the lock. If we get it, this means the function passed to `whileBusy` has simply finished, or the lock expired due to that function crashing. Just in case of the latter, we clear the busy flag in the global doc, which we can do because we hold the lock at this point. Then we release the lock and go on with our lives.

If we can't get the lock after several seconds, i.e. we're in danger of hitting the proxy server's time limit, and it's a GET/HEAD request, we send the user a simple 503 page with a "Refresh" header, lather rinse repeat until they get the page they were hoping for. Other request methods get a simple 503.

See unit tests in `test/global.js`.

### DIFF
--- a/lib/modules/apostrophe-global/index.js
+++ b/lib/modules/apostrophe-global/index.js
@@ -199,57 +199,84 @@ module.exports = {
 
       function findGlobal(callback) {
         return self.findGlobal(req, function(err, result) {
+          var lockName = 'apostrophe-global-busy';
+          var localeLockName;
+          var propName = 'globalBusy';
+          var localePropName;
           if (err) {
             return callback(err);
           }
+
           req.data.global = result;
 
-          if (req.data.global.globalBusy) {
-            return self.apos.locks.lock('apostrophe-global-busy', {
-              wait: (req.res && req.res.send) ? self.options.whileBusyRequestDelay * 1000 : Number.MAX_VALUE,
-              waitForSelf: true
-            }, function(err) {
-              if (!err) {
-                // We got in after a period of the system being busy,
-                // or the process with the lock went away. Either way,
-                // we should clear globalBusy and give up our lock,
-                // then continue normal operation.
-                return async.series([ unmark, unlock, refind ], callback);
-                function unmark(callback) {
-                  return self.apos.docs.db.update({
-                    _id: req.data.global._id,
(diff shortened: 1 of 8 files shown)


lib/modules/apostrophe-global/index.js
```diff
@@ -209,6 +261,71 @@ module.exports = {
...
+    // this method for anything more routine would have a crippling
+    // performance impact.
+
+    self.whileBusy = function(fn) {
+      var locked = false;
+      var marked = false;
+      return Promise.try(function() {
+        return self.apos.locks.lock('apostrophe-global-busy');
+      }).then(function() {
+        locked = true;
```
bgantick: Need to add `multi: true` for all locales

lib/modules/apostrophe-global/index.js
```diff
@@ -209,6 +261,71 @@ module.exports = {
...
+        return self.apos.docs.db.update({
+          type: self.name
+        }, {
+          $set: {
+            globalBusy: true
+          }
+        });
+      }).then(function() {
+        marked = true;
```
bgantick: same - need to deal w/ locales

boutell:
Please review the latest update which addresses locale-specific locking properly. So tempted to just merge this but no, you should see that commit (:

bgantick:
Good to go once Travis is passing

boutell:
tnx I had to please eslint

PR #1617, opened 2018-09-04
autoReconnect should not be used with replica sets

boutell:


boutell:
See apostrophecms/apostrophe#1508 for discussion with Matt Broadstone ( @mbroadst ) from the MongoDB team.

For bc, we continue to use autoReconnect by default, however if the mongodb URI contains a comma in the hostname portion (the only way to connect to a replica set in the mongo 2.x driver),  we know the client will implement its built-in server selection behavior. autoReconnect just breaks that by forcing reconnection to the same node that just failed, which might never come back.

So in the presence of the comma, we clear the parameters intended for autoReconnect. Nothing else changes here.

See also corresponding PR to the apostrophe-db-mongo-3-driver module for cases where a replica set connection is being used with the newer 3.x MongoDB driver.

PR #1625, opened 2018-09-14
Update versions modal "Revert" labels to "Revert to"

fredrikekelund:
As discussed in #1515, "Revert" is somewhat ambigous compared to "Revert to" - does it mean "undo this change" or "go back to this change"? Well it does mean the latter, and labelling it "Revert to" clarifies that.

Also, I took the chance to run some more labels in there through the i18n `__` function.

Lastly, I added some additional pixels to the subtracted width of `.apos-changes-meta`, because in viewports wider than 1450px, `.apos-button--major` receives an increase in font-weight. This adds 1 additional pixel to the width of "Revert to" now that we have more characters in there, so we need to counteract that.

### DIFF
--- a/lib/modules/apostrophe-versions/public/css/user.less
+++ b/lib/modules/apostrophe-versions/public/css/user.less
@@ -25,7 +25,7 @@
     }
     .apos-changes-meta
     {
-      width: ~'calc(100% - 220px)';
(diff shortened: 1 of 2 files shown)


boutell:
Thanks.

PR #1635, opened 2018-09-21
refactored whileBusy so it can be optionally put in place as a separa…

boutell:
…te middleware function if heavy-duty full loading of the global doc is being cached or otherwise not 100% real and up to date. DGAD needs this because they load req.data.global from a cache, so it might be stale and not contain the very latest value of the `globalBusy` property. The alternative middleware still fetches the global doc but does so without any areas or joins so it's lightweight.


boutell:
I also threw in:

`APOS_LOG_ALL_QUERIES=1 node app`

As an overdue debugging convenience.

Issue #1111, opened 2017-11-14
Convert Apostrophe in-context admin UI from anchors to buttons


stuartromanek:
Nested anchors are not allowed, some real-world cases of folks doing that 

https://github.com/punkave/apostrophe/issues/797

fredrikekelund:
Thanks for alerting me to this existing issue, @stuartromanek! I'll see if I can write up a PR within a week or so

abea:
This might be resolved?

stuartromanek:
The main macros have been replaced and tested but there could be others in the wild. Closing

PR #1638, opened 2018-09-24
Make apostrophe-ui button macro render a <button>

fredrikekelund:
As discussed in #1111.

Previously it would render an `<a>`. `<button>`s are generally better suited for this purpose, but above all it solves the problem where `<a>` elements can't be nested.

@stuartromanek, if there are other places that need updating – I'm happy to address those as well!

### DIFF
--- a/lib/modules/apostrophe-ui/public/css/components/buttons.less
+++ b/lib/modules/apostrophe-ui/public/css/components/buttons.less
@@ -20,6 +20,8 @@
 
 .apos-button
 {
+  -moz-appearance: none;
+  -webkit-appearance: none;
   position: relative;
   .apos-inline-block;
   padding: @apos-padding-1 @apos-padding-2;
@@ -27,6 +29,7 @@
   .apos-button-border;
   // .apos-text-small;
   font-size: 14px;
+  line-height: inherit;
   letter-spacing: 1px;
   @media @apos-breakpoint-desktop-xl { .apos-text-small; }
   background-color: @apos-white;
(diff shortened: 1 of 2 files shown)


stuartromanek:
hey @fredrikekelund we have this on our calendar to review in a few projects that make custom use of the button UI to make sure it is backwards compatible.

Thanks for the PR!

fredrikekelund:
Great! :+1: It's a minor change, but happy to help push it through. Let me know if further changes are needed!

abea:
@fredrikekelund My primary concern would be that it previously used an `href` attribute with `options.url`. A button doesn't know how to use an `href`, so if Apos is using that natively right now it could be a problem. If Apos is using JS to hijack the anchor element behavior right now then this should at least be refactored to a data attribute instead.

That's one thing we'll plan to look at later in the week, but if you have already looked into that please let us know what your thinking was.

fredrikekelund:
@abea @stuartromanek I've updated the macro to be conscious of the `options.url` property! I've also removed the `href` attribute on the `<button>` element, which I overlooked previously. Hopefully it should be all good now!

abea:
Would it be over-engineering to avoid duplication with a conditional for the tag?

```
  {%- if options.url -%}
    {% set tag = 'a' %}
  {%- else -%}
    {% set tag = 'button' %}
  {%- endif -%}

<{{ tag }} {%- if options.url -%}href="{{ options.url }}"{%- endif -%} class="apos-button {{ classes }}" {{ options.attributes }} data-apos-{{ options.action | css }}{% if options.value %}='{{ options.value }}'{% endif %}><span class="apos-button-label">{{ __(label | d('')) }}</span>{% if options.icon %}<i class="fa fa-{{ options.icon }}"></i>{% endif %}</{{tag}}>
```

abea:
I think I prefer less code, but I'll leave that to @stuartromanek to make the call ([I'm a 3 out of 10](http://blog.capwatkins.com/the-sliding-scale-of-giving-a-fuck)). Overall this is a great improvement.

stuartromanek:
I prefer it the way @fredrikekelund has it now. Will get this merged.

Thanks again!

PR #1640, opened 2018-09-26
Returns title as part of apos.images.all and apos.images.first

ngranahan:
building off of PR #1600 returns title too


boutell:
👍

PR #1647, opened 2018-09-28
Need to wrap text in a paragraph tag when converting to rich text

bgantick:


boutell:
Is a `p` tag the most likely not to cause problems when it hits ckeditor?

bgantick:
`p` tag is the default I thought?

boutell:
It is yeah. OK.

boutell:
Hmm, I just reverted this because I realized it's a bc break. People may be using this as-is and adding their own wrapper tag. Now they will have two.

Please address by adding an optional second argument, an `options` object with a `wrapper` property that, if present, should be the wrapper tag.

PR #1660, opened 2018-10-04
allow direct import of CSS files via import flags of LESS

boutell:


boutell:
To use this, you would write:

`self.pushAssets('stylesheet', 'myfilename', { import: { inline: true } })`

Or, in the options of apostrophe-assets:

`stylesheets: [ { name: 'myfilename', import: { inline: true } } ]`

You could also set other import flags supported by LESS:

http://lesscss.org/features/#import-atrules-feature-import-options

`Not all are necessarily sensible or wise to use in conjunction with Apostrophe, but `inline` makes sense when you want no LESS parsing because you are pushing existing CSS that should not get compiled (even though you would be using a `.less` filename), and `css` might be interesting if you want css validation (?) but no LESS interpretation.

This comes up for certain clients when they need to push assets compiled by other tools like SASS that are sometimes complex enough to stump LESS even though LESS ought to accept all valid CSS.

PR #1670, opened 2018-10-16
Respect return values of event handlers in jQuery.fn.onSafe

fredrikekelund:
`jQuery.fn.onSafe` would previously not pass the return value of event handlers to jQuery. That meant that even if the event handler would `return false`, event.stopPropagation and event.preventDefault would not be called.

This came up as an issue for me after the recent merging of #1638, where I now had a set of .apos-buttons inside of an area, that was wrapped in an `<a>` element. Clicking the .apos-buttons would trigger a click on the link as well, and so the browser would navigate away from the page.


boutell:
Thanks!

PR #1671, opened 2018-10-16
ability to pass cookie options for the csrf cookie, such as secure: true

boutell:
This addresses security scanner concerns about the cookie being potentially sent in response to HTTP as well as HTTPS requests.

### DIFF
--- a/lib/modules/apostrophe-express/index.js
+++ b/lib/modules/apostrophe-express/index.js
@@ -147,6 +147,21 @@
 // is unnecessary because the password itself is unknown to the
 // third party site; it effectively serves as an XSRF token.
 //
+// You can also pass options to the Express [`res.cookie`](https://expressjs.com/en/api.html#res.cookie) call that sets the cookie:
+//
+// ```javascript
+// csrf: {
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-express/index.js
```diff
@@ -491,12 +506,12 @@ module.exports = {
         }
         // Reset the cookie so that if its lifetime somehow detaches from
         // that of the session cookie we're still OK
-        res.cookie(self.apos.csrfCookieName, token);
+        res.cookie(self.apos.csrfCookieName, token, (self.options.csrf && self.options.csrf.cookie) || {});
```
mtthwmnc: should this also be checking for `self.options.csrf.cookie.secure` per your comments above?

boutell: No, `secure` is only one example of the properties you can give to `cookie`. Our job just to get `cookie` to `res.cookie` as a third argument, they might add `secure: true` or some other thing or nothing and that's not our biz

mtthwmnc: @boutell ah, implementation example. noted! good to go ✅

Issue #1680, opened 2018-10-23
Switch of "a" to "button" broke drag and drop of widgets


boutell:

abea:
It looks like this might be due to the default behavior of the button. [Suggestion here to use preventDefault to allow for dragging](https://stackoverflow.com/a/6875184/888550).

I'm also seeing suggestions of using `.draggable({cancel:false});` in that SO and [the jquery UI forum](https://forum.jquery.com/topic/draggable-button).

Also, an interesting demo for accessible drag-drop linked from the w3c: https://dev.opera.com/articles/accessible-drag-and-drop/example.html, maybe more for 3.x, phase 2.

PR #1682, opened 2018-10-24
Hotfix/drag btns

abea:
Disables canceling dragging/sorting in jQuery UI on drag buttons. Addresses #1680 

Ref:
- http://api.jqueryui.com/sortable/#option-cancel
- http://api.jqueryui.com/draggable/#option-cancel


abea:
I my testing I'm pretty sure the up/down were still working (do we need/have a manual regression test list?). But these are applied via the draggable/sortable methods, so shouldn't have any relationship to the clickable buttons, right?

boutell:
Thanks!

PR #1684, opened 2018-10-25
do not crash if format parameter to pieces list route is not present

boutell:
See #1683 for details.


boutell:
The cursor object was added to `result` late in the game after we'd stopped invoking this route without `format`, but there is supposed to be a fallback "provide all the info" format, and including cursor does not make sense sense for the reasons explained in #1683.

PR #1686, opened 2018-10-25
Adding field mixin

bobclewell:


lib/modules/apostrophe-schemas/src/apos/components/ApostropheFieldMixin.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
```
boutell: This should live in /src/apos/mixins, not /src/apos/components, use a relative path.

lib/modules/apostrophe-schemas/src/apos/components/ApostropheFieldMixin.js
```diff
@@ -0,0 +1,29 @@
...
+    field: Object
+  },
+  data() {
+    return {
+      next: (this.value.data !== undefined) ? this.value.data : (this.field.def || ''),
+      error: false
+    };
+  },
+  mounted() {
+    this.updateErrorAndEmit();
```
boutell: should be validateAndEmit.

lib/modules/apostrophe-schemas/src/apos/components/ApostropheFieldMixin.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+  watch: {
+    next(value) {
+      this.updateErrorAndEmit();
+    }
+  },
+  methods: {
+    updateErrorAndEmit() {
```
boutell: should be validateAndEmit.

lib/modules/apostrophe-schemas/src/apos/components/ApostropheFieldMixin.js
```diff
@@ -0,0 +1,29 @@
...
+      const value = this.validate();
```
boutell: this.validate does not return anything, do not modify value here... you will just pass on this.value... this.validate sets this.error.

Maybe it should return an error (or false) and you should assign it to this.error here though.

lib/modules/apostrophe-schemas/src/apos/components/ApostropheStringField.vue
```diff
@@ -8,28 +8,13 @@
...
-    this.updateErrorAndEmit();
-  },
-  watch: {
-    next(value) {
-      this.updateErrorAndEmit();
-    }
-  },
   methods: {
-    updateErrorAndEmit() {
+    validate() {
```
boutell: this should return error (or false) rather than setting this.error directly.

bobclewell: Returns `this.error`. See below.

lib/modules/apostrophe-schemas/src/apos/mixins/ApostropheFieldMixin.js
```diff
@@ -0,0 +1,28 @@
...
+  },
+  watch: {
+    next(value) {
+      this.validateAndEmit();
+    }
+  },
+  methods: {
+    validateAndEmit() {
+      this.$emit('input', {
+        data: this.validate(),
```
boutell: Should return this.next

lib/modules/apostrophe-schemas/src/apos/mixins/ApostropheFieldMixin.js
```diff
@@ -0,0 +1,28 @@
...
+        error: this.error
```
boutell: should be error: this.validate()

lib/modules/apostrophe-schemas/src/apos/components/ApostropheStringField.vue
```diff
@@ -8,28 +8,13 @@
...
       this.error = false;
```
boutell: should never modify this.error, should just return the desired value for it

PR #1689, opened 2018-10-26
concats rather than replaces html-widget fields

abea:
I realize this risks a BC break, but it seems like an oversight. Was there a reason that the HTML widget doesn't concat its `addFields` with the `options.addFields` now?

I think this might only be a BC break if someone has an HTML widget extension that intentionally leaves out the `code` field for the embed code. Which would be odd.


boutell:
It's a bug fix, and our semantic versioning policy specifically permits that. I don't think this will irritate anyone. Thanks.

PR #1694, opened 2018-10-30
Ui/area controls compete with add content

stuartromanek:
prevents this insanity insane


lib/modules/apostrophe-areas/public/js/user.js
```diff
@@ -45,7 +45,7 @@ apos.define('apostrophe-areas', {
           // an add content menu is active, we have UX problems (and
           // don't pass our tests) on smaller screens otherwise. -Tom
           $('.apos-context-menu-container').addClass('apos-content-menu-active');
-
+          $area.data('editor').disableAreaControls()
```
boutell: eslint error due to missing semicolon. Please enable eslint in your editor, or before submitting, just run:

```
npx eslint .
```

Or even better:

```
npx eslint . --fix
```

PR #1699, opened 2018-11-06
video field converter should not crash if value is missing, also impl…

boutell:
…ement required flag


abea:
I'm not sure how I should be testing this. I set a piece schema field to `video` and not required, but the field still requires a value. I might not be understanding how this is intended to work.

abea:
Okay, this looks good then. On a process point, what should we expect as far as tests being added for things like this?

boutell:
@abea in this case I was fixing a crashing bug, and manually tested the case where it didn't crash before. (Our test suite does that too because it includes video widgets I believe.)

Is that ideal? Well... it's not really no. I should attempt to add a test of the behavior I fixed. I will try to slip that in before merging.

PR #1709, opened 2018-11-19
feat(date): add locale parameter to date filter

sambadevi:
default locale is `en`, date filter can be called as `date(format, locale)`

### DIFF
--- a/lib/modules/apostrophe-templates/index.js
+++ b/lib/modules/apostrophe-templates/index.js
@@ -418,7 +418,7 @@ module.exports = {
         if (!date) {
           return '';
         }
-        var dateLocale = locale || 'en';
+        var dateLocale = locale || self.apos.templates.contextReq.locale;
         moment.locale(dateLocale);
         var s = moment(date).format(format);
         return s;
--- a/lib/modules/apostrophe-templates/index.js
+++ b/lib/modules/apostrophe-templates/index.js
@@ -419,7 +419,7 @@ module.exports = {
           return '';
         }
         moment.locale(locale);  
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-templates/index.js
```diff
@@ -411,13 +411,14 @@ module.exports = {
       // Format the given date with the given momentjs
       // format string.
 
-      env.addFilter('date', function(date, format) {
+      env.addFilter('date', function(date, format, locale='en') {
```
boutell: Please avoid ES6 in this version of Apostrophe, you can write:

`locale = locale || 'en'`

Inside the function.

sambadevi: Added the requested changes

boutell:
Much better. How about defaulting to `req.locale` if it is set? You can access `apos.templates.contextReq` in this situation.

boutell:
`self.apos.templates.contextReq` rather.

sambadevi:
Seems like `master` was updated in the last few days. I updated the PR accordingly

sambadevi:
To be honest I don't see any additional options to pass to the filter. And I don't really see any benefit in adding an options object just for two possible parameters (date format and date locale) especially if one of the parameters is just optional (locale).

fredrikekelund:
Hmm, as I understand it, calling `moment.locale()` changes the moment locale globally and permanently. But calling `moment().locale()` changes it only for the current instance ([docs](https://momentjs.com/docs/#/i18n/instance-locale/)). Perhaps that would be more appropriate in this situation?

PR #1712, opened 2018-11-26
Bless relationship nested objects

falkodev:
In a `joinByArray` field, a "relationship" property cannot have a nested object. 

For example, this field will trigger the warning ("unblessed array") and will not save the "relationship" data:
```
{
  name: '_whatever-join',
  type: 'joinByArray',
  filters: {
    projection: {
      title: 1,
      slug: 1
    }
  },
  relationship: [
    {
      name: 'nestedStructure',
      label: 'Nested Structure',
      type: 'array',
      schema: [
        {
          name: 'nestedField1',
          label: 'Nested Field 1',
          type: 'string',
          required: true
        },
        {
          name: 'nestedField2',
          label: 'Nested Field 2',
          type: 'string',
          required: true
        }
      ]
    }
  ]
}
```

This is because the `nestedStructure` array has not been blessed. To overcome this problem, I propose the following solution: bless every nested property in the `joinByArray` definition in `apostrophe-schemas`.

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -2014,6 +2014,11 @@ module.exports = {
       },
       bless: function(req, field) {
         self.apos.utils.bless(req, field, 'join');
+        _.each(field.relationship, function (elem) {
+          if (elem.schema) {
+            self.bless(req, elem.schema);
+          }
+        });
       },
       join: function(req, field, objects, options, callback) {
         return self.joinDriver(req, joinr.byArray, false, objects, field.idsField, field.relationshipsField, field.name, options, callback);


falkodev:
Hi @boutell,

I will fix this linter error.

Regarding `apos.schemas.bless`, I tried `self.apos.utils.bless(req, field.relationship, 'array');` and a few other things like `self.apos.utils.bless(req, field.relationship, 'object');` or `self.apos.utils.bless(req, field.relationship, 'joinByArray');` without success.

Any clue about what is failing?

On the other hand, this is working:
```
        _.each(field.relationship, function (elem) {
          self.apos.utils.bless(req, elem, elem.type);
        });
```

boutell:
You do not mention schemas.bless, which you are re-creating (:

boutell:
Thanks!

Issue #1717, opened 2018-11-30
Spinner appears forever when you attempt to save an object field with an empty required subfield


boutell:

boutell:
Fixed and in master (not released yet).

PR #1718, opened 2018-11-30
field types may optionally supply a showError method to intercept the…

boutell:
… normal error display process, providing a chance to show an error in a nested field of an object, for instance. Implementation for the object field type. Arrays do not need one because you cannot save the array modal with errors.


boutell:
Resolves #1717

PR #1719, opened 2018-11-30
no more relationship subform for permissions, single "copy all permis…

### LINKED ISSUE #1413: It's unclear if setting user edit permissions worked
In the Permissions tab of a page, when you use the "These Users can Edit" field to add editors for the page, it's unclear what happens when you also enable "Apply to Subpages." When an admin revisits that field it's reset back to "No."

I understand that this is an action taken that updates the pages lower in the page tree and not saved on the piece. Maybe that should change? Regardless, some indication of what actually happens would resolve confusion. I'm thinking either:

1. The option is saved to the page's document, recording that they have lower page permissions. The big challenge with this seems to be that if editing lower pages is manually unset, it could be a heavy task to update this page (though I can imagine how that'd work). This would be ideal so that editors can be confident that the right permissions are still in place later on.
2. A notification pops up on page refresh with something like "Editing permissions for [username] have been enabled for subpages."
3. Probably paired with the second option, add help text to that "Apply to Subpages" field indicating this is a one-time action updating child pages with  each use.

boutell:
…sions to subpages" action, no confusing sorting or editing controls while trying to pick people and groups, no irrelevant "draft/live" toggle for the "copy" action

Resolves #1413

### DIFF
--- a/lib/modules/apostrophe-pager/public/css/user.less
+++ b/lib/modules/apostrophe-pager/public/css/user.less
@@ -29,3 +29,7 @@
   }
 
 }
+
+.apos-field-apply-to-subpages .apos-workflow-field-controls {
+  display: none;
+}
--- a/lib/modules/apostrophe-doc-type-manager/index.js
+++ b/lib/modules/apostrophe-doc-type-manager/index.js
@@ -67,14 +67,18 @@ module.exports = {
         type: 'joinByArray',
         withType: 'apostrophe-user',
         label: 'These Users can View',
-        idsField: 'viewUsersIds'
+        idsField: 'viewUsersIds',
+        sortable: false,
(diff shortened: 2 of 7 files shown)


boutell:
See #1413 for screenshot of new, simplified page permissions interface 🎉

lib/modules/apostrophe-pager/public/css/user.less
```diff
@@ -29,3 +29,7 @@
   }
 
 }
+
+.apos-field-apply-to-subpages .apos-workflow-field-controls {
```
abea: Can we get a comment explaining this? It's not clear why we're removing something from view. If this is specific to sites using apos-workflow then this should be there, not in core.

PR #1723, opened 2018-12-05
contextual and skipInitialModal editing support for areas and singlet…

boutell:
…ons nested in fields of type object

### DIFF
--- a/lib/modules/apostrophe-schemas/public/js/user.js
+++ b/lib/modules/apostrophe-schemas/public/js/user.js
@@ -456,11 +456,16 @@ apos.define('apostrophe-schemas', {
       };
     };
 
+    // Return a new object with all default settings
+    // defined in the schema
     self.newInstance = function(schema) {
       var def = {};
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -821,6 +829,52 @@ apos.define('apostrophe-schemas', {
         var nestedErrors = $fieldset.data('nestedErrors');
         $fieldset.data('nestedErrors', null);
         self.showError($fieldset, nestedErrors[0]);
+      },
+      contextualConvert: function(data, name, $el, field) {
```
agilbert: Where is this method called?

boutell: When serializing a widget to be saved in the area editor, this method is invoked for any schema properties that have it, allowing nested areas/singletons to get saved too. It's present for arrays but was missing for objects.

PR #1730, opened 2018-12-11
modulesReady for groups was overriding a universal modulesReady for a…

boutell:
…ll pieces that was introduced later, grr, this is one of the reasons why we will use async events exclusively for this kind of thing in 3.x


boutell:
This resulted in end users seeing "Batch Commit" options for groups when the workflow module was enabled, because the batch operations hadn't been filtered properly due to the lack of a call to this method.

PR #1732, opened 2018-12-12
apos.notify is available on the server

boutell:
With exactly the same syntax as in the browser, except with `req` or a user `_id` as first argument and an optional callback if you really need to be sure it has landed in the database.

Also apos.notify is now persistent for that user until they dismiss the notification (dismiss: true still works and does dismiss it permanently).

Persistence applies whether the notification was born in browser or server land.

Long polling is used for a prompt response although with some compromise to avoid beating up mongodb too much.

There have been no changes to the user experience other than persistence (refreshing the page does not make all notifications "dismiss", which happened only because we did not have a real implementation of keeping them around until they were properly dismissed).

This was implemented because debugging yet another quasi-implementation of it for yet another module (in this particular case, apostrophe-favicons for our demo) did not make sense. The server needs a simple way to send the user notifications, which we can improve and revise the UX of anytime — I'm just opting the server in to what the browser already has access to.

### DIFF
--- a/lib/modules/apostrophe-notifications/index.js
+++ b/lib/modules/apostrophe-notifications/index.js
@@ -1,28 +1,250 @@
-// This module provides a framework for triggering notifications within the Apostrophe admin UI.
+// This module provides a framework for triggering notifications
+// within the Apostrophe admin UI. Notifications may be triggered
+// either on the browser or the server side, via `apos.notice`.
+
+var _ = require('@sailshq/lodash');
+var Promise = require('bluebird');
 
 module.exports = {
   extend: 'apostrophe-module',
 
   construct: function(self, options) {
+
     self.pushAssets = function() {
       self.pushAsset('script', 'user', { when: 'user' });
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-notifications/index.js
```diff
@@ -1,28 +1,250 @@
...
+    self.route('post', 'trigger', function(req, res) {
+      var type = self.apos.launder.select(req.body.type, [ 'error', 'warn', 'success', 'info' ], 'info');
       var message = self.apos.launder.string(req.body.message);
       var strings = self.apos.launder.strings(req.body.strings);
-      var args = [ message ].concat(strings);
-      message = self.apos.i18n.__.apply(self.apos.i18n, args);
-      res.send(self.render(req, 'notification', { message: message }));
+      var dismiss = self.apos.launder.integer(req.body.dismiss);
+      var pulse = self.apos.launder.boolean(req.body.pulse);
+      // TODO what is this one for?
```
stuartromanek: did we figure out what this one is for

PR #1744, opened 2019-01-07
fix: change usage of momentjs to local instance instead of global

sambadevi:
A small addition to my previous PR

This changes the momentjs usage to avoid changing the global locale with the date filter. 

As mentioned in #1709


sambadevi:
Oops I didn't update my local branches, seems like this is already fixed.
Please ignore this PR

PR #1745, opened 2019-01-07
1728 undo remove widget

boutell:
@stuartromanek I am assigning review to you because it looks good in the sandbox but I can't figure out how to use apos-text-normal (?) without getting an undefined error. Scoped it in apos-ui and still got one of those. Not sure if I'm barking up the right tree to make sure these get a reasonable font in any project context.

I went with the original @carlykaras suggestion of apos-purple and avoided the notification styles because the border on those turns out to be jarring when in context in a stream of widgets. This is more of an inline thing than an "unrelated context layered on top" thing.

### DIFF
--- a/lib/modules/apostrophe-areas/public/css/user.less
+++ b/lib/modules/apostrophe-areas/public/css/user.less
@@ -1,7 +1,6 @@
 
 @apos-translucent-opacity: 0.4;
 
-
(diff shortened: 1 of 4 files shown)


stuartromanek:
@boutell this is all styled and @carlykaras approved

PR #1746, opened 2019-01-07
emit textchange too when overriding title due to first or last name c…

boutell:
…hange, fixes 1738

I had trouble reproducing this if I populated both first and last name before moving on, but it does happen if I populate only first name, or populate things in a weird order. This fix makes sure all the relevant events fire.


grdunn:
thanks @bgantick and @boutell

PR #1769, opened 2019-01-18
bless relationship in joinByArray

falkodev:
In https://github.com/apostrophecms/apostrophe/pull/1712, I intended to solve the problem of a nested structure not blessed. But I forgot one part in `joinByArray` bless function.

Here is the fix.


boutell:
Thanks!

PR #1775, opened 2019-01-25
fix missing "enableShowFields" for checkboxes

falkodev:
In the PR #1540 "Enable showFields on checkboxes", I forgot to add the line to hide/show fields in user.js for checkboxes.

Here is the fix.


boutell:
Thanks (again) (:

PR #1778, opened 2019-01-29
do not crash in the notification polling route if apos has been destr…

boutell:
…oyed

This is a heavily used, long-polling route that usually has a request going at the moment apos is destroyed by apostrophe-monitor, apostrophe-multisite, etc. Fix it by making it possible to detect that the database connection has been closed and return an empty response gracefully.


mtthwmnc:
Just needs a rebase with `master`

PR #1784, opened 2019-02-05
Fix 1783 - update apostrophe-pages api.js

Mayank694:
This commit updates the docAfterDenormalizePermissions
method and makes it propagate the docPermissions property
as part of the the “Apply to sub-pages” functionality.


boutell:
Thanks, this is a correct fix.

boutell:
Would you please pop me an email at tom@punkave.com? I'd like to hear more about what you're doing with Apostrophe on your team. Thanks!

PR #1785, opened 2019-02-06
Migration progress and fixes

boutell:
* Progress display for `apostrophe-migrations:migrate` when individual migrations take time. If node is not running in a TTY (an interactive terminal) this automatically goes away to avoid flooding log files with progress information or creating problems for people who have already scripted Apostrophe.
* Previously if a migration function returned a promise it was executed, but its success was not recorded, so it got re-run each time. This has been fixed.

### DIFF
--- a/lib/modules/apostrophe-migrations/lib/api.js
+++ b/lib/modules/apostrophe-migrations/lib/api.js
@@ -1,6 +1,7 @@
 var broadband = require('broadband');
 var async = require('async');
 var Promise = require('bluebird');
+var ProgressBar = require('progress');
 
 module.exports = function(self, options) {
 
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-migrations/lib/api.js
```diff
@@ -81,20 +87,66 @@ module.exports = function(self, options) {
...
+      var progress;
+
+      return async.series([ count, execute ], callback);
+
+      function count(callback) {
+        if (!progressDisplay) {
+          return callback(null);
+        }
+        if (progressDisplay) { 
+          console.error('Determining total...');
```
boutell: console.error because the progress display goes there too, leaving console.log clean.

lib/modules/apostrophe-migrations/lib/api.js
```diff
@@ -81,20 +87,66 @@ module.exports = function(self, options) {
...
-          }).catch(function(err) {
+
+      var progressDisplay = process.stdout.isTTY && (self.options.progressDisplay !== false);
+      var progress;
+
+      return async.series([ count, execute ], callback);
+
+      function count(callback) {
+        if (!progressDisplay) {
+          return callback(null);
```
boutell: In this case we don't need to calculate a percentage, and counting the relevant documents can sometimes take a few seconds in its own right.

PR #1790, opened 2019-02-10
automatically include idField or idsField in a reverse join, as appro…

### LINKED ISSUE #1710: JoinByArrayReverse not return an array if the field have projection.
In one piece I have this
```
{
      name: '_library',
      type: 'joinByArray',
      withType: 'library',
      label: 'Library',
      idsField: 'libraryIds',
      filters: {
        projection: {
          _id: 1,
          title: 1,
          description: 1,
          slug: 1,
          _url: 1,
          image: 1,
        }
      }
    }
```
and in another pieces have this:
```
    {
      name: '_book',
      type: 'joinByArrayReverse',
      withType: 'book',
      reverseOf: '_library'',
      // filters: {
      //   projection: {
      //     _id: 1,
      //     title: 1,
      //     description: 1,
      //     slug: 1,
      //     _url: 1,
      //     image: 1,
      //   }
      // }
    }
```
if you uncomment the filters, in data.piece._booknot return the array of _book, but if I comment the filters, return the array with all fields of object

boutell:
…priate, closes #1710

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -848,6 +848,21 @@ module.exports = {
           _.extend(options.hints, join.hints);
         }
 
+        // For reverse joins with a projection, it is common for
+        // developers to leave out the idField or idsField without
+        // which the join can't work, especially since `reverseOf`
+        // was introduced, making it less likely that the developer
+        // is thinking about these automatic properties
+
+        if (join.type === 'joinByOneReverse') {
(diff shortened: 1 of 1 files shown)


boutell:
This resolves issue #1710 which is probably the main reason people sometimes feel reverse joins are broken: they didn't automatically include the idsField necessary to actually complete the join. The need to do this is not at all clear to the developer since idsField is implicit these days. Easily solved by automatically adding it to the projection if there is one.

PR #1793, opened 2019-02-11
the home page works properly as a pieces page, with unit tests to pre…

boutell:
…vent regression

### DIFF
--- a/test/pieces-pages.js
+++ b/test/pieces-pages.js
@@ -50,15 +50,6 @@ describe('Pieces Pages', function() {
         }
       },
       afterInit: function(callback) {
-        console.log('>> afterInit');
-        // In tests this will be the name of the test file,
-        // so override that in order to get apostrophe to
-        // listen normally and not try to run a task. -Tom
-        apos.argv._ = [];
-        return callback(null);
-      },
-      afterListen: function(err) {
-        console.log('>> afterListen');
         // In tests this will be the name of the test file,
         // so override that in order to get apostrophe to
         // listen normally and not try to run a task. -Tom
--- a/test/pieces-pages.js
+++ b/test/pieces-pages.js
@@ -1,7 +1,7 @@
 var t = require('../test-lib/test.js');
 var assert = require('assert');
 var request = require('request');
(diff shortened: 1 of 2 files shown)


test/pieces-pages.js
```diff
@@ -50,6 +50,15 @@ describe('Pieces Pages', function() {
         }
       },
       afterInit: function(callback) {
+        console.log('>> afterInit');
```
bobclewell: Do you want to keep this log output around?

test/pieces-pages.js
```diff
@@ -50,6 +50,15 @@ describe('Pieces Pages', function() {
...
+        // In tests this will be the name of the test file,
+        // so override that in order to get apostrophe to
+        // listen normally and not try to run a task. -Tom
+        apos.argv._ = [];
+        return callback(null);
+      },
+      afterListen: function(err) {
```
bobclewell: Same function with the same name, right below (line 68)?

test/pieces-pages.js
```diff
@@ -162,4 +171,36 @@ describe('Pieces Pages', function() {
...
+    });
+  });
+
+  it('pieces-page as home page: correct permalinks on index page', function(done) {
+    return request('http://localhost:7900/', function(err, response, body) {
+      assert(!err);
+      // Is our status code good?
+      assert.equal(response.statusCode, 200);
+      // Only page one events should show up
+      assert(body.match(/event-001"/));
```
bobclewell: perhaps a silly question, but what is the unpaired double quote in this line and also in the line below?

test/pieces-pages.js
boutell: Oops, fixed.

test/pieces-pages.js
boutell: Oops, fixed (caused no problems because last one wins, but still).

test/pieces-pages.js
boutell: This one is OK, it is just that I don't care about matching the entire URL here.

PR #1794, opened 2019-02-11
the docIds and trashDocIds properties of attachments were wrong becau…

boutell:
…se they included docs merely referencing the attachment via joins. They now reference direct ownership only, and the migration that originally populated these has been renamed to run again and rewritten to be faster thanks to mongodb batch operations. But the really important line of code is the clonePermanent call to make sure self.all does not find any attachments in joined docs.

This fix ensures that images and files are made inaccessible when the last one goes in the trash, and also enables the new, optional apostrophe-secure-attachments module to calculate permissions properly.

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -688,7 +688,7 @@ module.exports = function(self, options) {
     // incorrect and leads to failure to make it
     // unavailable at the proper time
 
-    self.apos.migrations.add(self.__meta.name + '.docReferencesCorrected', function(callback) {
+    self.apos.migrations.add(self.__meta.name + '.docReferencesContained', function(callback) {
       var attachmentUpdates = {};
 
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -681,26 +681,31 @@ module.exports = function(self, options) {
...
   self.addDocReferencesMigration = function() {
-    self.apos.migrations.add(self.__meta.name + '.docReferences', function(callback) {
+
+    // This migration is needed because formerly,
+    // docs that only referenced this attachment via
+    // a join were counted as "owning" it, which is
+    // incorrect and leads to failure to make it
+    // unavailable at the proper time
+
+    self.apos.migrations.add(self.__meta.name + '.docReferencesCorrected', function(callback) {
```
abea: I suggest `docReferencesContained` or the like so as to not tie naming to whether it was or wasn't at one point working.

PR #1798, opened 2019-02-13
sped up "select all that match this search" by (1) preventing the eve…

boutell:
…nt from firing twice, and (2) eliminating expensive DOM manipulation for the vast majority of ids that are not part of the current page of results


grdunn:
@boutell sans those linter errors -- looks good

boutell:
tnx

PR #1809, opened 2019-02-26
if boolean schema field is required and false, throw error

stuartromanek:


lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -1176,7 +1176,7 @@ apos.define('apostrophe-schemas', {
       convert: function(data, name, $field, $el, field, callback) {
         data[name] = $field.val();
         // Seems odd but sometimes used to mandate an "I agree" box
-        if (field.required && !data[name]) {
+        if (field.required && (!data[name]) || data[name] === '0') {
```
boutell: I think you need more parenthesis grouping here, it's ambiguous what's going on between the `&&` and `||`.

boutell:
This change is correct in accordance with the documentation. However, I had a doubt while preparing for release, and it turns out that enterprise clients are using "required: true" incorrectly for booleans quite a bit in their code. So presumably others are as well.

I think we may be forced to revert to the existing semantics here in which required unfortunately isn't very useful for booleans.

Obviously the frontend code that doesn't really enforce this should at least go away so it's not confusing, and the documentation should be fixed. And we could introduce a `mandatory` flag to do what we used to claim it would do. ☹️ 

What do you think @stuartromanek?

stuartromanek:
A `mandatory` or similar flag would be much clearer

Issue #1810, opened 2019-02-26
minSize not working in apostrophe-images


mtthwmnc:
```
"apostrophe": "^2.79.0",
```

Image manager is bringing up images that do not meet the required `minSize`. Allows you to save the field and it renders an empty singleton.

boutell:
To be precise, it blocks me from uploading a too-small image, but it still lets me pick one that's already present. So it's not a complete failure to pass on the hints to the chooser. Hmm.

boutell:
See #1814

PR #1814, opened 2019-02-28
pass the minSize hint, and other hints, the modern way so we cannot p…

boutell:
…ick images the wrong size


boutell:
closes #1810

PR #1817, opened 2019-02-28
Add existence check when insensitive sorting

fawzisf:


fawzisf:
Yeah my bad forgot to run eslint, all well now

Issue #1821, opened 2019-03-05
def is not respected for properties of the global doc when starting up a new site, or when starting up with newly defined fields that have def


boutell:
When starting up a new site that has not contained a global doc before, the `def` property should be respected for properties of the global doc.

boutell:
This should also do the right thing when restarting with additional properties added that have a def, with or without workflow, and should consult getDefault if needed.

PR #1822, opened 2019-03-05
def property should be respected for fields of global doc, both at in…

boutell:
…sert time and if there are new fields at restart time


boutell:
Closes #1821

PR #1834, opened 2019-03-19
Select multiple checkboxes with shift key.

bharath6365:
Add the ability to select multiple checkboxes by holding on to left shift (Something like Gmail) on the Reorganize Pages Module.

### DIFF
--- a/index.js
+++ b/index.js
@@ -491,7 +491,7 @@ module.exports = function(options) {
   }
 
   function lint(s) {
-    self.utils.warn('\n⚠️  It looks like you may have made a mistake in your code:\n\n' + s + '\n');
+    self.utils.warnDev('\n⚠️  It looks like you may have made a mistake in your code:\n\n' + s + '\n');
   }
 
   function afterInit(callback) {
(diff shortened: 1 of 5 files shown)


boutell:
Nice! How about adding this to the manage modal of pieces as well while you're in the neighborhood?

Also, are you aware of the "select all" checkbox, and that after you do that it invites you to select *everything* matching the current filters? This isn't necessarily the same thing, I know; sometimes a person wants to select a lot of things, but not everything. Just checking.

bharath6365:
> Nice! How about adding this to the manage modal of pieces as well while you're in the neighborhood?
> 
> Also, are you aware of the "select all" checkbox, and that after you do that it invites you to select _everything_ matching the current filters? This isn't necessarily the same thing, I know; sometimes a person wants to select a lot of things, but not everything. Just checking.

Hey Boutell. Will add this functionality to manage modal of pieces as well and then I will notify you.
Yes I noticed the select all checkbox on the top left but like you said I wanted to select a lot of things but not everything :)

bharath6365:
Hey Tom. I will rise the changes in a new PR. Somehow I managed to mess with GIT History.

PR #1846, opened 2019-03-25
Active array item in last to be resetted when deleting other items

anwarhussain93:
**Steps to reproduce :-**

1. In schema, add a two item back to back in an array
2. Ensure the active item in an array to be a last item
3. Delete any one item in an array other than last item
4. Now click add new item in an array. It’ll throw an error and the whole saved items in an array will become lose.


boutell:
Thanks!

PR #1848, opened 2019-03-25
Adds select image exif data to the image document if available.

abea:
This adds select exif metadata to apostrophe-images when available. There's a ton that could be pulled from EXIF data, but I'm keeping it limited at the moment. Location was a good next candidate, but there are privacy implications to suddenly having that stored in newly updated Apostrophe sites.

Regarding the schema update, one option would be to start a new "Metadata" tab group, but "Info" currently overlaps with that pretty significantly.

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -127,6 +127,9 @@ module.exports = function(self, options) {
                 }
 
                 const data = exifData;
+
+                if (!data) { return callback(null); }
+
                 info.description = data.ImageDescription ? data.ImageDescription.trim() : null;
 
                 if (data.Make && data.Model) {
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -122,21 +122,21 @@ module.exports = function(self, options) {
             if (info.extension === 'jpg') {
               exif.parse(file.path, (err, exifData) => {
                 if (err) {
-                  console.log(err);
-                } else {
-
-                  const data = exifData;
-                  info.description = data.ImageDescription ? data.ImageDescription.trim() : null;
-
-                  if (data.Make && data.Model) {
-                    info.camera = data.Make.trim() + ' ' + data.Model.trim();
-                  } else if (data.Model) {
-                    info.camera = data.Model.trim();
-                  }
-
-                  info.captureDate = data.DateTime ? data.DateTime : null;
-                  info.credit = data.Copyright ? data.Copyright.trim() : null;
+                  self.apos.utils.error('jpeg-exif error:', err);
+                  return callback(null);
                 }
+
+                const data = exifData;
+                info.description = data.ImageDescription ? data.ImageDescription.trim() : null;
+
+                if (data.Make && data.Model) {
+                  info.camera = data.Make.trim() + ' ' + data.Model.trim();
+                } else if (data.Model) {
+                  info.camera = data.Model.trim();
+                }
+
+                info.captureDate = data.DateTime ? data.DateTime : null;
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-pieces/lib/routes.js
```diff
@@ -346,10 +346,24 @@ module.exports = function(self, options) {
...
+        'title',
+        'description',
+        'credit',
+        'camera',
+        'captureDate'
+      ];
+
       piece.slug = attachment.name;
+
+      for (let prop of properties) {
```
boutell: you can use const here. Not wrong though.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -115,13 +118,29 @@ module.exports = function(self, options) {
             } else {
               info.portrait = true;
             }
+
+            if (info.extension === 'jpg' || info.extension === 'jpeg') {
+              const data = exif.parseSync(file.path);
```
boutell: what if data is null? Makes sense not to crash then.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -115,13 +118,29 @@ module.exports = function(self, options) {
             } else {
               info.portrait = true;
             }
+
+            if (info.extension === 'jpg' || info.extension === 'jpeg') {
```
boutell: It can only be `jpg`. (We had an issue for a while where we didn't guarantee this but it was fixed some time ago. Tell me if I'm wrong...)

lib/modules/apostrophe-attachments/lib/api.js
boutell: Although it is annoying, you should use exif.parse, not exif.parseSync. I read the source and it loads the whole dern JPEG into a buffer. That's tolerable but not if we do it sync, forcing all other web requests to wait for it.

abea: Null data will come back as an empty object, so I don't think we'll crash here.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -120,21 +120,29 @@ module.exports = function(self, options) {
...
-                info.camera = data.Make.trim() + ' ' + data.Model.trim();
-              } else if (data.Model) {
-                info.camera = data.Model.trim();
-              }
-
-              info.captureDate = data.DateTime ? data.DateTime : null;
-              info.credit = data.Copyright ? data.Copyright.trim() : null;
+              exif.parse(file.path, (err, exifData) => {
+                if (err) {
+                  self.apos.utils.error('jpeg-exif error:', err);
```
boutell: Is your goal here to fall back gracefully to our default behavior? That's why you invoke the callback with null? That's... a pretty good plan I think.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -120,21 +120,29 @@ module.exports = function(self, options) {
...
-
-              if (data.Make && data.Model) {
-                info.camera = data.Make.trim() + ' ' + data.Model.trim();
-              } else if (data.Model) {
-                info.camera = data.Model.trim();
-              }
-
-              info.captureDate = data.DateTime ? data.DateTime : null;
-              info.credit = data.Copyright ? data.Copyright.trim() : null;
+              exif.parse(file.path, (err, exifData) => {
```
boutell: ABR (Always Be Returning) when you invoke a callback-based function like this, so you don't get confused later and think you can put code to run "after this"... after this.

lib/modules/apostrophe-attachments/lib/api.js
abea: Yep! Bail.

PR #1849, opened 2019-03-26
should do the trick, eliminating these from showFields when inappropr…

boutell:
…iate

Please test. Thanks.


boutell:
Thanks @mtthwmnc for testing!

PR #1863, opened 2019-04-02
if you can edit something on the page it should not ever cache, ensur…

boutell:
…es you can hit "back" and still see your edits


boutell:
See #1840

PR #1868, opened 2019-04-08
move all ensureIndex calls to the new "migrate" phase, also work around con…

boutell:
…sole.trace producing nothing in some environments. This is part of ongoing enterprise work to move all database operations at apostrophe startup to a phase that can be skipped if desired (for instance, you know you already ran the migrate task). The greater part of that is already done, here we're cleaning up a loose end.


boutell:
I'll put the console.trace calls back, we don't need extra guff just to support bleeding-edge node 11 bugs

PR #1879, opened 2019-04-19
651 cleanup unneeded sessions when flag is set

stuartromanek:


stuartromanek:
can be tested in https://github.com/apostrophecms/apostrophe-enterprise-testbed/tree/redis-uninitialized-resave-test

lib/modules/apostrophe-express/index.js
```diff
@@ -500,14 +495,20 @@ module.exports = {
...
+            // that of the session cookie we're still OK
+            res.cookie(self.apos.csrfCookieName, token, (self.options.csrf && self.options.csrf.cookie) || {});
+          }
         }
-        // Reset the cookie so that if its lifetime somehow detaches from
-        // that of the session cookie we're still OK
-        res.cookie(self.apos.csrfCookieName, token, (self.options.csrf && self.options.csrf.cookie) || {});
       } else {
         // All non-safe requests must be preceded by a safe request that establishes
         // the CSRF token, both as a cookie and in the session. Otherwise a user who is logged
```
boutell: I don't see code here to check for the fallback token value in the appropriate situation. As it is I'm pretty sure this will reject any POST requests from a logged out user if disableAnonCsrf is set, because the browser side code has been updated to send "csrf-fallback" but the server side code has not been updated to accept that value.

boutell: Please make sure a POST request from jQuery can succeed for both logged in and logged out users with the change in place and the option activated. Logged in is easy (opening any manage dialog is sufficient proof). Logged out... hmm. You might have to create a test route and hit it from jquery. Any $.post or $.ajax or $.jsonCall will do (GET is not checked, by design).

boutell:
I was mistaken, you're 100% checking it the right way here

PR #1882, opened 2019-04-23
eliminate unnecessary arrangeFields warnings

boutell:


boutell:
This starts up with no grumping about the users type, and no grumping about the trash field for various types if workflow is turned on. Also in general no complaints about not arranging reverse joins that do not have an editing interface at all.

Issue #1886, opened 2019-04-24
Widgets lost when dragged to a different area, also you can't edit a widget after dragging it, period (at least rich text)


boutell:
I have a partial fix, it covers the first issue but not yet the second.

boutell:
cc @abea

boutell:
see the 1886 branch for work in progress.

boutell:
Got it now. PR coming.

PR #1887, opened 2019-04-24
1886: widgets lost when dragged to a different area, also you can't edit a rich text widget after dragging it

boutell:


boutell:
Closes #1886

boutell:
@grdunn this shows how badly we need regression tests of drag and drop, repeated editing operations, etc. This stuff breaks, sometimes for years, and tends to go unreported because it falls into the category of "oh well I'll refresh." We only heard about this one because the first half of it was more dire (you could lose work).

boutell:
@abea this is now fully ready for your attention. Includes a strong unit test ensuring that concurrent save-area api calls won't drop any information on the floor as long as they are saving different areas.

In conjunction with the fixes specific  to the client project we're working on that I sent you a PR for previously, this should resolve everything. If not we will just have to ask them for more specifics about how to reproduce it.

lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -644,25 +645,40 @@ apos.define('apostrophe-areas-editor', {
...
-        $(event.target).after($item);
+        var $oldWidget = $item.find('[data-apos-widget]');
+        var oldEditor = self;
+        var $newArea = $(event.target).closest('[data-apos-area]');
+        var newEditor = $newArea.data('editor');
+        $oldWidget.data('areaEditor', newEditor);
         self.disableDroppables();
-        self.reRenderWidget($item);
-        self.changeOwners($item);
+        oldEditor.startAutosavingThen(function() {
```
abea: Is it worth checking here if `oldEditor` and `newEditor` are different before doing extra work?

lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -644,25 +645,40 @@ apos.define('apostrophe-areas-editor', {
...
+          $(event.target).after($item);
+          newEditor.startAutosavingThen(function() {
+            newEditor.reRenderWidget($item, function(err) {
+              if (err) {
+                apos.notify('An error occurred.', { type: 'error' });
```
abea: Can we also log a more descriptive error on the server?

lib/modules/apostrophe-areas/public/js/editor.js
abea: It'd be a matter of performance hit, which might only be noticeable if a really beefy and nesting widget.

lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -673,13 +689,18 @@ apos.define('apostrophe-areas-editor', {
...
           // various situations in which jquery is
           // picky about HTML
           var $newWidget = $($.parseHTML($.trim(html), null, true));
           self.replaceWidget($widget, $newWidget);
+          return callback && callback(null);
           // TODO should this take a callback?
           // return callback(null);
+        }, function(err) {
+          apos.ui.globalBusy(false);
+          return callback && callback(err);
```
abea: If there's no callback this will error very quietly. Maybe we should log here as well.

lib/modules/apostrophe-areas/public/js/editor.js
boutell: We have always rerendered in this situation, so I'm going to leave it alone for this release rather than reopen the kettle of worms.

lib/modules/apostrophe-areas/public/js/editor.js
boutell: The server-side code hasn't been altered, so I'm going to open an issue for that but move on.

lib/modules/apostrophe-areas/public/js/editor.js
boutell: I will make a note to add this as a follow-on.

PR #1892, opened 2019-04-29
useful csrf developer warning; refactored to create a warnDevOnce met…

boutell:
…hod rather than duplicating that logic from schemas


lib/modules/apostrophe-utils/lib/api.js
```diff
@@ -616,13 +616,38 @@ module.exports = function(self, options) {
...
+  // the developer until they stop paying attention.
+
   self.warnDev = function(msg) {
     if (process.env.NODE_ENV === 'production') {
       return;
     }
     self.warn.apply(self, arguments);
   };
 
+  self.warnedDev = {};
```
abea: I was confused here thinking you were overriding `warnDev`. Optional suggested change is to make this something like `sentWarning` or `devWarned` for clarity and decreased human error.

PR #1898, opened 2019-05-02
email tests pass, fixed bug: email method did not return info

boutell:


test/email.js
```diff
@@ -56,40 +47,14 @@ describe('Email', function() {
...
-      let message = info.message.toString();
-      assert(message.match(/Fred Astaire/));
-      assert(message.match(/Subject: Welcome Aboard/));
-      assert(message.match(/From: test@example\.com/));
-      assert(message.match(/To: recipient@example\.com/));
-      assert(message.match(/\[http:\/\/example\.com\/\]/));
-      return true;
-    });
+    assert(info);
+    let message = info.message.toString();
```
abea: This should be a `const`.

PR #1900, opened 2019-05-04
3/docs test

abea:
doc.js tests conversion and fixes.


lib/modules/apostrophe-docs/lib/api.js
```diff
@@ -686,7 +688,7 @@ module.exports = function(self, options) {
       }
       if (info.advisoryLock.username === req.user.username) {
         // TODO add locked (423) to http module when I merge
-        // code that is at the office
+        // code that is at the office //
```
abea: Just wanted to call out this todo in case it's a good time to do this.

boutell:
thank you for pushing this forward so much!

PR #1901, opened 2019-05-04
3/events tests

abea:
Converted events tests to async-await. Removed `docBeforeInsert` step from the second since we're removing callAll functions.

### DIFF
--- a/test/events.js
+++ b/test/events.js
@@ -1,33 +1,37 @@
-let t = require('../test-lib/test.js');
-let assert = require('assert');
-let Promise = require('bluebird');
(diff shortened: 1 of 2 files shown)


boutell:
i left this one in a wip state here:

https://github.com/apostrophecms/apostrophe/tree/feature/event-tests

I think it is about the same for events however i also tackled events2 which has one more issue.

metging your events! but dont waste time on events2 i will get that merged

boutell:
i merged this prematurely, you changed the test so it no longer verifies that the string order reversing beforeInsert handler works.

PR #1908, opened 2019-05-08
Converts image tests

abea:
`assert.equal` is deprecated, hence that change.

### DIFF
--- a/test/images.js
+++ b/test/images.js
@@ -1,11 +1,8 @@
-
-let t = require('../test-lib/test.js');
-let assert = require('assert');
-let async = require('async');
-
+const t = require('../test-lib/test.js');
+const assert = require('assert');
 let apos;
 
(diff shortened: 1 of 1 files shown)


test/images.js
```diff
@@ -50,84 +47,85 @@ describe('Images', function() {
...
     assert(apos.images.insert);
-    return async.each(mockImages, function(image, callback) {
-      return apos.images.insert(apos.tasks.getReq(), image, callback);
-    }, function(err) {
-      assert(!err);
-      done();
+
+    const req = apos.tasks.getReq();
+
+    const insertPromises = mockImages.map(async (image) => {
```
boutell: This is a very faithful translation! I am tempted to suggest simplifying it to a `for...of` loop with an `await` inside, but your way we continue to know it can handle parallelism.

PR #1913, opened 2019-05-13
getSchemaOptions is not a good candidate to throw errors because it i…

boutell:
…s sometimes called in contexts without a good doc and that is not an error or even a warning-worthy situation, it has to do with alternate syntaxes for apos.area. Just politely return an empty object, in these cases the options must be restated

### DIFF
--- a/lib/modules/apostrophe-areas/lib/api.js
+++ b/lib/modules/apostrophe-areas/lib/api.js
@@ -350,10 +350,16 @@ module.exports = function(self, options) {
 
   self.getSchemaOptions = function(doc, name) {
     if (!(doc && doc.type)) {
-      throw new Error('You did not pass a doc as the first argument to getSchemaOptions, which expects (doc, name)');
+      // Probably an area or singleton helper call without a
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-areas/lib/api.js
```diff
@@ -350,10 +350,16 @@ module.exports = function(self, options) {
 
   self.getSchemaOptions = function(doc, name) {
     if (!(doc && doc.type)) {
-      throw new Error('You did not pass a doc as the first argument to getSchemaOptions, which expects (doc, name)');
+      // Probably an area or singleton helper call without a
```
abea: If this is still a mistake (even if not a fatal one) then we should still log a warning.

boutell: It is not any kind of mistake. It's an apos.area call with an options object rather than (doc, name), which is less common but entirely valid.

boutell: The mistake lay in me not realizing it would be called in this situation and making it too picky.

PR #1920, opened 2019-05-16
Support ''apostrophe-templates': { viewsFolderFallback:' by accepting an array of views folder.

ammein:
This could be an optional to other devs that wanted to organize their views on multiple project views. This could be done as example :

```js
// in app.js
const path = require('path');
module.exports = {
   // other modules ...
   'apostrophe-templates': { viewsFolderFallback: [path.join(__dirname, 'main-views'), path.join(__dirname, 'lib/modules/widget-views')] }
}
```


ammein:
@boutell Okay. Documentation `option` updated at the top of the file.

boutell:
Thanks!

PR #1926, opened 2019-05-21
1923 dynamic select

boutell:
New functionality implemented in this PR:

* Any `select` or `checkboxes` element in a widget's schema can be moved to the widget's inline controls via the `widgetControls: true` option. Either way a compact representation is used, based on dropdowns.

* Any `select` or `checkboxes` element in *any* module's schema can fetch its choices dynamically at the time they are needed. To do this, you pass a method name for `choices`, rather than an array. The method receives `req` and must return a promise for the array. This means you can just use an `async` function.

See https://github.com/apostrophecms/apostrophe-enterprise-testbed/tree/1923-dynamic-select for an example of use.

Initial application will be implementing salesforce "experiences" similar to personas while still allowing Salesforce, not hardcoded Apostrophe configuration, to decide what the choices are.

But, other applications abound.

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -260,7 +260,7 @@ module.exports = {
         // contain a method name. For bc don't mess with possible
         // existing usages in custom schema field types predating
         // this feature
-        field.moduleName = field.moduleName || module.__meta.name;
+        field.moduleName = field.moduleName || (module && module.__meta.name);
       });
 
       return schema;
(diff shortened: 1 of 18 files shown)


lib/modules/apostrophe-schemas/index.js
```diff
@@ -1414,20 +1443,28 @@ module.exports = {
             }
           },
           choices: function(callback) {
-            return cursor.toDistinct(field.name, function(err, values) {
-              if (err) {
-                return callback(err);
+            return Promise.try(function() {
+              if ((typeof field.choices) === 'string') {
+                return self.apos.modules[field.moduleName][field.choices](req, field);
```
abea: This `req` is undefined.

lib/modules/apostrophe-schemas/index.js
```diff
@@ -1539,20 +1593,28 @@ module.exports = {
...
```
abea: This `req` is also undefined

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -1240,21 +1288,71 @@ apos.define('apostrophe-schemas', {
...
-            $option.attr('value', choice.value);
-            $field.append($option);
+        var choices;
+        return async.series([
+          getChoices,
+          renderChoices
+        ], callback);
+        function getChoices(callback) {
+          if ((typeof field.choices) !== 'string') {
+            choices = field.choices;
```
abea: We don't do anything with `choices`, so the linter is angry. The `var choices;` on line 1291 is then also nada. I think we should be able to just nix this line and move on. Yes?

boutell:
Thanks, fixed those things, pls review

PR #1929, opened 2019-05-28
Converts locks tests to async/await.

abea:


test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
-    function lock(callback) {
-      return locks.lock('test', callback);
-    }
-    function unlock(callback) {
-      return locks.unlock('test', callback);
+  it('cleanup', async function() {
+    try {
+      const response = await apos.locks.db.remove({});
+
+      assert(response.result.ok === 1);
```
boutell: an error in remove() would throw an exception so this assert is probably not necesary.

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
+    } catch (e) {
+      assert(false);
```
boutell: try/catch not necessary because mocha will catch and treat as an assert failure.

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
-      return locks.unlock('test2', callback);
+  it('should allow a single lock without contention uneventfully', async function() {
+    const locks = apos.modules['apostrophe-locks'];
+
+    try {
+      await locks.lock('test');
+
+      await locks.unlock('test');
+    } catch (e) {
+      assert(false);
```
boutell: see above re unnecessary try/catch. With mocha you should typically only need a try/catch in a situation where you *expect* an exception in the "test passing" case.

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
     }
+
+    // function lock(callback) {
+    //   return locks.lock('test', callback);
+    // }
```
boutell: remove dead code.

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
-      return locks.lock('test', function(err) {
-        // SHOULD fail
-        assert(err);
-        return callback(null);
-      });
+  it('should allow two differently-named locks uneventfully', async function() {
+    const locks = apos.modules['apostrophe-locks'];
+
+    try {
+      await locks.lock('test1');
```
boutell: look at the async.series call you're replacing... the idea here is to lock both, then unlock both (both locks exist simultaneously).

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
+
+  it('should flunk a second lock by the same module', async function() {
+    const locks = apos.modules['apostrophe-locks'];
+
+    await locks.lock('test');
+
+    try {
+      await locks.lock('test');
+    } catch (e) {
+      assert(e);
```
boutell: It's OK to do this assert but you haven't proven that an exception definitely *did* get thrown. The way to do that is to put assert(false) immediately after the `await locks.lock` call, i.e. where it should never be reached if the exception is thrown.

With that you actually shouldn't need the try/catch at all because what you really want to know is that you were stopped from continuing (by an exception).

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
-      return locks.unlock('test', function(err) {
-        // SHOULD fail
-        assert(err);
-        return callback(null);
-      });
+
+    await locks.unlock('test');
+
+    try {
+      await locks.unlock('test');
```
boutell: same issue, you need an assert(false) after this to stop the show if the code is allowed to continue past the (intentionally) bad second unlock.

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
-    function attempt(locks) {
-      return locks.lock('test', function(err) {
-        assert(!err);
-        active++;
-        assert(active === 1);
-        setTimeout(release, 75 + Math.random() * 50);
-      });
-      function release() {
+
+    await Promise.all([
```
boutell: nice.

test/locks.js
```diff
@@ -39,231 +47,219 @@ describe('Locks', function() {
...
-      return locks.unlock('test');
-    }).then(function() {
-      return locks.unlock('test')
-        .catch(function(err) {
-          // SHOULD fail
-          assert(err);
-        });
-    });
-  });
+  it('with promises: should flunk a second lock by the same module', async function() {
```
boutell: I think this is just a duplicate test now, no?

test/locks.js
abea: My goal was to have a simple assertion that things went okay. Without this we don't have an assertion for this test. Is that okay?

test/locks.js
abea: ✔️ I'll add the `assert(false)`. If I don't catch anything it'll break subsequent tests, though.

test/locks.js
abea: seems okay

test/locks.js
```diff
@@ -39,231 +47,177 @@ describe('Locks', function() {
...
+  it('should flunk a second lock by the same module', async function() {
+    const locks = apos.modules['apostrophe-locks'];
+
+    await locks.lock('test');
+
+    try {
+      await locks.lock('test');
+      assert(false);
+    } catch (e) {
+      assert(e);
```
boutell: This is a nice way to show intent although it is not strictly necessary to have this assert, since the only way assert(false) can happen is if the desired exception is not thrown.

test/locks.js
```diff
@@ -39,231 +47,177 @@ describe('Locks', function() {
...
-        // SHOULD fail
-        assert(err);
-        return callback(null);
-      });
+
+    await locks.unlock('test');
+
+    try {
+      await locks.unlock('test');
+      assert(false);
```
boutell: Good

test/locks.js
```diff
@@ -39,231 +47,177 @@ describe('Locks', function() {
...
-    function lock2(callback) {
-      return locks.lock('test2', callback);
-    }
-    function unlock2(callback) {
-      return locks.unlock('test2', callback);
-    }
+  it('should allow two differently-named locks uneventfully', async function() {
+    const locks = apos.modules['apostrophe-locks'];
+
+    await locks.lock('test1');
```
boutell: Good.

PR #1930, opened 2019-05-29
3/login test

abea:


abea:
Note to self: try permissions tests again after merging this. Lots of `self.generic/specific is not a function` errors over there.

lib/modules/apostrophe-login/index.js
```diff
@@ -160,25 +160,26 @@ module.exports = {
...
           { email: username }
         ],
         disabled: { $ne: true }
-      });
+      }).toObject();
+
       if (!user) {
         await Promise.delay(1000);
-        throw 'invalid';
+        throw 'No user found.';
```
boutell: Please leave this as `invalid`, string exceptions are used to communicate specific meaningful error codes to the browser side.

lib/modules/apostrophe-login/index.js
```diff
@@ -160,25 +160,26 @@ module.exports = {
...
       }
       try {
         await self.apos.users.verifyPassword(user, password);
-        return user;
+        return done(null, user);
       } catch (err) {
         await Promise.delay(1000);
-        throw 'invalid';
+        throw 'User verification error.';
```
boutell: See above.

lib/modules/apostrophe-login/index.js
```diff
@@ -160,25 +160,26 @@ module.exports = {
     // If the user's login SUCCEEDS, the return value is
     // the `user` object.
 
-    self.verifyLogin = async function(username, password) {
+    self.verifyLogin = async function(username, password, done) {
```
boutell: I understand this change was needed to make it work, but please instead change the code that invokes this method to handle the icky callback-ness, allowing this method to follow the standard async pattern and just return a user or throw an exception nicely. So far as written an exception would probably not behave as expected.

boutell: Right now this method is probably passed directly to passport, but instead a wrapper function that invokes it and handles the callback-itude should be passed.

abea:
I changed the one `throw` to follow the passport pattern: https://www.npmjs.com/package/passport#strategies. We could leave it and return null user as an error as well.

lib/modules/apostrophe-login/index.js
```diff
@@ -168,10 +180,11 @@ module.exports = {
           { email: username }
         ],
         disabled: { $ne: true }
-      });
+      }).toObject();
+
       if (!user) {
         await Promise.delay(1000);
-        throw 'invalid';
+        return false;
```
boutell: Fine with this, it is passporty. But the doc comment should be fixed to cover this case.

abea: 👍 I added updated the comment.

PR #1933, opened 2019-05-30
3/schema filters tests

abea:


abea:
Should we close this, @boutell ? I would think so much has happened in test world.

boutell:
Yep! Far too much has happened. Closing in on these tests ere long.

PR #1944, opened 2019-06-07
apos.utils.post now accepts a FormData object as the data parameter a…

boutell:
…nd does the right thing so you can do a file upload back to IE11 without a lot of fuss and still get our CSRF protection. Also: database connection to localhost now uses 127.0.0.1 to avoid a DNS lookup that might not work properly while passing through a dodgy wifi area; should not matter but I saw it several times today.


boutell:
In that situation you are probably an xsrf attacker and apostrophe sends
you the big fat 403 you deserve.

boutell:
I promise to fix the lint (our release process requires it).

PR #1945, opened 2019-06-08
Holy crap! async-await has landed, plus other notable changes

boutell:
5000000 thanks to @abea for making so many tests pass.

In the end we did not make 100% of tests pass yet because the appropriate criterion is making the async-await branch work as well as the 3.0 branch, not 100% functionality, before we can merge back to 3.0 and work together going forward. For instance pieces and pages need a lot of reimplementation at the route level before we can pass those tests and that's OK because they weren't done in the 3.0 branch either.

So: now is the time!

Other important changes besides exterminating callbacks from all APIs exposed to the developer:

* methods like `docs.insert` and `docs.update` now emit `beforeInsert`, `afterSave`, etc. via the manager for the appropriate doc type. The corresponding "template pattern" placeholder methods have been eliminated from `apostrophe-pieces` and friends. This simplifies event handlers which don't have to start by checking the type anymore. You just listen for these events on the appropriate module.

However, since promise events support inheritance in 3.x, you can listen for events on `apostrophe-doc-type-manager:beforeInsert` if you are interested in everything. Or `apostrophe-pieces:beforeInsert`. And so on.

Code that listens for `apostrophe-docs:beforeInsert` will need to change to take this into account in 3.x.

* async components have landed. You can now write this in a template:

```
{% component "products:list" with { limit: 5 } %}
```

In the products module you back it with:

```
self.addComponent('list', async function(req, data) {
  return {
    products: await self.find(req).limit(data.limit).toArray()
  };
});
```

And in `products/views/list.html`, you can then write:

```
{% for product in data.products %}
  <h4>{{ product.title }}</h4>
{% endfor %}
```

This means we can "fetch stuff in a template" for the first time.

* As a necessary consequence of async components, **the syntax for areas and singletons has changed.** The new area syntax is:

```
{% area data.page, "body", { ... options like always .... } %}
```

Similarly `{% singleton %}`, which is not implemented yet, but the criterion for merging this branch was matching the functionality in `a3-test-1` that we had previously with the 3.0 branch, and that criterion has been met. So we can implement `singleton` when we get a minute.

* To support this I added support for easily creating Nunjucks custom tags with 50% less Nunjucks inside baseball. Refer to the `apostrophe-templates` and `apostrophe-areas` modules, which both set up the `customTags` option.

* `customTags` has been implemented following the pattern we're discussing as a general overhaul of module configuration on Wednesday. If we decide we don't want to go there, I can add an `addCustomTag` method instead at that time.


boutell:
I don't expect this PR to be reviewed in detail. Nearly all of this code has already been through review in smaller PRs against `async-await`. More an opportunity to sanity check and take stock before hitting the button.

boutell:
To facilitate live testing, I have pushed an `async-await` branch of the `a3-test-1` project. 

To refresh your memory, these things should work:

* Adding, editing, removing widgets
* The "test" widget, which has one text field, plus the rich text widget
* widget edits should "stick" on refresh

Frontend code has not changed so minor bugs there like drag and drop are probably not new in async-await.

With regard to the admin bar, the "manage" views for pieces (not pages) should pop up and you should be able to initiate the "insert" process, but you will only get "title" and "slug" and the save button is not supposed to work yet.

PR #1956, opened 2019-06-26
Fix open redirect vulnerability during trailing slash redirect.

bharath6365:


bharath6365:
Steps to Reproduce

1) Go to any website powered by Apostrophe. In this case I'm making use of the demo site http://dashboard.apostrophecmsdemo.org/

2) Paste this url http://dashboard.apostrophecmsdemo.org/%2f%2fgoogle.com%2f
As you can see its url encoded. (%2f is decoded as /). Now there will be a redirect because apostrophe doesnt accept trailing slashes. 

3) Check the location header sent for the redirect. 
https://www.evernote.com/l/ApTvRWGFoRBArb2F71MkT0XYmjer-lKP1BQB/image.png

boutell:
You're right. I think replacing any series of multiple slashes with a single slash would be sufficient and would have a lesser impact on innocent code that could be out there; what do you think? Can you see a way to exploit that?

boutell:
I pushed a fix for this that has a lesser impact, it just replaces multiple-slash sequences with a single slash. I'm working on releasing that now. Thank you for the report.

boutell:
This was npm published yesterday. Thanks again for the report.

bharath6365:
Thank you so much :)

PR #1957, opened 2019-06-26
Updates the major version of diff

abea:
This is in response to a vulnerability alert: https://github.com/apostrophecms/apostrophe/network/alert/package.json/diff/open. The only reported v3 breaking change reported is on a method we're not using and all tests are passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "credential": "^2.0.0",
     "cuid": "^1.3.8",
     "deep-get-set": "^0.1.1",
-    "diff": "^2.2.3",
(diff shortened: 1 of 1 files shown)


abea:
Looks like we could go to 4.x if we wanted, as well. https://github.com/kpdecker/jsdiff/blob/master/release-notes.md#v400---january-5th-2019

boutell:
I would go to the newest major version that doesn't break the "Page Versions" interface.

abea:
Tested and it works!

PR #1958, opened 2019-06-27
Color picker spectrum options

woodbrearlham:
These changes will allow for a person to pass in spectrumOptions for a field type of color. Spectrum is the plugin used for the color picker.

On a widget I was able to configure a field like this to show color pallets.
```
{
		      name: 'bgColor',
		      label: 'Background Color',
				type: 'color',
				spectrumOptions: {
				showPalette: true,
				    palette: [
				        ['black', 'white', 'blanchedalmond'],
				        ['rgb(255, 128, 0);', 'hsv 100 70 50', 'lightyellow']
				    ]
				}
			},
```

### DIFF
--- a/lib/modules/apostrophe-schemas/public/js/user.js
+++ b/lib/modules/apostrophe-schemas/public/js/user.js
@@ -1475,7 +1475,7 @@ apos.define('apostrophe-schemas', {
       name: 'color',
       populate: function (data, name, $field, $el, field, callback) {
         $field.val(data[name]);
-        apos.ui.enhanceColorpicker($field);
+        apos.ui.enhanceColorpicker($field, field.spectrumOptions);
         return setImmediate(callback);
       },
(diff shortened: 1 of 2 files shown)


stuartromanek:
Hey @woodbrearlham this PR rocks, but it dredges up the deeper issue of validation and sanitization of the color value (however it might need to be represented).

If an editor turns on `showAlpha`, their color value would most comfortably be saved as an rgba object rather than a hex string, which the current setup is hard coded enforcing.

So, In addition to letting the developer pass the Spectrum options to the field, we need to save and sanitize a more standard data format. Spectrum uses `tinycolor` already (think `moment` for color) and we already have the module on the backend so I think we should save the entire `tinycolor` object as the value and translate it back and forth.

I will probably start a new branch for this enhancement incase you need to depend on this one temporarily.

abea:
Closing this for inactivity.

PR #1961, opened 2019-06-28
integrate storybook into the 3.0 branch

stuartromanek:
This installs `apostrophe-storybook` as a dev dependency of Apostrophe and includes an npm task that reaches in and turns it on.

Does anyone oppose `apostrophe-storybook` remaining it's own repo?

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,7 @@
   "scripts": {
     "test": "mocha && eslint .",
     "lint": "eslint .",
-    "storybook": "(cd node_modules/apostrophe-storybook && start-storybook -p 6006)",
-    "build-storybook": "build-storybook"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,7 +5,9 @@
   "main": "index.js",
   "scripts": {
     "test": "mocha && eslint .",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "storybook": "(cd node_modules/apostrophe-storybook && start-storybook -p 6006)",
+    "build-storybook": "build-storybook"
```
boutell: How does this one work? Is build-storybook globally installed or something?

stuartromanek: oops this is a stump .. this is what we need to write, will remove

PR #1966, opened 2019-07-02
Allows test destroy util without done

abea:
Adds an option to use the `destroy` test utility method without `done`. This lets us use the method for a secondary Apostrophe instance in async tests.

### DIFF
--- a/test-lib/util.js
+++ b/test-lib/util.js
@@ -7,7 +7,7 @@ var async = require('async');
 // credentials per database.
 
 function destroy(apos, done) {
-  if (!apos) {
+  if (!apos && done) {
     done();
     return;
   }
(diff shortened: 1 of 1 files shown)


boutell:
You should instead pass an inline function as the "done" argument, and do your next step there.

Otherwise the next test might run with this secondary port still in use, and you don't know what that test might need, i.e. it might try to use that port. You should not call the final "done" until you are 100% sure everything is released.

abea:
The use case I came on was in an async test, which wouldn't be using `done()` at all (mocha throws a warning that it's over-specifying _something something_). In that case I'm trying to destroy a second instance of apostrophe at the end of its use. I could pass a function that does nothing, but that seems silly. My idea with this change is to prepare it to be used with the async tests we're writing more lately.

boutell:
any instance of apostrophe is holding onto a port number that the next test
file also being run by mocha, which you didn't write, might also need for
its own secondary instance of apostrophe. You can't know if that's going to
be the case.

So it is never okay to go on to the next test case in your file or end your
file without being sure that that port is closed.

if there is no done function on your step due to async/await then you can
promisify the destroy method and await it.

See the promisify method of node's built in util module.

PR #1968, opened 2019-07-03
1866 simple bundles for easier deployment in the cloud

boutell:
The underlying problem: when you have multiple servers/dynos, assets ought to be built once at deployment time and then served from a common location, like S3. 

The old solution: ... much too complicated. See our Heroku HOWTO. You had to build a named "asset bundle" and push it into your repo every time. This meant you couldn't "just merge to production to trigger Heroku," which is the right way to do Heroku, per our experienced team on the Crit project.

The new solution:

```
# Build the one and only asset bundle
APOS_BUNDLE=1 node app apostrophe:generation
# After deployment, launch taking advantage of the bundle
APOS_BUNDLE=1 node app
```

In Heroku that env var gets set the Heroku way, but that's the idea.

Under the hood it's pretty complicated - we have to stash the bundle in a tarball in the database, because all dynos/servers can see the database (unlike the filesystem). Those dynos/servers then extract it at startup so that any requests to access the files directly, rather than via S3, will still succeed too - a practical requirement in 2.x at least. We also have to handle the cleanup process for old bundles correctly.

I also changed the default delay before asset bundle cleanup in S3 to 5 hours because I was unrealistic about how long it might be before there are truly no servers expecting the old asset generation ID running.


boutell:
Thanks, fixed these issues and merged.

PR #1973, opened 2019-07-10
Required arrays cancelling arrays and dynamic select in arrays

### LINKED ISSUE #1969: Hidden, required array fields throw `required` errors when not populated
**To Reproduce**
1. Create a piece schema with a boolean or select field (e.g., `deciderField`) that activates a second field (`reliantField`). Make `reliantField` an array field and make it required.
2. Create a piece of that type, but do not activate the `deciderField`, keeping `reliantField` hidden.
3. Save the piece.

**Expected behavior**
Since the array field is hidden, it's requirement should not prevent saving.

**Describe the bug**
The hidden array field prevents saving, returning a `required` error.

boutell:
Fixed three array bugs:

(1) a required array hidden by showFields is in effect not required (same as other field types).

(2) Cancelling an edit operation on an array actually does revert everything, notably it does not leave you with a halfassed first entry in an array you never meant to create at all.

(3) Dynamic select and checkboxes fields work in arrays.

If you wish you can experience these fixes via the array-bug-tests branch of apostrophe-enterprise-testbed, which moves a bunch of fields of "Products" into a "Profiles" array field (look at the "Info" tab). The entire array becomes visible when you check "Has Profiles." (These are not actually regression tests, just a nice testbed for validating these fixes manually.)

closes #1969


boutell:
Crap I see this merges too much hang on.

PR #1976, opened 2019-07-10
additional logic for apos.utils.post and apos.utils.get for parsing json

grdunn:


lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,11 +87,46 @@
...
+      // If the response header is `application/json`, immediately try
+      // to JSON.parse(), and if it fails, report the error.
+      if (responseHeader === 'application/json; charset=utf-8') {
+        try {
+          data = JSON.parse(this.responseText);
+        } catch (e) {
+          return callback(e);
+        }
+      // If there is no explicit content-type header,
+      // test the trimmed response against our isJson() method.
```
boutell: isJSON is a function, not a method.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,11 +87,46 @@
...
+      // If it returns `true`, try to JSON.parse() it.
+      // And if it fails, try to JSON.stringify() it, and if that fails,
+      // report an error.
+      } else {
+        if (isJSON(trimResponse, callback)) {
+          try {
+            data = JSON.parse(this.responseText);
+          } catch (e) {
+            try {
+              data = JSON.stringify(this.responseText);
```
boutell: wat why would you stringify it? That would turn:

OK

Into (as a string): `"OK"`

at this point we just want to give up and treat data as a string, which it already is.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,11 +87,46 @@
...
+            } catch (e) {
+              return callback(e);
+            }
+            return callback(e);
+          }
+        // If the trimmed response returns false from our
+        // isJson() method, JSON.stringify it. If it fails,
```
boutell: Never stringify it. If JSON.parse doesn't like it, just deliver it as the string that it already is. This way if they ship us markup, we get the string with the markup, not a string with quotes around it.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -103,6 +138,18 @@
     });
   };
 
+  // Makes a reasonable determination that
+  // the data passed to the function is formatted
+  // JSON and returns `true`, if so. False, if not
+
+  function isJSON (json, callback) {
```
boutell: I'd call it `maybeJSON` as it's in no way definitive.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -103,6 +138,18 @@
...
+    if (!json) {
+      return callback(new Error('You forgot to pass JSON data as the first argument.'));
```
boutell: You should check that `json` is a string, not that it is falsy. The empty string is falsy and it is a plausible response.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -103,6 +138,18 @@
...
+    }
+    var cases = [ true, false, null, '"', '-', '[', '{' ];
```
boutell: You're looking for whether `json` is a string that begins with one of these strings - `'true'`, not `true`, for instance.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -103,6 +138,18 @@
...
+    return cases.some(function(element) { return json[0] === element; });
```
boutell: json[0] is going to be `'t'`, not `true`, so the test will fail for that case as written so far.

As for `.some`: back to IE9, so OK.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,42 +87,30 @@
...
             } catch (e) {
               return callback(e);
             }
             return callback(e);
           }
-        // If the trimmed response returns false from our
-        // isJson() method, JSON.stringify it. If it fails,
-        // report the error.
-        } else if (!isJSON(trimResponse, callback)) {
+        } else if (!maybeJson(trimResponse, callback)) {
```
boutell: This would just be "else", no need to duplicate the function call.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,42 +87,30 @@
...
-        if (isJSON(trimResponse, callback)) {
+        if (maybeJson(trimResponse, callback)) {
           try {
             data = JSON.parse(this.responseText);
           } catch (e) {
             try {
-              data = JSON.stringify(this.responseText);
+              data = this.responseText;
             } catch (e) {
               return callback(e);
```
boutell: Just `data = this.responseText` is very unlikely to throw an exception, I would not try/catch it. I think `responseText` is guaranteed to exist.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,42 +87,30 @@
...
             }
             return callback(e);
```
boutell: This is going to report an error after all and fail, remove this line; we set data to this.responseText to be tolerant and say "treat it as JSON only if it is valid as JSON."

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,42 +87,30 @@
...
           }
-        // If the trimmed response returns false from our
-        // isJson() method, JSON.stringify it. If it fails,
-        // report the error.
-        } else if (!isJSON(trimResponse, callback)) {
+        } else if (!maybeJson(trimResponse, callback)) {
           try {
-            data = JSON.stringify(this.responseText);
+            data = this.responseText;
```
boutell: Always exists, no try/catch required.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -138,16 +126,12 @@
...
   };
 
-  // Makes a reasonable determination that
-  // the data passed to the function is formatted
-  // JSON and returns `true`, if so. False, if not
-
-  function isJSON (json, callback) {
-    if (!json) {
-      return callback(new Error('You forgot to pass JSON data as the first argument.'));
+  function maybeJson (json, callback) {
```
boutell: This should not take `callback`. Functions that take `callback` should invoke it in every case, not only in the error case like here, it is confusing because it does not follow the node callback convention.

This function is just a sniffer so it should be tolerant and not throw exceptions at all. Just return false.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -138,16 +126,12 @@
...
+    if (typeof json !== 'string') {
+      return callback(new Error('You forgot to pass a string as the first argument.'));
```
boutell: Just return false. No callback invocation that makes the flow of the code confusing, and no need for an error object.

boutell: Unfortunately this didn't make this morning's release, but if it gets wrapped up soon it could make the next one...

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -87,11 +87,25 @@
...
   function monitor(xmlhttp, callback) {
     xmlhttp.addEventListener("load", function() {
+      var responseHeader = this.getResponseHeader("Content-Type");
+      var trimResponse = this.responseText.trim();
       var data;
-      try {
-        data = JSON.parse(this.responseText);
-      } catch (e) {
-        return callback(e);
+      if (responseHeader === 'application/json; charset=utf-8') {
```
boutell: Hmm, just make sure it either equals `application/json` in its entirety, or starts with `application/json;`. I say this because in this context `charset` is supposedly a cargo cult thing that has no meaning (JSON must be utf-8) and therefore might not be present. It took some research to figure that out just now. Weird.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -103,6 +117,14 @@
     });
   };
 
+  function maybeJson (json) {
+    if (typeof json !== 'string') {
+      return false;
+    }
+    var cases = [ 'true', 'false', 'null', '"', '-', '[', '{' ];
+    return cases.some(function(element) { return json.split(' ')[0] === element; });
```
boutell: Array.some is... oh, back to IE9. TIL.

This code is correct, but  you should use the second `limit` parameter to `split` to avoid doing a lot of extra work, since we'll use this a lot. We're only interested in the first result.

And, you should call `json.split(' ')[0]` just once and capture it in a variable rather than doing it on every pass through the loop. (Maybe the compiler figures this out for you, or maybe it can't.)

PR #1983, opened 2019-07-27
a simple and friendly solution for actually fixing slug conflicts in …

boutell:
…cases where you would like to move the other document out of the way

It took longer than I expected, but I was able to get this all the way nice, in that the user is given a clear explanation (it would be nice if in some future iteration this was not a confirm call), taken to the piece or page settings of the conflicting document, warped to the correct tab, and after saving, finds themselves where they need to be to hit "save" again

The "taken" error is, optimistically, removed once they click the wrench; if they haven't resolved it though it will safely come back when they try to "save" again.

I think this is, barring bugs, as nice as could be hoped for without going a little too crazy for a 2.x fix.


boutell:
slug-conflict-fix-button

stuartromanek:
Styled this up and put the field focus on a small timer, it was getting beat out by the default schema focus.

image

PR #1989, opened 2019-08-07
a simple bundle pushes assets to uploadfs too, but we were only check…

boutell:
…ing for the old command line flag for that, and thus missed the need to prefix asset URLs


boutell:
This matches a check elsewhere in the assets module where the actual copy to uploadfs takes place. Here it has to do with prepping assets with the right URLs.

PR #1993, opened 2019-08-13
apos.utils.post should report HTTP errors properly (XMLHTTPRequest do…

boutell:
…es not invoke an error handler for that on its own)


lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -110,7 +120,19 @@
...
+        // xmlhttprequest does not consider a 404, 500, etc. to be
+        // an "error" in the sense that would trigger the error
+        // event handler function (below). It looks like only network
+        // failures and/or browser problems can do that. So we need
+        // to recognize error status codes ourselves and correctly
+        // report an error.
+        if (xmlhttp.status < 400) {
+          return null;
+        }
+        return xmlhttp;
```
boutell: A simple way to get an object with a status property (the xmlhttprequest object is already such an object).

PR #1996, opened 2019-08-21
fixed and completed slugPrefix functionality for doc types, workaroun…

boutell:
…d for issue where migration to add the prefix could encounter a slug conflict (you fix those yourself, they should not be numerous and they do not break anything meanwhile)


boutell:
So, when you upload an image to a stock apostrophe project, the slug now has the prefix `image-`. When you upload a file, the slug has the prefix `file-`.

The slugs of images and files are little used (they are *not* part of the filename actually used to download the file, and people rarely make apostrophe-images-pages or apostrophe-files-pages), but they frequently conflict with slugs people care about. This eliminates that issue.

There was a prior attempt at this, which sped the work along, but it was incomplete.

PR #2002, opened 2019-09-02
clean up the context menu in several ways

### LINKED ISSUE #1982: UI error when using sendPage
Hi all! I translate the question from Discord here, as @grdunn suggested.

I'm having problems with using the sendPage method (https://docs.apostrophecms.org/apostrophe/modules/apostrophe-module#sendpage-req-template-data). When I use it, in the page appears this two buttons , neither of them worked. Of course the page is published. Is this a bug or is there anything that I missed? Thanks! I let you some example code to clarify what I am telling.

image

```
  self.addRoutes = function() {
    self.apos.app.get('/', async function(req, res) {
      return self.sendPage(req, '/pages/home', {message: 'test'});
    });
  }
}

```

### LINKED ISSUE #2000: Error when publishing pieces page
When publishing a piece through publish menu it gives error. Even though the article is being published.
image

Error log is:
```
POST /modules/articles/publish: 

Error: Error is not a string or an instance of Error, maybe this is not an error and you forgot to
pass null as the first argument to next()?
```

I found that it comes from this line:
https://github.com/apostrophecms/apostrophe/blob/fcf7ada3e962b0b19543e7f0aa03fa362a0ce148/lib/modules/apostrophe-jobs/lib/api.js#L103

In `apiResponder(req, err, result, extraError)` function the second argument is `err`, however in the code above it is being passed the result. As a solution the second argument should be `null` indicating there is no error so that the argument: 
```
{
          // legacy
          data: single || {},
          // preferred
          results: results
        }
``` 
will be the third one.

boutell:
Closes #1982 
Closes #2000 

TL;DR: the context menu is no longer popping up at dumb times, busted-feeling and generally confusing people.

In #1982 we see there is a context menu when sendPage is being used to render a page directly with no `data.page` or `data.piece`. There should be none. The context menu's visibility is now predicated on both `req.user` and at least one of `req.data.page._edit` and `req.data.piece._edit`. This eliminates a great many situations where people wind up overriding things they shouldn't just to get rid of the context menu.

We used to be afraid to do this because "Reorganize" had, wrongly, crept into the "context" menu and in theory you might want to "Reorganize" from a specialized template that has no `data.page`. However (1) "Reorganize" has also been available as "Pages" on the admin bar for a long time now and (2) this is a far-fetched scenario anyway.

In theory, someone could be using `apos.templates.prepend('aposContextMenu')` to set up some cool contextual UI of their own that doesn't require a page or a piece. I don't think this is  likely. However, I wrapped the "show it / don't show it" decision in the new method `apos.templates.showContextMenu`, which can be overridden to suit custom needs.

In #2000 there is a simple "oops, wrong number of parameters" error that prevents the correct UI feedback after successfully publishing a piece via the "Unpublished" dropdown on the context menu. I fixed that too as without it the other fix was a bit underwhelming.

### DIFF
--- a/lib/modules/apostrophe-jobs/lib/api.js
+++ b/lib/modules/apostrophe-jobs/lib/api.js
@@ -100,7 +100,7 @@ module.exports = function(self, options) {
             }
           });
         }
-        return self.apiResponder(req, {
+        return self.apiResponder(req, null, {
           // legacy
           data: single || {},
           // preferred
--- a/lib/modules/apostrophe-pages/views/publishMenu.html
+++ b/lib/modules/apostrophe-pages/views/publishMenu.html
@@ -1,5 +1,5 @@
 {% import 'apostrophe-ui:components/dropdowns.html' as dropdowns with context -%}
-{%- if not data.page.published or (data.piece and not data.piece.published) -%}
+{%- if not (data.page and data.page.published) or (data.piece and not data.piece.published) -%}
 <div class="apos-context-menu">
   {{ dropdowns.base({
       label: "Unpublished",
--- a/lib/modules/apostrophe-pieces-pages/index.js
+++ b/lib/modules/apostrophe-pieces-pages/index.js
@@ -77,14 +77,6 @@ module.exports = {
       }
     ];
 
-    self.publishMenu = options.publishMenu || [
-      {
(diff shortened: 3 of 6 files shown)


abea:
@boutell Do you know what's going on with the failing test: `should render pages successfully with prepend and append to locations`?

Issue #1845, opened 2019-03-22
Auto-dismiss save notification


scottbuscemi:
**Is your feature request related to a problem you are experiencing that is not a bug? Please describe.**
After saving a page or piece, a notification appears to let the admin know: "The page has been created and saved." However, this notification never goes away. Editing multiple pages/pieces results in these notifications stacking up and filling the screen.

**Describe the solution you'd like**
Auto-dismiss after a few seconds.

boutell:
The notifications go away when you click on them, they don't stack up forever.

We should use the auto-dismiss option for more of them, though. These sound like good examples of that so I'll keep the ticket open with these particular cases in mind. (Some notifications already auto-dismiss after a few seconds.)

boutell:
Fixed in commit c431756, will be published with the next release of apostrophe-workflow. Also there is a PR to add an obvious UI for closing notifications.

PR #2004, opened 2019-09-02
ui for closing notifications closes #1870

### LINKED ISSUE #1870: Error message should have close button to dismiss the message
- Apostrophe modal error messages currently doesn't have a close button to let the back end users dismiss these errors and it will never go away unless you reload the page

- Currently the only way to dismiss/acknowledge this error message is to click on the message itself but how will the back end user know that ? 

- Please add a close button to allow the back end user to dismiss this error message

- This is a sample override code to change the error messages in custom coded module/piece:

```
// Inside lib/modules/apostrophe-users/public/js/editor-modal.js
apos.define('apostrophe-users-editor-modal', {

  extend: 'apostrophe-pieces-editor-modal',

  construct: (self, options) => {
    self.getErrorMessage = function (err) {
      if (err === 'unique') {
        return 'Please make sure the username and email address fields are unique.';
      } else if (err === 'Invalid Password') {
        return 'Invalid Password!';
      } else if (err === 'Invalid Administrator Password') {
        return 'Invalid Administrator Password!';
      }
      else {
        return 'An error occurred. Please try again.';
      }
    };
  }
});
```

boutell:
I'm just a backend hacker but I think this came out surprisingly unterrible:

Screen Shot 2019-09-02 at 3 08 07 PM

What say you @stuartromanek?

### DIFF
--- a/lib/modules/apostrophe-notifications/views/notification.html
+++ b/lib/modules/apostrophe-notifications/views/notification.html
@@ -1,3 +1,3 @@
-<div class="apos-ui apos-notification apos-notification--hidden" data-apos-notification>
+<button class="apos-ui apos-notification apos-notification--hidden" data-apos-notification>
   <p class="apos-notification-message" data-apos-notification-message>{{ data.message }}<span class="apos-notification-message-close"><i class="fa fa-times-circle" aria-hidden="true"></i></span></p>
-</div>
+</button>
--- a/lib/modules/apostrophe-notifications/views/notification.html
+++ b/lib/modules/apostrophe-notifications/views/notification.html
@@ -1,3 +1,6 @@
 <button class="apos-ui apos-notification apos-notification--hidden" data-apos-notification>
-  <p class="apos-notification-message" data-apos-notification-message>{{ data.message }}<span class="apos-notification-message-close"><i class="fa fa-times-circle" aria-hidden="true"></i></span></p>
+  <p class="apos-notification-message" data-apos-notification-message>
+    {{ data.message }}<span class="apos-notification-message-close"><i class="fa fa-times-circle" aria-hidden="true"></i></span>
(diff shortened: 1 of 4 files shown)


boutell:
See also #1845, confusion about how to dismiss these is common.

abea:
We are missing some easy a11y here, including in the existing code. The `span` should definitely be a `button`, for this contrib.

We should also be using the `alert` role on the whole thing. That would mean moving from only an opacity change to a `display` change, though, so maybe to save for another PR.

lib/modules/apostrophe-notifications/views/notification.html
```diff
@@ -1,3 +1,3 @@
 <div class="apos-ui apos-notification apos-notification--hidden" data-apos-notification>
-  <p class="apos-notification-message" data-apos-notification-message>{{ data.message }}</p>
+  <p class="apos-notification-message" data-apos-notification-message>{{ data.message }}<span class="apos-notification-message-close"><i class="fa fa-times-circle" aria-hidden="true"></i></span></p>
```
abea: We should use a `button` here for semantics and the built-in benefits. Or maybe the `div` around the whole thing should be a `button` if clicking the whole thing closes it. That'd only be bad if auto-dismiss ones don't dismiss on click too.

abea: Holler if you want me to take a swipe at that.

boutell: They all dismiss on click. Feel free to make it a button and get the styles happy with that. Thanks.

boutell: Agree we should do what's expedient in this PR (fill the current UX hole) and continue to iterate. Button seems reasonable to do now.

abea:
Updated. @stuartromanek Could you take a look since I did a bunch in there?

PR #2006, opened 2019-09-06
Hotfix/intangible tooltip

abea:
The invisible tooltip was making it hard to target a small singleton in a modal. Since the tooltip isn't triggered directly this should be fine.


boutell:
LGTM

PR #2011, opened 2019-09-07
the registry of existing area editors was removed because dot paths a…

boutell:
…re not good unique identifiers (drag and drop changes them), however this registry is essential to figure out what to do when part of the DOM is rerendered due to creating a new piece during a join editing operation. Resolve it by bringing back the registry, but recomputing it on all inserts, removals, moves and drags. What a... drag. In 3.x consider having area _ids to solve this.

### DIFF
--- a/lib/modules/apostrophe-areas/public/js/editor.js
+++ b/lib/modules/apostrophe-areas/public/js/editor.js
@@ -10,7 +10,6 @@ apos.define('apostrophe-areas-editor', {
   },
 
   construct: function(self, options) {
-
     self.options = options;
     self.$el = self.options.$el;
     self.$body = $('body');
@@ -38,6 +37,7 @@ apos.define('apostrophe-areas-editor', {
         // and avoid trying to autosave on their own
         self.$el.attr('data-apos-area-virtual', '1');
       }
+      apos.areas.register(self.$el.attr('data-doc-id'), self.$el.attr('data-dot-path'), self);
       var canceled = false;
       self.$el.parents('[data-apos-area]').each(function() {
         if (canceled) {
@@ -188,6 +188,7 @@ apos.define('apostrophe-areas-editor', {
       } else if (direction === 'bottom') {
         $wrapper.parent().children(':last').after($wrapper);
       }
+      apos.areas.remapDotPaths();
       apos.emit('widgetMoved', $wrapper.find('[data-apos-widget]'));
     };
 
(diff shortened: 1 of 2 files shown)


lib/modules/apostrophe-areas/public/js/editor.js
```diff
@@ -468,7 +420,7 @@ apos.define('apostrophe-areas-editor', {
       // rule out areas that do not allow the widget type in question
       return $('[data-apos-area]').filter(function() {
         var editor = $(this).data('editor');
-        if ((editor && !editor.limitReached()) || ($widget.data('areaEditor') === editor)) {
+        if (editor && ((!editor.limitReached()) || ($widget.data('areaEditor') === editor))) {
```
boutell: I noticed this unrelated bug and fixed it while I was in the neighborhood.

lib/modules/apostrophe-areas/public/js/user.js
```diff
@@ -148,11 +148,95 @@ apos.define('apostrophe-areas', {
...
+    };
+
+    self.remapDotPaths = function() {
+      $('body').find('[data-apos-area]').each(function() {
+        var $area = $(this);
+        var docId = $area.attr('data-doc-id');
+        if (!docId) {
+          return;
+        }
+        if ($area.parents('[data-doc-id="' + docId + '"]').length) {
```
boutell: The recalculate method we'll call recurses through descendants, so avoid those that will be reached that way.

lib/modules/apostrophe-areas/public/js/user.js
```diff
@@ -148,11 +148,95 @@ apos.define('apostrophe-areas', {
...
+          return;
+        }
+        if (!$area.data('editor')) {
+          return;
+        }
+        self.register(docId, dotPath, $area.data('editor'));
+      });
+    };
+
+    self.recalculateDotPathsInArea = function($area) {
```
boutell: This method was moved from "editor.js", it is not new logic in this PR.

lib/modules/apostrophe-areas/public/js/user.js
```diff
@@ -148,11 +148,95 @@ apos.define('apostrophe-areas', {
...
+          self.recalculateDotPathsOfAreasInWidget($widget, docId, widgetDotPath);
+        } catch (e) {
+          // Do not fail outright if a widget has an unusual storage mode
+          apos.utils.warn(e);
+        }
+        ordinal++;
+      });
+    };
+
+    self.recalculateDotPathsOfAreasInWidget = function($widget, docId, dotPath) {
```
boutell: Also moved.

PR #2025, opened 2019-09-13
Fix incontext editing after modal save

bharath6365:
More Details About the Bug => https://www.notion.so/InContext-editing-errors-out-after-modal-editor-is-saved-a27cc8f0c2cc47a9b2c4eecca9eb640d

I'm not completely confident about the fix as I'm not sure what other modules this change would affect.

### DIFF
--- a/lib/modules/apostrophe-areas/public/js/editor.js
+++ b/lib/modules/apostrophe-areas/public/js/editor.js
@@ -937,6 +937,13 @@ apos.define('apostrophe-areas-editor', {
         // if found; they will recursively save our contents
         return callback(null);
       }
+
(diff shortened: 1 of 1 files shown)


boutell:
Thanks for the report! You've got the general idea right but possibly not exactly the right fix, I will have to take a closer look.

bharath6365:
> Thanks for the report! You've got the general idea right but possibly not exactly the right fix, I will have to take a closer look.

Knew something was fishy even though the fix worked fine. Let me know how to proceed.

boutell:
Much appreciated but there is a simpler fix.

Issue #2028, opened 2019-09-13
Required fields inside `object` fields are still required when the object is hidden by `showFields`


boutell:
Behavior should match `array` fields. This impacts SC.

boutell:
Closing in favor of #2029

PR #2029, opened 2019-09-13
fixed a case where required fields nested in an object or array field…

boutell:
… could cause an error even if the parent field is hidden via showFields


boutell:
Closes #2028

PR #2039, opened 2019-09-24
if you subclass apostrophe-rich-text-widgets-editor your subclass edi…

boutell:
…tor should be instantiated


boutell:
Without this you can't change the frontend editor behavior in a subclass of apostrophe-rich-text-widgets.

Issue #2042, opened 2019-09-26
Denial of service via initiating many consecutive batch jobs as a user with suitable privileges


boutell:
As reported by NPM, a DOS can occur when many consecutive batch jobs are initiated by a user with suitable privileges, because there is a `setInterval` leak. While this is not a security vulnerability and can only be carried out by someone who already has privileges to log in, resource leaks are always a bad thing.

PR #2043, opened 2019-09-26
fix setInterval leak leading to DOS if thousands of batch jobs are in…

boutell:
…itiated during process lifetime


boutell:
Closes #2042

PR #2051, opened 2019-10-04
behave reasonably if a scoped npm module name is also used as the nam…

boutell:
…e option of a pieces module

Background: since you can now have a pieces module like @company/articles, an enterprise client reasonably assumed they could set the `name` option of the module to `@company/article`. This didn't work because of issues with inconsistent conversion to a "cssName" friendly for use as an attribute name.

We now convert to a CSS name consistently both when outputting the button and when setting up a click handler for it, thus resolving the issue.

Also the `cssName` function itself now never outputs a leading `-`, which was never desired behavior.

### DIFF
--- a/lib/modules/apostrophe-browser-utils/public/js/always.js
+++ b/lib/modules/apostrophe-browser-utils/public/js/always.js
@@ -62,15 +62,19 @@ apos.define('apostrophe-browser-utils', {
       var i;
       var css = '';
       var dash = false;
+      var start = true;
       for (i = 0; (i < camel.length); i++) {
         var c = camel.charAt(i);
         var lower = ((c >= 'a') && (c <= 'z'));
         var upper = ((c >= 'A') && (c <= 'Z'));
         var digit = ((c >= '0') && (c <= '9'));
         if (!(lower || upper || digit)) {
-          dash = true;
+          if (!start) {
+            dash = true;
+          }
           continue;
         }
+        start = false;
         if (upper) {
           if (i > 0) {
             dash = true;
--- a/lib/modules/apostrophe-utils/lib/api.js
+++ b/lib/modules/apostrophe-utils/lib/api.js
@@ -136,15 +136,19 @@ module.exports = function(self, options) {
     var i;
     var css = '';
     var dash = false;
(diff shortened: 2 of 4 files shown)


lib/modules/apostrophe-browser-utils/public/js/always.js
```diff
@@ -62,15 +62,19 @@ apos.define('apostrophe-browser-utils', {
...
       var dash = false;
+      var start = true;
       for (i = 0; (i < camel.length); i++) {
         var c = camel.charAt(i);
         var lower = ((c >= 'a') && (c <= 'z'));
         var upper = ((c >= 'A') && (c <= 'Z'));
         var digit = ((c >= '0') && (c <= '9'));
         if (!(lower || upper || digit)) {
-          dash = true;
+          if (!start) {
```
abea: Can't we simplify this by checking if `i === 0`, removing the `start` variable?

lib/modules/apostrophe-utils/lib/api.js
```diff
@@ -136,15 +136,19 @@ module.exports = function(self, options) {
...
     var dash = false;
+    var start = true;
     for (i = 0; (i < camel.length); i++) {
       var c = camel.charAt(i);
       var lower = ((c >= 'a') && (c <= 'z'));
       var upper = ((c >= 'A') && (c <= 'Z'));
       var digit = ((c >= '0') && (c <= '9'));
       if (!(lower || upper || digit)) {
-        dash = true;
+        if (!start) {
```
abea: Same here with `i === 0`. Fewer things to keep track of.

lib/modules/apostrophe-browser-utils/public/js/always.js
boutell: That would not work if the string started with several non-alpha-digit characters that should not turn into leading `-`'s.

lib/modules/apostrophe-utils/lib/api.js
boutell: See above re: doing the right thing if there's more than one non-alphanumeric character at the start.

PR #2052, opened 2019-10-04
add apos obj as second argument of dynamic select function

stuartromanek:
should i be handling this another way?


stuartromanek:
ah yes, answer I wanted

PR #2063, opened 2019-10-25
Dgad 862: optional password complexity rules for apostrophe

boutell:
`passwordMinLength`

The minimum length for passwords. You should set this, as there
is no default for bc reasons (effectively the default is `1`).

`passwordRules`

An optional array of password rule names, as strings. The standard rules
available are `noSlashes`, `noSpaces`, `mixedCase`, `digits`, and
`noTripleRepeats`. The `noTripleRepeats` rule forbids repeating a
character three times in a row. By default no rules are in effect.

When this option is set, the rules are consulted when a password
is set or reset. Existing passwords that do not follow the rules
are tolerated. If you wish to enforce them for existing passwords
as well, see below.

`passwordRulesAtLoginTime`

By default, password rules are enforced only when a password is 
being set or reset. If you wish, you can set `passwordRulesAtLoginTime: true`
to apply them at login time so that even an existing password will
not work unless it passes the rules. This can be useful if you don't
mind a few irritated users and you have the `passwordReset` option
properly configured (see below).

### DIFF
--- a/lib/modules/apostrophe-login/index.js
+++ b/lib/modules/apostrophe-login/index.js
@@ -307,7 +307,7 @@ module.exports = {
             // failureFlash does not appear to work properly, let's just
             // use a query parameter
             failureRedirect: '/login?' + qs.stringify({
-              message: 'Incorret login or password',
+              message: 'Incorrect login or password',
               // bc
               error: '1'
             })
--- a/lib/modules/apostrophe-login/index.js
+++ b/lib/modules/apostrophe-login/index.js
@@ -274,17 +274,17 @@ module.exports = {
             return res.redirect('/');
           }
           req.scene = 'user';
-          // message is supported as a way of delivering all errors
-          // for bc but if present errors is shown by preference
-          // by the modern template
+          // message is supported as a simple way of delivering all errors
+          // but we also provide `errors` for templates that wish to do more
           return self.sendPage(req, 'login', { passwordReset: self.options.passwordReset, message: getMessage(), errors: req.query.errors });
           function getMessage() {
+            let message;
             if (req.query.message) {
-              return req.query.message;
+              message = req.query.message;
+            } else if (req.query.errors && req.query.errors.length) {
+              message = req.query.errors.join(', ');
             }
-            if (req.query.errors && req.query.errors.length) {
-              return req.query.errors.join(', ');
-            }
+            return self.apos.utils.capitalizeFirst(message);
           }
         });
         self.apos.app.post('/login',
@@ -375,12 +375,17 @@ module.exports = {
           });
 
           function getMessage() {
+            let message;
             if (req.query.message) {
-              return req.query.message;
+              message = req.query.message;
             }
             if (req.query.errors && req.query.errors.length) {
-              return req.query.errors.join(', ');
+              message = req.query.errors.join(', ');
             }
+            if (message === undefined) {
+              return undefined;
+            }
+            return self.apos.utils.capitalizeFirst(message);
           }
         });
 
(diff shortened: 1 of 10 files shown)


boutell:
Sample config in `data/local.js`:

```
module.exports = {
  modules: {
    'apostrophe-login': {
      passwordMinLength: 10,
      passwordRules: [ 'noSlashes', 'noSpaces', 'mixedCase', 'digits', 'noTripleRepeats' ],
      passwordRulesAtLoginTime: true
    }
  }
};
```

boutell:
Other new things of note in this PR:

* The `editor-modal.js` of pieces now has a displayError method invoked after a server-side save error, which, by default, does the existing thing: display a notification. This is a useful point at which to override.

* In `apostrophe-users`, an override of this creates and displays the full set of password rule errors when needed.

* To facilitate passing an array of error messages up from deeply nested code without a big bc break in how errors are propagated in Apostrophe, in the event of an error, the contents of `req.errorMessages` are now included in the `messages` property of the response from the APIs that save pieces. `req.errorMessages` can be populated anywhere, even if a simple string error must be passed up from that point for compatibility reasons.

* Password schema fields always use `autocomplete="off"`. Note this does not interfere with login, which does not use schema based forms. It's to prevent weirdness when editing an account as an admin, etc.

Care has been taken to preserve bc, even with existing template overrides.

lib/modules/apostrophe-login/index.js
```diff
@@ -315,22 +368,52 @@ module.exports = {
...
 
         self.apos.app.post('/password-reset', function(req, res) {
           var reset = self.apos.launder.string(req.body.reset);
           var email = self.apos.launder.string(req.body.email);
           var password = self.apos.launder.string(req.body.password);
-          if ((!reset.length) || (!password.length)) {
-            return res.redirect('/password-reset-request?error=missing');
+          if (!reset.length) {
+            return res.redirect('/password-reset?' + qs.stringify({
+              errors: [ 'The password reset code is missing from your link. Try resetting your password again.' ],
```
abea: Suggestion: "The password reset code is missing from your URL. Click the link in your reset email or try resetting your password again."

lib/modules/apostrophe-login/index.js
```diff
@@ -315,22 +368,52 @@ module.exports = {
...
             return res.redirect('/login?message=' + req.__('That reset code was not found. It may have expired. Try resetting again.'));
           });
+
+          function getMessage() {
+            let message;
+            if (req.query.message) {
+              message = req.query.message;
+            }
+            if (req.query.errors && req.query.errors.length) {
+              message = req.query.errors.join(', ');
```
abea: This seems like it'll read weirdly. Why not write the messages as sentences and join with a space?

lib/modules/apostrophe-login/index.js
```diff
@@ -454,6 +537,89 @@ module.exports = {
...
+      });
+      return errors;
+    };
+
+    self.passwordRules = {
+      noSlashes: {
+        test: function(password) {
+          return !((password.indexOf('/') !== -1) || (password.indexOf('\\') !== -1));
+        },
+        message: '/ and \\ characters are not allowed in password'
```
abea: I guess the assumption with these is that they would be overridden for non-English?

boutell: No, they are passed through `__` later.

lib/modules/apostrophe-login/index.js
boutell: That... would take less logic. Will do.

boutell:
Implemented your suggestion, merging.

Issue #2065, opened 2019-10-25
Need a way to build image URLs from attachment objects in lean mode


boutell:
Currently not possible to do this, and it is common to have various joined images come down the pipe from an API, so insisting it be done via templates and data attributes isn't really fair.

Issue #2064, opened 2019-10-25
apos.attachments.url doesn't work browser-side logged out


boutell:
`self.uploadsUrl` is not being properly set.

PR #2066, opened 2019-10-25
you should be able to determine an attachment url in lean mode with a…

boutell:
…pos.utils.attachmentUrl. Also fixed apos.attachments.url for the anon scene when lean is not in effect.


boutell:
Closes #2064 closes #2065

PR #2067, opened 2019-10-28
updated links for new documentation locations and to fix existing errors

stephenkostas:
@boutell - here are the updates for the links. If I missed any, they should be caught by the redirects.

### DIFF
--- a/lib/modules/apostrophe-admin-bar/index.js
+++ b/lib/modules/apostrophe-admin-bar/index.js
@@ -1,7 +1,7 @@
 // The admin bar module implements Apostrophe's admin bar at the top of the screen. Any module
 // can register a button (or more than one) for this bar by calling the `add` method of this
 // module. Buttons can also be grouped into dropdown menus and restricted to those with
-// particular permissions. [apostrophe-pieces](tutorials/core-concepts/reusable-content-pieces/README.md) automatically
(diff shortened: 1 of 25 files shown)


lib/modules/apostrophe-admin-bar/index.js
```diff
@@ -1,7 +1,7 @@
 // The admin bar module implements Apostrophe's admin bar at the top of the screen. Any module
 // can register a button (or more than one) for this bar by calling the `add` method of this
 // module. Buttons can also be grouped into dropdown menus and restricted to those with
-// particular permissions. [apostrophe-pieces](../apostrophe-pieces/README.md) automatically
+// particular permissions. [apostrophe-pieces](tutorials/core-concepts/reusable-content-pieces/README.md) automatically
```
abea: Why does this one not start with a leading slash but the others do?

abea: So probably typo there. If you've checked that all these work then it looks good to me.

stephenkostas: Yeah. That was a typo. I just fixed that one and visually verified everything again. I'm going to manually check everything after we merge it. I also updated the generated MD files in the docs repo so that they'll be correct from the start without having to regenerate them.

PR #2069, opened 2019-10-29
expose res.rawRedirect

boutell:
Provide access to the original version of res.redirect, which does not automatically add the sitewide URL prefix. This is needed if you are building a URL to redirect to starting with a `_url` property, which will already contain the prefix. See:

https://github.com/apostrophecms/apostrophe-workflow/issues/268

### DIFF
--- a/lib/modules/apostrophe-express/index.js
+++ b/lib/modules/apostrophe-express/index.js
@@ -254,9 +254,11 @@ module.exports = {
         // Use middleware to patch the redirect method to accommodate
         // the prefix. This is cleaner than patching the prototype, which
         // breaks if you have multiple instances of Apostrophe, such as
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-express/index.js
```diff
@@ -265,7 +267,7 @@ module.exports = {
             if (!url.match(/^[a-zA-Z]+:/)) {
               url = self.apos.prefix + url;
             }
-            return superRedirect.call(this, status, url);
+            return res.rawRedirect.call(res, status, url);
```
abea: It looks like this no longer needs to use `.call`. `res.rawRedirect(status, url)`. eslint is throwing on that.

Issue #1247, opened 2018-02-22
Updating an area in a piece should fire beforeSave, etc.


boutell:
Right now, the areas module doesn't attempt to use the update handler of the piece, it goes straight to docs.update. If a manager can be found and it has an update method it should be used. cc @tortilaman

boutell:
Also a good first issue for those curious about the node-y bits of the code. Hint: the file you'll be working in is `lib/modules/apostrophe-areas/lib/api.js`. The `saveArea` method calls `self.apos.docs.update`, but that is low level and it bypasses the special handling the developer might want for a particular type of piece. 

Instead find the manager for the type:

`var manager = self.apos.docs.getManager(doc.type)`

If the manager exists and has an `update` method, call that instead. If both conditions are not met, go ahead and call `docs.update`.

kalidis:
I was trying to do this pull request but I have no idea that which manager could have a update method, could you give my an example?

boutell:
You can use an if statement to see whether manager.update exists or not.

PR #2070, opened 2019-10-30
Updating an area in a piece

kalidis:
Solving the issue #1247

### DIFF
--- a/lib/modules/apostrophe-areas/lib/api.js
+++ b/lib/modules/apostrophe-areas/lib/api.js
@@ -232,7 +232,12 @@ module.exports = function(self, options) {
           type: 'area',
           items: items
         });
-        return self.apos.docs.update(req, doc, callback);
+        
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-areas/lib/api.js
```diff
@@ -232,7 +232,12 @@ module.exports = function(self, options) {
           type: 'area',
           items: items
         });
-        return self.apos.docs.update(req, doc, callback);
+        
+        if(self.apos.docs.getManager(doc.type) != undefined && self.apos.docs.getManager(doc.type).update != undefined){
```
boutell: Thanks!

A few quick things:

* Avoid redundant code, please set a manager variable rather than fetching it twice.
* A simple if without != undefined would be more appropriate.
* Also please follow the coding style, space after "if", space before "callback".

boutell: You can type `npx eslint .` to see what coding style issues are flagged automatically. `npx eslint . --fix` can fix some.

PR #2076, opened 2019-11-06
Only add footer height to modal if footer is present

paulisloud:
Previously this method would always undersize the modal by 50px. I believe 50px was a stand-in for footerHeight, so this commit replaces the hardcoded value with the calculated footer height. In my testing, this change does work to:
* Size the window appropriately if there is a footer present in the modal
* Size the window to leave no space for the footer if no footer is present


paulisloud:
Would it be possible to get a review on this? Fairly trivial change here.

stuartromanek:
sorry about that @paulisloud .. will test this today!

stuartromanek:
thanks @paulisloud !

paulisloud:
Great, thanks @stuartromanek!

PR #2078, opened 2019-11-11
fast edit virtual areas support via debouncing and a route to handle …

boutell:
…many at once


boutell:
cc @bharath6365

bharath6365:
> cc @bharath6365

Boutell, we can go ahead with this approach. Thanks.

boutell:
This PR accelerates opening a modal when there are hundreds of rich text areas.

PR #2080, opened 2019-11-12
Update README.md with the right link for doc

Guillaume-N:
The "ApostropheCMS documentation" link in the "Getting started" section currently points to a not found page. I edited it with what seems to be the right link.


boutell:
Thanks!

PR #2089, opened 2019-11-21
changed padding on page title column

kingsleytorlowei:


boutell:
I appreciate the PR but I don't think this fixes the issue fully. Here's a screenshot with your fix in a test situation. Note that the spacing between levels varies. There's something else messing up the nesting overall.

The trashcan here should be visually nested under "Child."

Screen Shot 2019-11-21 at 1 26 22 PM

PR #2090, opened 2019-11-22
Fast add user task

boutell:
cc @paulisloud this should do the trick for you, but I don't have write access in your environment this morning, so please apply this change yourself and let me know.

### DIFF
--- a/lib/modules/apostrophe-pages/public/js/reorganize.js
+++ b/lib/modules/apostrophe-pages/public/js/reorganize.js
@@ -145,7 +145,7 @@ apos.define('apostrophe-pages-reorganize', {
       // add page depth as data attr
       $li.attr('data-apos-reorganize-depth', node.getLevel());
       if (node.getLevel() !== 1) {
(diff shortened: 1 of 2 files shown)


paulisloud:
@boutell  Works beautifully! Thanks.

lib/modules/apostrophe-pages/public/js/reorganize.js
```diff
@@ -145,7 +145,7 @@ apos.define('apostrophe-pages-reorganize', {
       // add page depth as data attr
       $li.attr('data-apos-reorganize-depth', node.getLevel());
       if (node.getLevel() !== 1) {
-        $li.find('.apos-reorganize-title').css('padding-left', 15 * node.getLevel() + 'px');
+        $li.find('.apos-reorganize-title').css('padding-left', 30 * node.getLevel() + 'px');
```
abea: This looks unrelated. What bringing this in intentional?

boutell:
Fixed that issue, merging.

PR #2096, opened 2019-12-02
aspect ratio support for direct attachment properties in a doc schema

boutell:
Attachments can be direct properties of any doc type, not just images or files. And we use this feature, but we have mostly used it for PDFs and the like. Now I want to use it for an image in apostrophe-global to better implement apostrophe-favicons 3.x, and I run into the fact that we never fleshed out support for the aspectRatio option at the attachment level. Introduce support for this, as well as the necessary autocropping if the user doesn't manually crop (this is not new UX, it's taken directly from how we handle it when choosing images if the aspect ratio does not match and the user does not crop).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.100.2 (2019-12-02)
+
+* The `aspectRatio` option to the `attachments` schema field type is now fully implemented. We always had this for selecting images, e.g. in our `apostrophe-images-widgets` module, but it is now also available when directly using an `attachment` schema field as a property of your own doc. You can also set `crop: true` to allow manual cropping in that case. This is a useful technique when including the image in a reusable media library does not make sense.
+
 ## 2.100.1 (2019-11-21)
 
 * Must confirm when resetting password, since there are no do-overs if we do not have the email confirmation method available (with `resetLegacyPassword: true`) and since it's generally a pain not to have this.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apostrophe",
-  "version": "2.100.1",
+  "version": "2.100.2",
   "description": "The Apostrophe Content Management System.",
   "main": "index.js",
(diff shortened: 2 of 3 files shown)


boutell:
To clarify, when choosing an image for a widget, the crop is a property of the relationship between the widget and the image, and not a property of the attachment itself. That makes sense when the image is part of a reusable media library and may be cropped in many ways for many placements. But for a one-off image like a favicon, this does not make sense: we want the crop to be a property of the attachment itself.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "apostrophe",
-  "version": "2.100.1",
+  "version": "2.100.2",
```
abea: Wouldn't this be a minor version bump?

boutell: The feature always existed in a busted form, so I regard it as a bug fix. Semantics but I'm good with it.

boutell:
@abea any concerns with this one? I can still revise. Thanks.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.100.3 (2019-12-03)
+
+* The `aspectRatio` option to the `attachments` schema field type is now fully implemented. We always had this for selecting images, e.g. in our `apostrophe-images-widgets` module, but it is now also available when directly using an `attachment` schema field as a property of your own doc. You can also set `crop: true` to allow manual cropping in that case. This is a useful technique when including the image in a reusable media library does not make sense.
```
abea: The description here implies that you can use `aspectRatio` independently from `crop`. But if I only set the `aspectRatio` property it doesn't auto crop until I set `crop: true` as well. That makes sense with the conditional written on line 187 of user.js.

So we should either update this or edit the conditional to allow auto cropping. I gather the latter.

boutell: Good point. I think aspectRatio should full-on imply crop because there isn't anything to be gained by denying the user a little more control over what must happen anyway.

PR #2097, opened 2019-12-02
thanks to sergiodop92 for this cross browser fix

boutell:


boutell:
Prior to this fix the back button didn't work properly with our pieces pages AJAX functionality, at least in some browsers.

PR #2100, opened 2019-12-04
adding eps file type

livetodeliver:
Making this request per the instruction of @boutell to enable EPS files as a valid file attachment for `apostrophe-attachments`.


boutell:
Cool. This can go out in the next apostrophe release which is highly probable within a week.

PR #2110, opened 2019-12-18
schema field: property permission and readOnly also accept a callback

FerX:
added the possibility to use callbacks in the value of the permission and readOnly fields.

### DIFF
--- a/lib/modules/apostrophe-doc-type-manager/lib/api.js
+++ b/lib/modules/apostrophe-doc-type-manager/lib/api.js
@@ -254,6 +254,7 @@ module.exports = function(self, options) {
       return (!field.permission) || self.apos.permissions.can(req, field.permission);
     });
     schema = schema.map(function(field) {
+      field = _.cloneDeep(field);
       if (_.isFunction(field.readOnly)) {
         field.readOnly = !!(field.readOnly(self, req, field));
       }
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-doc-type-manager/lib/api.js
```diff
@@ -242,13 +242,23 @@ module.exports = function(self, options) {
...
     var type;
     var schema = _.filter(self.schema, function(field) {
+      if (_.isFunction(field.permission)) {
+        return !!(field.permission(self, req, field));
+      }
       return (!field.permission) || self.apos.permissions.can(req, field.permission);
     });
+    schema = schema.map(function(field) {
+      if (_.isFunction(field.readOnly)) {
+        field.readOnly = !!(field.readOnly(self, req, field));
```
boutell: You are modifying `field` so everyone will get the same answer in the future, which is not what you want. To be safe you would need to make a shallow clone of `field` before modifying it.

abea:
The method name approach would be consistent. I guess I don't have a strong opinion on allowing a function, but we _should_ allow the method name string _as well_ if we do this.

boutell:
We should go with the approach of accepting a string which is invoked as a method of the module, rather than an inline function that takes self directly.

boutell:
Un-staling this. Very close to acceptance, would be a shame if it were closed for staleness.

@FerX would you mind adding support for a method name as a string rather than a function? Functions in the schema that take self explicitly are avoided elsewhere, for instance for dynamic select choices. We like the idea that everything in the schema can be mirrored to the browser side in case it's useful information there in some situation. Inline functions tend to break that.

PR #2111, opened 2019-12-26
Disable account if not logged for a long period

falkodev:


lib/modules/apostrophe-login/index.js
```diff
@@ -296,12 +297,13 @@ module.exports = {
               return callback(null, false);
             }, 1000);
           }
-          return callback(err, user);
```
boutell: `err` is definitely null at this point (see lines 294 through 299), which was unclear because I passed it to the old callback rather than passing `(null, user)`, but let's not compound that by passing it on to the new `disableIfInactive` method.

lib/modules/apostrophe-login/index.js
```diff
@@ -311,6 +313,37 @@ module.exports = {
       });
     };
 
+    self.disableIfInactive = function(err, user, callback) {
```
boutell: Remove the `err` argument.

lib/modules/apostrophe-login/index.js
```diff
@@ -311,6 +313,37 @@ module.exports = {
...
+      if (self.apos.users.options.disableInactiveAccounts) {
+        var usersOptions = self.apos.users.options.disableInactiveAccounts;
+
+        if (user.lastLogin) {
+          // by default, "admin" users are not disabled
+          var isUserWhitelisted = user._groups.some(function(group) {
```
boutell: Let's change "whitelist" to "neverList".

lib/modules/apostrophe-login/index.js
```diff
@@ -311,6 +313,37 @@ module.exports = {
...
+          });
+
+          if (!isUserWhitelisted) {
+            var lastLogin = moment(user.lastLogin);
+            var now = moment();
+            var diff = lastLogin.diff(now, 'days');
+
+            if (diff >= usersOptions.inactivityDuration) {
+              var req = self.apos.tasks.getReq();
+              user = Object.assign({}, user, { disabled: true, trash: false, lastLogin: '' });
```
boutell: Use `apos.docs.db.update`, i.e. use the mongodb docs collection, so you can use `$set` and avoid a possible race condition as well as your Object.assign step.

lib/modules/apostrophe-login/index.js
```diff
@@ -311,6 +313,37 @@ module.exports = {
...
+              return self.apos.users.update(req, user)
+                .then(function() {
+                  return callback('Account disabled due to inactivity. Please, refer to the administrator of the site for assistance.', false);
+                }).catch(function(err) {
+                  self.apos.utils.error(err);
```
boutell: Report this error to the original callback too, otherwise it looks like success to the user.

lib/modules/apostrophe-login/index.js
```diff
@@ -311,6 +313,37 @@ module.exports = {
...
+                });
+            }
+          }
+        }
+      }
+      return callback(err, user);
```
boutell: This can just be `(null, user)` since we don't need to pass in an error argument anymore.

lib/modules/apostrophe-login/index.js
```diff
@@ -853,15 +886,23 @@ module.exports = {
         }
       }
 
-      return self.callAllAndEmit('loginAfterLogin', 'after', req, function(err) {
-        if (err) {
-          self.apos.utils.error(err);
-          return res.redirect('/');
-        }
+      var user = Object.assign({}, req.user, { lastLogin: new Date() });
```
boutell: This needs to become a mongodb `$set` operation.

test/login.js
```diff
@@ -142,4 +158,59 @@ describe('Login', function() {
...
+
+    apos.users.insert(apos.tasks.getReq(), user, function(err) {
+      assert(!err);
+      return request.post('http://localhost:7901/login', {
+        form: { username: 'random-test', password: 'crookshanks' },
+        followAllRedirects: true,
+        jar: loginLogoutJar
+      }, function(err, response, body) {
+        assert(!err);
+        assert.equal(response.statusCode, 500);
```
boutell: Don't return `500` for something that is not a failure of the system. Use an appropriate 400 series status code.

test/login.js
```diff
@@ -142,4 +158,59 @@ describe('Login', function() {
...
+  it('should disable an inactive user', function(done) {
+    var user = apos.users.newInstance();
+
+    user.firstName = 'Random';
+    user.lastName = 'Test';
+    user.title = 'Random Test';
+    user.username = 'random-test';
+    user.password = 'crookshanks';
+    user.email = 'randomtest@aol.com';
+    user.lastLogin = new Date();
```
boutell: If their last login was just a moment ago, shouldn't they be able to get in again (they are not not inactive)?

lib/modules/apostrophe-login/index.js
falkodev: Done

lib/modules/apostrophe-login/index.js
falkodev: Done

lib/modules/apostrophe-login/index.js
falkodev: Change to `neverDisabled`

lib/modules/apostrophe-login/index.js
falkodev: Done

lib/modules/apostrophe-login/index.js
falkodev: Done

lib/modules/apostrophe-login/index.js
falkodev: Done

test/login.js
falkodev: In this test, I configured the parameter `inactivityDuration` to 0 in purpose. This way, we check the user is really disabled after a login connection if the inactivity duration is reached.

test/login.js
falkodev: If we want to inform the user its accound has been disabled, we need to override the traditional error message "Login or password incorrect". I did not succeed. I wanted to discuss that with you to check if you know the right move to make here. Meanwhile, I put this to have something working (I will update the test afterwards).

test/login.js
boutell: Did you also include a test that it doesn't incorrectly time out users who have not timed out?

boutell: If so then the 0 approach for this test of a positive case is OK.

lib/modules/apostrophe-login/index.js
falkodev: Done

test/login.js
falkodev: Yes. As I configured the `apos` object with a `inactivityDuration` with 0, every existing "logging" tests in this file check if non-timed out users can log in (line 99 and line 115).

And I also check if a user belonging to `neverDisabledGroups` ("admin" in this file), even timed-out, can still log in (line 161)

test/login.js
falkodev: Ok with custom callback from `passport`. I updated the test too

test/login.js
falkodev: I added the missing test

lib/modules/apostrophe-login/index.js
```diff
@@ -296,12 +297,13 @@ module.exports = {
...
             }, 1000);
           }
-          return callback(err, user);
+          return self.disableIfInactive(user, callback);
         });
       });
     };
 
     self.verifyTotp = function(user, done) {
+      self.disableIfInactive(user, done);
```
boutell: Lack of a return statement here means callbacks will be double-invoked. This must be reconsidered so that if we do disable it we do not continue into the other code below it.

falkodev: Ok, here is a new attempt. What do you think @boutell ?

lib/modules/apostrophe-login/index.js
```diff
@@ -305,12 +306,51 @@ module.exports = {
...
+            key: userSafe && userSafe.totp && userSafe.totp.key,
+            period: userSafe && userSafe.totp && userSafe.totp.period
+          }
+        );
       }).catch(function(err) {
         return done(err);
       });
     };
 
+    self.disableIfInactive = function(user, callback, totp) {
```
boutell: This function now has a muddled purpose. It handles disabling if inactive, but it also handles returning the totp information, which is messy.

It should have a single purpose, and the piece of code that cares about both tasks should be responsible for handling the totp information when there is no error from it.

falkodev: Ok @boutell 
Here is the new proposal

lib/modules/apostrophe-login/index.js
```diff
@@ -305,12 +306,54 @@ module.exports = {
       return self.apos.users.safe.findOne({
         _id: user._id
       }).then(function(userSafe) {
-        return done(null, userSafe && userSafe.totp && userSafe.totp.key, userSafe && userSafe.totp && userSafe.totp.period);
+        return self.checkIfActive(userSafe, done);
```
boutell: Code for TOTP (Google Authenticator) was removed here and did not come back anywhere else.

You need to pass an inline function as the callback of checkIfActive, and in that callback, do what the original code did.

Do not move that code into checkIfActive, as that would be poor separation of concerns. Just make sure the callback given to checkIfActive does the original job fully.

boutell: Google Authenticator support should be retested with this code.

falkodev: Ok, tested with Google auth and working

lib/modules/apostrophe-login/index.js
```diff
@@ -309,12 +310,56 @@ module.exports = {
       return self.apos.users.safe.findOne({
         _id: user._id
       }).then(function(userSafe) {
-        return done(null, userSafe && userSafe.totp && userSafe.totp.key, userSafe && userSafe.totp && userSafe.totp.period);
+        return self.checkIfActive(userSafe, function() {
```
boutell: You are ignoring any errors from checkIsActive here.

lib/modules/apostrophe-login/index.js
```diff
@@ -309,12 +310,56 @@ module.exports = {
...
+          return done(null, userSafe && userSafe.totp && userSafe.totp.key, userSafe && userSafe.totp && userSafe.totp.period);
```
boutell: This part is good.

lib/modules/apostrophe-login/index.js
falkodev: Od, done

PR #2122, opened 2020-01-14
enable default values for pieces

falkodev:
For https://git.dgadteamdev.com/pa/apostrophe-enhancements/issues/910#note_386326

Not sure this simple modification does the job. I thought it would require many modifications in multiple files (as I did for pages). I looked into many files the way it worked previously but did not find anything. You might point me in the right direction if I forgot something.


boutell:
I don't think it could, no. For one, the property is `def`. Also, the data should already be in `piece` because it comes from `newInstance` with is aware of `def`. Has this fix been tested? Do you have a recommendation on how to demo the original bug and the fix?

falkodev:
Oh, you are right..It's `def` not `default`. I was misled by Jonathan's example. But it is wrong..

PR #2124, opened 2020-01-15
add option loginUrl

GiuseppeCM:
add option loginUrl to change the default login url.

### DIFF
--- a/lib/modules/apostrophe-login/index.js
+++ b/lib/modules/apostrophe-login/index.js
@@ -353,16 +353,9 @@ module.exports = {
       return next();
     };
 
-    // Check if the first character about the option `loginUrl` is `/`
-    // and if is not present, add the `/` in the first position.
-    // If the option is not present, return the default url `/login`
-    self.fetchLoginUrl = function() {
-      if (self.options.loginUrl) {
-        return self.options.loginUrl && self.options.loginUrl.startsWith('/') ? self.options.loginUrl : `/${self.options.loginUrl}`
-      }
-      else {
-        return '/login'
-      }
+    // return the loginUrl option
+    self.getLoginUrl = function() {
+      return self.options.loginUrl ? self.options.loginUrl : "/login"
     }
 
     // Add the `/login` route, both GET (show the form) and POST (submit the form).
@@ -370,7 +363,7 @@ module.exports = {
 
     self.addRoutes = function() {
       if (self.options.localLogin !== false) {
-        self.apos.app.get(self.fetchLoginUrl(), function(req, res) {
+        self.apos.app.get(self.getLoginUrl(), function(req, res) {
           if (req.user) {
             // User is already logged in, redirect to home page
             return res.redirect('/');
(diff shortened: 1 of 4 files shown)


lib/modules/apostrophe-login/index.js
```diff
@@ -349,12 +353,24 @@ module.exports = {
       return next();
     };
 
+    // Check if the first character about the option `loginUrl` is `/`
+    // and if is not present, add the `/` in the first position.
+    // If the option is not present, return the default url `/login`
+    self.fetchLoginUrl = function() {
+      if (self.options.loginUrl) {
+        return self.options.loginUrl && self.options.loginUrl.startsWith('/') ? self.options.loginUrl : `/${self.options.loginUrl}`
```
boutell: We should skip this mechanism, the developer is responsible for including the leading / on all other URL based options and I don't want to cause any confusion about that.

lib/modules/apostrophe-login/index.js
```diff
@@ -349,12 +353,24 @@ module.exports = {
...
+      }
+    }
+
     // Add the `/login` route, both GET (show the form) and POST (submit the form).
     // Also add the `/logout` route.
 
     self.addRoutes = function() {
       if (self.options.localLogin !== false) {
-        self.apos.app.get('/login', function(req, res) {
+        self.apos.app.get(self.fetchLoginUrl(), function(req, res) {
```
boutell: This URL appears in many other places: for CSRF protection, for redirects from other steps and for other things. You will need to switch the code to call getLoginUrl in each case.

lib/modules/apostrophe-login/index.js
```diff
@@ -349,12 +353,24 @@ module.exports = {
       return next();
     };
 
+    // Check if the first character about the option `loginUrl` is `/`
+    // and if is not present, add the `/` in the first position.
+    // If the option is not present, return the default url `/login`
+    self.fetchLoginUrl = function() {
```
boutell: getLoginUrl would be more consistent with the rest of the application.

lib/modules/apostrophe-login/views/passwordResetRequest.html
```diff
@@ -26,7 +26,7 @@
               <input type="submit" class="apos-button apos-button--major" value="{% block buttonLabel %}{{ __('Request Password Reset') }}{% endblock %}" />
             </div>
             <div class="apos-login-reset-cancel">
-              <a href="{{ apos.prefix }}/login">{{ __('Cancel') }}</a>
+              <a href="{{ apos.prefix }}{{data.loginUrl}}">{{ __('Cancel') }}</a>
```
boutell: Where is this getting into the template data?

GiuseppeCM: https://github.com/apostrophecms/apostrophe/pull/2124/commits/32337ca76f483931dd45935611858115cff82fa6#diff-297b2881cc29c9a42671c4bc64f95986R454 this?

boutell:
Please fix eslint issues:

```
npx eslint . --fix
```

And manually address anything that points out as not fixable automatically.

Issue #1396, opened 2018-05-18
Chooser choices: if `limit: 1`, don't display directional move UI


stuartromanek:

stuartromanek:
PR here https://github.com/apostrophecms/apostrophe/pull/2126

abea:
Should we reopen this since the PR is closed?

stuartromanek:
the issue is open, the pr is closed

PR #2126, opened 2020-01-17
if chooser is set to limit 1, remove reorder controls. also disable p…

stuartromanek:
…osition UI for items that dont make sense, ie the first item cant move up any further

address #1396 and it's comments


boutell:
Does it behave when you move the item so it is no longer first? Not sure we re-render the whole thing in that situation. Asking if the up button comes back or not.

stuartromanek:
it does

stuartromanek:
ah it doest when adding or removing stuff but not reordering stuff, putting this on blocks for the mo'

PR #2131, opened 2020-01-21
WIP: replace synchronous save on beforeunload with a warning and a subsequ…

boutell:
…ent indicator display until save is complete

Our existing "save when closing the page" mechanism does not work in most non-Chrome browsers anymore, and it is not reliable in Chrome. This is because all browsers are deprecating or ignoring attempts to make synchronous HTTP requests during beforeunload events.

This is a major concern for an enterprise client.

The resolution is to use a beforeunload prompt instead. These are still fully supported, although the actual text you return is not displayed in any modern browser (instead, you get a standardized prompt, which is fine).

That much is simple. However, after the prompt appears, the user needs some indication of whether enough time has passed to save the content and they can close the page.

As a WIP, I resolved this with a simple "Saving..." fake button in the context menu.

I put it down there because that is always on the screen. The admin bar and logo are not always on the screen so I am concerned that they are inappropriate if you are typing halfway down a long page, but I defer to you on the UX decision there.

Note that this "Saving..." indicator only appears if you try to leave to leave the page *and* there are unsaved changes (type a bunch and quickly try to close the tab). Of course it would be great to have "saving..." indicators all the time as well as a "saved" indicator that disappears when you start to type. However, that is much more technically ambitious for several reasons, so I would prefer to save it for 3.x.

You will note that "Saving..." usually only appears for a moment. That's because considerable time passes while you are looking at the prompt, and the next save operation arrives within five seconds, which may have already elapsed by the time you hit "cancel" to permit execution to continue in this way.

### DIFF
--- a/lib/modules/apostrophe-areas/public/js/editor.js
+++ b/lib/modules/apostrophe-areas/public/js/editor.js
@@ -755,6 +755,7 @@ apos.define('apostrophe-areas-editor', {
       if (!request) {
         return callback && callback(null);
       }
+      apos.areas.markUnsaved();
       $.jsonCall(
         self.options.action + '/save-area',
         {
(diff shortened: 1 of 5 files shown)


stuartromanek:
Added a Saved! message with delayed fadeout when save operation is complete.

I think it would be worthwhile to prototype the 'display this message during all save operations' to see if that is helpful or annoying in the current design system

boutell:
🔥😀

PR #2132, opened 2020-01-22
enable attachment to be non-required

falkodev:
I stumbled upon an error with schemas having an optional attachment. If I had no attachment, there was an error in the browser console saying "cannot read crop of undefined in user.js". I figured out it might be due to recent changes in this file.

If the attachment is null, the error was triggered. So, here is a fix to not enter in the `if` condition where no attachement added by the user in the apostrophe modal of a piece.


boutell:
Good catch.

PR #2133, opened 2020-01-22
WIP: get copy-page menu and modal

falkodev:


lib/modules/apostrophe-pages/lib/api.js
```diff
@@ -1168,6 +1168,12 @@ module.exports = function(self, options) {
           return item.action !== 'insert-page';
         });
       }
+
+      if (!args.edit && !args.contextMenu && self.apos.permissions.can(req, 'view-page')) {
+        args.contextMenu = _.filter(self.options.contextMenu, function(item) {
+          return item.action === 'copy-page'
```
boutell: Is this about allowing the user to start from the page they want to copy?

The idea is that you still start from the page you want to be the new parent, but there's a join inside the page settings modal which defaults to that page but can be used to pick a different source to copy from.

lib/modules/apostrophe-pages/lib/routes.js
```diff
@@ -292,7 +292,7 @@ module.exports = function(self, options) {
     if (!manager) {
       return next('notfound');
     }
-    return self.find(req, { _id: _id }).ancestors(true).permission('edit-apostrophe-page').toObject(function(err, page) {
+    return self.find(req, { _id: _id }).ancestors(true).permission('view-apostrophe-page').toObject(function(err, page) {
```
boutell: There should be no change in how we handle the parent page.

The change is that we now have a new join in the schema for picking the source to be copied from. Which defaults to the parent page, but can be browsed to pick another source.

PR #2134, opened 2020-01-22
Fixes urls for documentation

abea:
Changing these here to not be overwritten in documentation when generated as done here: https://github.com/apostrophecms/apostrophe-documentation/pull/216.

### DIFF
--- a/lib/modules/apostrophe-widgets/index.js
+++ b/lib/modules/apostrophe-widgets/index.js
@@ -102,7 +102,7 @@
 // were passed to the widget.
 //
 // For example, here is the `public/js/always.js` file for the
-// [apostrophe-video-widgets](../apostrophe-video-widgets/README.md) module:
+// [apostrophe-video-widgets](/modules/apostrophe-video-widgets) module:
 //
 // ```javascript
 // apos.define('apostrophe-video-widgets', {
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-widgets/index.js
```diff
@@ -102,7 +102,7 @@
 // were passed to the widget.
 //
 // For example, here is the `public/js/always.js` file for the
-// [apostrophe-video-widgets](../apostrophe-video-widgets/index.html) module:
+// [apostrophe-video-widgets](../apostrophe-video-widgets/README.md) module:
```
boutell: Why relative here and absolute above?

abea: It's a fair question. I don't know. I was matching against what we currently have in the live docs, which I [reverted to in the related PR](https://github.com/apostrophecms/apostrophe-documentation/pull/216/commits/ca0d44c8021b20e990f19ba4cd66957947c855d1).

abea: @stephenkostas Do you know the answer to that?

stephenkostas: I may have fixed the links at different times because there were already a mix of working and non-working links in this section. My later judgement was to use relative links for anything within the modules section, and absolute links for everything else, so module links that are relative are newer, but I didn't go through the readjust ones that were already good.

abea: So the absolute one here ("/tutorials/schema-guide/schema-guide.md") should be made relative?

stephenkostas: For the relative links, I meant only for links from one module to another. So if it's linking into another section of the doc, I would use an absolute link.

The reason being that from modules, you would need to ../../ to root anyway before getting to /tutorials or /other, so it's cleaner to just use the absolute link.

PR #2137, opened 2020-01-25
no edit pencil for pages in joins, since it cannot be supported witho…

### LINKED ISSUE #2125: Edit UI in Page Join chooser doesn't work
**To Reproduce**
1. Make a schema with a join to a **page**
2. Add a page to join
3. Click the edit pencil and see the nothing happens

error in the front end console is 
```
Uncaught TypeError: e.manager.edit is not a function
    at user-ck5edyrw40009cun2y6ohyvse.js:71
    at HTMLButtonElement.<anonymous> (user-ck5edyrw40009cun2y6ohyvse.js:48)
    at HTMLDivElement.dispatch (user-ck5edyrw40009cun2y6ohyvse.js:8)
    at HTMLDivElement.g.handle (user-ck5edyrw40009cun2y6ohyvse.js:8)
```

**Expected behavior**
Page Settings for given page's modal opens

**Note**
This works for pieces, just not pages

boutell:
…ut unwanted impacts on the chooser experience

Closes #2125

### DIFF
--- a/lib/modules/apostrophe-doc-type-manager/views/chooserChoices.html
+++ b/lib/modules/apostrophe-doc-type-manager/views/chooserChoices.html
@@ -19,9 +19,12 @@
           {{ buttons.inGroup('', { icon: 'list', action: 'relate-item' , value: choice._id, tooltip: 'Related Fields'}) }}
         {%- endif -%}
         {# By default you can edit the items themselves, sometimes #}
-        {# this is confusing #}
+        {# this is confusing. Pages can never be edited here because #}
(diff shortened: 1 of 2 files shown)


stuartromanek:
@boutell pieces can be edited in these scenarios, their editor modal's are opened on top of the chooser experience. This is not possible for pages?

boutell:
It is not practical for pages because the page settings modal isn't designed to be used without redirecting at the end and no consideration has been given to other weird eventualities, like editing the slug of the page you are currently sitting on. For now I think we should remove the open manhole cover. For 3.x we could really consider it properly.

PR #2143, opened 2020-02-06
update CHANGELOG and inline doc for apostrophe-users option "disableInactiveAccounts"

falkodev:
update CHANGELOG and inline doc for apostrophe-users option "disableInactiveAccounts"


boutell:
Thanks! Apologies for my initial misread.

PR #2150, opened 2020-02-28
Fail if joinByOne field is required and it's missing

rjrodriguezalvarez97:


lib/modules/apostrophe-schemas/index.js
```diff
@@ -2062,6 +2062,11 @@ module.exports = {
           });
         },
         form: function(req, data, name, object, field, callback) {
+
+          if (field.required && _.isEmpty(data[field.idField])) {
+            return callback("required");
+          }
+
           object[field.idField] = self.apos.launder.id(data[field.idField]);
```
boutell: Test should be after laundering.

boutell:
Thanks!

PR #2157, opened 2020-03-10
if there are no properties in data or data is null do not add a ? in …

### LINKED ISSUE #2156: apos.utils.get adds an extra "?" even when the query string is empty
**Is your feature request related to a problem you are experiencing that is not a bug? Please describe.**

In a project I'm working on, we expose GET routes that expect complex query string that make use of the nested object syntax. For example, I can end up with URLs that look like `/somewhere?some[data]=1`.

As I'm working in lean mode, I'd like to use the `apos.utils.get` method to make the request from a client script; however, looking into Apostrophe source code, it's clearly stated in a comment that nested properties are not supported in the query string. So in order to make it work, I'm currently building the query string myself and injecting it in the `uri` parameter, something like this:

```js
apos.utils.get('/somewhere?' + toQueryString(data), {}, callback)
```

However, the way `apos.utils.get` is implemented always adds an extra question mark `?` at the end of the URI, even if the query string is empty; in my case, it's thus modifying the value of the last property of the query string (e.g. `GET /somewhere?some[data]=1?` → `some.data === '1?'`).

**Describe the solution you'd like**
I don't think nested properties in query string should be supported by default, as it requires a middleware on server-side to parse them anyway. The minimal solution I'd advocate for would be to detect if the `data` object is empty, and then not add the extra `?` at the end of the `uri`. We could even pass `null` or `undefined` for this purpose, to make it explicit that we do not intend for a query string to be added at all.

This solution would keep the previous, simple behavior, while still allowing to create custom query strings for complex cases when needed.

**Describe alternatives you've considered**
Another solution would be to let the user provide a transformation function to convert the input data to a query string. It could be an additional optional parameter for example.

In my use case, I'm using the [`qs` package](https://www.npmjs.com/package/qs), which exposes a `stringify` method. Thus could the code look like:

```js
apos.utils.get('/somewhere', data, qs.stringify, callback)
```

I personally find this solution more elegant, and it aligns quite well with the API of other libraries like Lodash or JSON.stringify and such.

boutell:
…apos.utils.get

closes #2156

### DIFF
--- a/lib/modules/apostrophe-browser-utils/public/js/lean.js
+++ b/lib/modules/apostrophe-browser-utils/public/js/lean.js
@@ -60,7 +60,10 @@
   // Like `apos.utils.post` but uses a GET request, with the properties of `data`
   // added with query string encoding. Currently no support for nested properties
   // in `data`; intended for simple cases where you actually want the browser
-  // to cache. Like `jsonCall`, it invokes the callback Node.js style,
+  // to cache. If `data` is null or contains no properties then no query string
+  // is added.
+  //
+  // Like `jsonCall`, it invokes the callback Node.js style,
   // with an error if any, followed by the response as parsed JSON.
   //
   // If a browser error (status 400 or above) is reported, the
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -69,17 +72,20 @@
   // delivered as well, in the second argument.
 
   apos.utils.get = function(uri, data, callback) {
+    var keys, i;
```
abea: It doesn't seem like these are doing anything outside of the conditional block, but also not harming anything. I'd move them in, still.

boutell: var has no block scope, so it is considered harmful to pretend it does. Better to declare them at the top of the function because like it or not, that's the scope.

Hooray for not caring about this in 3.x.

boutell:
Huh, thought I sent this... the reason we don't move the variable declarations back is because it's a lie; var is always function scoped. We really should never use it except at the top of a function because that's how it actually behaves and nesting it implies a specificity that is not there. Thankfully a nonissue for non-legacy coding.

PR #2159, opened 2020-03-17
make it possible for overrides of apos.templates.i18n to determine wh…

boutell:
…at module called render(), as an inexpensive way to achieve namespacing and/or just prevent i18n of admin text when that is not desired


boutell:
Nah, wrong approach.

PR #2162, opened 2020-03-20
we do not invent our own event bus in lean mode, because DOM events w…

boutell:
…ork just fine, but we do now provide a convenience method to emit one, with the convention being to emit non-DOM-related events as custom event names on "body"

### DIFF
--- a/lib/modules/apostrophe-browser-utils/public/js/lean.js
+++ b/lib/modules/apostrophe-browser-utils/public/js/lean.js
@@ -17,6 +17,22 @@
 
   apos.utils = apos.utils || {};
 
+  // emit a custom event on the specified DOM element in a cross-browser way.
+  // If `data` is present, the properties of `data` will be available on the event object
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -17,6 +17,22 @@
...
+  // in your event listeners. For events unrelated to the DOM, we often emit on
+  // `document.body` and call `addEventListener` on `document.body` elsewhere.
+  //
+  // "Where is `apos.utils.on`?" You don't need it, use `addEventListener`, which is
+  // standard.
+
+  apos.utils.emit = function(el, name, data) {
+    // IE-compatible custom DOM event
+    const event = document.createEvent('Event');
+    event.initEvent(name, true, true);
```
abea: Since [this method is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent), I think we should check for `CustomEvent` and use that if we can.

PR #2170, opened 2020-04-06
Fixes `def: null` handling for `type: 'date'` in schemas

timashev:
`def: null` is not working for `type: 'date'`, it is set to current date always if left empty or cleared in editor, because this is default `launder` behavior fo dates when no default value is provided to `launder.date` call.

This simple fix resolves the issue.

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -1867,7 +1867,7 @@ module.exports = {
       name: 'date',
       converters: {
         string: function(req, data, name, object, field, callback) {
-          object[name] = self.apos.launder.date(data[name]);
+          object[name] = self.apos.launder.date(data[name], field.def);
           return setImmediate(callback);
         },
         form: 'string'


boutell:
I take your point, but this is not quite how `field.def` works either. Written this way, it would be impossible to leave the date null on purpose if the default were something other than null. (This used to be true for all field types and we fixed it, precipitating the issue you're seeing now).

So a correct fix would be to specifically look for a blank value and accept that as null.

boutell:
To be clear, I'm talking about the case where `def` is, say, `2025-01-01`, and the user tries to clear the date on purpose (intentionally discarding the default). They would not be able to with this fix.

timashev:
@boutell OK, maybe it is easier for you to fix it then. Meanwhile I'll keep using this fix for my projects.

boutell:
Bea, can you look into this? The thing is we don't want to use launder's default mechanism every time we sanitize something, but launder's "default default" is current date, and that's not what people mean when they clear a date field — a thing that used to work. So this code needs a special case outside of the launder call for intentionally setting the date blank/null, resulting in a null value.

abea:
Closing to use #2171 . Thanks for the tip off for this, @timashev!

PR #2171, opened 2020-04-06
Allows for empty date fields

abea:
Alternate implementation of #2170 to allow for null date values. Currently you can't leave date fields empty, or else they'll submit the current date.

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -1867,6 +1867,12 @@ module.exports = {
       name: 'date',
       converters: {
         string: function(req, data, name, object, field, callback) {
+          if (data[name] === '') {
+            object[name] = null;
+
+            return setImmediate(callback);
+          }
+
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-schemas/index.js
```diff
@@ -1867,6 +1867,12 @@ module.exports = {
       name: 'date',
       converters: {
         string: function(req, data, name, object, field, callback) {
+          if (data[name] === '') {
```
boutell, line 1870: This is good, but actually it could also be `null` as it could be a JSON submission. Indeed a simple check for a falsy `date[name]` would work much better than usual here.

lib/modules/apostrophe-schemas/index.js
```diff
@@ -1867,6 +1867,12 @@ module.exports = {
...
+            object[name] = null;
```
boutell, line 1871: Canonicalizing to `null` like this is great.

PR #2174, opened 2020-04-07
docs: Fix simple typo, sesions -> sessions

timgates42:
There is a small typo in lib/modules/apostrophe-express/index.js.

Should read `sessions` rather than `sesions`.


abea:
Thanks for the fix, @timgates42!

PR #2178, opened 2020-04-08
Removes goof file

abea:


abea:
This was a goof, right?

PR #2179, opened 2020-04-09
portable event emit and same site test

boutell:
Awful IE11 truths from ECG:

* CustomEvent exists, but is not a constructor, so our old sniff for it was not valid on IE11. Ow.
* The trick of using the host property of an anchor tag to check if it is the same site as location.host is not valid on IE11. Ow ow.

### DIFF
--- a/lib/modules/apostrophe-browser-utils/public/js/lean.js
+++ b/lib/modules/apostrophe-browser-utils/public/js/lean.js
@@ -447,7 +447,7 @@
   // which is generally the case.
 
   apos.utils.sameSite = function(uri) {
-    var matches = uri.match(/^(https?:)?\/\/([^\/]+)/);
+    var matches = uri.match(/^(https?:)?\/\/([^/]+)/);
     if (!matches) {
       // If URI is not absolute or protocol-relative then it is always same-origin
       return true;
--- a/lib/modules/apostrophe-browser-utils/public/js/lean.js
+++ b/lib/modules/apostrophe-browser-utils/public/js/lean.js
@@ -27,13 +27,14 @@
 
   apos.utils.emit = function(el, name, data) {
     var event;
-    if (!window.CustomEvent) {
-      // bc
+    try {
+      // Modern. We can't sniff for this, we can only try it. IE11
+      // has it but it's not a constructor and throws an exception
+      event = new window.CustomEvent(name);
+    } catch (e) {
+      // bc for IE11
       event = document.createEvent('Event');
       event.initEvent(name, true, true);
-    } else {
-      // modern browsers
-      event = new CustomEvent(name);
     }
     apos.utils.assign(event, data || {});
     el.dispatchEvent(event);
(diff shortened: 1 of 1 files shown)


boutell:
With these changes, our lean video widgets really REALLY work in IE11 now.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -445,4 +440,19 @@
...
   };
 
+  // Returns true if the uri references the same site (same host and port) as the
+  // current page. Cross-browser implementation, valid at least back to IE11.
+  // Regarding port numbers, this will match as long as the URIs are consistent
+  // about not explicitly specifying the port number when it is 80 (HTTP) or 443 (HTTPS),
+  // which is generally the case.
+
+  apos.utils.sameSite = function(uri) {
+    var matches = uri.match(/^(https?:)?\/\/([^\/]+)/);
```
abea, line 450: This is throwing an eslint error due to unnecessary escaping.

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -27,13 +27,14 @@
 
   apos.utils.emit = function(el, name, data) {
     var event;
-    if (!window.CustomEvent) {
-      // bc
+    try {
```
abea, line 30: 🤯 TIL `try...catch` isn't only for Promises. I guess that makes sense!

lib/modules/apostrophe-browser-utils/public/js/lean.js
```diff
@@ -445,4 +440,19 @@
     return path + '.' + file.extension;
   };
 
+  // Returns true if the uri references the same site (same host and port) as the
+  // current page. Cross-browser implementation, valid at least back to IE11.
+  // Regarding port numbers, this will match as long as the URIs are consistent
+  // about not explicitly specifying the port number when it is 80 (HTTP) or 443 (HTTPS),
+  // which is generally the case.
+
+  apos.utils.sameSite = function(uri) {
```
abea, line 449: We should have a test for this, right?

boutell: It's frontend code, so not easily tested in our current harness, but I tested it at the console with several distinct cases in IE11 and in Chrome.

lib/modules/apostrophe-browser-utils/public/js/lean.js
boutell: Hee! This is probably a common impression among JavaScript developers because we got "it's useless because async" beaten into our heads for so long. Now it's back because async functions allow it wrapped around `await`, but folks still forget it is perfectly fine for synchronous code too (where it was born!)

lib/modules/apostrophe-browser-utils/public/js/lean.js
abea: Oh true. Sounds good.

PR #2191, opened 2020-04-22
add return false to preventi href redirect

GiuseppeCM:
With the <base> tag in html, the href='#' in button to open the array modal redirect in home, return false in this function prevents this redirect.


boutell:
Thanks!

PR #2193, opened 2020-04-23
nice password complexity errors for command line tasks, instead of ju…

boutell:
…st the word "rules"

Surface the nice errors shown for password complexity errors in our frontend when using CLI tasks to change or set passwords.


abea:
Why is this still trying to run Travis?

PR #2196, opened 2020-05-05
fix problem with broken descendants when rescuing from trash via reor…

boutell:
…ganize with trashInSchema: true

Please verify we didn't break the trash when trashInSchema is *not* true.


abea:
Resolves #2194

lib/modules/apostrophe-pages/public/js/reorganize.js
```diff
@@ -345,21 +345,25 @@ apos.define('apostrophe-pages-reorganize', {
...
         }
-        // Reflect changed slugs as needed
-        _.each(data.changed, function(info) {
-          var node = self.$tree.tree('getNodeById', info.id);
-          if (node) {
-            node.slug = info.slug;
-          }
-        });
         e.move_info.do_move();
+        if (trashing !== undefined) {
```
abea, line 358: @boutell I think we want this to run regardless of whether `trashing` is set. That gets us the right link in the tree right away when moving things out of the trash. Any concern about that?

abea: Otherwise the link, when `trashInSchema` is undefined, is prepended with `/trash`

boutell: You are correct, pls change

bmauger:
@boutell @abea - I'm happy to report that I updated to the latest version (2.106.3) on my local instance and tested it. Twice. In both tests, I created a top-level page with a couple of subpages. I trashed them and then rescued them. In both cases, all pages were restored to their original location. Nice.

However, I pushed the fix to my live instance and, as we suspected, it did not rescue the missing pages. How do we go about getting those back? Should I open a new bug ticket?

Thanks!

PR #2201, opened 2020-05-06
object notation at project level breaks apostrophe admin strings, thi…

boutell:
…s best-effort workaround solves that


boutell:
Tested with objectNotation active via apostrophe-i18n-static in apostrophe-enterprise-testbed.

PR #2206, opened 2020-05-14
Allow empty values in time fields

mikko-paukkonen-solita:
Time fields do not currently seem to support empty values, defaulting to current time. Included is a fix using the approach that was used to allow empty values for dates in PR #2171. I've also included a test for the handling of empty values for date fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.106.4 (2020-05-20)
+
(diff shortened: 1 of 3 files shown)


abea:
Shades of #2171 here. Clearly we had something against empty fields.

abea:
You are a true hero, @mikko-paukkonen-solita. Please add an entry for this in the `CHANGELOG.md` file, then it's ready to merge. You can start a new date with `## 2.106.4 (2020-05-20)` (the next release will be next Wednesday).

mikko-paukkonen-solita:
Thank you 😄 I formatted an entry similar to the date fix. As far as I can see, the regression was introduced in the same version as the date issue, so I copied that bit over too.

PR #2213, opened 2020-05-27
apos.utils.post and apos.utils.http should tolerate no content type o…

boutell:
…n the response

Don't crash when there is no content type. Backport from A3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-.nyc_output
-coverage
 *.swp
 package-lock.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+.nyc_output
```
abea, line 1: Should this be here? Istanbul isn't in 2.x, right?

abea: `coverage` as well, I think

PR #2222, opened 2020-06-11
Add "busy" and "saving..saved" indicators

falkodev:
I added an overlay for HTTP requests when `options.busy` is passed.

I used an arbitrary color (to be checked with @stuartromanek later).

## How to test

Check out this branch, and run `npm link`.
Clone `a3-test-1` project , switch to `1940-new-module-format` branch, run `npm i` if not done previously, then run `npm link apostrophe`.

Go to a Vue component using a `http` request. For example, `lib/modules/@apostrophecms/areas/src/apos/components/ApostropheAreaEditor.vue` and modify the `up` method this way:

```js
   await apos.http.patch(`${apos.docs.action}/${this._docId}`, {
        busy: true,
        body: {
          $move: {
            [`@${this._id}.items`]: {
              $item: this.next[i].widget._id,
              $before: this.next[i - 1].widget._id
            }
          }
        }
      });
```

Add two widgets in the same area, and click on the "Up" button. As the overlay will be quickly removed and therefore hard to see, you can comment this part in `lib/modules/@apostrophecms/utils/src/apos/always/http.js` for test purposes and see the overlay is active (line 201):

```js
    xmlhttp.addEventListener('loadend', function () {
      if (options.busy) {
        apos.bus.$emit('apos-busy', false);
      }
    });
```

### DIFF
--- a/defaults.js
+++ b/defaults.js
@@ -49,6 +49,6 @@ module.exports = {
     '@apostrophecms/files': {},
     '@apostrophecms/files-widgets': {},
     '@apostrophecms/soft-redirects': {},
-    '@apostrophecms/overlay': {}
+    '@apostrophecms/busy': {}
   }
 };
--- a/defaults.js
+++ b/defaults.js
@@ -48,6 +48,7 @@ module.exports = {
     '@apostrophecms/images-widgets': {},
     '@apostrophecms/files': {},
     '@apostrophecms/files-widgets': {},
-    '@apostrophecms/soft-redirects': {}
+    '@apostrophecms/soft-redirects': {},
+    '@apostrophecms/overlay': {}
   }
 };
--- a/lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
+++ b/lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
@@ -1,12 +1,12 @@
 <template>
-  <div class="apos-overlay">
+  <div class="apos-busy">
     <div v-if="busy" class="busy"></div>
   </div>
 </template>
 
 <script>
 export default {
-  name: 'TheApostropheOverlay',
+  name: 'TheApostropheBusy',
   data() {
     return {
       busy: false
--- a/lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
+++ b/lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
@@ -13,7 +13,11 @@ export default {
     }
   },
   mounted() {
-    apos.bus.$on('apos-busy', state => (this.busy = state.active));
+    apos.bus.$on('apos-busy', state => {
+      if (state.name === 'busy') {
+        this.busy = state.active
+      }
+    });
   }
 };
 </script>
--- a/lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
+++ b/lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
@@ -13,7 +13,7 @@ export default {
     }
   },
   mounted() {
-    apos.bus.$on('apos-busy', state => (this.busy = state));
+    apos.bus.$on('apos-busy', state => (this.busy = state.active));
   }
 };
 </script>
--- a/lib/modules/@apostrophecms/doc-type/index.js
+++ b/lib/modules/@apostrophecms/doc-type/index.js
@@ -1450,7 +1450,6 @@ module.exports = {
                   }
 
                   area._docId = doc._id;
-                  let i = 0;
                   for (const item of area.items) {
                     if (area._edit) {
                       // Keep propagating ._edit so a widget can be passed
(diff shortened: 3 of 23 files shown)


lib/modules/@apostrophecms/utils/src/apos/always/http.js
```diff
@@ -120,36 +124,21 @@
...
-    if (sendJson) {
-      data = JSON.stringify(options.body);
-    } else {
-      data = options.body;
-    }
-    xmlhttp.send(data);
-    xmlhttp.addEventListener("load", function() {
-      var responseHeader = this.getResponseHeader("Content-Type");
+
+    xmlhttp.addEventListener('progress', function() {
```
falkodev, line 134: I moved all the listeners here (before the `open` and `send` methods) in order to have the `progress` listener work correctly (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest?redirectlocale=en-US&redirectslug=DOM/XMLHttpRequest/Using_XMLHttpRequest#Monitoring_progress)

lib/modules/@apostrophecms/doc-type/index.js
```diff
@@ -1450,7 +1450,6 @@ module.exports = {
                   }
 
                   area._docId = doc._id;
-                  let i = 0;
```
falkodev, line 1453: eslint made me do it..

defaults.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
     '@apostrophecms/images-widgets': {},
     '@apostrophecms/files': {},
     '@apostrophecms/files-widgets': {},
-    '@apostrophecms/soft-redirects': {}
+    '@apostrophecms/soft-redirects': {},
+    '@apostrophecms/overlay': {}
```
falkodev, line 52: Is it better to name it "overlay" or "busy"?

lib/modules/@apostrophecms/overlay/src/apos/components/TheApostropheOverlay.vue
```diff
@@ -0,0 +1,32 @@
...
+<script>
+export default {
+  name: 'TheApostropheOverlay',
+  data() {
+    return {
+      busy: false
+    }
+  },
+  mounted() {
+    apos.bus.$on('apos-busy', state => (this.busy = state));
```
falkodev, line 16: I used `apos.bus.$on` and `apos.bus.$emit` in `lib/modules/@apostrophecms/utils/src/apos/always/http.js` because it felt less odd than doing:

`apos.utils.emit(document.body, 'apos-busy', false)` and here:

```js
document.body.addEventListener('apos-busy', state => (this.busy = state));
```

But both worked. So no problem to switch to the `addEventListener` approach if required.

(BTW, is `apos.bus` a Vue thing done by Apostrophe?)

defaults.js
boutell: "busy". Name it for the function, not its current appearance.

boutell:
One more little thing: please name the event just "busy", it is already namespaced by using apostrophe's private `bus` object. You will see that other events we emit on the bus do not use a prefix.

`apos.bus` is a headless Vue instance which exists just to serve as an event bus, something all apos vue apps can emit and listen on to talk to each other.

I had considered getting rid of it in favor of events on the body element, but that is just awkward enough that I think for now the bus should live on.

lib/modules/@apostrophecms/utils/src/apos/always/utils.js
```diff
@@ -264,4 +264,8 @@
     }
     return window.location.host === matches[2];
   };
+
+  apos.utils.generateId = function() {
```
boutell, line 268: We should not be adding anything to the "always" scene if we only need it in the "user" scene. You can import the cuid module and use that directly.

lib/modules/@apostrophecms/utils/src/apos/always/http.js
```diff
@@ -148,13 +152,24 @@
       data = options.body;
     }
     xmlhttp.send(data);
-    xmlhttp.addEventListener("load", function() {
-      var responseHeader = this.getResponseHeader("Content-Type");
+    xmlhttp.addEventListener('load', function() {
+      if (options.busy) {
```
boutell, line 156: I have an idea for a cleaner resolution. let's have the `busy` option take a value (a string), and emit that string as the payload of the event. For a contextual rich text save operation, it can be "save-in-context", and then the frontend team can worry about displaying it in an appropriate vue component. No need for us to hack on the DOM.

If `options.busy` is just `true`, we can pass `busy` as the payload of the event (the general case).

lib/modules/@apostrophecms/busy/src/apos/components/TheApostropheBusy.vue
```diff
@@ -0,0 +1,32 @@
...
+<script>
+export default {
+  name: 'TheApostropheBusy',
+  data() {
+    return {
+      busy: false
+    }
+  },
+  mounted() {
+    apos.bus.$on('apos-busy', state => (this.busy = state.active));
```
boutell, line 16: This is the global, blocking busy indicator right? So it should pay attention to `state.name` and ignore the event if it is not `busy`.

boutell:
Thanks! Needs a semicolon, but the linter will catch that soon.

PR #2233, opened 2020-06-24
V3/ui input

abea:
Mostly a test example
- adding the AposButton component to an existing usage
- updating eslint for Vue
- moving over styles

TODO:

- [x] Update webpack build to process SCSS theme and load it for Vue components.


lib/modules/@apostrophecms/ui/scss/_storybook.scss
```diff
@@ -0,0 +1,8 @@
+// Storybook specific override styles
```
gregvanbrug, line 1: Should this be here? If this is storybook specific, couldn't it have a negative affect on the the UI?

PR #2234, opened 2020-06-26
generate a file we could use to import components into storybook

gregvanbrug:


boutell:
Looks good but how about reading the json filename as argv.component-map so we're not hardcoding Storybook paths in this task.

boutell:
Superseded by our work on the other branch.

PR #2236, opened 2020-06-29
V3/ui webpack

gregvanbrug:
This PR  moves the internal webpack config out of the assets index file and follows other patterns we've established for webpack configs making it easier to maintain our config. 

This pretty much what was there with the addition of autoprefixer and stylelint and updates to scss to fix issues identified by the initial lint. 

Now that this is in place, do we want to allow webpack settings based on the environment, ex: watch if we're in development, or min if we're in production. I can see the watch being useful for integration.


boutell:
Regarding environment-dependent settings, we should think about whether we want to embrace the `config` npm module officially, as some of our clients have. Or on the most minimal end, we could just check for whether `NODE_ENV` equals `production` or not.

gregvanbrug:
`config` looks nice. Could be helpful for a bunch of things. For now, I was just thinking of checking the `NODE_ENV`.

PR #2248, opened 2020-07-17
Login throttle dgad 1005

boutell:
In addition to the unit test coverage, you can test it with this configuration:

```
    'apostrophe-login': {
      throttle: {
        attempts: 3,
        per: 1,
        lockout: 1
      }
    },
```

Now three invalid login attempts in one minute for the same account will lock you out for one minute, even if your credentials for the fourth attempt are correct. The issue is disclosed to the user so they know why they still can't get in. As discussed this feature was implemented with an async function and it's ever so much more readable as a result, but since for bc it must have a callback based interface, I plumbed it carefully to avoid returning any promises that might generate warnings or confuse developers.

### DIFF
--- a/lib/modules/apostrophe-login/index.js
+++ b/lib/modules/apostrophe-login/index.js
@@ -306,7 +306,12 @@ module.exports = {
           if (err) {
             // Slow down and keep 'em hanging to make brute force attacks less easy
             return setTimeout(function () {
-              return callback(null, false);
+              // string error = OK to pass on to browser (like "throttle")
+              if ((typeof err) === 'string') {
+                return callback(err);
+              } else {
+                return callback(null, false);
+              }
             }, 1000);
           }
           return self.checkIfActive(user, callback);
@@ -363,10 +368,13 @@ module.exports = {
         try {
           await verify(user, password);
         } catch (e) {
+          let err = e;
           await addAttempt();
-          await lockoutIfNeeded();
+          if (await lockoutIfNeeded()) {
+            err = 'throttle';
+          }
           await cleanUpAttempts();
-          throw e;
+          throw err;
         }
       }
 
@@ -396,6 +404,7 @@ module.exports = {
               lockout: new Date()
             }
           });
+          return true;
         }
       }
 
@@ -561,6 +570,9 @@ module.exports = {
                 // allow them to fix it
                 return self.verifyLogin(req.body.username, password, function(err, user) {
                   if (err || (!user)) {
+                    if (err === 'throttle') {
+                      errors.unshift(req.__ns_n('apostrophe', 'Too many login attempts. You may try again in %s minute(s).', self.options.throttle.lockout));
+                    }
                     return fail(errors);
                   }
                   req.session.resetLegacyPasswordId = user._id;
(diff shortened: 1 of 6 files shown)


lib/modules/apostrophe-login/index.js
```diff
@@ -74,6 +74,14 @@
 // (see above for concerns). You should bear in mind that this option is not as
 // secure as requiring confirmation via email with `passwordReset.
 //
+// `throttle`
+//
+// If the `throttle` option is set to `{ attempts: 3, per: 1, lockout: 10 }` for
```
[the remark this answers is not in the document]
abea: Also `allowedAttempts`. Down further I lost track of whether `attempts` was the number of times the user had tried for a minute.

lib/modules/apostrophe-login/index.js
```diff
@@ -445,6 +570,9 @@ module.exports = {
                 // allow them to fix it
                 return self.verifyLogin(req.body.username, password, function(err, user) {
                   if (err || (!user)) {
+                    if (err === 'throttle') {
+                      errors.unshift(req.__ns_n('apostrophe', 'Too many login attempts. You may try again in %s minute(s).', self.options.throttle.lockout));
```
abea, line 574: "Too many failed login attempts" for clarity.

lib/modules/apostrophe-login/index.js
```diff
@@ -482,14 +610,21 @@ module.exports = {
           function(req, res, next) {
             self.passport.authenticate('local', function(err, user, info) {
               if (err) {
+                if (err === 'throttle') {
+                  // verifyPassword does not take req, so we i18n the error here
+                  //
+                  // Use of __ns_n means that the "minute(s)" string can be pluralized
+                  // more nicely if desired via i18n json config, even for English
+                  err = new Error(req.__ns_n('apostrophe', 'Too many login attempts. You may try again in %s minute(s).', self.options.throttle.lockout));
```
abea, line 618: Also adding "failed" here.

test/login-throttle.js
```diff
@@ -0,0 +1,151 @@
...
+      }
+    });
+  });
+
+  it('should be able to insert test user', function(done) {
+    assert(apos.users.newInstance);
+    const user = apos.users.newInstance();
+    assert(user);
+
+    user.firstName = 'Harry';
```
abea, line 69: I mean 'crookshanks' should go with Hermione, but I digress.

lib/modules/apostrophe-login/index.js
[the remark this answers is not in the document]
boutell: It's not `perMinute`, it's the number of minutes. It's `attempts` over `per` minutes. Not sure how to improve the naming for that.

test/login-throttle.js
boutell: And I should replace this well-worn example with an author who hasn't proven herself a stubbornheaded transphobe. ):

lib/modules/apostrophe-login/index.js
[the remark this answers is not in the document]
abea: Right right right. `trialMinutes`?

test/login-throttle.js
abea: Also true. "Lilith Iyapo" and "nikanj" for a semi-deep cut (from a superb book).

lib/modules/apostrophe-login/index.js
```diff
@@ -350,8 +350,8 @@ module.exports = {
 
       async function throttleAttempts() {
         if (self.options.throttle) {
-          if (!attempts) {
-            throw new Error('If the throttle option is set for apostrophe-login, then the attempts subproperty is required.');
+          if (!allowedAttempts) {
```
abea, line 353: Is this still necessary with the default set now?

PR #2249, opened 2020-07-17
3.0 media tags

falkodev:
https://apostrophecms.atlassian.net/browse/A30U-115

### DIFF
--- a/lib/modules/@apostrophecms/files-tags/index.js
+++ b/lib/modules/@apostrophecms/files-tags/index.js
@@ -1,14 +1,6 @@
 module.exports = {
   extend: '@apostrophecms/piece-type',
   options: {
-    label: 'Files Tag',
-    fields: {
-      add: {
-        title: {
-          type: 'string',
-          required: true
-        }
-      }
-    }
+    label: 'Files Tag'
   }
 };
--- a/lib/modules/@apostrophecms/files-tags/index.js
+++ b/lib/modules/@apostrophecms/files-tags/index.js
@@ -1,0 +1,13 @@
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    fields: {
+      add: {
+        title: {
+          type: 'string',
+          required: true
+        }
+      }
+    }
+  }
+};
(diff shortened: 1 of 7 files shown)


lib/modules/@apostrophecms/files-tags/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    fields: {
+      add: {
+        title: {
```
boutell, line 6: This is correct, but is it necessary? There is already a title field configuration for all doc types by default.

lib/modules/@apostrophecms/files/index.js
```diff
@@ -45,6 +45,12 @@ module.exports = {
         type: 'url',
         name: 'creditUrl',
         label: 'Credit URL'
+      },
+      {
+        type: 'joinByArray',
+        name: '_tags',
+        label: 'Tags',
+        withType: '@apostrophecms/files-tags'
```
boutell, line 53: 👍

lib/modules/@apostrophecms/files-tags/index.js
falkodev: Oh ok, I thought I had to add this field. Will remove it.

lib/modules/@apostrophecms/files-tags/index.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extend: '@apostrophecms/piece-type',
+  options: {
+    label: 'Files Tag'
```
falkodev, line 4: Should it be `File Tag`? @boutell

boutell: I think "File Tag" reads better, yes.

lib/modules/@apostrophecms/schemas/index.js
```diff
@@ -769,12 +769,12 @@ module.exports = {
         if (Array.isArray(field.withType)) {
           _.each(field.withType, function (type) {
             if (!_.find(self.apos.docs.managers, { name: type })) {
-              fail('withType property, ' + type + ', does not match the "name" property of any doc type. Hint: this is not the same thing as a module name. Usually singular.');
+              fail('withType property, ' + type + ', does not match the "name" property of any doc type. You have to set it as the name of the joined module.');
```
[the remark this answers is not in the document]
boutell: How about: "in most cases this is the same as the module name."

PR #2252, opened 2020-07-20
eslint and tests passing

boutell:


lib/modules/@apostrophecms/assets/index.js
```diff
@@ -154,24 +154,6 @@ ${appImports.invokeCode}
         fs.writeFileSync(`${apos.rootDir}/public/apos-frontend/user-bundle.js`, fs.readFileSync(`${apos.rootDir}/public/apos-frontend/anon-bundle.js`) + fs.readFileSync(`${apos.rootDir}/public/apos-frontend/user-only-bundle.js`));
       }
 
-      function getIcons() {
```
abea, line 157: This is still referenced on L89

abea: Nevermind. It's a dupe.

PR #2255, opened 2020-07-20
V3/misc

abea:
Migrates additional Storybook stories


abea:
I did update the FilterMenu story since the `menu` prop was changed to `filters` and is structurally different. I also went with this `getData` function patter instead of having a separate file with story data so we could still have stories for some of these molecule-level components.

lib/modules/@apostrophecms/ui/src/apos/components/AposAreaMenu.vue
```diff
@@ -0,0 +1,295 @@
...
+      if (this.$refs[name]) {
+        this.$refs[name][0].querySelector('button').focus();
+      }
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+// @import '../../scss/_mixins';
```
gregvanbrug, line 184: This should be able to be safely removed.

lib/modules/@apostrophecms/ui/src/apos/components/AposAreaMenuItem.vue
```diff
@@ -0,0 +1,91 @@
...
+  methods: {
+    click() {
+      this.$emit('click');
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+// @import '../../scss/_mixins';
```
gregvanbrug, line 51: This should be able to be safely removed.

lib/modules/@apostrophecms/ui/src/apos/components/AposTag.vue
```diff
@@ -0,0 +1,71 @@
...
+  methods: {
+    click() {
+      this.$emit('click', this.slug);
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+// @import '../../scss/_mixins';
```
gregvanbrug, line 35: This should be able to be safely removed.

lib/modules/@apostrophecms/ui/src/apos/scss/_zindex.scss
```diff
@@ -1,3 +1,4 @@
+$z-index-under: -1;
 $z-index-base: 0;
 $z-index-default: 1;
```
gregvanbrug, line 3: Extremely minor but the default value for `z-index` corresponds to `0`. Maybe base should be called default and default should be something else?

lib/modules/@apostrophecms/ui/src/apos/components/AposAreaMenu.vue
boutell: +1

lib/modules/@apostrophecms/ui/src/apos/components/AposAreaMenuItem.vue
```diff
@@ -0,0 +1,91 @@
...
+    item: {
+      required: true,
+      type: Object
+    },
+    tabbable: {
+      type: Boolean,
+      default: true
+    }
+  },
+  emits: ['click', 'up', 'down'],
```
boutell, line 32: A Vue 3 ism, but doesn't seem to hurt in Vue 2, so let's start getting ready I guess!

abea: The linter yells about it. The alternative would be to change the rule to disable the requirement.

PR #2257, opened 2020-07-21
V3/modal stories

abea:
This should be the last of the stories from the storybook repo.


abea:
I kept it under 40 files, so... you're welcome?

lib/modules/@apostrophecms/doc-type/src/apos/components/AposDocEditor.vue
```diff
@@ -0,0 +1,150 @@
...
+            @input="update" :modifiers="['small', 'inverted']"
+          />
+        </div>
+      </AposModalRail>
+    </template>
+  </AposModal>
+</template>
+
+<script>
+import AposModalParentMixin from '../../../../modals/src/apos/mixins/AposModalParentMixin';
```
abea, line 50: The `'Modules/@apostrophecms'` alias wasn't working for me for script imports.

boutell: We gotta fix this. I'll get in.

boutell:
I pushed up two commits with the correct module relative paths, which look like:

`'Modules/@apostrophecms/modals/mixins/AposModalParentMixin'`

src/apos is implied, because what we're accessing here is the build folder containing the final frontend files for this module after all overrides are taken into account.

No other concerns!

PR #2262, opened 2020-07-22
Adds build comments

abea:
These tripped me up as I was trying to "tidy up" the files and then broke the build.


abea:
I might suggest we replace those with more explicit "COMPONENTS INSERTED HERE" or something.

PR #2264, opened 2020-07-22
Adds the admin bar component

abea:
Originally reviewed in the storybook repo: https://github.com/apostrophecms/apostrophe-storybook/pull/110/files


boutell:
OK I think I was mistaken about the admin bar items being hardcoded, there is some hardcoded stuff related to another menu?

abea:
It does still work, yes. That one element does not have an active equivalent and needs to be built out.

PR #2267, opened 2020-07-23
V3/storybook deploy

abea:


lib/modules/@apostrophecms/storybook/template/.storybook/theme.tmpl.js
```diff
@@ -0,0 +1,9 @@
+import { create } from '@storybook/theming/create';
+
+export default create({
+  base: 'light',
+
+  brandTitle: 'ApostropheCMS 3',
```
abea, line 6: This is only in the logo alt text from what I can tell.

PR #2268, opened 2020-07-23
ensure knobs work in storybook

gregvanbrug:
Updated the knobs addon, the `knobStore` state object was empty without knobs being "registered". This looked like a pretty much direct copy from the `apostrophe-storybook` repo so not totally sure why it just "worked" there. 

Also, re-added the decorators to the `AposInputString` so those should also work now too.


boutell:
🙏

PR #2270, opened 2020-07-23
3/tree component

abea:
Originally from https://github.com/apostrophecms/apostrophe-storybook/pull/117

Changes from that PR:
- In the first commit I updated how the data is generated in the stories file to avoid the separate data.js file.
- In the first commit I changed `AposTree` to get the data and draggable state via props rather than pulling from the external data file directly. This is more like how it will be in a modal. Also I was clearly brain fried since I was both passing in the prop in the stories file and not using it in the component.
- In the final commit, I switched drag handle icons since I realized there was a better matching one already. Style tweaks to support that.


.eslintrc
```diff
@@ -9,6 +9,7 @@
   "rules": {
     "no-var": "error",
     "no-console": 0,
+    "vue/no-deprecated-destroyed-lifecycle": 0,
```
abea, line 12: Maybe we should be using the vue eslint plugin v6...

Issue #2276, opened 2020-07-27
Support configuring an alternate collection name or entire alternate URL for the REST API of a module


boutell:
For sane and simple communication between its own browser UI code and the backend, a module will always have a REST API where Apostrophe expects it:

```
/api/v1/module-name
```

However, our module names are now singular (by popular demand). If the module name is `product` this will look like:

```
/api/v1/product
```

While the most common REST convention would be:

```
/api/v1/products
```

So, we should implement a `restAlias` option, which may be set like this:

```
restAlias: 'products'
```

Now the API will respond at BOTH of these URLs:

```
/api/v1/product
/api/v1/products
```

`restAlias` may also start with `/`, in which case it is understood to be a rooted URL, so it can be anywhere, not just `/api/v1`.

**This rule will also apply to `apiRoutes`, `htmlRoutes`, etc.** because the same issue exists when these are viewed as extensions of the REST paradigm to cover actions without a clear pattern in REST URLs. *All* routes for a given module will be impacted by `restAlias`.

abea:
@boutell Do you think we should get this in the backlog still?

boutell:
I would say no, I would say this will be reopened by popular demand if necessary. We're just not hearing it as a need right now.

PR #2277, opened 2020-07-27
Great renaming

### LINKED ISSUE #2244: Design and implement new module asset directory structure and naming in A3
As we punch ahead through a3, we've started talking through some of our well-established naming conventions, and want to surface the following two proposals for open discussion.

## Pushing Assets to the Browser Automatically: Scenes vs Builds

The good news is that A3 pushes assets to the browser automatically if you put them in the right folder in any Apostrophe module. This is a nice win over A2, where you had to call `pushAsset` for each one.

But, we’d like your input on how to name these folders.

We’ve considered both this:

```jsx
// browser js for all users, example: widget player
lib/modules/articles-widgets/builds/everyone/player.js
// vue components to register for apostrophe admin UI
lib/modules/articles-widgets/builds/user/components/*.vue
```

And this:

```jsx
// browser js for all users, example: widget player
lib/modules/articles-widgets/scenes/everyone/player.js
// vue components to register for apostrophe admin UI
lib/modules/articles-widgets/scenes/user/components/*.vue
```

We like the name “builds” because it’s clearly about building assets for the frontend.

We like the name “scenes” because it relates to Apostrophe’s internal concept of an asset “scene”. But, we’re thinking of renaming that concept “builds” altogether.

You’ll also note that the rules for `/user` are a little different from `/everyone`. Because Apostrophe’s admin UI is built on Vue, we provide a specific folder to automatically register Vue components from any module. We have other special folder names as well. `/everyone` is different: any .js files directly in that folder simply get pushed all the time, as part of the bundle for site visitors or editors.

Your input is welcome!

## Singular vs Plural: Singularizing our module names

For a long time now we have had plural names for modules that manage many instances of something. For instance, an "articles" module that extends pieces might manage thousands of articles, so we gave the module a plural name. It seemed like a good idea, but in practice it has caused a lot of confusion. When is a module singular, and when is it plural? If it takes more than a second to explain, it's not really helping.

So even though we are very far along with 3.x and the change is a little bit challenging, we are looking at moving to strictly singular module names. Here are some examples of the change we're talking about, as folder paths, so you can visualize it in the context of a project:

```jsx
lib/modules/articles —> lib/modules/article
lib/modules/events —> lib/modules/event
lib/modules/apostrophe-pages —> lib/modules/@apostrophecms/page
lib/modules/articles-widgets -> lib/modules/article-widget
```

We think this will be less confusing in the long run, but **we would love to get your input before we make this change.**

A special case: there are a few modules in Apostrophe that are never actually "instantiated" (you don't turn them on in `app.js`, and we don't turn them on in `defaults.js`). These modules are "abstract base classes:" you extend them to make your own modules. For these, we already made the change to add `-type` to the name:

```jsx
lib/modules/apostrophe-pieces --> lib/modules/@apostrophecms/piece-type
lib/modules/apostrophe-widgets --> lib/modules/@apostrophecms/widget-type
```

### LINKED ISSUE #2256: RFC: Apostrophe 3 modules directory name/nesting
In Apostrophe 2.x, the directory containing individual module directories is `lib/modules`. Having been involved with many Apostrophe sites over the last few years, I've only seen other files or directories directly in this root level `lib` directory a couple of times. To my understanding, nesting `modules` within the `lib` directory was originally done due to concern that developers might be confused by having a `modules` directory as a sibling of the common `node_modules` directory. There was also desire to keep it feeling like a Node application, rather than a CMS that's simply using Node, and so the structure was kept more generic.

With Apostrophe 3.x we have a chance to reconsider this. Given that the `node_modules` directory is such a common concept for Javascript developers and, secondarily, that IDEs typically will fade directories like `node_modules` that are ignored by git, I'm not totally convinced that developers would have much trouble managing the two directories.

## A case for `lib/modules`:
- There is no change for developers used to the Apostrophe 2.x structure
- This structure will not need to change when upgrading 2.x projects to 3.x.
- If the concern about confusing `modules` and `node_modules` is warranted, that confusion will be avoided

## A case for `modules` at project root level:
- Less unnecessary directory nesting
- We're already requiring much more significant changes to modules when upgrading (e.g., naming, actual code structure), so simply shifting the parent directory to the project root seems trivial
- Arguably cleaner code structure, with the root level telling a clearer "story" of the application, with `app.js` already on the root level.

### LINKED ISSUE #2261: Plural vs. singular for certain core module names that aren't piece types in A3
Question about our upcoming renaming plans. We said we would give *all* modules singular names. Some examples that make me think follow.

* `apos.areas`: this is a module that is in charge of all areas, but that excuse isn’t good enough for a pieces module to be plural, so it’s not good enough here. So it becomes `apos.area`. Or I’m going crazy. Sure, fine.

There is no conflict with inserting an area in a template because that is `{% area %}` now. However it does mean that we might have to find another way of linting for projects with legacy 2.x `apos.area` calls in them. Because we can't just replace `apos.area` with a helpful warning function if it is the name of a module. At least they will get an error ("not a function").

* `apos.utils`: this literally contains more than one utility. “Literally contains” is @localghost443 's suggested trigger for a plural. But, nodejs itself has a util module, which I’ve often found confusing because apostrophe contains a utils module. Do we rename our module to match?

* `apos.caches`: in charge of all caches - yes, yes, not good enough - but the syntax is gonna be weird if we make it singular. `apos.cache.get` really, really sounds like it gets a value out of a cache! But it doesn't. It gets you *a cache, that you can get values from*.

OK, big 👾 🧠 moment: I propose we fix this forever like so...

```
// Remember an oembed response from youtube
await apos.cache.set('oembed', 'https://youtube.com/v/something', { data... });
// Get it back later
await apos.cache.get('oembed', 'https://youtube.com/v/something');
```

We **kill the idea of multiple cache objects as part of the API completely,** and just always include a namespace name when getting and setting things. Having multiple caches, one per namespace, as part of the implementation under the hood depends on how the cache is implemented but an apostrophe developer should not care. They should have a simple, normal apostrophe module method they can call to get something. This also fixes our documentation about how to use the module, which is currently written in a weird way because of the "call get to get a cache object" issue.

Can I get a hell yeah?

* I have been over the modules list in core and I think these are the only ones I'm unsure about.

cc @stuartromanek @abea @falkodev

boutell:
closes #2261
closes #2256 
closes #2244

Does not introduce `restAlias`, but I opened a public ticket here:

#2276

And we'll track that internally too as regards the 3.x release.

Corresponding PR for a3-demo coming up in a minute.


.eslintrc
```diff
@@ -46,7 +46,7 @@
   "overrides": [
     {
       "files": [
-        "**/src/**/*.js"
```
abea, line 49: I don't really understand this as an equivalent change. `src` directories were across modules, but it's now only targeting the UI module.

modules/@apostrophecms/cache/index.js
```diff
@@ -0,0 +1,114 @@
+// A general purpose cache implementation for improved performance in all areas
```
abea, line 1: Could this be an a different PR where it could be more of the focus?

abea: Or was this previously in another file?

boutell: This was part of the decision taken to rename the module to the singular, while fixing its API so that, among other things, that would not be super confusing.

boutell: Also it's contained to one file and its test file, so I don't think it greatly burdens the PR. 🙏

boutell:
Reconciled your commit moving storybook.scss and fixed oversights with storybook build.

PR #2279, opened 2020-07-28
3/tree select

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-309, adding checkbox action to the tree component.


modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -36,7 +36,23 @@
         >
           <drag-icon
             v-if="draggable && index === 0" class="apos-tree__row__handle"
-            size="20"
+            :size="20"
+            :fill-color="null"
```
gregvanbrug, line 40: Does `fill-color` need to be bound if the values don't change?

abea: I think this was required because otherwise it's interpreted as a string?

PR #2280, opened 2020-07-28
Adds collapsing to the tree sections

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-308


stuartromanek:
@abea just pushed some style changes to this branch .. creating new child lists is still finicky but we can refine

abea:
@stuartromanek Are you still good to merge this as-is?

abea:
style changes moved to separate branch: https://github.com/apostrophecms/apostrophe/pull/2284

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -162,13 +164,39 @@ export default {
...
+      const branches = this.$el.querySelectorAll('[data-apos-tree-branch]');
+
+      branches.forEach(branch => {
+        // Add padding to the max-height to avoid needing a `resize`
+        // event listener updating values.
+        const height = branch.clientHeight + 20;
+        branch.setAttribute('data-apos-tree-branch', `${height}px`);
+        branch.style.maxHeight = `${height}px`;
+      });
+    });
```
gregvanbrug, line 179: Could this be simplified using a `ref`? https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements

Might save you from having to store the attribute in two places.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -162,13 +164,39 @@ export default {
...
+  },
   methods: {
     startDrag() {
       this.$emit('busy', true);
     },
     endDrag(event) {
       this.$emit('update', event);
     },
+    toggleSection(event) {
```
gregvanbrug, line 188: This probably also could leverage the `ref` if used.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
boutell: +1 for refs when this kind of wizardry can't be avoided.

abea: I'm not sure how that would work in this case since they are generated.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
abea: Same here. We don't know the level that the item is on, and it's all generated, so I'm not sure how that works.

abea:
From testing, it doesn't look like setting a background color will help. The overlap appearance is related to the height and overflow.

PR #2281, opened 2020-07-28
Moves AposPager

abea:
This was in the wrong position since it was merged around the same time as the Great Renaming.


abea:
There are also `moog.js` and `moog-require.js` files still in `lib/modules`. I'm less sure about what to do with those.

PR #2283, opened 2020-07-29
3.0 login screen

falkodev:
Following these designs: https://www.figma.com/file/RSVt4z4ZVMsqIYbyjYWcpN/a3-interface?node-id=0%3A3657

login

Development mode:
image


lib/modules/@apostrophecms/login/src/apos/components/TheApostropheLogin.vue
```diff
@@ -1,50 +1,118 @@
...
-        <input v-model="password" required type="password" />
-      </fieldset>
-      <fieldset>
-        <button type="submit">Log In</button>
-      </fieldset>
-    </form>
+  <div class="apos-login apos-theme-dark">
+    <div class="apos-login__overlay"></div>
+    <div class="apos-login__menu-overlay"></div>
+    <AposLoginBackground />
```
falkodev, line 5: Background image in data-uri format. Don't know if it is a good idea. I could have `import` a JPG file (works with webpack's `file-loader`).

lib/modules/@apostrophecms/login/src/apos/components/TheApostropheLogin.vue
```diff
@@ -1,50 +1,118 @@
...
+        />
+        <AposInputString
+          @input="fill($event, 'passwordField')"
+          :field="passwordField.field"
+          :type="passwordField.type"
+          :status="passwordField.status"
+          :value="passwordField.value"
+          :modifiers="['dark']"
+        />
+        <a href="#" class="apos-login__link">Forgot Password</a>
```
falkodev, line 36: I don't know what this is supposed to do (open a modal? go to another page?)

abea:
It looks like you need to merge in the latest `3.0` branch for the Great Renaming changes.

lib/modules/@apostrophecms/login/src/apos/components/TheApostropheLogin.vue
```diff
@@ -1,50 +1,117 @@
...
+      this[field].value = value
+    },
     async submit() {
       const self = this;
+      self.busy = true;
       self.error = false;
       try {
+        if (!self.usernameField.value.data || !self.passwordField.value.data) {
+          throw new Error('Username and password required.');
+        }
```
boutell, line 100: You could use an AposSchema component to avoid reimplementing validation etc. I think that is probably a good idea.

lib/modules/@apostrophecms/login/src/apos/components/TheApostropheLogin.vue
```diff
@@ -1,50 +1,107 @@
...
+          required: true
+        }
+      ],
+      context: {},
+      contextLoading: true
     };
   },
+  async beforeCreate () {
+    try {
+      const context = await apos.http.get(`${apos.modules['@apostrophecms/login'].action}/context`, {});
```
boutell, line 78: busy: true?

lib/modules/@apostrophecms/schemas/src/apos/components/AposInputWrapper.vue
```diff
@@ -55,10 +55,17 @@ export default {
     };
   },
   computed: {
+    type () {
+      if (this.field.type) {
```
boutell, line 59: Why is this necessary, when is a field not passed to AposInputWrapper?

lib/modules/@apostrophecms/login/src/apos/components/TheApostropheLogin.vue
```diff
@@ -1,50 +1,107 @@
...
+      >
+        {{ context.env }}
+      </label>
+      <label class="apos-login__project apos-login__project-name">{{ context.name }}</label>
+      <label class="apos-login--error">{{ error }}</label>
+    </div>
+
+    <div class="apos-login__body">
+      <form>
+        <AposSchema
```
boutell, line 21: AposSchema is still not following the pattern from ApostropheSchemaEditor where it implements support for v-model and the value it emits as `input` contains both an overall `hasErrors` property and an object with all of the field values. That should be fixed before this goes forward.

boutell: I see it is fixed in the string field PR, so when that is finalized this PR can be updated, and it should rely on the `hasErrors` property of the modeled value to decide if the form is ready to submit or not.

stuartromanek:
@falkodev I'm not able to get this branch up and running .. is this branch all up to date with `3.0` and plays nicely with the `a3-demo` repo?

modules/@apostrophecms/login/ui/apos/components/TheApostropheLogin.vue
```diff
@@ -1,55 +1,277 @@
...
+        {
+          name: 'password',
+          label: 'Password',
+          placeholder: 'Enter password',
+          type: 'string',
+          required: true
+        }
+      ],
+      context: {},
+      contextLoading: true
```
falkodev, line 75: I put this to have a loading screen before displaying the actual login screen. Now, there are infos to get from http://localhost:3000/api/v1/@apostrophecms/login/context (environment, version, project name). So it takes a few ms.

If I put `busy: true`, the login screen is still displayed (maybe a spinner will appear in the future), but the view is re-rendered when the `context` info arrive.

And I find this re-rendering ugly:
login2

Here is my suggestion (but idk if there is a design from @stuartromanek that could be better):
login3

falkodev:
> @falkodev I'm not able to get this branch up and running .. is this branch all up to date with `3.0` and plays nicely with the `a3-demo` repo?

You can try again @stuartromanek. I have updated my branch.

modules/@apostrophecms/login/ui/apos/components/TheApostropheLogin.vue
```diff
@@ -1,55 +1,277 @@
 <template>
```
boutell, line 1: Rename file to "TheAposLogin.vue" to match the new pattern.

modules/@apostrophecms/login/ui/apos/components/TheApostropheLogin.vue
```diff
@@ -1,55 +1,277 @@
...
+      <label class="apos-login__logo-name">ApostropheCMS</label>
+      <label class="apos-login__project-version">Version {{ context.version }}</label>
+    </div>
   </div>
 </template>
 
 <script>
-
 export default {
   name: 'TheApostropheLogin',
```
boutell, line 49: Rename

modules/@apostrophecms/login/ui/apos/components/TheApostropheLogin.vue
boutell: What is wrong with using `TheApostropheBusy` in the normal way? If it is not finished then it should get finished, we should not reinvent it here.

boutell: I am saying that `busy: true` should be passed to apos.http.get. There should be no need for another busy screen implementation.

modules/@apostrophecms/schema/index.js
```diff
@@ -2343,16 +2343,18 @@ module.exports = {
   extendMethods(self, options) {
     return {
       getBrowserData(_super, req) {
-        if (!req.user) {
+        if (req.scene !== 'apos') {
           return;
         }
-        const browserOptions = _super(req);
```
boutell, line 2349: You can keep this code if you fix `module/index.js` to also look at `req.scene`. You should do that to avoid missing out on standard things.

modules/@apostrophecms/module/index.js
```diff
@@ -393,7 +393,7 @@ module.exports = {
       // with editing privileges.
 
       getBrowserData(req) {
-        if (req.user) {
+        if (req.user || req.scene === 'apos') {
```
boutell, line 396: Just `req.scene === 'apos'` should be enough.

modules/@apostrophecms/login/index.js
```diff
@@ -196,6 +197,20 @@ module.exports = {
             return user;
           }
         } : {})
+      },
+      get: {
+        async context () {
+          try {
+            const aposPackage = await fs.readJson('./node_modules/apostrophe/package.json');
```
falkodev, line 204: @boutell Stuart wanted the Apostrophe version, not the project one. Is this correct or should I use `self.apos.rootDir`?

modules/@apostrophecms/login/index.js
```diff
@@ -196,6 +197,22 @@ module.exports = {
             return user;
           }
         } : {})
+      },
+      get: {
+        async context () {
+          let aposPackage = {};
+          try {
+            aposPackage = await fs.readJson('./node_modules/apostrophe/package.json');
```
falkodev, line 205: @boutell Stuart wanted the Apostrophe version, not the project one. Is this correct or should I use self.apos.rootDir?

boutell: This code is inside the apostrophe module, therefore you can use relative paths within the module. Don't try to read from node_modules.

boutell: Use require instead. require can read json files, and it is always relative to the file making the require call.

modules/@apostrophecms/module/index.js
boutell: This gets cleaned up permanently in my PR.

PR #2286, opened 2020-08-03
Modal area editing take 2

boutell:
Enables area editing inside a modal. To see that you'll need the corresponding branch of a3-demo, which has a "blurb" area for "products."

Does not break contextual area editing.

I removed the "status" prop from all components because it duplicated the function of the `error` property of the `value` prop. I updated stories to account for this. The error property now actually works with the new fields, although we should do more with it (currently it's just error: true but our validator functions could return an error string to do more with the plan I see laid out in AposInputWrapper). That would be a subsequent ticket, if you don't mind, as this is already a progressive improvement from errors not being hooked up at all.

I almost removed `modifiers` too but that is also a topic for a new ticket. Basically I think the modifiers should be taken from `field`, and all remaining direct uses of the field components should be fixed to use AposSchema.

When changes are made in an area they are emitted up the component hierarchy, as well as saving them on the fly if `docId` is known. Note that `docId` will **never** be known to AposSchema and its children when editing in a modal, not even for an existing piece, because that would modify things without waiting to see if the user clicks `cancel` or not. Patching on the fly is only appropriate outside of the modals.

I also reworked `ApostropheAreaEditor` to always create a new `next` array on changes in order to simplify the `v-model` implementation.


modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -53,34 +47,27 @@ export default {
...
     });
-    if (this.doc._id) {
-      next.data._id = this.doc._id;
-    }
 
     return {
       next,
       fieldState,
-      fieldComponentMap: window.apos.schemas.components.fields || {}
+      fieldComponentMap: window.apos.schema.components.fields || {}
```
falkodev, line 58: I have this error.

```js
[Vue warn]: Error in data(): "TypeError: window.apos.schema is undefined"

found in

---> <AposSchema> at apostrophe/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
       <ApostropheModal> at apostrophe/modules/@apostrophecms/modal/ui/apos/components/ApostropheModal.vue
         <ApostrophePiecesInsertModal> at apostrophe/modules/@apostrophecms/piece-type/ui/apos/components/ApostrophePiecesInsertModal.vue
           <AposModal> at apostrophe/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
             <AposPiecesManager> at apostrophe/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
               <TheAposModals> at apostrophe/modules/@apostrophecms/modal/ui/apos/components/TheAposModals.vue
                 <Root>
```

modules/@apostrophecms/schema/index.js
```diff
@@ -2351,6 +2351,7 @@ module.exports = {
         for (const name in self.fieldTypes) {
           fields[name] = 'AposInput' + self.apos.util.capitalizeFirst(name);
         }
+        console.log(fields);
```
falkodev, line 2354: Forgotten `console.log`

modules/@apostrophecms/area/ui/apos/components/ApostropheAreaEditor.vue
```diff
@@ -47,74 +47,124 @@ export default {
...
   methods: {
     async up(i) {
-      await apos.http.patch(`${apos.doc.action}/${this.docId}`, {
-        busy: true,
-        body: {
-          $move: {
-            [`@${this.id}.items`]: {
-              $item: this.next[i].widget._id,
-              $before: this.next[i - 1].widget._id
+      if (this.docId) {
```
falkodev, line 61: It works in the modal, but this button appears on the page, and when clicking it, there is this error: 

```js
[Vue warn]: Error in v-on handler (Promise/async): "[object XMLHttpRequest]"

found in

---> <ApostropheAreaEditor> at apostrophe/modules/@apostrophecms/area/ui/apos/components/ApostropheAreaEditor.vue
       <Root>
```

Maybe remove the button on the page if it is intended to be modified only in the modal

boutell: Both the up and the down button work for me "on the page" as well, please clarify how to reproduce this issue with steps. Thanks!

boutell: Tried a new project to be sure, still unable to reproduce an issue with the up or down button "on the page."

boutell:
@falkodev I think I fixed the bug you found, it seems specific to nested widget situations and had to do with a prop name change that threw off Apostrophe's ability to track the true `_docId` into sub-widgets.

In addition, I implemented a mechanism to make sure other contextual areas on the page know when any patch is made via contextual editing, as there were situations where they could mistakenly believe they had complete information about sub-areas. That closes another issue on our board.

PR #2287, opened 2020-08-03
styles and transitions changes

stuartromanek:
style changes


modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -1,52 +1,56 @@
 <template>
-  <div class="apos-login apos-theme-dark">
-    <div class="apos-login__overlay"></div>
-    <div class="apos-login__menu-overlay"></div>
-    <AposLoginBackground />
+  <transition name="fade-stage">
+    <div class="apos-login apos-theme-ssdark" v-show="loaded">
```
falkodev, line 3: `apos-theme-ssdark` is it a typo? or do you want just to disable the dark theme?

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -109,21 +116,47 @@ export default {
...
-    &:after {
-      position: absolute;
-      content: '';
-      top: 0;
-      left: 0;
-      display: block;
-      width: 100%;
-      height: 100%;
-      background: linear-gradient(38.7deg, rgba(179, 39, 191, 0.3), rgba(30, 30, 76, 0.3), rgba(0, 192, 154, 0.3));
+    &.apos-theme-dark {
```
falkodev, line 153: I think this rule is not needed anymore

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -215,10 +247,8 @@ export default {
     form {
       z-index: $z-index-manager-toolbar;
```
falkodev, line 248: Every `z-index` rule was necessary with a background image and a filter. They can be removed now there is a white background.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -245,12 +275,14 @@ export default {
     &__footer {
       z-index: $z-index-manager-display;
       position: fixed;
+      right: 0;
       bottom: 32px;
-      left: 32px;
+      left: 0;
       display: flex;
+      width: fit-content;
```
falkodev, line 282: In FF, this rule is invalid, causing the logo and the version to be stuck on the right and on the left of the screen without padding/margin. I would change `position: fixed` to `position: absolute` and remove `right` and `left`rule. 

Also, the version is the project's version not Apostrophe's. If the original idea was to put forward the version of Apostrophe used in the project, I have to make a code change. If it was for the version of the project (label "version" in package.json), I suggest to move the version under or next to the project's name (in `apos-login__header`)  and to leave in the footer only the logo and the label "Apostrophe CMS", both centered in the same width as the body (320px).

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
stuartromanek: oh yes, was testing light/default

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
stuartromanek: Yes, my thought was for it to be the current version of Apostrophe, which feels like it makes sense with the CMS brand footer.

PR #2293, opened 2020-08-05
fix patch subobject bug so areas do not disappear

boutell:


boutell:
This fixes a lingering bug stemming from my area editing / edit preserving PR that was merged. Without it stuff disappears, so it is fairly urgent as 3.0 PRs go 😆

PR #2295, opened 2020-08-06
Removes superfluous macro

### LINKED ISSUE #2135: Remove superfluous `areaBody` macro.
This macro does not look to be used anywhere: https://github.com/apostrophecms/apostrophe/blob/master/lib/modules/apostrophe-schemas/views/macros.html#L215-L218

```
{%- macro areaBody(field) -%}
  {# js adds this area to the dialog #}
  <div data-{{ field.name }}-edit-view></div>
{%- endmacro -%}
```

abea:
Closes #2135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.110.1 (2020-08-12)
+
+* Removes the `aposBody` template macro, which was unused.
+
 ## 2.110.0 (2020-07-29)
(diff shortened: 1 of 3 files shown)


boutell:
LGTM. I can find no evidence in `git log -p` that we ever used it anywhere, so it is very unlikely that someone else... somehow does. But if they did we would just bring it back marked "deprecated, no functional use, retained for bc."

PR #2309, opened 2020-08-14
3/piece editor

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-121, connecting AposDocEditor to the piece creation API.

I had to refactor AposDocEditor since it wasn't built to work this way before. This trickled down to edits to the schema component and some others. I tested and TheAposLogin component still works with the AposSchema changes.


modules/@apostrophecms/doc-type/index.js
```diff
@@ -97,9 +97,15 @@ module.exports = {
...
         label: 'Basics',
         fields: [
-          'title',
-          'slug',
-          'published'
+          'title'
+        ]
+      },
+      {
+        name: 'utility',
```
abea, line 104: These fields should be in the right sidebar in the editor modal.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.stories.js
```diff
@@ -2,19 +2,15 @@ import { storiesOf } from '@storybook/vue';
...
       AposDocEditor
     },
     data () {
       return {
-        schema: data.schema,
-        groups: data.groups,
-        doc: data.doc,
-        typeLabel: 'Article',
+        moduleName: 'products',
+        id: null, // TODO: Add a demo document to the mock API.
```
abea, line 13: This will be relevant in the "Edit a piece" ticket.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -112,27 +85,92 @@ export default {
   },
 
   computed: {
+    moduleOptions() {
+      return window.apos.modules[this.moduleName] || {};
+    },
+    schema() {
+      return this.moduleOptions.schema || [];
+    },
+    groups() {
```
abea, line 94: Creating field groups ("arrangeFields")

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -112,27 +85,92 @@ export default {
...
       }
-      return fields;
     }
   },
   async mounted() {
-    // TODO: Get data here.
     this.modal.active = true;
+
+    if (this.docId) {
+      // TODO: Get data here.
```
abea, line 153: This will be done in the "edit a piece" story as well.

modules/@apostrophecms/schema/index.js
```diff
@@ -1155,11 +1155,11 @@ module.exports = {
 
         // always make sure there is a default group
         const defaultGroup = self.options.defaultGroup || {};
-        let groups = [{
-          name: defaultGroup.name || 'default',
-          label: defaultGroup.label || 'Info',
+        let groups = [ {
+          name: defaultGroup.name || 'other',
```
abea, line 1159: I'm differentiating here between default fields and unarranged fields. Maybe this should be `unarranged` for the name property instead?

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
```diff
@@ -0,0 +1,120 @@
...
+          class="apos-input-icon"
+          :is="icon"
+        />
+      </div>
+    </template>
+  </AposInputWrapper>
+</template>
+
+<script>
+// NOTE: This is a temporary component, copying AposInputString. Base modules
```
abea, line 36: See here. I kept getting errors, so I stubbed this in.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -112,27 +85,92 @@ export default {
...
+    moduleOptions() {
+      return window.apos.modules[this.moduleName] || {};
+    },
+    schema() {
+      return this.moduleOptions.schema || [];
+    },
+    groups() {
+      const groupSet = {
+        basics: {
+          label: 'Basics',
```
boutell, line 97: Making sure everything is in a fallback group by the end of "compose" should be the server's problem because it's easier to configure, i18n and change things there. But, maybe a fallback here can't hurt.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
boutell: I am not sure if you are repurposing this for "page settings," but if you are, start thinking ahead to what happens when the user changes the page type. That can change the schema and the tabs on the fly.

modules/@apostrophecms/schema/index.js
boutell: `unarranged` sounds smart.

modules/@apostrophecms/schema/ui/apos/components/AposInputBoolean.vue
```diff
@@ -9,7 +9,7 @@
           class="apos-sr-only"
           type="checkbox" :id="`${uid}-true`"
           :value="true" @change="setValue(true)"
-          :checked="value.data === true"
+          :checked="value && value.data === true"
```
boutell, line 12: This kind of thing is going to wind up all over the components. Let's not do that. Instead the component that owns the AposSchema should always make a document using `field.def` as appropriate before handing it to `AposSchema`.

boutell: If you look at the 2.x code this is how it's done, there's a browser-side newInstance method.

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
boutell: Makes sense.

modules/@apostrophecms/schema/index.js
abea: I'm sure people with put things in this intentionally at some point. I guess that's my hesitation at the moment.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: True. Though if I'm treating this as the fallback it should be "Other".

abea: Maybe this is a bad idea. We allow people to set their own `defaultGroup`.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: I moved the schema population into a method to make it reusable in this context and future contexts. I'm not going to do too much business logic for the page settings type case right now since it's not the subject of the current ticket, but I'll make sure to create a ticket for that.

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
```diff
@@ -0,0 +1,120 @@
+<template>
```
falkodev, line 1: I don't know if this is related to this new component, but you type in the "slug" field, the "title" field becomes `undefined` (the value in the Vue component, not the rendered input in the DOM) and vice versa.

abea: You saw that using this branch? I'm not sure about that. This is simply a placeholder component to avoid all of the errors since it was missing. This component should not be considered complete at all.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -171,6 +162,11 @@ export default {
       } finally {
         apos.bus.$emit('busy', false);
       }
+    },
+    poplateSchema () {
```
boutell, line 166: populate

abea: gracias

PR #2311, opened 2020-08-17
3/doc versions

abea:
## What's in this?
- Display the versions in a tree component with no header
- Each day is grouped in a collapsing branch with the last version of the day as the parent.

Other features are coming in other PRs, for the sake of the reviewer.


modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -30,8 +30,8 @@
           v-for="(col, index) in headers"
           :key="`${col.name}-${index}`"
           :is="col.type === 'link' ? 'a' : col.type === 'button' ? 'button' : 'span'"
-          :href="col.name === '_url' ? row[col.name] : false"
-          :target="col.name === '_url' ? '_blank' : false"
+          :href="col.type === 'link' ? row[col.name] : false"
```
abea, line 33: This seems to make more sense than looking for a specific row name.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocHistory.vue
```diff
@@ -0,0 +1,144 @@
...
+        }
+      )).versions;
+
+      docVersions = this.formatVersions(docVersions);
+
+      if (docVersions) {
+        this.versions = docVersions;
+      }
+    },
+    openEditor(event) {
```
boutell, line 109: Is this connected to anything? I don't see it being invoked.

Is editing the document an action that makes sense from the doc history modal? If you did, what relationship would that action bear to a particular version...?

I have thoughts on implementation but I think these two questions should be answered first so I know what I'm talkin' bout.

abea: No, that was from the component I copied to start this one. I'll remove this method.

PR #2313, opened 2020-08-17
3/versions select

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-347, "Page History: As a user, I can select a row by clicking on it, expanding the row." This does not include UI to actually restore the version. After clicking a row, the `selected` property on `AposDocHistory` will have the ID of the selected row in it.

### DIFF
--- a/modules/@apostrophecms/page/ui/apos/components/AposPagesManager.stories.js
+++ b/modules/@apostrophecms/page/ui/apos/components/AposPagesManager.stories.js
@@ -1,8 +1,8 @@
 export default {
-  title: 'Pages Organize'
+  title: 'Pages Manager'
 };
 
-export const pagesOrganize = () => {
+export const pagesManager = () => {
   return {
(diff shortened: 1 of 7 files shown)


modules/@apostrophecms/page/ui/apos/components/AposPagesManager.stories.js
```diff
@@ -1,8 +1,8 @@
 export default {
-  title: 'Pages Organize'
+  title: 'Pages Manager'
```
abea, line 2: This was a bit extracurricular, but tangentially related.

falkodev:
@abea How to test? I cannot access the pages entry in the admin bar

abea:
Right now this is only a Storybook story, so you can run it using the `a3-demo` site with `npm run storybook`

falkodev (approved):

PR #2314, opened 2020-08-18
Lower default height for tabs in array selection

RyamBaCo:
When selecting entries from an pieces array the last entry is not visible when the browser window is only a few pixels smaller than the last entry. Here the last entry can be selected, but the entry itself is not visible. To fix this issue i lowered the height by 61px (default size of one array entry) so that the scroll bars appear in time.

Before no scrollbar is visible an it is not clear that there is another element to select:
before
After my fix the scrollbars appear sooner, therefore the user can scroll down to the last entry:
after

### DIFF
--- a/lib/modules/apostrophe-modal/public/css/components/modal-tabs.less
+++ b/lib/modules/apostrophe-modal/public/css/components/modal-tabs.less
@@ -5,7 +5,7 @@
     position: fixed;
     top: 0;
     float: left;
-    height: 100%;
+    height: calc(100% - 61px);
     width: 25%;
     background-color: @apos-light;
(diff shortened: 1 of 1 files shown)


stuartromanek:
@RyamBaCo What browser are you experiencing this bug in? 

Things appear normal in 
- macos chrome/brave chromium 84.0.4147.125
- macos firefox 79
- macos safari 13.1.1
- windows 10 Edge 84.0
- windows 10 IE11 (has some rendering issues but not this one)
- windows 8.1 IE11 (again, IE11 rendering issues but not this)

abea:
Close for inactivity.

PR #2315, opened 2020-08-18
Moves notifications story

abea:
This had not been moved to the correct directory.


modules/@apostrophecms/notification/ui/apos/components/AposNotification.stories.js
```diff
@@ -5,15 +5,13 @@ import {
   boolean,
   number
 } from '@storybook/addon-knobs';
-import AposNotification from './AposNotification.vue';
```
abea, line 8: We don't need to import them anymore.

PR #2317, opened 2020-08-18
3/arborist

abea:
Cleans up AposTree. Get it?

Resolves https://apostrophecms.atlassian.net/browse/A30U-332


modules/@apostrophecms/ui/ui/apos/components/AposTree.vue
```diff
@@ -72,7 +72,7 @@ export default {
   data() {
     return {
       // Copy the `rows` property to mutate with VueDraggable.
-      myRows: [],
+      myRows: this.rows,
```
boutell, line 75: Should probably slice() that puppy, if it's in `data` you are presumably going to mutate it, and you should never mutate props.

boutell: If you're going to mutate subproperties of it too, you should deeply clone it, but maybe that's not something you do.

abea: Got klona. Use klona.

PR #2319, opened 2020-08-19
Date field

falkodev:
## How to test

- In a3-demo project, add this to `modules/product/index.js` under `fields/add` object:
```js
date: {
   type: 'date'
}
```
- Create a new product and go to the date field

image


abea:
How is this different from the string field with `type="date"`? It is already ready to use the calendar icon: https://github.com/apostrophecms/apostrophe/blob/3.0/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue#L59. If we're going to have a separate field component we should remove that line.

falkodev:
> How is this different from the string field with `type="date"`? It is already ready to use the calendar icon: https://github.com/apostrophecms/apostrophe/blob/3.0/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue#L59. If we're going to have a separate field component we should remove that line.

`AposInputString` is used here with the type `date`. The line you mentioned is then used to display the calendar icon.

stuartromanek:
@falkodev I think what @abea is asking is why did this get broken out into a separate component when we have already solved this one way?

modules/@apostrophecms/schema/index.js
```diff
@@ -381,6 +381,7 @@ module.exports = {
 
     self.addFieldType({
       name: 'integer',
+      frontComponent: 'string',
```
falkodev, line 384: Not sure about the naming of this prop. Open to new ideas..

boutell: `vueComponent`, and it should be the complete vue component name, it shouldn't go through any more automatic name picking logic. If we are bypassing it then we are bypassing it completely.

abea:
🎉

PR #2321, opened 2020-08-19
3/open piece

abea:
Resolves:
- [Opening an existing piece from the manager](https://apostrophecms.atlassian.net/browse/A30U-371)
- [Editing an existing piece](https://apostrophecms.atlassian.net/browse/A30U-122)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -27,6 +33,7 @@
           <AposModalTabsBody>
             <div class="apos-doc-editor__body">
               <AposSchema
+                v-if="docReady"
```
abea, line 36: If we get this up immediately the populated doc isn't ready yet and it's all blank.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -163,10 +192,11 @@ export default {
         apos.bus.$emit('busy', false);
       }
     },
-    populateSchema () {
+    buildEmptyDoc () {
       this.schema.forEach(field => {
-        this.doc.data[field.name] = {};
```
abea, line 168: This was a bug, but we didn't know because the schema jawn was being generated before it was done building this.

modules/@apostrophecms/storybook/template/public/mocks.js
```diff
@@ -223,55 +223,63 @@
 
   apos.http = {};
 
+  const grape = {
```
abea, line 226: 🍇

modules/@apostrophecms/storybook/template/public/mocks.js
```diff
@@ -223,55 +223,63 @@
...
+    'highSearchWords': [
+      'grape'
+    ],
+    'lowSearchText': 'grape grape',
+    'searchSummary': '',
+    'docPermissions': [],
+    '_edit': true
+  };
+
+  const strawberry = {
```
abea, line 249: 🍓

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -143,18 +151,39 @@ export default {
...
 
     if (this.docId) {
-      // TODO: Get data here.
+      let docData;
+      try {
+        const getOnePath = `${this.moduleOptions.action}/${this.docId}`;
+        docData = await apos.http.get(getOnePath, {
+          busy: true
+        });
+      } finally {
```
boutell, line 160: Does `finally` make sense here? Can it really cope if `this.doc.data` is undefined? It is good not to just get wedged, it's true. Maybe we need a catch block too to display an error, if that's a thing we're ready to do?

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -39,6 +46,7 @@
       <AposModalRail type="right">
         <div class="apos-doc-editor__utility">
           <AposSchema
+            v-if="docReady"
```
boutell, line 49: clever way to defer this.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -163,10 +192,11 @@ export default {
         apos.bus.$emit('busy', false);
       }
     },
-    populateSchema () {
+    buildEmptyDoc () {
       this.schema.forEach(field => {
-        this.doc.data[field.name] = {};
+        this.doc.data[field.name] = '';
```
boutell, line 197: This is wrong for lots of field types and also doesn't respect `def` and... yeah we need to figure that out. I think we'll wind up wanting to get a placeholder doc back from the server, which can work out exactly what this particular user should get in a new empty doc of this particular type at this particular time, using the existing `newInstance` method.

What should that URL look like though? Hmmm... I think you `POST` like you're going to make a new piece, but you do it with `?template=1`, and that does *not* insert anything in the database, it just calls `newInstance` server side and returns that object to you: a "template" for a new doc. What do you think?

boutell: Actually I'm not sure why we're doing this at all in the case where we're editing an existing doc. My comments above make sense for a *new* doc. And in certain situations when switching types.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: There should be a catch. We'd probably want to deliver an error an then exit out of the modal.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: This is only used when you are creating a new doc. I get your point about `def` and the property type issues. What this is doing is building an empty piece, so `newInstance` is perfect for this. I forgot about that method. As for the route, it sounds like a `GET` situation since you're _getting_ the blank doc from the server ("Excuse me waiter, would you get me a new place setting for my friend?").

abea: @boutell If it's alright with you, I'd like to handle this in a separate PR. It's not about editing existing pieces. 

What I'm imagining, to be RESTful, is something like `/api/v1/product/new` hitting `getOne`, seeing the `new` "_id" and returning a new instance.

boutell: OK this is horribly pedantic but... I just saw the words "new" and "GET" in the same sentence and my gut said "but that's not RESTful! The two words that belong in the same sentence are 'new' and 'POST'!"

I get (cough) that we're not actually putting it in the database, but we *are* making a new thing, even if only to return it. And POST does always return the new object (because it'll have properties you weren't sure of until apostrophe inserted it, like slug).

Yes this can go in a separate ticket.

abea: Ticket created: https://apostrophecms.atlassian.net/browse/A30U-375. 🙇

PR #2323, opened 2020-08-20
Adds filter labels for piece manager

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-295

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposFilterMenu.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposFilterMenu.vue
@@ -66,7 +66,7 @@ export default {
           field: {
             name: filter.name,
             type: 'radio',
-            label: filter.label,
+            label: filter.label || filter.name,
             choices: filter.choices
           },
           value: {
--- a/modules/@apostrophecms/ui/ui/apos/components/AposFilterMenu.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposFilterMenu.vue
@@ -22,7 +22,7 @@
 import AposHelpers from '../mixins/AposHelpersMixin';
 
 export default {
-  mixins: [AposHelpers],
+  mixins: [ AposHelpers ],
   props: {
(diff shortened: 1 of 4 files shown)


falkodev (approved):

falkodev:
I don't understand what was wrong or missing before. But no harm in this PR.

modules/@apostrophecms/ui/ui/apos/components/AposFilterMenu.vue
```diff
@@ -65,9 +66,7 @@ export default {
           field: {
             name: filter.name,
             type: 'radio',
-            // TODO: Remove `|| filter.name.toUpperCase()` once filters include
```
boutell, line 68: Should we support a more graceful automatic naming if label is missing? We do for schema fields. It's a bit of a parlor trick, but devs do appreciate that "save me all the time" stuff.

```
        if ((!field.label) && (!field.contextual)) {
          field.label = _.startCase(field.name.replace(/^_/, ''));
        }
```

Does a reasonably pretty job.

abea: At the moment it fails to have the filter label missing, but doesn't break the app. I'd be alright with defaulting to the `name` values. I'm hesitant to install a utility for this kind of thing. I'd rather make it clear that something needs fixing and how to fix it.

abea:
@falkodev The filter labels were displaying incorrectly because we didn't have `label` values for them and so they would end up displaying all-caps rather than title-case as in the designs.

modules/@apostrophecms/ui/ui/apos/components/AposFilterMenu.vue
boutell: Yeah, if we do it, it should be done server side where lodash is already available, and the configuration that reaches Vue should be ready to rock with a label. Which, puts it outside the scope of your PR. Approving.

PR #2329, opened 2020-08-23
simple joins

boutell:
The model layer of apostrophe now expects you to just pass an array of docs to be joined as a property with the name of the join field. For instance, `_products` can be set by literally passing in an array of product pieces. Same thing you get back later when you fetch the doc. Same thing you probably just got from find() and were hoping to painlessly join with. Intuitive.

Transition to the idsField and relationshipsField is handled at a lower level in a promise event handler. Thus both front and backend developers no longer have to think about them at all, and the REST APIs also automatically work in this simple and intuitive way.

This is NOT a way to *create new docs* of the type you're joining with, or update the joined docs themselves. You should still do that separately.  For updating the join all Apostrophe really cares about is the _id property of each joined doc, and the _relationship property if appropriate. Supporting updates would be a performance & race conditions nightmare, have lousy error handling, etc.

This is still a draft because I need to add tests that cover the relationship feature, and update all existing tests to work this way. However `npx mocha test/pieces` passes at the moment, notably including a bunch of RESTful tests of joins (not with relationships though).

I'm mindful of the proposal to rename `relationship` to `fields`, etc., and I'm in favor of it, but for purposes of this PR, one thing at a time.

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -255,8 +255,8 @@ module.exports = {
   handlers(self, options) {
     return {
       beforeSave: {
-        async joinsToStorage(req, doc) {
-          self.apos.schema.joinsToStorage(doc);
+        async prepareJoinsForStorage(req, doc) {
+          self.apos.schema.prepareJoinsForStorage(req, doc);
         }
       },
       afterSave: {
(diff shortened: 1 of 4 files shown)


boutell:
All tests passing now, including tests covering PATCHing with a relationship, so this is no longer a draft.

boutell:
For @falkodev the big win here is that his Vue components for joins can just use v-model normally like other fields do, and in fact can work the way he's already written them: the value they provide is an array of joined docs.

falkodev (approved):

PR #2333, opened 2020-08-25
fix story, fix interaction states

stuartromanek:
this got stepped on when moving over from standalone storybook


stuartromanek:
forgot that this has to be squared with new join work.

@falkodev there is not a standalone join field storybook component .. do the mock apis exist for such a thing? We should have one in storybook

abea:
I don't think we have mock APIs for slats or joins yet.

abea:
@stuartromanek Do you want a review on the draft?

abea:
Removing reviewers until it's ready to go.

abea:
@stuartromanek This can be retired now, right?

PR #2335, opened 2020-08-25
3/pieces search

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-128

Adds working text search to pieces manager modal.

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
@@ -25,7 +25,8 @@
         @input="filter"
       />
       <AposInputString
-        @input="search" :field="searchField.field"
+        @input="search" @return="search($event, true)"
+        :field="searchField.field"
         :status="searchField.status" :value="searchField.value"
         :modifiers="['small']"
       />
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
```diff
@@ -25,7 +25,8 @@
         @input="filter"
       />
       <AposInputString
-        @input="search" :field="searchField.field"
+        @input="search" @return="search($event, true)"
```
abea, line 28: Here's where we set the `force` argument to trigger search regardless of the number of characters entered.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
```diff
@@ -74,7 +76,8 @@ export default {
         field: {
           name: 'search',
           placeholder: 'Search Images',
-          icon: 'magnify-icon'
+          icon: 'magnify-icon',
+          enterSubmittable: true
```
abea, line 80: The string already had a return event for AposTagApply that didn't return the value. There wouldn't be any harm in always including the value formatted like this (it's the same format as the `input` emission), but this make it an option.

abea:
Back to draft. I need to store the search query for pagination.

falkodev (approved):

PR #2336, opened 2020-08-25
Adds pagination to piece manager

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-126


abea:
This is true.

abea:
@boutell You're right. I was worried about a flash of the first page with old filters before it was refreshed with new filters, but that wouldn't happen without calling `getPieces` twice. I'll switch this up.

stuartromanek:
change Menu Swap Icon (up/down arrows in pagination select) size 18

stuartromanek:
image

change spacing rhythm here and add **Page** to the select box, making the UI a consistent series of boxes

PR #2337, opened 2020-08-25
AposAttachments UI & story

stuartromanek:


stuartromanek:
This is only a partial storybook implementation .. there are a number of things that need to be cleaned up about Slats to make a unified experience across the components that use SlatList.. After that happens I can do a pt 2 here where we emit events up when changing the field value (removing items). We also need to decide on a file drop approach.

abea:
Linties.

falkodev (approved):

modules/@apostrophecms/schema/ui/apos/components/AposAttachment.vue
```diff
@@ -0,0 +1,104 @@
...
+          />
+        </div>
+      </div>
+    </template>
+  </AposInputWrapper>
+</template>
+
+<script>
+import AposInputMixin from '../mixins/AposInputMixin.js';
+import AposSlatList from '../../../../ui/ui/apos/components/AposSlatList';
```
abea, line 25: We should use the `Modules` webpack alias for both of these.

modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
```diff
@@ -18,11 +18,24 @@
...
+            {{ item.ext }}
+          </span>
+        </div>
         <div class="apos-slat__label">
           {{ item.title }}
         </div>
       </div>
       <div class="apos-slat__secondary">
+        <div class="apos-slat__size" v-if="item.size">{{ item.size }}</div>
         <close-icon
```
abea, line 39: This close icon needs to be a button. You could use the `AposButton` and the icon props, or wrap this directly here.

modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
```diff
@@ -150,4 +167,55 @@ export default {
...
+  .apos-slat__extension--numbers,
+  .apos-slat__extension--csv {
+    background-color: var(--a-spreadsheet);
+  }
+
+  .apos-slat__extension--key,
+  .apos-slat__extension--ppt,
+  .apos-slat__extension--pptx {
+    background-color: var(--a-presentation);
+    // color: var(--a-primary-text);
```
abea, line 209: comment here.

modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
```diff
@@ -90,8 +91,11 @@ export default {
       this.engaged = null;
     },
     remove(item, focusNext) {
+      console.log('removing');
```
abea, line 94: Three logs here.

abea (changes requested):
When I remove one to be less than the max, the "Limit Reached" message doesn't disappear. It probably needs to refresh that on the `remove` event.

modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
stuartromanek: made a new `inline` button modifier that is basically an AposButton with no chrome

stuartromanek:
@abea better limit watching

abea (approved):

PR #2339, opened 2020-08-26
Moves storybook dependencies to main dependencies

abea:


abea:
I'm not sure why Circle is suddenly finding tons of ESLint errors.

boutell:
The spaces-around-arrays thing is indeed our past standard come back because we got more specific to avoid having it flip the other way due to a changed default upstream.

Automatic detection of of `let`s that should have been `const`s sounds like a new feature, but a totally fricking awesome one that should have been there sooner.

abea:
We've sort of let that array spacing thing happen. I don't think we have a rule for it. We could switch it if we want.

boutell:
I agree strongly with the change that array spacing rule is making. It's what the upstream default behavior used to be, so we didn't have an explicit rule, then the upstream behavior changed. Greg and I (and others I think, it was in channel) talked about it and I believe Greg added the explicit rule to deal with it and make the spaces come back. So what we're seeing is probably the other shoe dropping.

PR #2341, opened 2020-08-27
3/unpublished piece

falkodev:
https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?assignee=5ddbce7e47baa50cfd05eed2&selectedIssue=A30U-377

https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?assignee=5ddbce7e47baa50cfd05eed2&selectedIssue=A30U-378

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRadio.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRadio.vue
@@ -39,8 +39,7 @@ export default {
     getChoiceId(uid, value) {
       // Convert any boolean values for this purpose.
       if (typeof value !== 'string') {
-        value = !value && value === null ? 'null'
-          : !value ? 'undefined' : value.toString();
+        value = !value ? 'undefined' : value.toString();
       }
 
       value = value.toString();
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -84,7 +84,7 @@ module.exports = {
             label: 'Draft'
           },
           {
-            value: null,
+            value: 'any',
             label: 'Both'
           }
         ],


modules/@apostrophecms/schema/ui/apos/components/AposInputRadio.vue
```diff
@@ -39,7 +39,7 @@ export default {
     getChoiceId(uid, value) {
       // Convert any boolean values for this purpose.
       if (typeof value !== 'string') {
-        value = !value && value === null ? 'null'
+        value = !value && value === null ? 'any'
```
abea, line 42: It is working for the pieces manager, but this is the general radio input. Could there be any problems from always making the radio input have `'any'` value when it's originally `null`?

boutell: 'any' is a plausible value (not meaning "null") for the general case of the radio input, so something may have to wrap the radio input for purposes of serving as a "boolean plus null" component. Alternatively, the `launder` module could be updated to accept the string `'null'` as another "null equivalent" suitable for use with query strings, which don't have a native representation of null. That's how `'any'` was originally chosen. The nice thing about `'any'` is that it looks good in user visible URLs, but this isn't one, and launder doesn't have to give up support for `'any'`.

abea: Maybe if the piece type filters aren't going to use `null` for the all-inclusive option then we don't need a case in the radio input for `null` at all. Falsy values could always return `'undefined'`, forcing us all to set some representative value, such as `'any'` like we're doing in `piece-type` here.

abea: Okay, my suggestion is to simply remove the `null`case completely. So this line ends up simply as `value = !value ? 'undefined' : value.toString()`.

We might want to revisit whether `'undefined'` is the right thing to use for falsy situations, but that's not our immediate concern.

boutell:
This won't work because the default value of the filter is not `null`. So we must pass something to say `null`. I suggest the word `null`, i.e. let's patch `launder`. I'll submit a PR.

abea:
@boutell What's the issue? The filter no longer has a null option. It's `'any'`.

boutell:
OK, gotcha. Passing `any` explicitly is what we need, yes.

PR #2343, opened 2020-08-27
3/join storybook

stuartromanek:
Adds joins to SB, adds some preliminary knobs.

@falkodev I changed some internal workings to separate the search input from the value of field (which will be pieces or ids or something). Please review this under the circumstances you originally developed the join field to make sure everything is still working!

Could be some growing pains on this one


modules/@apostrophecms/schema/ui/apos/components/AposInputJoin.stories.js
```diff
@@ -0,0 +1,138 @@
...
+        field: {
+          name: 'myJoin',
+          label: 'Link to A Page',
+          help: 'Choose a page from the joiner to link to it',
+          type: 'join',
+          withType: 'product',
+          max
+        },
+        value: {
+          data: getData(prepoulate)
```
abea, line 52: I think this will need to be an array of `_id` values. Then in the component it hits an API to get the pieces. I assume that'll be the `getOne` route once for each _id. Later we'll need a way to send up multiple IDs.

modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -1,11 +1,18 @@
...
-        <div class="item__slug">{{ item.slug }}</div>
+        <div class="item__title">
+          {{ item.title }}
+        </div>
+        <div class="item__slug">
+          {{ item.slug }}
+        </div>
       </div>
-      <div class="item__type">{{ getTypeLabel(item.type) }}</div>
+      <!-- <div class="item__type">{{ getTypeLabel(item.type) }}</div> -->
```
abea, line 15: Cancelled.

modules/@apostrophecms/schema/ui/apos/components/AposInputJoin.stories.js
abea: It won't need more than the title and the _id back from the request, though.

abea: Okay, so that's wrong. You'll have the full documents. Forget this part.

modules/@apostrophecms/schema/ui/apos/components/AposInputJoin.stories.js
```diff
@@ -0,0 +1,138 @@
...
+        :status="status"
+      />`
+  };
+};
+
+function getData(n) {
+  const data = [
+    {
+      '_id': 'htcuoykl9012j38ecrjghrn0c',
+      'published': true,
```
abea, line 72: We could clear out most of this mock data other than `title` and `_id`. Not critical, but nicer.

modules/@apostrophecms/schema/ui/apos/components/AposInputJoin.vue
```diff
@@ -24,7 +25,11 @@
           />
         </div>
         <AposSlatList @update="updated" :initial-items="items" />
-        <AposSearchList :list="searchList" @select="selected" :selected-items="items" />
+        <AposSearchList
+          :list="searchList"
+          @select="selected"
```
falkodev, line 30: The selection of searched items (to include them in the SlatList) is not working anymore

falkodev (changes requested):

abea:
@stuartromanek It seems like you might need to do a fresh install. ESLint is erroring on unnecessary quotes on object keys and whatnot.

PR #2347, opened 2020-08-28
3/media mgr

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-173.

Reviewable in Storybook for the media manager. You can also generate image pieces with `node app @apostrophecms/image:generate 20` to run it in the app (though it won't generate actual images).


abea:
@boutell This is simply for https://apostrophecms.atlassian.net/browse/A30U-173, "As a user I can see thumbnails of images while browsing the image library in the main pain" (in the description).

PR #2352, opened 2020-08-31
Fixes Both filter and select input

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-372 with the recent published filter update.

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
@@ -194,7 +194,7 @@ export default {
   },
   created() {
     this.options.filters.forEach(filter => {
-      this.filterValues[filter.name] = filter.choices[0].value;
+      this.filterValues[filter.name] = filter.def || filter.choices[0].value;
     });
   },
   async mounted() {
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
@@ -43,12 +43,13 @@ export default {
     };
   },
   mounted() {
-    if (!this.field.choices.find(choice => {
+    // Add an empty option if there isn't one already or an "any" option.
+    if (!this.field.required && !this.field.choices.find(choice => {
       return !choice.value || choice.value === 'any';
     })) {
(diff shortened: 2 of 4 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
```diff
@@ -37,13 +37,29 @@ export default {
...
+    }
+    this.choices = this.choices.concat(this.field.choices);
+  },
   methods: {
     validate(value) {
       if (this.field.required && !value.length) {
         return 'required';
       }
 
-      if (!this.field.choices.includes(value)) {
```
abea, line 46: This was erroring on every value since it wasn't looking at the `value` property.

modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
```diff
@@ -37,13 +37,29 @@ export default {
...
+  data() {
+    return {
+      choices: []
+    };
+  },
+  mounted() {
+    if (!this.field.choices.find(choice => {
+      return !choice.value || choice.value === 'any';
+    })) {
+      this.choices.push({
```
abea, line 49: Should every select field have this? Maybe it should be `null`? This is to allow people to have no response or clear the field.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -194,7 +194,7 @@ export default {
   },
   created() {
     this.options.filters.forEach(filter => {
-      this.filterValues[filter.name] = filter.choices[0].value;
+      this.filterValues[filter.name] = filter.def || filter.choices[0].value;
```
abea, line 197: We weren't using the `def` value in the filters.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -125,8 +125,14 @@ export default {
           this.next.hasErrors = true;
         }
 
-        this.next.data[field.name] = this.fieldState[field.name].data ||
```
abea, line 128: This was preventing us from emptying schema fields.

falkodev (approved):

PR #2358, opened 2020-09-01
3/widget controls

stuartromanek:
Step towards implementing new area/widget UX

https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?selectedIssue=A30U-387

This PR
- implements modify and move controls in separate components
- adds a wrapper button group for creating a2-style toolbars with `<AposButton />` components
- various style accommodations


stuartromanek:
implemented a11y roles

modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
```diff
@@ -71,6 +72,13 @@ export default {
     };
   },
   computed: {
+    role () {
+      if (this.type === 'group') {
+        return 'menuitem'
```
abea, line 77: I think it would be better to put the role attribute on the `<AposButton>` component than to tie it to a prop that's generally used for styling.

PR #2359, opened 2020-09-01
In the presence of middleware, we should still call the wrapper function for this type of route properly

boutell:
I was calling it without the method name, resulting in the error Bea saw.

I can now get to a polite 404 when I send a bogus file upload attempt, due to the nonexistence of req.files, which makes sense (throwing an "invalid" error could be more appropriate perhaps).

Also fixed a tiny bug in the `finally` cleanup code.


abea:
I'll merge this into a local branch I have going.

PR #2361, opened 2020-09-02
3/rescue

abea:
Confirms and fixes some bugs for https://apostrophecms.atlassian.net/browse/A30U-124, "As a user I can rescue a piece from trash." Most of the work for this was in #2354

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
@@ -80,7 +80,6 @@
       @close="close"
       @insert="insert"
       :options="addWidgetOptions"
-      :doc-id="docId"
     />
   </div>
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -106,10 +105,6 @@ export default {
     widgetOptions: {
       type: Object,
       required: true
-    },
-    docId: {
```
abea, line 110: This stuff was unused and was causing an error in the doc editor since there was no doc _id yet.

falkodev (approved):

falkodev:
@abea I understand it is just a status change for the "trash" property. Will there be another issue for the pieces manager button "delete" to trash/rescue a piece?

PR #2363, opened 2020-09-02
3/media upload

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-407.

This is for the basic ability to upload and image file to create an image piece.

Also resolves https://apostrophecms.atlassian.net/browse/A30U-388 for limiting to the acceptable file types.


modules/@apostrophecms/attachment/index.js
```diff
@@ -294,7 +296,10 @@ module.exports = {
           }
         }
         info.used = true;
-        await self.db.updateOne({ _id: info._id }, info);
+
+        await self.db.updateOne({ _id: info._id }, {
```
boutell, line 300: Oops, my bad, I forgot that `replaceOne` is a better replacement for updateOne here. It does *not* require `$set`, which is good because `$set` would not remove properties that should no longer exist.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -95,6 +105,72 @@ export default {
...
+    async uploadMedia (event) {
+      const file = event.target.files[0];
+
+      const emptyDoc = await apos.http.post(this.moduleOptions.action, {
+        body: {
+          _newInstance: true
+        }
+      });
+
+      // While the upload is working, set an uploading animation.
```
boutell, line 131: Does this comment say what this is going to do, or is it a TODO?

boutell: Actually I think you can remove this comment because the two `busy: true` flags should suffice.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -95,6 +105,72 @@ export default {
...
+        return;
+      }
+
+      const imageData = Object.assign(emptyDoc, {
+        title: attachment.title,
+        attachment
+      });
+
+      try {
+        await apos.http.post(this.moduleOptions.action, {
```
boutell, line 164: Doesn't look like you did that "open in right rail" thing, do we have a ticket for it to come next?

boutell (changes requested):
Looks great, minor stuff mostly having to do with migrating things from comments to tickets.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -95,6 +105,72 @@ export default {
...
       }
     }
+  },
+  mounted() {
+    // Get the acceptable file types, if set.
+    const imageGroup = apos.modules['@apostrophecms/attachment'].fileGroups
+      .find(group => group.name === 'images');
+
+    if (imageGroup && this.$refs['apos-upload-input']) {
+      const acceptTypes = imageGroup.extensions.map(type => `.${type}`)
```
abea, line 115: This is the bit for https://apostrophecms.atlassian.net/browse/A30U-388

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
abea: That is more of a TODO. I can make that clearer. There will be visual animations.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
abea: Also a TODO

abea:
Please see the updated description with another ticket connected to this (accepted types).

boutell:
Would you mind changing `updateOne` to `replaceOne` so we can lose `$set` and the risk of failing to delete a removed property when we should?

abea:
That should be all set.

boutell (approved):

falkodev (approved):

PR #2364, opened 2020-09-02
Updates image getAll to include img paths

abea:
Fixes the media manager to get actual image paths to display images.

### DIFF
--- a/modules/@apostrophecms/image/index.js
+++ b/modules/@apostrophecms/image/index.js
@@ -78,6 +78,24 @@ module.exports = {
       }
     }
   },
+  extendRestApiRoutes: (self, options) => ({
+    async getAll (_super, req) {
+      const pieces = await _super(req);
+
+      console.info(pieces.results.map(piece => ({
+        title: piece.title,
+        _id: piece._id
+      })));
+
+      for (const piece of pieces.results) {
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/image/index.js
```diff
@@ -78,6 +78,24 @@ module.exports = {
...
+    async getAll (_super, req) {
+      const pieces = await _super(req);
+
+      console.info(pieces.results.map(piece => ({
+        title: piece.title,
+        _id: piece._id
+      })));
+
+      for (const piece of pieces.results) {
+        piece._path = await self.apos.attachment.url(piece.attachment, {
```
boutell, line 91: Yes we need this, but see `apostrophe-headless` for the way to do it to match our current practice for including it in RESTful responses. We use the `annotate: true` feature of `apos.attachment.all`.

```
            // Attach `_url` and `_urls` properties
            self.apos.attachment.all(pieces, { annotate: true });
```

boutell (changes requested):

boutell (approved):
Cool! I think we should probably do this in `getAll` by default for both pieces and pages, for parity with apostrophe-headless, but that's another ticket.

stuartromanek (approved):
lgtm

PR #2367, opened 2020-09-03
Don't try to fetch notifications when there is no user; don't fetch notifications redundantly

boutell:
Also cc @falkodev, it's not necessary that he be one of the two reviewers but he'll be interested in the fix:

1. The notifications module was trying to fetch notifications because it had no way of knowing no user was logged in. You can now look in `apos.login.user` for a minimal amount of info about the current user. If that object does not exist, there is no logged-in user. (See `apos.user` in 2.x browserland.)

2. The notifications module was displaying duplicates because of a dumb mistake of mine in `apos.http.remote`. This is fixed, `options.qs` now works like it is supposed to.

NOTE: I am not sure exactly how this will play with old notifications that predate Anthony's work. So if you still get redundant notifications, try `db.aposNotifications.remove({})` once before opening a ticket. A migration would be bloat since nobody but us will have that early version of the code.

### DIFF
--- a/modules/@apostrophecms/login/index.js
+++ b/modules/@apostrophecms/login/index.js
@@ -350,7 +350,15 @@ module.exports = {
 
       getBrowserData(req) {
         return {
-          action: self.action
+          action: self.action,
+          ...(req.user ? {
+            user: {
+              title: req.user.title,
+              _id: req.user._id,
+              username: req.user.username,
+              email: req.user.email
+            }
+          } : {})
         };
       }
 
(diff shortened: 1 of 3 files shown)


abea (approved):

stuartromanek:
@boutell does this mean we cannot use the notifications module to send anonymous users messages? 

I think this worked in 2.0 but don't know how intentional that was

boutell:
We could allow it, but it would only work when `scene` is `user` or someone builds a different frontend to it, so I don't know how useful it would be to write additional plumbing to support it. So for now I suggest the answer is "notifications are only for actual users," with no bias against eventually supporting it for anons if we need to.

stuartromanek (approved):

PR #2368, opened 2020-09-03
V3/observe upload progress

boutell:
Filling in the gap in `apos.http.remote` so it becomes possible to code up progress meters for uploads at some point.

I tested successfully by adding two lines of code here in the media manager:

```
        attachment = await apos.http.post('/api/v1/@apostrophecms/attachment/upload', {
          busy: true,
          body: formData,
          downloadProgress(received, total) { console.log(`< ${received} ${total}`); },
          uploadProgress(sent, total) { console.log(`> ${sent} ${total}`); }
        });
```

### DIFF
--- a/modules/@apostrophecms/util/ui/public/http.js
+++ b/modules/@apostrophecms/util/ui/public/http.js
@@ -12,7 +12,7 @@
   // `options.body`. See `apos.http.remote` for details.
   // You do NOT have to pass a callback unless you must support IE11
   // and do not want to include a promise polyfill in your build.
- 
+
   apos.http.post = function(url, options, callback) {
     return apos.http.remote('POST', url, options, callback);
   };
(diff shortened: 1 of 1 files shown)


boutell:
The browser does not guarantee these functions will be called, or called at 0%, or called at 100%, etc. depending on the timing and the size of the upload or download in question.

Effort could be put towards guaranteeing those things, but I'm guessing it is not a big deal for the person coding the progress display to just pass 0% before starting and 100% at the end.

stuartromanek (approved):

abea (commented):
If multiple are going at once do we have a way to tell them apart with this? I can't tell if so.

boutell:
Yes, it's a callback not an event on the bus, you can put it right in your function that makes the apos.http.* call so it can see that scope and no other.

boutell:
(i.e. use a closure.)

abea (approved):

PR #2370, opened 2020-09-04
3/bar fixes

abea:
Resolves bug notes from https://apostrophecms.atlassian.net/browse/A30U-379:

- The height of the buttons in Figma is 52px, with the border being centered (for a total of 54 visual px)
  - **Note**: In Figma the whole set of admin rows are 102px. I got to 50px height by subtracting the border heights.
- The right borders are missing on both buttons

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -41,7 +41,7 @@
         :user="user" :avatar-url="userAvatar"
       />
     </div>
-    <div class="apos-admin-bar__row">
+    <div class="apos-admin-bar__row apos-admin-bar__row--utils">
       <AposButton
         type="default" label="Page Settings"
         icon="cog-icon" class="apos-admin-bar__btn"
@@ -117,6 +117,7 @@ $menu-item-height: $menu-row-height - (2 * $menu-v-pad);
 $menu-h-space: 16px;
 $menu-v-space: 25px;
 $admin-bar-h-pad: 20px;
+$admin-bar-border: 1px solid var(--a-base-9);
 
 body {
   margin-top: $menu-row-height * 2;
@@ -136,7 +137,7 @@ body {
   align-items: center;
   height: $menu-row-height;
   padding: 0 $admin-bar-h-pad 0 0;
-  border-bottom: 1px solid var(--a-base-9);
+  border-bottom: $admin-bar-border;
 }
 
 .apos-admin-bar__items {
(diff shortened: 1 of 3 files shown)


gregvanbrug (approved):

abea:
@gregvanbrug FYI - one more fix thrown in there if you want to take a look.

gregvanbrug (approved):

abea:
Okay, that should be it.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -111,18 +121,15 @@ export default {
...
 <style lang="scss" scoped>
-$menu-row-height: 68px;
+$menu-row-height: 50px;
 $menu-v-pad: 18px;
-$menu-item-height: $menu-row-height - (2 * $menu-v-pad);
 $menu-h-space: 16px;
 $menu-v-space: 25px;
 $admin-bar-h-pad: 20px;
-
-body {
```
abea, line 121: This does nothing since the styles are scoped.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -84,7 +84,17 @@ export default {
     }
   },
   mounted() {
-    this.menuItems = [ ...this.items ];
+    this.menuItems = [
+      ...this.items.map(item => {
+        if (item.items) {
+          item.items.forEach(subitem => {
+            // The context menu needs an `action` property to emit.
```
abea, line 91: ☝️

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -84,7 +84,17 @@ export default {
     }
   },
   mounted() {
-    this.menuItems = [ ...this.items ];
+    this.menuItems = [
+      ...this.items.map(item => {
```
gregvanbrug, line 88: Not totally sure the spread is necessary. A map returns a new array so I think you could do something like:
```
this.menuItems = this.items.map(item => { ... });
```
and get the same result.

gregvanbrug (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
abea: true.

gregvanbrug (approved):

stuartromanek (changes requested):
image

paddings are still off here, 20px and 16px, respectively

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -184,8 +189,6 @@ body {
 }
 
 .apos-admin-bar__sub /deep/ .apos-context-menu__btn {
-  padding-left: 20px;
```
abea, line 187: Aligning the second row button set me off looking at this 20px value as well. These are overridden and not actually used, it turns out.

stuartromanek (approved):

PR #2371, opened 2020-09-04
fixed duplicates, i18n, string substitution, replaced apply with ES2020 syntax

boutell:
The string substitution thing doesn't even work in 2.x I think? 😆 Well, it works properly now in 3.x...

### DIFF
--- a/modules/@apostrophecms/notification/index.js
+++ b/modules/@apostrophecms/notification/index.js
@@ -16,6 +16,8 @@
 // Until it times out the request will keep making MongoDB queries to
 // see if any new notifications are available (long polling).
 
+const Promise = require('bluebird');
+
 module.exports = {
   options: {
     alias: 'notification'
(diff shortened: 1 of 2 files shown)


boutell:
I also dropped the `id` option because it only existed in frontend code for reasons that don't apply to our 3.x notifications technical design.

falkodev (approved):

modules/@apostrophecms/notification/index.js
```diff
@@ -16,6 +16,8 @@
 // Until it times out the request will keep making MongoDB queries to
 // see if any new notifications are available (long polling).
 
+const Promise = require('bluebird');
```
abea, line 19: Maybe we should only bring in the `delay` method.

abea (changes requested):
Little bit. Otherwise good. Holler and I'll give a quick re-review.

abea (approved):

PR #2373, opened 2020-09-07
3/join browse button

falkodev:
# Browse button

## How to test

- in A3 demo, `npm run dev`
- have existing products, create or edit an article
- go to the `ungrouped` tab in the article, and click on "Browse"
- add, remove, reorder joined products
- check if `max` limit is respected
- check if existing joins are updated and saved correctly


boutell (approved):
Excellent! I tried hard to break this and I could not (:

I have opened a new ticket for editing the custom fields of the relationship, but what you have done here is very compatible with that.

stuartromanek:
Should this baked directly into `AposPiecesManager`? .. That means we're going to have to similarly bake it into the media manager.

Should chooser be its own modal subclass that imports either AposPiecesManager OR the media manager into it's modal body? If there were to be a third chooser interface a pattern like that would better accommodate it.

falkodev:
> Should this baked directly into `AposPiecesManager`? .. That means we're going to have to similarly bake it into the media manager.
> 
> Should chooser be its own modal subclass that imports either AposPiecesManager OR the media manager into it's modal body? If there were to be a third chooser interface a pattern like that would better accommodate it.

@stuartromanek `AposPiecesManager` is finally used here. But I wonder how to display limits (`min`, `max`, `Limit reached` message) here. In case you have a suggestion to respect what we did for the splat list.

stuartromanek:
closing the loop on https://github.com/apostrophecms/apostrophe/pull/2373#issuecomment-688452699 In our standup we agreed to make a follow up ticket to normalize the chooser pattern while integrating it with the media manager cc @abea @falkodev

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -142,7 +164,9 @@ export default {
       editing: false,
       editingDocId: '',
       queryExtras: {},
-      holdQueries: false
+      holdQueries: false,
+      selectedItems: this.items,
+      checked: this.items.map(item => item._id)
```
abea, line 169: Please add a comment `// NOTE:` here for the refactor work: "`checked` is originally set in `AposTableMixin.js`"

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -203,6 +227,11 @@ export default {
     this.modal.active = true;
     this.getPieces();
   },
+  watch: {
+    checked: function() {
+      this.generateUi();
```
abea, line 232: Is this needed to disable them when the max is reached? The table mixin is also doing this when watching `rows`.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -124,6 +132,20 @@ export default {
     moduleName: {
       type: String,
       required: true
+    },
+    join: {
+      type: Boolean,
+      default: false
+    },
+    items: {
```
abea, line 140: This needs to be named more specifically. We already have a `pieces` property that's populated in `getPieces`. If this property is specific to the relationship context then let's be clear about that (especially to help with refactoring).

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
abea: Please add a `// NOTE:` comment to revisit this in the refactor.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -280,6 +309,20 @@ export default {
       this.currentPage = 1;
 
       this.getPieces();
+    },
+    updateSlatList(items) {
```
abea, line 313: This isn't necessarily updating the slat list, as the name suggests. It's emitting an updated `items` array.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
```diff
@@ -3,6 +3,7 @@
     <template #leftControls>
       <AposButton
         label="Select" :icon-only="true"
+        :disabled="disableSelection"
```
abea, line 6: It doesn't look like you used this in your implementation. Same for the prop below.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
```diff
@@ -146,8 +151,11 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+  .apos-pieces-manager-toolbar {
+    max-height: 37px;
```
abea, line 155: The problem is this:
```
.apos-schema[data-v-4d5681c9] .apos-field {
    margin-bottom: 30px;
    letter-spacing: 0.5px;
}
```

Because the manager is opening inside the schema component of a piece this is getting applied to the search field. Let's disable that bottom margin when inside the toolbar instead.

abea (changes requested):

boutell:
@falkodev the renaming of joins to relationships has landed, so this will need a manual merge with the upstream work in 3.0.

modules/@apostrophecms/modal/ui/apos/mixins/AposTableMixin.js
```diff
@@ -113,7 +113,8 @@ export default {
             name: row._id,
             type: 'checkbox',
             hideLabel: true,
-            label: `Toggle selection of ${row.title}`
+            label: `Toggle selection of ${row.title}`,
+            disabled: this.checked.length >= this.maxItems && !this.checked.includes(row._id)
```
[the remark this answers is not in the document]
falkodev: You're right. Done

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -280,6 +309,20 @@ export default {
...
+      this.selectedItems = items;
+      this.checked = items.map(item => item._id);
+      this.$emit('updated', items);
+    },
+    updateSelectedItems(event) {
```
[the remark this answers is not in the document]
falkodev: `checked` only contains IDs of checked items. I needed to pass the checked items to the `AposInputRelationship`. As for the naming, if this is used for another type in the future, the naming would be too specific.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
falkodev: Done

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
falkodev: `rows` are not updated after every click to check/uncheck items. `checked` is. Done for the note.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
falkodev: `pieces` are only the 10 pieces displayed in the current "page" of pieces in the modal. `items` is a prop, becoming the data `selectedItems` (`selectedItems` are the ones used throughout the component) and holds the list of selected pieces in all "pages" of pieces. I don't think it will be only useful in a `relationship` context (although it is at the moment). Do you prefer `initiallySelectedItems`?

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
falkodev: You're right. In fact, it should be `updateSelectedItems`. And the current `updateSelectedItems` should be `updatedCheckbox` and should call the new `updateSelectedItems` to update `this.selectedItems` and emit the event. What do you think?

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
falkodev: Right. A mistake when I merged my work for `AposJoinsManager` and `AposPiecesManager`. Fixed and improved.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
falkodev: I don't see the relationship between the field's height and the button. But it works. At least when you disable it through the browser. But you have the information you are in a toolbar much deeper than the field. 
I tried to remove the "scope" but no success. I will need a CSS expert on that one :grimacing:

abea:
It looks like there's a small eslint error.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
abea: Yes, `initiallySelectedItems` sounds good, thanks. That will help if it needs to be part of the refactor.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
abea: I pushed an update to this branch to fix this. You'll need to `git pull` before doing additional work on it.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
abea: This is getting confusing. We definitely need the refactor to keep things separate. I'll do that as my next project.

For now, please add a note before both `updateSelectedItems` and `updateSlatList` to move them into the new `AposRelationshipManager` in the refactor.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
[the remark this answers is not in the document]
abea: We can temporarily resolve this by adding the comment I mentioned on the other method.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
abea: 👍

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
```diff
@@ -97,7 +102,8 @@ export default {
         },
         status: {},
         value: { data: '' }
-      }
+      },
+      disable: this.disableSelection
```
abea, line 106: I don't think this one is used now.

abea (changes requested):

abea:
@falkodev You'll need to pull before doing anything. I updated the CSS issue: https://github.com/apostrophecms/apostrophe/pull/2373#discussion_r485731299

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
falkodev: Done

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
falkodev: Done

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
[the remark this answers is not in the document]
falkodev: Done

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
falkodev: :facepalm:

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerToolbar.vue
falkodev: Noice

modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
```diff
@@ -156,4 +188,16 @@ export default {
...
+
+  .apos-modal__rail .apos-slat-list {
+    padding: 16px;
+  }
+
+  .apos-modal__rail .apos-slat-list /deep/ .apos-slat {
+    margin-bottom: 8px;
+  }
+
+  .apos-slat-status {
```
falkodev, line 200: Not sure about the styling here

abea (approved):
Thank you for dealing with all my comments!

falkodev:
> Thank you for dealing with all my comments!

@abea Thank you for your thorough review and for the CSS rule.

PR #2375, opened 2020-09-08
Email field

falkodev:
# Email field

## Content

- Backend validation
- Frontend style

## How to test

- In a3-demo, add this to `product/index.js`:

```js
// in `fields/add object`
    email: {
        type: 'email'
      }
```

- `npm run dev`
- go to a product or create one, and try to add a wrong email address, and then a valid one

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -468,6 +468,27 @@ module.exports = {
     });
 
     self.addFieldType({
+      name: 'email',
+      vueComponent: 'AposInputString',
+      convert: function (req, field, data, object) {
+        object[field.name] = self.apos.launder.string(data[field.name], undefined, field.min, field.max);
+        if (!data[field.name].length) {
+          if (field.required) {
+            throw self.apos.error('required');
+          }
+        } else {
+          // regex source: https://emailregex.com/
+          // eslint-disable-next-line no-useless-escape
+          const matches = data[field.name].match(/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/);
+          if (!matches) {
+            throw self.apos.error('invalid');
+          }
+        }
+        object[field.name] = data[field.name];
+      }
+    });
+
+    self.addFieldType({
       name: 'color',
       convert: async function (req, field, data, object) {
         const test = self.apos.launder.string(data[field.name]);
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -478,8 +478,7 @@ module.exports = {
           }
         } else {
           // regex source: https://emailregex.com/
-          // eslint-disable-next-line no-useless-escape
-          const matches = data[field.name].match(/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/);
+          const matches = data[field.name].match(/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/);
           if (!matches) {
             throw self.apos.error('invalid');
           }
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
@@ -97,6 +97,7 @@ export default {
         }
       }
       if (this.field.type === 'email') {
+        // regex source: https://emailregex.com/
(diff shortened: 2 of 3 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -467,6 +467,27 @@ module.exports = {
...
+      vueComponent: 'AposInputString',
+      convert: function (req, field, data, object) {
+        object[field.name] = self.apos.launder.string(data[field.name], undefined, field.min, field.max);
+        if (!data[field.name].length) {
+          if (field.required) {
+            throw self.apos.error('required');
+          }
+        } else {
+          // regex source: https://emailregex.com/
+          // eslint-disable-next-line no-useless-escape
```
abea, line 481: This "useless-escapes" should be fixable rather than disabling the linter.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -96,6 +96,12 @@ export default {
           return 'max';
         }
       }
+      if (this.field.type === 'email') {
+        const matches = value.match(/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/);
```
abea, line 100: Let's note where the regex came from here as well.

modules/@apostrophecms/schema/index.js
abea, line 481: This "useless-escapes" should be fixable rather than disabling the linter.

abea (changes requested):

modules/@apostrophecms/schema/index.js
```diff
@@ -467,6 +467,27 @@ module.exports = {
       }
     });
 
+    self.addFieldType({
+      name: 'email',
+      vueComponent: 'AposInputString',
+      convert: function (req, field, data, object) {
+        object[field.name] = self.apos.launder.string(data[field.name], undefined, field.min, field.max);
```
stuartromanek, line 474: what do `min` and `max` do for an email field? Should these be shut out to the developer?

abea: `min` could have a purpose, but not a big one. Good point. They aren't hurting, but aren't doing a lot either. If we remove them we just need to make a note to include that in the docs. There's a space in our main 3.0 how-to Notion doc where that could go.

falkodev: I used the code from 2.x that used `min`  and `max`.

modules/@apostrophecms/schema/index.js
falkodev: I used the code from 2.x, event. I remember a conversation with Tom about it when I tried to adapt the regex to the linter. We agreed to keep the ignore rule instead of modifying the regex.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
falkodev: Done

modules/@apostrophecms/schema/index.js
boutell: As I recall this lint just plain wasn't right, the regexp did not cooperate.

boutell: (When modified to please lint.)

modules/@apostrophecms/schema/index.js
boutell: I see a value for `max`. It's not unreasonable to side-eye an 1800 character email address.

stuartromanek: `max` could be reasonable i guess, but `min` is really dictated by the possible domain/tld combinations

modules/@apostrophecms/schema/index.js
abea: I don't think it's wrong. It's reacting to `[^<>()\[\]\\.,;:\s@"]`, which is excluding all the characters in that set. The closing bracket in the middle (`/]`) needs to be escaped to not end the set, but the opening bracket is inside the set (`[]`) so it's interpreted as the actual bracket character.

Changing it to `/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/` resolves the linter error and does the same thing.

modules/@apostrophecms/schema/index.js
abea: It seems like this is a case where the options don't hurt. At worst, they are not useful very often, but might occasionally be appreciated. I vote to leave them until we have a good case for removing.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
abea: This isn't checked by eslint for some reason, but could be updated to match the passing pattern from the other instance. 

`/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/`

abea (changes requested):

modules/@apostrophecms/schema/index.js
falkodev: Done on front and back. Seems to work as expected during my tests.

boutell (approved):

modules/@apostrophecms/schema/index.js
stuartromanek: ¯\_(ツ)_/¯

abea (approved):

PR #2376, opened 2020-09-08
Join story, styles and UX

stuartromanek:
A take-two on #2343, this adds a storybook story for Joins, re works some style cruft and a little bit of UX .. works in both SB and apos 3 demo


modules/@apostrophecms/schema/ui/apos/components/AposInputJoin.vue
```diff
@@ -69,6 +85,12 @@ export default {
         return { message: 'required' };
       }
 
+      // if the original status was disabled, no validation should change that
+      if (this.originalDisabled) {
```
abea, line 89: is this really an issue for all fields?

modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -72,20 +68,21 @@ export default {
 .item {
   display: flex;
   justify-content: space-between;
-  padding: 15px 25px;
+  margin: 10px;
+  padding: 10px 20px;
   letter-spacing: 0.5px;
-
+  border-radius: 5px;
```
abea, line 74: We have this stored as `--a-border-radius`.

abea (changes requested):
A question and a minor thing.

modules/@apostrophecms/schema/ui/apos/components/AposInputJoin.vue
stuartromanek: I don't think so .. joins can be disabled by reaching their `max` limit, but then can be re-enabled by removing an item from the slat list. This two-ways-to-affect-state is unique to things that use the slat list.

A normal input is just going to be disabled from the get go and won't get re validated

boutell:
@falkodev the renaming of joins to relationships has landed, so this will need a manual merge with the upstream work in 3.0.

boutell:
@stuartromanek  the renaming of joins to relationships has landed, so this will need a manual merge with the upstream work in 3.0.

stuartromanek:
updated to new Relationship name

abea (approved):

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.stories.js
```diff
@@ -0,0 +1,89 @@
...
+        field: {
+          name: 'myJoin',
+          label: 'Link to A Page',
+          help: 'Choose a page from the joiner to link to it',
+          type: 'join',
+          withType: 'product',
+          max
+        },
+        value: {
+          data: getData(prepoulate)
```
boutell, line 51: prepopulate, not prepoulate

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.stories.js
```diff
@@ -0,0 +1,89 @@
...
+    'Limit', {
+      None: null,
+      One: 1,
+      Two: 2,
+      Three: 3
+    },
+    null
+  );
+
+  const prepoulate = select(
```
boutell, line 26: typo

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -44,17 +47,30 @@ export default {
...
   },
   mounted() {
     this.validateAndEmit();
   },
+  computed: {
+    pluralLabel() {
+      return apos.modules[this.field.withType].pluralLabel;
+    },
+    placeholder() {
+      return this.field.placeholder || `Search ${this.pluralLabel}`;
```
boutell, line 68: I think that if we're going to do this "auto-guess the verbiage" thing we should push it as config from the server side where i18n is easy. I would be open with that being a followup ticket though as long as it gets opened.

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -44,17 +47,30 @@ export default {
...
+    },
+    browseLabel() {
+      return `Browse ${this.pluralLabel}`;
```
boutell, line 71: Same comment about the origin of browser-side text and i18n.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
stuartromanek: opened an issue https://apostrophecms.atlassian.net/browse/A30U-436

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.stories.js
stuartromanek: fixec

boutell (approved):

PR #2378, opened 2020-09-09
Throw errors in await compatible way

boutell:
Old school callbacks can happily take multiple arguments but when promisifying them, you can only give (err, value), and value is ignored if there is an error. Send a single error argument with all interesting info about the error.


abea (approved):

abea:
I'll merge this into a feature branch that needs it and and merge them together.

PR #2379, opened 2020-09-10
Fields: integer and float

falkodev:


modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -67,6 +79,11 @@ export default {
       }
     }
   },
+  watch: {
```
abea, line 82: If the `type` property is not going to change while the component is in place, I don't think we need to watch it. `defineStep` could run on `mounted` and look at the `field.type` directly.

abea: Oh, I see you are doing that already. Are we expected it would also change? It doesn't seem like we need to watch this.

abea (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
falkodev: You are right. No need to watch.

abea (approved):

boutell (approved):

boutell (changes requested):
What about `min` and `max`? Their behavior must change when the type is numeric.

falkodev:
> What about `min` and `max`? Their behavior must change when the type is numeric.

Done

boutell (approved):

PR #2384, opened 2020-09-10
3/upload noties

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-393

This delivers notifications to the end user for uploading images:
- starting an upload
- successful upload
- error from uploading an unacceptable file type (one apostrophe doesn't know, e.g., json)
- error from uploading an incorrect file type for that field (e.g., pdf)
- general error messages if there's not something more specific.


modules/@apostrophecms/attachment/index.js
```diff
@@ -285,9 +285,18 @@ module.exports = {
           return;
         }
         _.assign(info, _.omit(dbInfo, 'crop'));
-        if (!self.acceptableExtension(field, info)) {
-          throw self.apos.error('invalid');
+
+        // Check if the file type is acceptable of if there are
+        const correctedExtensions = self.checkExtension(field, info);
```
abea, line 290: I wanted to get the list of acceptable extensions from this, so I changed how this works.

modules/@apostrophecms/attachment/index.js
```diff
@@ -308,29 +317,34 @@ module.exports = {
...
+      checkExtension(field, attachment) {
+        const groups = field.fileGroups ||
+          (field.fileGroup && [ field.fileGroup ]);
         let extensions;
+
         if (groups) {
           if (!_.includes(groups, attachment.group)) {
             extensions = [];
             _.each(groups, function (group) {
-              const groupInfo = _.find(self.options.fileGroups, { name: group });
```
abea, line 318: Fix here. Please confirm.

modules/@apostrophecms/attachment/index.js
```diff
@@ -308,29 +317,34 @@ module.exports = {
...
               extensions = extensions.concat(groupInfo.extensions);
             });
-            return false;
+
+            return extensions;
           }
         }
-        extensions = field.extensions || (field.extension && [ field.extension ]);
+        extensions = field.extensions ||
+          (attachment.extension && [ attachment.extension ]);
```
abea, line 340: Fix here. Please confirm.

modules/@apostrophecms/schema/index.js
```diff
@@ -1427,6 +1425,9 @@ module.exports = {
             try {
               await convert(req, field, data, object);
             } catch (e) {
+              if (e instanceof Error) {
```
abea, line 1428: If we get an `Error` send it on.

modules/@apostrophecms/attachment/index.js
boutell: 👍

modules/@apostrophecms/attachment/index.js
boutell: attachment.extension has no business on a list of what's allowed; it's what was actually sent.

The idea here was to support attachment fields specifying a single extension to allow or more than one, as an alternative to specifying a group. The existing code seems correct for that.

boutell: A method that returns either `true` or an array of things that would have been OK is unusual, so please document it.

modules/@apostrophecms/attachment/index.js
```diff
@@ -947,7 +964,10 @@ module.exports = {
       canUpload(req, res, next) {
         if (!self.apos.permission.can(req, 'edit-attachment')) {
           res.statusCode = 403;
-          return res.send('forbidden');
+          return res.send({
+            type: 'forbidden',
+            message: req.__('You do not have enough permissions to upload')
```
boutell, line 969: You do not have permission to upload a file.

boutell: "Enough permissions" seems needlessly technical.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -132,13 +132,21 @@ export default {
...
       const file = event.target.files[0];
 
       const emptyDoc = await apos.http.post(this.moduleOptions.action, {
         body: {
           _newInstance: true
         }
       });
+      await apos.notify(
+        // TODO: i18n
+        `Uploading ${fileCount} image${fileCount > 1 ? 's' : ''}`,
```
boutell, line 145: I was about to tell you how to do this with %s placeholders and more arguments to apos.notify, but it involves pluralization and we don't actually have a wrapper for that yet. Hmm. OK for now.

modules/@apostrophecms/notification/ui/apos/apps/ApostropheNotification.js
```diff
@@ -121,6 +122,7 @@ export default function() {
           :key="notification._id"
           :label="notification.message"
           :type="notification.type"
+          :icon="notification.icon"
```
boutell, line 125: Do we need icon in addition to type? Should we just map types to icons or something? Is this part of Stu's design?

modules/@apostrophecms/schema/index.js
boutell: This would send on pretty much everything, because self.apos.error does construct an Error.

boutell: Not sure what you're trying to do.

boutell (changes requested):

modules/@apostrophecms/attachment/index.js
abea: True. For some reason that seemed necessary, but the `self.options.fileGroups` fix was probably the important one.

modules/@apostrophecms/notification/ui/apos/apps/ApostropheNotification.js
abea: they are passed in separately right now. The current implementation increases flexibility, but I'd be happy for it to be more limited to simplify the implementation. Out of scope of this PR, though.

modules/@apostrophecms/schema/index.js
abea: The problem was that an error thrown in the field's `convert` method was buried here, including the helpful message. I couldn't see the `message` property on it by logging, but I see it's there now to check and use.

modules/@apostrophecms/schema/index.js
```diff
@@ -1425,8 +1425,8 @@ module.exports = {
             try {
               await convert(req, field, data, object);
             } catch (e) {
-              if (e instanceof Error) {
-                throw e;
+              if (e instanceof Error && e.message) {
```
boutell, line 1428: This isn't going to work either, but it's my fault. The error module doesn't currently set any property that doesn't exist on a generic Error, or potentially other people's third party subclasses of Error. We need to duck type this.

In the error module, please set `.aposError = true` on errors constructed by the error module. Then you can test for it here successfully. We could also include an `isOurs` method that tests it, for calling `apos.error.isOurs(error)`.

Let me know if you'd rather I did a separate PR for that first but I'm trying to speed you along.

boutell (changes requested):

boutell (approved):

falkodev (approved):

falkodev (approved):

boutell (approved):

PR #2385, opened 2020-09-10
PoC: mechanism for child areas to emit an event to their parent area

boutell:
This contains a PoC demo, so it makes sense to get two approvals, but not to merge it directly. Once approved, Stuart should merge it into his branch in progress that requires the functionality, implement `onChildAreaEvent` for real, and then remove the PoC test button etc.

Stuart is familiar with this, but for Bea's benefit:

## The problem

When editing, Stu needs to emit events from a nested child area to its parent area.

Normal Vue events will not work because an AposAreaEditor does not render the entire DOM of its contents.

To review: the widgets inside get rendered by Apostrophe via AJAX calls, for 100% faithfulness to what you get when logged out. Then additional, totally independent AposAreaEditor instances are created to manage any sub-areas that turn out to be in those widgets. This part isn't new.

So we need an elegant way to bridge the gap.

## The solution

Let's say AposAreaEditor wants to tell its parent that it took the focus.

`AposAreaEditor` may now call `this.emitToParentArea('focus')`.

In the **parent** AposAreaEditor, `onChildAreaEvent` will now be called, with these arguments:

`(childArea, inWidget, 'focus')`

**Implementing that function so it does something for each event name is up to Stuart.** The current implementation just logs to show you it's working, and working only in the immediate parent area, not all the areas in the page.

## Passing extra arguments

You can pass more arguments to `emitToParentArea`. If you do, they will all show up at the end of the arguments list of `onChildAreaEvent`. Since there could be different events with different requirements, I'd probably leave that as `...args` to catch them all in an array.

## Can't we just emit events to the parent somehow as if they were normal events arriving in the parent?

In other words: "why do we need onChildAreaEvent?" I looked but couldn't find a way to get a component to emit an event *to itself*, as opposed to its parent, which is what would have to happen for that to work.

## Can we register nice event handlers by name somehow, rather than if/else if/else in onChildAreaEvent?

We could, but I'm pretty sure the number of events is going to be rather small, and I didn't want to drag along a boatload of code for this feature if I'm overthinking it.

## Implementation notes

The implementation relies on the Apostrophe event bus, then uses `apos.util.closest` to figure out if this area is the right one to receive the event and to identify the widget in which the area was nested.

## How do I test this?

a3-demo includes two-column areas, so add one of those to the home page, then try the "emit demo event" button in the left or right child area. You should get logging for one (1) parent area, plus some messages from areas that saw it wasn't their business and ignored it.

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
@@ -1,5 +1,6 @@
 <template>
-  <div class="apos-area">
+  <div data-apos-area class="apos-area">
+    <button @click="emitToParentArea('test', { data: 5 })">Emit Demo Event</button>
     <AposAreaMenu
       @add="insert"
       :menu="choices"
@@ -43,6 +44,7 @@
           :id="widget._id"
           :area-field-id="fieldId"
           :value="widget"
+          data-apos-widget
           @edit="edit(i)"
         />
         <AposAreaMenu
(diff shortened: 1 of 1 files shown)


abea (approved):

stuartromanek:
implementing some of this over in #2400 but don't need this pr anymore

PR #2386, opened 2020-09-10
Adds development note about scoped styles

abea:
Clarifies the standard rule for scoping component styles.

CC @falkodev @gregvanbrug @breyell


stuartromanek (approved):

boutell (commented):
I think there's more to this that I'm working through here:

https://apostrophecms.atlassian.net/browse/A30U-445

boutell (approved):
Correct for now, may need an update if we come up with a new name for the "just variables and mixins" import file and also have a selection of import files for actual utility classes if the component will truly need them.

PR #2388, opened 2020-09-10
search the dependencies of apostrophe for modules

boutell:


modules/@apostrophecms/storybook/template/.storybook/webpack.js
```diff
@@ -37,6 +37,13 @@ module.exports = async (config) => {
     }
   );
 
+  // Allow for an npm linked apostrophe module to work
+  config.resolve.modules = [
```
abea, line 41: Should this involve some environment variable (like excluding the production env)?

boutell: We want to resolve apostrophe asset build dependencies like apostrophe does, and apostrophe always does this. In production it is unlikely to be necessary, but it *could* come up if the project level needs a different version and npm resolves it, as it will, by putting them in different places.

stuartromanek (approved):

abea (approved):

PR #2389, opened 2020-09-11
3.0 fix duplicate scss

boutell:
I fixed the "bundle bloated by CSS" problem.

The total size of our direct Vue components and styles in the bundle dropped by 57%. That tipped the balance such that our frontend npm dependencies are now the majority of our bundle size. Which is still big. But this is a major improvement.

Here's what I did:

* I migrated all "meta" scss, i.e. mixins, variables, and functions that do *not* generate any rules if not used, into a `/mixins` folder and imported them via `/mixins/import-all.scss`, which I think is clear naming.
* I migrated all non-component-specific scss rules into `/global` and imported them via `/global/import-all.scss`.
* I configured the vue loader to implicitly import `/mixins/import-all.scss`.
* I put an import of `/global/import-all.scss` right smack at the top of the bundle. Nowhere else. We don't want duplicate rules.
* I declared every vue module's styles to be scoped, and then backed off on that only when I saw it created a problem. The only real example I think: `AposButton`, which gets styled by the admin bar in a way that breaks due to CSS specificity if the AposButton component styles are scoped. I wasn't able to figure that one out with `/deep/`, but maybe someone else can, as a followup?

This branch also incorporates my work yesterday to get storybook running again when apostrophe is npm linked, as I need that to test storybook.

Thanks to @localghost443 for spotting the issue in the first place.


modules/@apostrophecms/ui/ui/apos/scss/mixins/_theme_mixins.scss
```diff
@@ -0,0 +1,42 @@
+$font-default: -apple-system, system-ui, 'BlinkMacSystemFont', 'Segoe UI',
```
abea, line 1: It would be clearer to have a separate variables file. We could probably clean up files by moving more variables scattered about into that file and then get more of the disparate actual SCSS mixins together as well.

abea: They could be different files if that makes sense, but imported into a main mixins file to get into the full import.

modules/@apostrophecms/storybook/template/.storybook/webpack.js
```diff
@@ -28,15 +28,22 @@ module.exports = async (config) => {
...
+`@import "Modules/@apostrophecms/ui/scss/mixins/import-all.scss";
+`
           }
         }
       ]
     }
   );
 
+  // Allow for an npm linked apostrophe module to work
+  config.resolve.modules = [
```
abea, line 41: isn't this in another branch and PR?

abea (commented):
One chunk seems duplicated across two PRs. Then there's some SCSS organization I'm suggesting. That could be a new ticket as well.

modules/@apostrophecms/storybook/template/.storybook/webpack.js
boutell: It is, but I mentioned that; it's necessary here too to be able to verify the correctness of this new PR for storybook. I am not sure why it shows up as a change still since this is merged in 3.0.

modules/@apostrophecms/ui/ui/apos/scss/mixins/_theme_mixins.scss
boutell: I think that might require some deliberation among the frontend crew. Can we accept this PR as an iterative improvement and open a new ticket for that?

modules/@apostrophecms/storybook/template/.storybook/webpack.js
boutell: OK I merged the 3.0 branch into this branch and as you can see this is no longer an issue.

modules/@apostrophecms/ui/ui/apos/scss/mixins/_theme_mixins.scss
abea: Yep.

abea (approved):

boutell:
This is paused for a fix to the button scoped issue by @stuartromanek

stuartromanek (approved):
added a deep selector for secondary admin bar buttons

PR #2390, opened 2020-09-14
Ties modal lip width to right rail width

abea:
Fixes the funky right rail lip. Width didn't match the right rail. https://apostrophecms.atlassian.net/browse/A30U-438

💋

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
@@ -383,7 +383,7 @@ export default {
   }
 
   .apos-modal__main--with-rails {
-    grid-template-columns: 18% 1fr minmax(250px, 22%);
+    grid-template-columns: 18% 1fr minmax(250px, $modal-rail-right-w);
   }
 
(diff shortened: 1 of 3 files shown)


gregvanbrug (approved):

modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
```diff
@@ -383,14 +383,14 @@ export default {
   }
 
   .apos-modal__main--with-rails {
-    grid-template-columns: 18% 1fr minmax(250px, 22%);
+    grid-template-columns: 18% 1fr minmax(250px, $modal-rail-right-w);
```
falkodev, line 386: I have this when visualizing an image:
image

The button "published/unpublished" is is wider than the rail (might depend on the window's width). Only the "Save" button is present (and is not doing anything, but that might be done in another PR I guess).

abea: That is true. Yes, this stuff would be separate issues. The save button should work in the 3.0 branch right now.

stuartromanek (approved):

PR #2393, opened 2020-09-15
3/fix finished

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-428 and https://apostrophecms.atlassian.net/browse/A30U-427.

AND it generally resolves the fact that you couldn't save edits on an image before.


abea:
For some reason the _Storybook_ editor doesn't want to populate the mock image data. The real one is fine.

boutell:
This looks good, but why is the attachment removed from the schema, isn't replacing the attachment one of your options?

abea:
That attachment field will eventually be done through another custom interface (an "Edit" button/link, I think). That is a separate user story, though. This was focused on fixing what is there.

boutell:
OK.

boutell (approved):
I assume the separate story for editing attachments that already exist is in Jira. It's not a schema field really until it can be edited.

falkodev (approved):

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -95,6 +95,14 @@ export default {
       handler() {
         this.updateNextAndEmit();
       }
+    },
+    value: {
+      deep: true,
+      handler(newVal, oldVal) {
+        if (newVal.data.slug !== this.fieldState.slug.data) {
```
boutell, line 102: Not every schema has a slug field at all, for example relationships that have schemas. This is hardcoding a special case in a way that will crash if there isn't a slug. I'm also not sure why it's being done.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
boutell: In the original PoC components, slug fields were set up to watch title fields without polluting ApostropheSchema with special-case code. This was done via the `context` prop passed to all fields and implemented by the ApostropheFieldMixin. It looks like this feature got lost in the new Vue components.

boutell: (I'm not sure if this code is related to slugs watching titles, or something else.)

abea: I don't know if the `context` bit is related to this. The situation here is that in the image editor the document in the editor might be completely swapped out for another image document at any time. The schema needs to recognize this and update the field values.

Once the doc is passed into `AposSchema` it is updated to reflect the schema fields. This means that `_id` is removed, among others. So the only unique field I had left was `slug`.

boutell: I see. Wouldn't we always want to refresh the whole thing when any part of `value` is modified, not just slug? That's how the v-model pattern works: watch changes to the `value` prop to accept change from above, emit the `input `event to push change from below.

boutell (approved):
This removes my concern about `slug`, but I think we should be watching for any change at all inside `value` and rebuilding fieldState, in order to fully implement the v-model pattern and its support for two-way binding.

However I fear that could lead to infinite loops and pushing changes "down" is not something we actually need at this moment for this component, so I'm removing my objection; your fix with `_id` covers the territory for now.

abea:
Right. I think that would result in an infinite loop. That's why I wanted a unique field. `slug` actually wasn't great for that anyway since it could be edited. The v-model connection does work here, though.

PR #2396, opened 2020-09-16
Prevents relationship methods when not needed

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-462. The involvement of `selectedItems` when it was not needed in the pieces manager was mutating `checked` when it shouldn't have been. I also removed an argument from `toggleRowCheck` that wasn't actually used.

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
@@ -17,7 +17,10 @@
 
     <template v-if="relationship" #leftRail>
       <AposModalRail>
-        <AposSlatList @update="updateSlatList" :initial-items="selectedItems" :field="field" />
+        <AposSlatList
+          @update="updateSlatList"
+          :initial-items="selectedItems" :field="field"
+        />
       </AposModalRail>
     </template>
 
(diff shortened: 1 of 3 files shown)


boutell (approved):

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -230,12 +239,6 @@ export default {
     this.modal.active = true;
     this.getPieces();
   },
-  watch: {
-    // NOTE: revisit this during refactoring
-    checked: function() {
```
falkodev, line 235: this is needed for the relationship editor

falkodev (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
abea: It's still there. The linter wants these in a particular order, so it's moved up to line 226.

falkodev: Ok

falkodev (approved):

PR #2397, opened 2020-09-16
Fixes the height of the media library drop zone

abea:
https://apostrophecms.atlassian.net/browse/A30U-439


abea:
It should be two image squares high.

gregvanbrug (approved):

stuartromanek (approved):

PR #2400, opened 2020-09-17
3/area hover

stuartromanek:
Implements UX found here https://jsfiddle.net/stuartromanek/rxLaw1ct/18/show

@boutell this editing state flag came over from the original AposAreaEditor but is now in AposAreaWidget with `editing` as a prop.. in vue devtools it seemed like AposAreaEditor was still managing the `editing` flags correctly because of AposAreaWidget emitting editing/saving events .. does AposAreaWidget need to manage this state at all?


abea (changes requested):
A few odd bits so far. Let me know if any of these are not relevant yet.
- the drop down from the big plus button isn't aligned.
Screen Shot 2020-09-18 at 9 52 12 AM

- One of these widgets is in focus, then I hover over the other. Maybe this is alright?
Screen Shot 2020-09-18 at 9 58 34 AM

- The new widget modal I added for the Test 1 widget is far to the right so the save/cancel buttons are hidden.
Screen Shot 2020-09-18 at 9 53 26 AM

modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
```diff
@@ -424,6 +434,10 @@ export default {
   }
 }
 
+.apos-button--icon.apos-button--small {
+  padding: 7.5px 10px;
```
abea, line 438: 7.5? Does the half px really do anything?

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
```diff
@@ -138,7 +145,7 @@ export default {
         this.unbind();
       }
     },
-    buttonClicked() {
+    buttonClicked($event) {
```
abea, line 148: It doesn't look like `$event` is used. Does that mean the addition in `AposButton` is not needed?

modules/@apostrophecms/ui/ui/apos/scss/mixins/_theme_mixins.scss
```diff
@@ -37,6 +37,7 @@ $boolean-padding: $spacing-double;
   // stylelint-disable
   --a-primary-button-hover: #{mix(black, $color, 10%)};
   --a-primary-button-active: #{mix(black, $color, 15%)};
+  --a-primary-child: #{mix(red, $color, 50%)};
```
abea, line 40: I don't understand the semantics of this naming.

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -122,17 +123,25 @@ export default {
...
+        modifiers: [ 'round', 'tiny' ],
+        iconSize: this.empty ? 20 : 11
       }
     };
   },
   computed: {
     moduleOptions() {
       return window.apos.area;
     },
+    contextOpts() {
```
abea, line 135: The naming could get confusing. `extendedContextOptions`?

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -0,0 +1,468 @@
...
+<template>
+  <div class="apos-area-widget-wrapper" :data-area-widget="widgetId">
+    <div
+      class="apos-area-widget-inner"
+      :class="ui.container"
+      @mouseover="widgetMouseover($event)"
+      @mouseleave="widgetMouseleave"
+      @click="widgetFocus($event)"
+    >
+      <div
```
abea, line 11: seems like this should be a heading of some kind.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -0,0 +1,468 @@
...
+      required: true
+    },
+    contextOptions: {
+      type: Object,
+      required: true
+    }
+  },
+  emits: [ 'up', 'down', 'remove', 'edit', 'update', 'insert', 'changed' ],
+  data() {
+    const s = {
```
abea, line 154: Let's use some kind of name on this.

abea (changes requested):
A few more notes.

modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
stuartromanek: don't be mean, we're already using this padding set

modules/@apostrophecms/ui/ui/apos/scss/mixins/_theme_mixins.scss
stuartromanek: changed this to full fledged `--a-secondary` 🤞

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
stuartromanek: changed to the even more verbose `extendedContextMenuOptions`

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
stuartromanek: removed, unneeded

stuartromanek:
@abea responding to your screenshots

1) This should be solved by https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?selectedIssue=A30U-447 🤞 
2) Yeah that is a valid state to be in, pretty chaotic in that screenshot though
3) These widget editors should be using the right side schema rail, I made a ticket https://apostrophecms.atlassian.net/browse/A30U-526

boutell (changes requested):
Um. Well it can't work like this, this code is modifying subproperties of a prop passed to it, which is an antipattern in Vue. You should never modify props, they should be treated as read-only.

I would say that just a boolean `editing` flag for this one widget, *not* the entire `editing` object with information about completely different widgets, should be passed as a prop to `AposAreaWidget`, so it knows when to show and hide things.

And then `AposAreaWidget` should emit events sufficient for `AposAreaEditor` to update `editing[widget._id]`.

Props and their subproperties must never be modified by our Vue components.

stuartromanek:
@boutell i'm now only checking the editing state being passed down from the parent. The widget has several ways to signal we're done editing (save, cancel, 'done') but it never tries to manage that state.

I noticed in the 3.0 branch the rich text editor never gets dismissed.. i don't think i broke that but let me know

boutell (changes requested):
This works, but please make the `editing` prop to AposAreaWidget just a boolean and pass that particular widget's current value to it, so it doesn't have more information than it needs. Thanks. 🙏

boutell (approved):
It's... it's so beautiful 😭 Thank you

abea (approved):

PR #2402, opened 2020-09-17
3/fix relation input

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-395, "As a user, I can tag a single image through the schema," by making the relationship field fully styled.

I commented out some styles that were set too broadly. I didn't fix them fully because the fix will need to come when the pieces manager is refactored to split it apart from the relationships manager.

To test, click to edit an image in the media library. Or look at the relationship input field in the storybook.

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -17,6 +17,7 @@
             tabindex="0"
           >
           <AposButton
+            class="apos-input-relationship__button"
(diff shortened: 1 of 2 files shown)


stuartromanek (approved):

stuartromanek (approved):

falkodev:
Credit URL is oddly small

image

falkodev (commented):
Credit URL

abea:
@falkodev The url field is outside the scope of this ticket. This is only about styling the relationship input.

falkodev (approved):

PR #2411, opened 2020-09-21
3/disable pager

abea:
Disables the pager when there is only one page of results.


abea:
@stuartromanek Fixing the color made me realize that the `AposButton` disable color is different from the one used on the inputs. I think they should both work from a single variable set in the theme file. We'd need to decide between base-4 and base-7.

stuartromanek:
@abea testing in storybook, it looks like we should make the button disabled state follow the input disabled state

stuartromanek (approved):

falkodev (approved):

PR #2415, opened 2020-09-22
Fixes scoped tree styles

abea:


boutell (approved):
LGTM (your relative import paths seem cool to me since you're not crossing a module boundary here)

stuartromanek (approved):

PR #2420, opened 2020-09-23
3/pieces views relations mgr

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-435, "Create AposRelationshipManager modal component to use AposPiecesManagerView and AposMediaManagerView"

It really reflects a different approach altogether, using the manager modals without a built-in slat list.


modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -41,32 +71,46 @@ export default {
...
+      if (!this.checkedDocs) {
+        return;
+      }
+
+      // If in a relationship input context, keep `checkedDocs` in sync with
+      // `checked`, first removing from `checkedDocs` if no longer in `checked`
+      this.checkedDocs.filter(doc => this.checked.includes(doc._id));
+      // then adding to `checkedDocs` if not there yet. These should be in
+      // `items`.
+      // TODO: Once we have the option to select all docs of a type even if not
```
abea, line 85: This is a significant TODO, but I don't think we're getting "select all things of this type in the db" in alpha version. It's one of those situations that will likely never get too crazy, but if someone did want to choose 400 products to relate to another document we would need to set a `busy` state for sure while we got all those docs. That's why I thought it might be better done a level up.

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -41,32 +71,46 @@ export default {
...
+      // database... maybe better to do that in the relationship input?
+      this.checked.forEach(id => {
+        if (this.checkedDocs.findIndex(doc => doc._id === id) === -1) {
+          this.checkedDocs.push(this.items.find(item => item._id === id));
+        }
+      });
     }
   },
   methods: {
-    toggleRowCheck(id) {
```
abea, line 47: This wasn't doing a lot for us.

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
boutell: If this means it doesn't work right if you select two things from one text search or other filtering scenario and then two things from another one, which i.e. are not on the screen at the same time in the list view, then it's not ready. "Select Everything" is not necessary for this scenario to apply, the selection view must be fully independent of what happens to be on the current page or the architecture is still unsettled.

But I'm not sure if that's what you're saying or not.

boutell (changes requested):

boutell (approved):
Thanks.

PR #2423, opened 2020-09-24
Checks for valid widget browser data in area editor

abea:
Ensures that we have browser data for widget modules before using them in the area editor. If you have a widget in your document that is no longer in the application then you won't have the browser data to use it.


stuartromanek (changes requested):
Console warn the offending widget name in addition to the server log

falkodev (approved):

stuartromanek (approved):

PR #2424, opened 2020-09-24
Open newly uploaded image in media editor

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-425, "As a user, when done uploading a single image in the media manager, I can view and edit image piece properties in the right rail"

To test, upload an image to the media editor. It should open in the right rail editor. Then check a few items and upload an image. The checked items should become un-checked and the new one should be checked and in the editor.

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
@@ -163,9 +163,14 @@ export default {
         dimensions
       });
     },
-    async completeUploading () {
+    async completeUploading (imgId) {
       this.uploading = false;
       await this.getMedia();
+
+      if (imgId) {
+        this.checked = [ imgId ];
+        this.updateEditing(imgId);
+      }
     },
     clearSelected() {
       this.checked = [];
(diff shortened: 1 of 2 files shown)


stuartromanek:
testing this in a3 demo does not do what it advertises

stuartromanek (changes requested):
not working

stuartromanek (changes requested):
If we already have images selected the newly uploaded images should be added to the `selected` array and not reset it

abea:
@stuartromanek Scenario there here says to clear it https://apostrophecms.atlassian.net/browse/A30U-425

> Scenario 3: User uploads new image(s) while currently selecting multiple existing images
> Clear the selected array and pop new items onto it as they are uploaded

stuartromanek:
> @stuartromanek Scenario three here says to clear it https://apostrophecms.atlassian.net/browse/A30U-425
> 
> > Scenario 3: User uploads new image(s) while currently selecting multiple existing images
> > Clear the selected array and pop new items onto it as they are uploaded

Updated scenario 3 to reflect this request ^

I think this direction works for two reasons:
1. You can always jump to Edit for those new images once they're in the Selected rail. The opposite of recovering a selection is not true.
2. If you are mixing select and upload (like in relationships) this preserves your previous work.

abea:
https://media.giphy.com/media/fsnzYr1pgvAQM/giphy.gif

boutell (approved):

stuartromanek (approved):

PR #2425, opened 2020-09-25
3/save new doc

falkodev:
# New doc validation

Resolves https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?assignee=5ddbce7e47baa50cfd05eed2&selectedIssue=A30U-448

## How to test

- Create a new product in a3-demo, click "Save" without touching any input.
- Try to add a slug and stay on the first tab (not containing the required title) : it should not save the doc (because the title is still empty). Go to the "ungrouped" tab and check title is in error.

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -34,7 +34,10 @@
             <div class="apos-doc-editor__body">
               <AposSchema
                 v-if="docReady"
-                :schema="schema" :current-fields="currentFields"
+                :key="'schema-body-validation-' + triggerValidation"
+                :schema="schema"
+                :current-fields="currentFields"
+                :trigger-validation="triggerValidation"
                 v-model="doc"
               />
             </div>
@@ -47,7 +50,10 @@
         <div class="apos-doc-editor__utility">
           <AposSchema
             v-if="docReady"
-            :schema="schema" :current-fields="utilityFields"
+            :key="'schema-rail-validation-' + triggerValidation"
+            :schema="schema"
+            :current-fields="utilityFields"
+            :trigger-validation="triggerValidation"
             v-model="doc"
             :modifiers="['small', 'inverted']"
           />
@@ -98,7 +104,8 @@ export default {
         active: false,
         type: 'overlay',
         showModal: false
-      }
+      },
+      triggerValidation: false
     };
   },
   computed: {
@@ -184,30 +191,34 @@ export default {
     }
   },
   methods: {
-    async submit() {
-      apos.bus.$emit('busy', true);
+    submit() {
+      this.triggerValidation = true;
+      this.$nextTick(async () => {
+        if (!this.doc.hasErrors) {
+          apos.bus.$emit('busy', true);
+          let route;
+          let requestMethod;
+          if (this.docId) {
+            route = `${this.moduleOptions.action}/${this.docId}`;
+            requestMethod = apos.http.put;
+          } else {
+            route = this.moduleOptions.action;
+            requestMethod = apos.http.post;
+          }
 
-      let route;
-      let requestMethod;
-      if (this.docId) {
-        route = `${this.moduleOptions.action}/${this.docId}`;
-        requestMethod = apos.http.put;
-      } else {
-        route = this.moduleOptions.action;
-        requestMethod = apos.http.post;
-      }
+          try {
+            await requestMethod(route, {
+              busy: true,
+              body: this.doc.data
+            });
+            this.$emit('saved');
 
-      try {
-        await requestMethod(route, {
-          busy: true,
-          body: this.doc.data
-        });
-        this.$emit('saved');
-
-        this.modal.showModal = false;
-      } finally {
-        apos.bus.$emit('busy', false);
-      }
+            this.modal.showModal = false;
+          } finally {
+            apos.bus.$emit('busy', false);
+          }
+        }
+      });
     },
     async newInstance () {
       const newInstance = await apos.http.post(this.moduleOptions.action, {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -34,7 +34,6 @@
             <div class="apos-doc-editor__body">
               <AposSchema
                 v-if="docReady"
-                :key="'schema-body-validation-' + triggerValidation"
                 :schema="schemaOtherFields"
                 :current-fields="currentFields"
                 :trigger-validation="triggerValidation"
@@ -51,7 +50,6 @@
         <div class="apos-doc-editor__utility">
           <AposSchema
             v-if="docReady"
-            :key="'schema-rail-validation-' + triggerValidation"
             :schema="schemaUtilityFields"
             :current-fields="utilityFields"
             :trigger-validation="triggerValidation"
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -14,7 +14,7 @@
     <template #primaryControls>
       <AposButton
         type="primary" label="Save"
-        :disabled="doc.hasErrors"
+        :disabled="docOtherFields.hasErrors || docUtilityFields.hasErrors"
         @click="submit"
       />
     </template>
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -34,7 +34,10 @@
             <div class="apos-doc-editor__body">
               <AposSchema
                 v-if="docReady"
-                :schema="schema" :current-fields="currentFields"
+                :key="'schema-body-validation-' + triggerValidation"
```
falkodev, line 37: When the `triggerValidation` (a boolean) is activated (after clicking on "submit"), the key changes, and this forces the schema to re-render its sub-components, forcing them to validate the value they hold.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -116,7 +119,7 @@ export default {
           return 'max';
         }
       }
-      if (this.field.type === 'email') {
+      if (this.field.type === 'email' && value) {
```
falkodev, line 122: Fixes an error when a empty non-required email was in error.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -3,13 +3,15 @@
 -->
 <template>
   <div class="apos-schema">
-    <div v-for="field in currentSchema" :key="field.name">
+    <div v-for="field in schema" :key="field.name">
```
falkodev, line 6: Previously, only the fields that needed to appear on the current tab were validated. Now, every field is used for validation. So need to hide fields belonging to other tabs through `displayComponent` hereunder.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -110,7 +94,7 @@ export default {
       }
     }
   },
-  mounted() {
+  created() {
```
falkodev, line 97: I don't know why, but on "mounted" some fields were `undefined`. With `created` everything field has a value with `{ data: undefined, error: false }` so no problem anymore. If anyone has an explanation..

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
boutell: OK, but why not just pass it as a boolean prop?

boutell: Not sure why the string concatenation.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -184,30 +191,34 @@ export default {
...
-      if (this.docId) {
-        route = `${this.moduleOptions.action}/${this.docId}`;
-        requestMethod = apos.http.put;
-      } else {
-        route = this.moduleOptions.action;
-        requestMethod = apos.http.post;
-      }
+          try {
+            await requestMethod(route, {
+              busy: true,
```
boutell, line 211: Pretty sure this busy: true is enough and we don't need the manual emitting of busy events here, but this is not new in this PR so I won't be a pain about it (:

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
boutell: Weird. Mounted is actually called later.

boutell (changes requested):

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
falkodev: Did you mean having `:key="triggerValidation"`? That means two components would have the same key in `AposDocEditor` because there are 2 `AposSchema`s.  I don't know if this would be a problem.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
falkodev: I can remove `apos.bus.$emit('busy', true);` if you think it it is useless.

abea (changes requested):
This isn't working consistently for me. I click "save" and the errors appear. Then I fill in those fields, but the "Save" button is still disabled. 

Then if I edit an existing piece, clear a required field (triggering an error) and click into another field, validation is run and the save button is enabled, even though there's still an error in the first field.

abea:
It looks like that second bug is in 3.0 as well.

abea:
Maybe `hasErrors` on the AposDocEditor level should be an object with a key for each field name.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: This feels like tricking the schema into validating rather than explicitly telling it to.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
abea: Why is every field in the doc needed for validation?

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
falkodev: `triggerValidation` was not enough to really trigger the validation because the component has already been mounted. It won't validate until it's updated and Vue won't consider the component is updated by only changing the value of `triggerValidation` but it will if the key is changed. So, yes, it is tricky. But I did not find a good alternative. Unless you have a solution in mind.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -93,12 +101,21 @@ export default {
         data: {},
         hasErrors: false
       },
+      docUtilityFields: {
```
falkodev, line 104: `doc` is split into 2 parts, each one will be validated separately, avoiding having 2 `AposSchema` with the same fields but different data (see `v-model` of each `AposSchema`).

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
falkodev: You were right. Removed

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
falkodev: The fielgs belonging to other tabs are not displayed but still need to be validated before save.

boutell (approved):
Approving this, it does what it promises, I will work iteratively from here

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
```diff
@@ -33,6 +34,9 @@ export default {
   },
   mounted () {
     this.$el.addEventListener('focusout', this.validateAndEmit);
+    if (this.triggerValidation) {
```
abea, line 37: We would want to `watch` this instead, right?

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
```diff
@@ -19,7 +19,8 @@ export default {
         return [];
       },
       type: Array
-    }
+    },
+    triggerValidation: Boolean
```
abea, line 23: This should be formatted like the others with `default: false`.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -53,25 +57,11 @@ export default {
...
-      });
-
-      return fields;
-    },
     fields() {
       const fields = {};
-      this.currentSchema.forEach((item) => {
+      this.schema.forEach(item => {
         fields[item.name] = {};
-        fields[item.name].field = { ...item };
```
abea, line 74: 👍

abea (changes requested):

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
falkodev: Yes, and no need for a `key` in this case 👍

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
falkodev: Done

abea (approved):

PR #2429, opened 2020-09-28
renames AposPiecesManagerView

abea:
renames AposPiecesManagerView to renames AposPiecesManagerDisplay. We discussed this to maintain consistency with the media manager display component.


localghost443 (approved):
I like this approach a lot.

falkodev (approved):

PR #2442, opened 2020-10-02
fix risk of the list projection for unrelated piece types being alter…

boutell:
…ed due to copy by reference

See DC 1016

### DIFF
--- a/lib/modules/apostrophe-pieces/lib/api.js
+++ b/lib/modules/apostrophe-pieces/lib/api.js
@@ -616,6 +616,8 @@ module.exports = function(self, options) {
         // add the new field to the columns
         self.columns.push(newColumn);
       });
+      // So we do not inadvertently modify it for other subclasses
+      self.options.listProjection = Object.assign({}, self.options.listProjection);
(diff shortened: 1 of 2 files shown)


boutell:
cc @jogarijo

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -616,6 +616,8 @@ module.exports = function(self, options) {
         // add the new field to the columns
         self.columns.push(newColumn);
       });
+      // So we do not inadvertently modify it for other subclasses
```
falkodev, line 619: Is it to be sure `listProjection` will always be an object?

abea (approved):

lib/modules/apostrophe-pieces/lib/api.js
boutell: Yes there is a default setting for it.

boutell:
@falkodev yes it is always an object, there is a default setting for it.

falkodev (approved):

PR #2444, opened 2020-10-02
Fixes AposDocEditor story

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-573: As a user I see title/slug synchronization in the media editor the same as in the regular doc editor

Prepares us to share more functionality between editor components.

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.stories.js
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.stories.js
@@ -9,7 +9,7 @@ storiesOf('Doc Editor', module)
     },
     data () {
       return {
-        moduleName: 'products',
+        moduleName: 'product',
         id: null, // TODO: Add a demo document to the mock API.
         active: true
       };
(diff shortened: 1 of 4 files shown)


boutell (changes requested):
Shouldn't AposDocEditor also be using the mixin? Also the media editor doesn't have docOtherFields etc. so I don't see how this could work as-is right now.

abea:
Yep. I didn't commit one of my changes...

stuartromanek (approved):

boutell:
That's better, but since the media editor still has no docOtherFields or docUtilityFields, does this actually work — does the slug actually follow the title in the media editor?

Since (if I recall properly) the media editor does group all the fields together, this can work with all of the fields in docOtherFields and an empty docUtilityFields, but those things do need to be part of the plumbing and working the way they do in AposDocEditor.

Also, does the media editor accommodate tabs and extra fields? People are going to have custom fields added via extending `doc-type` or `piece-type` that should be on all pieces.

abea:
@boutell `docOtherFields` and `docUtilityFields` are simply `doc` broken into groups. So using `doc` directly works fine.

There's no design I'm aware of for tabs in a slide out modal.

boutell:
Take a look at the mixin, it references utilityFields and docUtilityFields and so on, but it doesn't create them. Looking at the media editor component, they don't exist on it. Also I don't see how `followingValues` is getting called at all by the media editor component.

Maybe I am taking crazy pills and this is working?

abea:
It is working since there's a conditional looking for `followedBy` in the method. There's an argument for building `docUtilityFields` in the mixin because of this for sure. Or to only have the final `for` loop of the method (below) be in the mixin, with it accepting the fields and source objects as arguments.

```
      const followingValues = {};

      for (const field of fields) {
        if (field.following) {
          followingValues[field.name] = source.data[field.following];
        }
      }

      return followingValues;`
```

boutell:
OK yes, I see it now. This is a good way for it to fall back.

boutell (approved):

PR #2446, opened 2020-10-02
adds New Piece UI to relationship modal

stuartromanek:
Satisfies https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?selectedIssue=A30U-455

Nast

#### Uh oh! That context menu isn't in the right place?
As modals get nested v-tooltip starts to misplace them, I've confirmed placing it higher up works as expected.
Not spending too much time wrestling them now but we need to make sure they get sorted after vue3/teleport

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
@@ -29,9 +29,7 @@
                 </span>
                 {{ modalTitle }}
               </h2>
-              <div
-                class="apos-modal__controls--primary" v-if="hasPrimaryControls"
-              >
+              <div class="apos-modal__controls--primary" v-if="hasPrimaryControls">
                 <slot name="primaryControls" />
               </div>
             </div>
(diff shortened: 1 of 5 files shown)


boutell (approved):

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -4,23 +4,34 @@
...
+        v-else
+        type="default" label="Finished"
+        @click="cancel"
       />
     </template>
-    <template v-else #primaryControls>
+    <template #primaryControls>
+      <!-- TODO Make sure this gets positioned correctly after Vue3/teleport -->
+      <AposContextMenu
+        v-if="relationshipField"
```
abea, line 22: This needs to be conditional on `relationshipMore.menu.length`

modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
```diff
@@ -362,9 +360,18 @@ export default {
...
+    display: flex;
+    align-items: center;
+  }
+
   .apos-modal__controls--primary {
+    justify-content: flex-end;
     flex-grow: 1;
-    text-align: right;
+  }
+  .apos-modal__controls--primary /deep/ > * {
```
abea, line 373: Why not put `.apos-button` instead of the star?

abea (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
stuartromanek: generalized the More menu to be independent of the whether the modal has a relationship field .. This should allow other scenarios to add to it should they need to

abea (approved):

PR #2448, opened 2020-10-05
3/multi upload

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-415

Adds multi-image uploading. Also breaks the uploader into its own component.

💁‍♀️Use this to view only the PRs _after_ breaking the component out: https://github.com/apostrophecms/apostrophe/pull/2448/files/db15939dedde1a14a3811c38dd146c4a690ed6ad..711fb16d399573ae4f43044d5fe68472a79826c1

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
@@ -25,6 +25,7 @@
       type="file" class="apos-sr-only"
       ref="apos-upload-input"
       @input="uploadMedia"
+      multiple="true"
     >
   </label>
 </template>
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
```diff
@@ -0,0 +1,258 @@
...
+      });
+
+      try {
+        const imgPiece = await apos.http.post(this.action, {
+          body: imageData
+        });
+
+        return imgPiece;
+      } catch (error) {
+        await this.notifyErrors(error, 'Upload Error');
```
boutell, line 162: I think this method doesn't exist in the uploader component, it was left behind in the parent component.

boutell (changes requested):
Looking good, think I found a minor bug.

boutell (approved):

stuartromanek (approved):
Looks good, seems to work .. the experience is a little sluggish .. the success notifications seem out of sync with when the thumbnail actually get populated (in my test all 4 image thumbs got populated at once, after all uploads were complete)

Ideally thumbnails would not wait for each other and populate the preview as they are ready

@localghost443 when you test on staging happy to help write a follow up ticket

Issue #2173, opened 2020-04-07
Groups should not have a "published" column in the manage view


boutell:
There is no "published" field for groups (or users) but groups, at least, still have the column configured in the manage view, always showing "draft."

PR #2449, opened 2020-10-06
Remove published columns in apostrophe-users and apostrophe-groups

Jose96GIT:
Solves #2173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.112.2 (2020-10-07)
+* Remove published columns in apostrophe-users and apostrophe-groups, the modules where this field does not exist.
+
 ## 2.112.1 (2020-10-07)
 * When configuring columns in the pieces manager, the `listProjection` option was accidentally altered in a way that would impact other subclasses of pieces. This has been fixed.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.112.2 (2020-10-07)
+## 2.113.0 (2020-10-07)
 * Remove published columns in apostrophe-users and apostrophe-groups, the modules where this field does not exist.
 
 ## 2.112.1 (2020-10-07)
--- a/lib/modules/apostrophe-groups/index.js
+++ b/lib/modules/apostrophe-groups/index.js
@@ -64,8 +64,7 @@ module.exports = {
     options.removeFilters = [ 'published' ]
       .concat(options.removeFilters || []);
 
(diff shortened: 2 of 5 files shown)


lib/modules/apostrophe-groups/index.js
```diff
@@ -64,6 +64,9 @@ module.exports = {
     options.removeFilters = [ 'published' ]
       .concat(options.removeFilters || []);
 
+    options.removeColumns = (options.defaultRemoveColumns || [ 'published' ])
```
abea, line 67: I don't know if this one needs the `options.defaultRemoveColumns` option since there isn't a similar option for `removeFilters`. None of these seems to be documented either, so it seems safe to make this simpler.

lib/modules/apostrophe-users/index.js
```diff
@@ -185,6 +185,9 @@ module.exports = {
     options.removeFilters = (options.defaultRemoveFilters || [ 'published' ])
       .concat(options.removeFilters || []);
 
+    options.removeColumns = (options.defaultRemoveColumns || [ 'published' ])
```
abea, line 188: I'm torn on whether `options.defaultRemoveColumns || ` is needed here as well. I guess it makes sense give the others here and that users might more reasonably be used as content pieces.

abea (changes requested):
One change note. Also please update the changelog.

boutell (changes requested):
This needs to respect any existing `options.removeFields` and concatenate, otherwise it will break project-level use of `removeFields` for these modules.

As for `defaultRemoveColumns`, the logical thing would be to introduce that as a new option whose preconfigured default *is* `[ published ]`.

Jose96GIT:
I guess that last changes should match @boutell suggestion.

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -628,6 +628,8 @@ module.exports = function(self, options) {
         }
       });
     }
+    options.removeColumns = (options.defaultRemoveColumns || [ 'published' ])
```
boutell, line 631: This is in apostrophe-pieces so you are removing published for everything. But I don't think you need it, since the subclass will have set this correctly already in beforeConstruct.

boutell (changes requested):

boutell (approved):

abea (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.112.2 (2020-10-07)
```
abea, line 3: Would this be a patch or minor version bump?

boutell: Minor version bump, new feature.

PR #2452, opened 2020-10-07
Implemented the max option for areas

boutell:


abea (approved):

stuartromanek:
testing instructions would be helpful here!

boutell:
Excerpt from my `modules/home-page/index.js`:

```
module.exports = {
  fields: {
    add: {
      top: {
        type: 'area',
        options: {
          widgets: {
            'test1': {},
            '@apostrophecms/rich-text': {
              toolbar: [ 'bold', 'italic', 'link' ]
            }
          },
          max: 2
        }
      }
    }
  }
};
```

I also had to add the `top` area to the template since it existed in the index.js but not the template.

stuartromanek (approved):
lgtm .. @localghost443 we need a follow up ticket for adding a tooltip in the case of area's limit reached

PR #2455, opened 2020-10-08
fix the bug that broke keyboard focus when editing a nested rich text…

boutell:
… widget. In the process, reduced our dependency on deep watching. Also fixed: raw HTML widgets, and a bug where area event handlers were never removed.

You may note that we're not setting `this.next` to a new array anymore when a change propagates from a widget nested within. That's deliberate, the nested widget has already updated itself and so all that code was doing was causing an unnecessary performance hit and the loss of keyboard focus due to a complete rebuild of all nested areas.

In theory there is a scenario where the same area appears on the page twice, and it would be nice if the one you're not editing also reflected your edits. I experimented in that direction but encountered a fairly gnarly Vue bug that we don't need to be wranglin' right now, so I went with a simple fix that doesn't move the goalposts.

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
@@ -177,7 +177,8 @@ export default {
       });
 
       if (changeFound) {
-        this.$emit('input', this.next);
+        // ... removes need for deep watch at parent level
+        this.$emit('input', { ...this.next });
       }
     },
     findRealChange(oldData, newData) {
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
@@ -140,17 +140,10 @@ export default {
   mounted() {
     if (this.docId) {
       this.areaUpdatedHandler = (area) => {
-        let patched = false;
         for (const item of this.next) {
           if (this.patchSubobject(item, area)) {
-            patched = true;
             break;
           }
-        }
-        if (patched) {
-          // Make sure our knowledge of the change is reflected
-          // everywhere via a refresh
-          this.next = this.next.slice();
         }
       };
       apos.bus.$on('area-updated', this.areaUpdatedHandler);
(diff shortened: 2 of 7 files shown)


boutell:
I want to do more validation of editing nested areas inside a modal with this fix in place, but due to the current z-index bugs it isn't really practical until the teleport PR lands. At that point I'll take another look at nested areas in modals.

boutell:
(It is worth landing this PR in the meantime because Stu is actively testing editing of this sort in context on the page.)

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -177,7 +177,8 @@ export default {
       });
 
       if (changeFound) {
-        this.$emit('input', this.next);
+        // ... removes need for deep watch at parent level
+        this.$emit('input', { ...this.next });
```
abea, line 181: This seems to be breaking the media manager's ability to switch between images active in the editor. That was the initial use case of this.

abea (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
abea: I'm not sure I see how this relates to this issue as well.

boutell: I have to get rid of re-slicing the entire `next` array on changes by subwidgets. But without that, I was encountering a nasty deep Vue bug supposedly fixed in 2018 until I removed the deep watch.

Uh... that's my story, but I'm not quite sure if I'm sticking to it. I can try to reproduce it again.

I would like to know though why this breaks the media manager's ability to switch between images. it makes me wonder if the media manager is following the v-model pattern properly when using AposSchema, which promises to tell you about changes only by emitting an object with an `input` event (which triggers v-model accepting that new value). It doesn't and shouldn't promise never to emit a new object.

stuartromanek (approved):
Area editing is improved (!) but I'm not sure I understand what from this PR is fixing that 🙃

abea (approved):
nevermind it works.

PR #2458, opened 2020-10-08
3/new page from tree

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-130: "As a user I can create a subpage of the current page"

This lets you create a new page from the page manager, making it the child of the current page. Also an update to the slug-follow mechanism in there for pages.

A demo site branch is in PR to add another page option: https://github.com/apostrophecms/a3-demo/pull/25

Look here to avoid the portal commit: https://github.com/apostrophecms/apostrophe/pull/2458/files/1071e7f4508a58f0262ab7660ec5140fde5b00b3..0592564d48533efb47ec2d1684b6b00f691e2c27

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -173,7 +173,7 @@ export default {
       return tabs;
     },
     modalTitle () {
-      return `Edit ${this.moduleOptions.label}`;
+      return `Edit ${this.moduleOptions.label || ''}`;
     },
     currentFields: function() {
       if (this.currentTab) {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -256,7 +256,7 @@ export default {
           return;
         }
 
-        this.doc.data = {
+        const body = {
           ...this.doc.data,
           ...this.docUtilityFields.data,
           ...this.docOtherFields.data
@@ -269,11 +269,16 @@ export default {
         } else {
           route = this.moduleAction;
           requestMethod = apos.http.post;
+
+          if (this.moduleName === '@apostrophecms/page') {
+            body._targetId = apos.page.page._id;
+            body._position = 'lastChild';
+          }
         }
 
         await requestMethod(route, {
           busy: true,
-          body: this.doc.data
+          body
         });
         this.$emit('saved');
         this.modal.showModal = false;
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -71,6 +71,7 @@
 import AposModalParentMixin from 'Modules/@apostrophecms/modal/mixins/AposModalParentMixin';
 import AposModalTabsMixin from 'Modules/@apostrophecms/modal/mixins/AposModalTabsMixin';
 import AposEditorMixin from 'Modules/@apostrophecms/modal/mixins/AposEditorMixin';
+import { defaultsDeep } from 'lodash';
 
 export default {
   name: 'AposDocEditor',
@@ -101,6 +102,7 @@ export default {
   emits: [ 'saved', 'safe-close' ],
   data() {
     return {
+      docType: this.moduleName,
       doc: {
         data: {},
         hasErrors: false
@@ -119,6 +121,7 @@ export default {
         type: 'overlay',
         showModal: false
       },
+      splittingDoc: false,
       schemaUtilityFields: [],
       schemaOtherFields: [],
       triggerValidation: false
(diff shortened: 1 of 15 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -165,7 +173,7 @@ export default {
       return tabs;
     },
     modalTitle () {
-      return `Edit ${this.moduleOptions.label}`;
+      return `Edit ${this.moduleOptions.label || ''}`;
```
abea, line 176: If they didn't set a label on a custom module you'd have `Edit undefined`.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -119,14 +121,20 @@ export default {
...
   },
   computed: {
     moduleOptions() {
-      return window.apos.modules[this.moduleName] || {};
+      return window.apos.modules[this.docType] || {};
+    },
+    moduleAction () {
+      // Use moduleName for the action since all page types use the
+      // `@apostrophecms/page` module action.
+      return (window.apos.modules[this.moduleName] || {}).action;
```
abea, line 137: this was the main thing that we can't get from the `docType` module.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -179,13 +187,37 @@ export default {
...
+    'docUtilityFields.data': {
+      deep: true,
+      handler(newVal, oldVal) {
+        if (this.moduleName !== '@apostrophecms/page' || this.splittingDoc) {
+          return;
+        }
+
+        if (this.docType !== newVal.type) {
+          // Return the split data into `doc.data` before splitting again.
+          this.doc.data = defaultsDeep(this.docOtherFields.data, this.docUtilityFields.data, this.doc.data);
```
abea, line 200: The new schema may have a field with the same in name in a different tab, so get it all back together before splitting up again.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -224,49 +256,89 @@ export default {
...
+          body._position = 'lastChild';
         }
-      });
+        const newDoc = await apos.http.post(this.moduleAction, {
+          body
+        });
+
+        return newDoc;
+      } catch (error) {
+        await apos.notify('Error while creating a new, empty item. Please check configured content types.', {
```
abea, line 303: this happened to me because the page types was misconfigured in the project. Maybe unnecessary?

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -224,49 +256,89 @@ export default {
...
+          type: 'danger',
+          icon: 'alert-circle-icon',
+          dismiss: true
+        });
+
+        this.cancel();
```
abea, line 309: bail.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -224,49 +256,89 @@ export default {
...
+
+      this.docUtilityFields.data = {};
+      this.docOtherFields.data = {};
+      this.schemaUtilityFields = [];
+      this.schemaOtherFields = [];
+
       this.schema.forEach(field => {
         if (field.group.name === 'utility') {
-          this.docUtilityFields.data[field.name] = this.doc.data[field.name];
+          this.docUtilityFields.data[field.name] = this.doc.data[field.name] || field.def;
```
abea, line 334: There might be a better way or place to use the default value.

modules/@apostrophecms/page/index.js
```diff
@@ -202,7 +203,7 @@ module.exports = {
       // `_targetId` is the _id of another page, and `_position` may be `before`, `after`,
       // `firstChild` or `lastChild`.
       //
-      // If you do not specify these properties they default to the homepage and `inside`,
```
abea, line 205: this seemed like a 2.x leftover

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -145,7 +167,7 @@ export default {
             all: 1
           }
         }
-      )).results;
```
abea, line 148: I'm not sure when this changed.

stuartromanek (commented):
wouldn't mind doing a ☎️ for this one

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
boutell: Probably a buggy borrowing from pieces, which return metadata related to pagination etc. as well as the pieces.

boutell (changes requested):

boutell (approved):

stuartromanek (approved):
Two overall concerns to existing Apostrophe patterns here:
- The intermittent module checking in AposDocEditor feels like it could get unwieldy, hard to read etc. Not a show stopper.

- All the doc splitting and remerging feels brittle. We should separate these UI presentation groupings from the actual doc we need to read/write/send so we're always sure we have what we think we have and can skip transforming and flag setting. We might want a follow up ticket to refactor this

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: I do think we should revisit this splitting pattern in beta phase. It doesn't feel great. CC @localghost443

PR #2460, opened 2020-10-08
3/replace img

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-396, "As a user, I can replace an image in the schema preview," and https://apostrophecms.atlassian.net/browse/A30U-404, "As a user, I can view an image in the schema preview"

To test, jump on into the media library and open an image in the editor. Click "View" to open that image in a new tab to download. Then click "Replace" to view the attachment input. You can clear that, then upload a new file. You can try a non-image and it should be safely rejected with a warning.


boutell (approved):

stuartromanek (approved):
The UX on this is a little rough, making a follow up jira ticket to smooth out this experience later

PR #2464, opened 2020-10-09
adds alt text field to images

abea:
resolves https://apostrophecms.atlassian.net/browse/A30U-179, primarily adding an explicit alt text field.

### DIFF
--- a/modules/@apostrophecms/image/index.js
+++ b/modules/@apostrophecms/image/index.js
@@ -41,7 +41,7 @@ module.exports = {
         fileGroup: 'images',
         required: true
       },
-      altText: {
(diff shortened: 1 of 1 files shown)


boutell:
It sounds like this should not be `textarea: true`:

https://developer.paciellogroup.com/blog/2015/09/short-note-on-coding-alt-text/

boutell (changes requested):
Alt should not be textarea.

localghost443:
@boutell I don't believe that textareas preserve line breaks

boutell (approved):

stuartromanek (approved):

PR #2466, opened 2020-10-10
adding a widget was throwing an error per A30U-583 and leaving the context menu on the page

boutell:
I know passing a function as a prop is generally not preferred in node but since emitting events from slot to parent is forbidden it is the most frequently cited (and thumbsed-up) solution to this problem, see discussion:

https://github.com/vuejs/vue/issues/4332


stuartromanek:
closing this in favor of simpler #2467

PR #2475, opened 2020-10-13
Simple image widget, also use correct chooser for images, numerous fixes to piece…

boutell:
Piece manager now supports standard configured filters via select element (parity with A2), REST API supports null values for filters, radio button and select elements directly support non-string values making them more suitable for use as filters.

Previously we used hacks like the string `any`, which was problematic because it can be a legitimate value.

Now null values are represented in the query string as `?_tags` rather than `?_tags=` (note the lack of a =). This is a query string convention that started with Java and was embraced in other frameworks. It is also supported by the qs module used in Express, but not turned on by Express. So I found out how to configure our own query parser (i.e. qs with the right option passed) the approved Express way.

But oh yeah, image widgets! I wrote a dead simple single image widget here, as simple as it could possibly be, but it does use srcset and alt as specified.

I also elected to put a class on the img tag so that styling it is at least possible, and put `max-width: 100%` on `apos-image-widget` in my project level code, which had the desired effect. Gotta do something, or it breaks out of the box.

### DIFF
--- a/modules/@apostrophecms/image-widget/views/widget.html
+++ b/modules/@apostrophecms/image-widget/views/widget.html
@@ -1,6 +1,6 @@
 {% set attachment = apos.image.first(data.widget._image) %}
 {% if attachment %}
-  <img {% if data.manager.options.className %} class="{{ data.manager.options.className }}"{% endif %}
+  <img{% if data.manager.options.className %} class="{{ data.manager.options.className }}"{% endif %}
     srcset="{{ apos.image.srcset(attachment) }}"
     src="{{ apos.attachment.url(attachment, { size: data.options.size or 'full' }) }}"
     alt="{{ data.widget._image[0].alt or '' }}"
--- a/modules/@apostrophecms/image-widget/views/widget.html
+++ b/modules/@apostrophecms/image-widget/views/widget.html
@@ -1,0 +1,8 @@
+{% set attachment = apos.image.first(data.widget._pieces) %}
+{% if attachment %}
+  <img class="apos-image-widget"
+    srcset="{{ apos.image.srcset(attachment) }}"
+    src="{{ apos.attachment.url(attachment, { size: data.options.size or 'full' }) }}"
+    alt="{{ data.widget._piece.altText or '' }}"
+  />
+{% endif %}
--- a/modules/@apostrophecms/image-widget/views/widget.html
+++ b/modules/@apostrophecms/image-widget/views/widget.html
@@ -1,8 +1,8 @@
-{% set attachment = apos.image.first(data.widget._pieces) %}
+{% set attachment = apos.image.first(data.widget._image) %}
 {% if attachment %}
   <img class="apos-image-widget"
     srcset="{{ apos.image.srcset(attachment) }}"
     src="{{ apos.attachment.url(attachment, { size: data.options.size or 'full' }) }}"
-    alt="{{ data.widget._piece.altText or '' }}"
+    alt="{{ data.widget._image[0].alt or '' }}"
   />
 {% endif %}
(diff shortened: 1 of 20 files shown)


modules/@apostrophecms/image/index.js
```diff
@@ -19,11 +19,6 @@ module.exports = {
     sort: { createdAt: -1 },
     insertViaUpload: true,
     searchable: false,
-    browser: {
```
boutell, line 22: Something like this is desirable, but options don't merge (they never have), a new setting for a top level option blows out an inherited one. So for now this pattern doesn't work, we must extend the getBrowserData method. We could discuss a feature to provide option merging at some point. It's never as simple as it seems...

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -170,7 +175,11 @@ export default {
   },
   created() {
     this.options.filters.forEach(filter => {
-      this.filterValues[filter.name] = filter.def || filter.choices[0].value;
+      this.filterValues[filter.name] = filter.def;
```
boutell, line 178: There might not be any choices yet. Better to require explicit defaults, or do the "first value as fallback default" dance more carefully.

boutell:
I had initially suggested that I fixed a bug with filter values not sticking, but that was a side effect of passing in the server-supplied choices as props which required some extra plumbing only in this new situation. Honor is preserved.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -365,20 +366,27 @@ module.exports = {
...
+              filter.def = null;
+              filter.choices.push({
+                value: null,
+                label: 'None'
+              });
+            }
+          } else {
+            // Dynamic choices from the REST API, but
+            // we need a label for "no opinion"
+            filter.nullLabel = 'Choose One';
```
abea, line 389: This is only in play when there are no set choices? I'm not clear on why "choose one" makes sense then.

abea (changes requested):

modules/@apostrophecms/piece-type/index.js
boutell: The REST API will supply the choices, but it doesn't supply an "I don't care" choice, so we must supply that. "Choose One" is a good label because it is then understood that you can start filtering by changing it.

boutell:
@abea I fixed the issues you mentioned except for "choose one", see my note above.

I also added a `className` module option per our discussion today, and eliminated offering any class on the img tag by default, but I still wonder if this is a bit much. Now you have to configure the module whereas formerly you could just use it if you didn't mind the default class name. We could give it a default value; you could still set it `false`.

abea (dismissed):

modules/@apostrophecms/image-widget/views/widget.html
```diff
@@ -0,0 +1,8 @@
+{% set attachment = apos.image.first(data.widget._image) %}
+{% if attachment %}
+  <img {% if data.manager.options.className %} class="{{ data.manager.options.className }}"{% endif %}
```
stuartromanek, line 3: Widget configuration doesn't feel like the right place to set `className` .. I think I would mirror rich text and just give it a class of `apos-image`

stuartromanek (changes requested):

boutell:
I added classes on the appropriate element for each widget. It is possible to override them by setting the className option of that module, but I expect most people who don't override the template will just use the class name we're providing to style.

I included a `-widget` suffix on them just for good namespacing.

stuartromanek (approved):

PR #2479, opened 2020-10-13
Allows AposButton to be an anchor and adds download link

abea:
Controversy...

Resolves https://apostrophecms.atlassian.net/browse/A30U-403, "As a user, I can download an image in the schema preview"


boutell (approved):

boutell (approved):

stuartromanek (approved):
i like when the button component asks itself 'am i button?'

PR #2480, opened 2020-10-13
Slats array editor

boutell:
Acceptance of this PR does not mean the whole ticket is done because I'm waiting on Stu's design to deal with things like the add button better, but accepting the PR soon would be good because it fixes issues for both slats and attachments too (slat related).

Now uses slat list Also, refactored slat list to support selection and removed logic that modified props (not permitted in Vue).

Slat list now supports the v-model pattern, although we are currently choosing to pass in "value" and accept the "input" event directly in the three components that use it (that is also what v-model does).

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
@@ -29,6 +29,7 @@
           @remove="remove"
           @engage="engage"
           @disengage="disengage"
+          @select="select"
           @move="move"
           @item-clicked="$emit('item-clicked', item)"
           :key="item._id"
@@ -76,7 +77,7 @@ export default {
       }
     }
   },
-  emits: [ 'update', 'item-clicked' ],
+  emits: [ 'update', 'item-clicked', 'select' ],
   data() {
     return {
       isDragging: false,
@@ -115,6 +116,7 @@ export default {
     }
   },
   mounted() {
+    console.log('sanity check');
     this.updateMessage();
   },
   methods: {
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
```diff
@@ -110,8 +114,25 @@ export default {
         this.delayedDragging = false;
       });
     },
-    items() {
-      this.updateMessage();
+    value() {
+      this.next = this.value.slice();
```
boutell, line 118: Guarantees we don't modify the original.

modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
```diff
@@ -110,8 +114,25 @@ export default {
...
+    },
+    next(newValue, oldValue) {
+      let equal = true;
+      if (newValue.length === this.value.length) {
```
boutell, line 122: This logic prevents infinite loops while still allowing the parent to send a decorated version of its current list into the slat list. This makes it easy for the array editor to send in temporary fake titles like `#5`.

We only consider changes to the list length, the ids (order), and the title to be interesting because those are the legitimate concerns of the slat list.

boutell:
I made a separate ticket for just this part so we can close something when it is merged.

stuartromanek (changes requested):
When opening an array field with items already inside, default to open the first element in the modal body

stuartromanek (approved):

abea (changes requested):
I'm getting a styling bug where if there's only one in the slat list and I click it, then hover away, the bg color goes back to light grey but the text color stays white, making it hard to read. It looks like it's `.is-selected` with `:focus` styles but not `:hover`.

boutell:
@abea I addressed the thing; good to go?

abea (approved):

PR #2481, opened 2020-10-14
Register icons

boutell:
I added the ability to register material design icons from any module. I also thew in the ability to register an icon to be imported from somewhere else, by starting the `importFrom` argument with a `~`, for symmetry with the way you escape the normal import rules and get to "real" webpack imports in SASS.

This PR includes an example: the rich text widget now registers an appropriate icon.

I have verified that the icon appears in the "add widget" dropdown (plumbing for this mostly existed already but needed a little lashing-together).

### DIFF
--- a/modules/@apostrophecms/area/index.js
+++ b/modules/@apostrophecms/area/index.js
@@ -125,10 +125,14 @@ module.exports = {
         const field = self.apos.schema.getFieldById(area._fieldId);
         const options = field.options;
         _.each(options.widgets, function (options, name) {
-          choices.push({
-            name: name,
-            label: options.addLabel || (self.widgetManagers[name] && self.widgetManagers[name].label) || '*ERROR*'
-          });
+          const manager = self.widgetManagers[name];
+          if (manager) {
+            choices.push({
+              name: name,
+              icon: manager.options.icon,
+              label: options.addLabel || manager.label || `No label for ${name}`
+            });
+          }
         });
         // Guarantee that `items` at least exists
         area.items = area.items || [];
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -28,72 +28,111 @@ module.exports = {
   },
 
   init(self, options) {
-
     self.restartId = self.apos.util.generateId();
-
-    self.addTask('build', 'Build Apostrophe frontend javascript master import files', async function (apos, argv) {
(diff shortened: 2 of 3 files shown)


modules/@apostrophecms/asset/index.js
```diff
@@ -28,72 +28,111 @@ module.exports = {
   },
 
   init(self, options) {
-
     self.restartId = self.apos.util.generateId();
-
-    self.addTask('build', 'Build Apostrophe frontend javascript master import files', async function (apos, argv) {
```
boutell, line 34: I refactored the body of the task to a method so it can be properly overridden. No code changes beyond that except those specific to icons.

stuartromanek:
Follow up ticket to refactor old pattern, if more detail is needed please edit https://apostrophecms.atlassian.net/browse/A30U-607

stuartromanek (approved):

modules/@apostrophecms/area/index.js
```diff
@@ -125,10 +125,14 @@ module.exports = {
...
         _.each(options.widgets, function (options, name) {
-          choices.push({
-            name: name,
-            label: options.addLabel || (self.widgetManagers[name] && self.widgetManagers[name].label) || '*ERROR*'
-          });
+          const manager = self.widgetManagers[name];
+          if (manager) {
+            choices.push({
+              name: name,
+              icon: manager.options.icon,
```
abea, line 132: We're sure there's always an `options` object?

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -8,6 +8,7 @@ const jsDiff = require('diff');
 module.exports = {
   extend: '@apostrophecms/widget-type',
   options: {
+    icon: 'format-text-icon',
```
abea, line 11: I'm a little worried that we don't seem to have a safety net if they choose an icon here and it isn't registered. It would probably be done in AposAreaMenuItem. The dev should be catching this while building their site, but it'd be nice if we could catch it. Nothing actionable yet, maybe, but something to think about. CC @localghost443

abea (approved):

modules/@apostrophecms/rich-text-widget/index.js
boutell: Yeah that's catchable. Would make a good followup ticket.

PR #2483, opened 2020-10-14
When editing users, full name now follows first and last name

boutell:
* Full name now follows first and last name input when editing a user
* `following` may now be a single field name or an array of field names
* `followingValue` prop of input components is now `followingValues`, an object with sub-properties for all the fields your field is following
* AposInputString will join the followed values with spaces and use that, provided the old value was already a match for the old followed value (in sync, not deliberately different)
* AposInputSlug now supports multiple source fields in the same way, because why not
* fixed the slug prefix for users and groups
* removed the published column and filter for users and groups


modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -4,33 +4,40 @@
...
+          for (const name of following) {
+            followingValues[field.name][name] = get(name);
+          }
         }
       }
-
       return followingValues;
+
+      function get(name) {
+        if (self.docUtilityFields && (self.docUtilityFields.data[name] !== undefined)) {
```
abea, line 34: We do have some editors that don't have these split groups (e.g., media editor) that do the following dance.

abea: If we have an `else` in here using `self.doc` that should do it. Also another argument for us to undo this splitDoc business.

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
```diff
@@ -81,15 +81,23 @@ export default {
...
-        } else {
-          this.next = this.slugify(newValue);
+    followingValues: {
+      // We are usually interested in followingValue.title, but a
+      // secondary slug field could be configured to watch
+      // one or more other fields
+      deep: true,
+      handler(newValue, oldValue) {
+        oldValue = Object.values(oldValue).join(' ');
+        newValue = Object.values(newValue).join(' ');
```
abea, line 91: Should we bail of the values are objects or arrays?

abea (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
boutell: The only values we will ever receive are those we are `following`. This input field type should be documented as having support for `following` fields that convert themselves sensibly to a string. I don't think we need to do more than that.

boutell: (Documentation being a task we all know is forthcoming.)

abea:
@boutell Did you see the other comment about the media editor not splitting the doc?

boutell:
@abea I addressed this with some changes to the mixin and the use of the method override pattern as a way to migrate any weirdness to things that are actually weird, i.e. AposDocEditor.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -103,10 +103,6 @@ export default {
   data() {
     return {
       docType: this.moduleName,
-      doc: {
```
boutell, line 106: Now docFields, which is configured for us by the mixin.

boutell: I prefer docFields because (1) it's analogous to the split field properties, and (2) doc should be a doc, not a thing with a data property and an error property, plus it's not uncommon for the actual existing doc to be a prop in other potential users of the mixin.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -82,6 +79,25 @@ export default {
...
   },
+  watch: {
+    followingValues: {
+      // We may be following multiple fields, like firstName and lastName,
+      // or none at all, depending
+      deep: true,
+      handler(newValue, oldValue) {
+        // Follow the value of the other field(s), but only if our
+        // previous value matched the previous value of the other field(s)
+        oldValue = Object.values(oldValue).join(' ').trim();
```
boutell, line 90: Trim solved a problem when first or last name is blank.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -13,7 +13,9 @@
         <template #bodyMain>
           <div class="apos-widget-editor__body">
             <AposSchema
-              :schema="schema" v-model="docInfo"
+              :schema="schema"
+              v-model="docFields"
+              ref="schema"
```
abea, line 18: This `ref` isn't used, right? Can we nix that before merge?

boutell:
Marked as a draft because I merged the as yet unapproved esc key branch into it, so I'm not blocked on that later when the esc key branch is approved.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
abea: It's used in the mixin

boutell: Yep

abea (approved):

PR #2484, opened 2020-10-14
Removes the transition for the sake of nesting

abea:
Needs the designer blessing.

Our transition group put a `span` in between the draggable component and the target list. This looks to have got in the way when the list was empty. No one is happy, the sign if a good compromise.

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
@@ -310,9 +310,6 @@ export default {
     &.is-dragging {
       opacity: 0.5;
     }
-    &.apos-flip-list-move {
-      transition: transform 0.4s;
-    }
   }
   .apos-tree__list {
     transition: max-height 0.3s ease;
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -7,90 +7,86 @@
...
-          <!-- {{ options.startCollapsed }} -->
-          <button
-            v-if="row.children && row.children.length > 0"
-            class="apos-tree__row__toggle" data-apos-tree-toggle
-            aria-label="Toggle section" :aria-expanded="!options.startCollapsed"
-            @click="toggleSection($event)"
-          >
-            <chevron-down-icon :size="16" class="apos-tree__row__toggle-icon" />
-          </button>
+    <li
```
abea, line 10: Hide whitespace changes.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -7,90 +7,86 @@
...
-              v-if="col.icon" :is="icons[col.icon]"
-              class="apos-tree__cell__icon"
-            />
-            <span v-show="!col.iconOnly">
-              {{ row[col.name] }}
-            </span>
-          </component>
-        </div>
-        <AposTreeRows
-          v-if="row.children"
```
abea, line 71: This does nothing. `children` is always an empty array at least.

stuartromanek (approved):
😚 👌 💥

abea:
... I think there may be a way to bring it back by setting it to be a div.

PR #2485, opened 2020-10-15
handle server side errors well

boutell:
Handle server side errors well, regardless of nested depth in the schema; bring them to light on the front end; removed status prop that was not well used and replaced it with a data field where used.

You can test by altering a schema field type to throw a server side error in some easily reproduced situation. For instance, I did it by making this change to string fields in schema/index.js.

```
object[field.name] = self.apos.launder.string(data[field.name], field.def);
if (object[field.name].length === 5) {
  throw self.apos.error('invalid', 'This server hates five character strings.');
}
```


boutell:
The way server side errors are reported in our REST APIs changed a bit, I believe I addressed the places where we were paying attention to that and tested, in particular attachments.

boutell:
I am refactoring a little to merge most of this into the mixin so that the media editor has it too.

boutell:
Supported in the media editor too now. Now it's a CMS.

stuartromanek (approved):
Approving these Great Enhancements, @localghost443 and I need to work out tickets that complete the user experience (error counts on schema tabs, clearing errors while fixing, more hand holding)

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -84,10 +91,7 @@ export default {
         fields[item.name].value = {
           data: this.value[item.name]
         };
-        // What is this TODO supposed to be? We have error and value already. -Tom
```
abea, line 87: 😆

abea (approved):

Issue #1365, opened 2018-05-10
No warning for piece types adding a field called "type"


abea:
Right now there's no warning for creating a piece type with new added field with `name: 'type'`. This is used by Apostrophe already and any data users attempt to store there will be overridden.

[I attempted to solve this](https://github.com/apostrophecms/apostrophe/pull/1363) by adding it to the `forbiddenFields` array in doc-type-manager, but that was too far-reaching. There might need to be solution closer tied to `apostrophe-pieces`.

boutell:
Hint: this could be checked for in `apostrophe-pieces` and `apostrophe-widgets` in `afterConstruct`, where `self.schema` should already be fully baked. It should not be checked for in the universal case in `apostrophe-schemas` because for pages, a `type` field is appropriate.

itsajay1029:
Hey @boutell  can you assign me this issue so that I can get started working on it ?

boutell:
Hi @itsajay1029, we generally don't use the assign feature, but you are welcome to work on it, and we'll let this comment stand as an indication that we're aware of your intention to do so.

sajdakabir:
is this issue open?

boutell:
@sajdakabir offhand it looks like this still has not been addressed, at least not in A3. It's easy to check by trying it.

If you're interested in creating a PR, you would need to put that check in the `@apostrophecms/piece-type` and `@apostrophecms/widget-type` modules (for A3). See earlier notes for where to put it in A2 if you're interested in A2 (maintained in the `2.0` branch).

Ayushparui:
> @sajdakabir offhand it looks like this still has not been addressed, at least not in A3. It's easy to check by trying it.
> 
> If you're interested in creating a PR, you would need to put that check in the `@apostrophecms/piece-type` and `@apostrophecms/widget-type` modules (for A3). See earlier notes for where to put it in A2 if you're interested in A2 (maintained in the `2.0` branch).

Hello @boutell I'm new to open source. This codebase is a little overwhelming to me. What I understood by the issue is it's not providing any warning for creating a piece with `name: 'type'`. Would you mind elaborating on this?

BoDonkey:
[...]

PR #2489, opened 2020-10-16
Check that schema field does not contain any forbidden field

Jose96GIT:
Following the #1365 issue, don't you think that type should be considered as a forbbiden field for apostrophe-pieces.


boutell (changes requested):
Good idea. Please name this method `validateSchema` though for consistency, and refer to it as a "piece type" rather than a "doc type" just to make it that much easier to find. Thanks!

boutell (approved):

PR #2494, opened 2020-10-20
ESC key, confirm modals

boutell:
* Esc key closes only the topmost modal
* Simplified modal handling: `await apos.modal(confirmContent)` returns either `true` or `false`
* `AposModalMixin` automatically provides a confirm modal before canceling if `this.modified` is true, labels can be overridden by setting simple data properties
* All major document editors have a custom cancelDescription
* Various uses of v-model replaced with explicit :value prop and @input handler in order to be able to detect when a change comes directly from the user and set `this.modified` without setting it every time like a watch would


modules/@apostrophecms/modal/ui/apos/mixins/AposModalParentMixin.js
```diff
@@ -7,15 +7,37 @@ parent components should initiate that transition on mount.
 They should also emit a `safe-close` event after the modal is completely
 closed (`modal.active` is set to `false`), at which point they themselves can
 be destroyed however necessary.
+
+If the modal parent sets `this.modified` to true, a confirmation prompt will
+appear on attempts to click cancel or use the esc key to close the modal.
+The labels can be overridden by overriding data properties as shown below.
 */
 export default {
+  data() {
```
abea, line 16: This all should be in the `AposEditorMixin` file since (almost) only editor modals need it. The media manager modal is the exception, which should be served via a Vue Composition API-type approach or some temporary code duplication until we're using VCA.

modules/@apostrophecms/modal/ui/apos/mixins/AposModalParentMixin.js
```diff
@@ -7,15 +7,37 @@ parent components should initiate that transition on mount.
...
+  },
   methods: {
     async cancel() {
-      if (this.editing) {
-        // If the modal is a manager with an open editor, close the editor and
-        // keep the manager open.
-        await apos.notify('Finish or cancel editing before closing');
-        return;
+      if (this.modified) {
+        const discard = await apos.confirm({
```
abea, line 28: 😚👌

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
```diff
@@ -104,7 +105,9 @@ export default {
       next: klona(this.items),
       triggerValidation: false,
       minError: false,
-      maxError: false
+      maxError: false,
+      editing: false,
```
abea, line 109: We don't seem to be using `editing` anymore.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
```diff
@@ -131,8 +131,6 @@ export default {
         negativeLabel: 'Resume Editing',
         affirmativeLabel: 'Discard Changes'
       },
-      confirmingDiscard: false,
```
abea, line 134: I think we can remove `confirmContent` as well, fwiw

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -146,14 +137,7 @@ export default {
         message: 'Uploaded media will appear here',
         emoji: '🖼'
       },
-      confirmingDiscard: false,
-      discardConfirmed: false,
```
abea, line 150: This is still part of a conditional on L255

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -97,15 +97,6 @@
           />
         </div>
       </AposModalRail>
-      <portal to="modal-target">
```
abea, line 100: 🔥

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -282,8 +264,12 @@ export default {
         this.editing = this.items.find(item => item._id === id);
         this.editsInProgress = false;
```
abea, line 265: This and the declaration on L 262 should also use `updateEditsInProgress`

modules/@apostrophecms/piece-type/ui/apos/components/AposRelationshipEditor.vue
```diff
@@ -79,13 +80,16 @@ export default {
...
     this.modal.active = true;
     this.docReady = true;
-    this.doc.data = this.value || {};
   },
   methods: {
     async submit() {
-      this.$emit('input', this.doc.data);
-      this.cancel();
+      this.$emit('input', this.docInfo.data);
+      this.modal.showModal = false;
```
abea, line 87: This should go back to `cancel` when the other logic is moved into the `AposEditorMixin` file.

abea (changes requested):
The big

modules/@apostrophecms/modal/ui/apos/mixins/AposModalParentMixin.js
boutell: Modals have other features that not all modals need. I don't think we should duplicate code or push it back to VCA over a few lines. How about it stays where it is for now, but migrates to VCA at a future date?

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
boutell: Removed.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
boutell: Actually the purpose is to distinguish between updates coming from AposSchemas (or otherwise from the user) and those not (which would trigger false updates of `modified`). That's why I needed the event handler and couldn't just `watch`.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
boutell: Thanks.

modules/@apostrophecms/modal/ui/apos/mixins/AposModalParentMixin.js
abea: I'm not saying it should be in VCA now. But this is an editor thing, so this should all be in the editor mixin. The `cancel` method should maybe look for a method to call that is in the editor mixin wrapping all the logic added in this file currently.

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
abea: We should remove that here.

boutell: Oops pushed that fix up for real this time.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
abea: `editsInProgress` is semantically the same as `modified` and served that purpose in the initial implementation of the confirmation modal. `editsInProgress` should probably be retired with this PR.

boutell:
Think I've got the right fixes and simplifications here, even for the media manager, at least in my tests... 🙏

abea (approved):

PR #2499, opened 2020-10-22
Array editor per design

boutell:


abea (changes requested):
- I have an added array item, but the top level input still says "0 Added"
- When creating a new array item after the first, or when switching from a populated item to an empty/new one, I get console errors from the string field (assuming a string field), `Error in callback for watcher "next": "TypeError: can't access property "length", value is undefined"`
- Assuming an array minimum of 1, if I open the array editor and try to save, I get an error (👍). If I then create an array item or items, the error persists and I can't save the array.

boutell:
I don't think any of these issues are new, exactly, except arguably the first one, but this is as good a PR as any for you to stop me and get me to fix them on. 😆

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
```diff
@@ -329,4 +363,30 @@ export default {
...
+    margin: 10px;
+  }
+
+  .apos-modal-array-items__heading {
+    position: relative;
+    margin: 20px 4px;
+  }
+
+  // Specificity needed due to AposButton rules
+  .apos-modal-array-items__heading .apos-modal-array-items__add {
```
abea, line 377: Small thing, but since it's an `.apos-button` style that's conflicting, I might do this as `.apos-modal-array-items__add.apos-button` instead to communicate that relationship a bit more.

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
```diff
@@ -329,4 +363,30 @@ export default {
...
+    position: absolute;
+    top: -5px;
+    right: 0;
+  }
+
+  .apos-modal-array-items__label {
+    // Consistent with appearance of same information in input field
+    letter-spacing: 1.5px;
```
abea, line 385: I don't like that we have so many "bespoke" text styles, but this didn't create that problem, so proceed...

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -4,20 +4,18 @@
...
             'is-disabled': field.disabled
           }"
         >
-          <p class="apos-array-count">
-            {{ next.length }} Items
-          </p>
-          <button
+          <AposButton
+            :label="editLabel"
+            :modifiers="[]"
```
abea, line 16: we don't need this if it's empty.

abea (changes requested):

abea:
> I don't think any of these issues are new, exactly, except arguably the first one, but this is as good a PR as any for you to stop me and get me to fix them on. 😆

Yeah, I thought on that for a minute. I figured since you know this better than anyone and this a sort of wrap up stage I'd throw them out.

boutell:
All should be tidied up.

abea (approved):

PR #2500, opened 2020-10-22
3/area bugs

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-648, "Area Bugs"


boutell (approved):

boutell:
You could have used `areaModuleOptions` in case it were mistaken for the options of the individual area, but mu.

abea:
I wasn't making that connection. I reverted and added a note for us slow folks.

boutell (approved):

PR #2502, opened 2020-10-22
3/global admin

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-666, making the global document available in an editor from the admin bar (unless it has no custom schema).

This also sets up all piece types to be opened directly in an editor as well.


modules/@apostrophecms/admin-bar/index.js
```diff
@@ -84,8 +85,10 @@ module.exports = {
 
       add(name, label, permission, options) {
         let index;
+
         const item = {
-          name: name,
+          name: name.indexOf(':') === -1 ? name : name.split(':')[0],
```
abea, line 90: Maybe the argument in `add` should be `action` instead of `name`?

modules/@apostrophecms/admin-bar/index.js
```diff
@@ -100,17 +103,6 @@ module.exports = {
...
       },
 
-      // Group several menu items together in the interface (currently
-      // implemented as a dropdown menu). If `items` is an array of menu
-      // item names, then the group's label is the same as the label of
-      // the first item. If you wish the label to differ from the label
-      // of the first item, instead pass an object with a `label` property
-      // and an `items` property.
-
-      group(items) {
```
abea, line 110: I could not find this used. Items group in the menu and tests pass without it.

modules/@apostrophecms/admin-bar/index.js
```diff
@@ -128,7 +120,7 @@ module.exports = {
         // admin bar items with menus.
         const groupedItems = [];
         let menu = false;
-        _.each(items, function (item, i) {
+        items.forEach(function (item, i) {
```
abea, line 123: Slowly reducing reliance on lodash.

modules/@apostrophecms/admin-bar/index.js
```diff
@@ -196,17 +188,21 @@ module.exports = {
...
-        // `menuLeader` property.
-        _.each(self.options.groups || self.groups.concat(self.options.addGroups || []), function (group) {
-          if (group.label) {
-            self.groupLabels[group.items[0]] = group.label;
-            group = group.items;
+        // Implement the groups and addGroups options. Mark the grouped items
+        // with a `menuLeader` property.
+        const groups = self.options.groups || self.groups.concat(self.options.addGroups || []);
+
+        groups.forEach(function (group) {
```
abea, line 195: This ended up not being critical to my grouping problems, but setting `group = group.items` was confusing and seemed like odd obfuscation.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -100,11 +103,13 @@ export default {
...
     this.$refs.spacer.style.height = `${this.$refs.adminBar.offsetHeight}px`;
-    this.menuItems = this.items.map(item => {
+    const itemsSet = this.filterEmptySchema(klona(this.items));
+
+    this.menuItems = itemsSet.map(item => {
       if (item.items) {
         item.items.forEach(subitem => {
           // The context menu needs an `action` property to emit.
-          subitem.action = subitem.name;
+          subitem.action = subitem.action || subitem.name;
```
abea, line 112: we're actually setting an action property now.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -129,6 +134,25 @@ export default {
   methods: {
     emitEvent: function (name) {
       apos.bus.$emit('admin-menu-click', name);
+    },
+    filterEmptySchema(menu) {
```
abea, line 138: The global doc has no editable schema by default. It's unlikely but possible for other modules to be in the same situation. There could be some edge case where someone wants a manager for a type that you can't actually edit... maybe. But is there harm in not supporting that by default?

abea:
Merge #2503 into this before merging into 3.0 branch.

modules/@apostrophecms/admin-bar/index.js
boutell: `action` could get confused with the action URL of a module. I kinda like name.

boutell: But I'm flexible on it.

modules/@apostrophecms/admin-bar/index.js
boutell: The workflow module uses it to programmatically create a reasonable grouping of its options. Without this there isn't a good way to do that in a module that doesn't "own" top level configuration.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
boutell: I'm OK with it if you want to Action All The Things.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
boutell: I would go so far as to say it will not happen, but not so far as to say you're wrong for preparing for it.

boutell (changes requested):

modules/@apostrophecms/admin-bar/index.js
boutell: It's possible this should be left in place for now, but a beta ticket opened to think more about merging admin bar configuration versus the current "if the dev sets it up workflow's suggestions are ignored anyway" situation.

modules/@apostrophecms/admin-bar/index.js
abea: yeah, that's true. The main case for it here is that it's going to get passed down into AposContextMenu components, which emit the items' `action` as well. There is potential for confusion, though.

modules/@apostrophecms/admin-bar/index.js
abea: Gotcha. I'll create a ticket in the workflow epic related to this.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
abea: I'm going to go about this differently instead.

modules/@apostrophecms/global/index.js
```diff
@@ -122,7 +122,14 @@ module.exports = {
       // to set the permission requirement to admin-@apostrophecms/global.
       // This is called for you.
       addToAdminBar() {
-        self.apos.adminBar.add(`${self.__meta.name}:editor`, self.pluralLabel, 'admin-' + self.name);
+        console.info('👚', self.schema);
```
boutell, line 125: Ready to remove this?

boutell (changes requested):

modules/@apostrophecms/global/index.js
abea: I guess...

boutell (dismissed):

abea:
Fixed a test since the global item is not there without a custom schema.

boutell (approved):

PR #2503, opened 2020-10-22
Sets up quick create menu in admin menu

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-669, populating the quick-create menu with real piece types.

This merges into the global admin menu item branch (#2502) since it uses the piece editor actions set up there.

### DIFF
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -5,7 +5,8 @@ module.exports = {
   cascades: [ 'filters', 'columns', 'batchOperations' ],
   options: {
     manageViews: [ 'list' ],
-    perPage: 10
+    perPage: 10,
+    quickCreate: true
     // By default there is no public REST API, but you can configure a
     // projection to enable one:
     // publicApiProjection: {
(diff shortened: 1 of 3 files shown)


boutell (approved):

abea:
Gah, I forgot to hold the other PR for this. Merging into 3.0 instead.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -683,6 +683,7 @@ module.exports = {
         browserOptions.contextual = self.contextual;
         browserOptions.batchOperations = self.batchOperations;
         browserOptions.insertViaUpload = self.options.insertViaUpload;
+        browserOptions.quickCreate = true;
```
boutell, line 686: I was thinking more like actually setting:

```
options: {
  quickCreate: true
}
```

And here where you build browserOptions, pulling that in as self.options.quickCreate.

Similarly global can shut it off cleanly via configuration.

boutell: This way devs can opt out for piece types that are not so primary.

boutell (changes requested):

boutell (approved):

PR #2504, opened 2020-10-22
never show a "confirm before you cancel" button if we do not have to

boutell:


modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -146,6 +149,32 @@ export default {
...
+        return false;
+      }
+      if (this.chosen.length !== this.checkedDocs.length) {
+        return true;
+      }
+      for (let i = 0; (i < this.chosen.length); i++) {
+        if (this.chosen[i]._id !== this.checkedDocs[i]._id) {
+          return true;
+        }
+        if (this.relationshipField.schema) {
```
boutell, line 172: Currently you can't edit the relationship in our manager-as-chooser modals, but this could change, so this is good futureproofing.

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
```diff
@@ -0,0 +1,23 @@
...
+
+import { isEqual } from 'lodash';
+
+export function detectDocChange(schema, v1, v2) {
+  return schema.some(field => {
+    return detectFieldChange(field, v1[field.name], v2[field.name]);
+  });
+}
+
+export function detectFieldChange(field, v1, v2) {
```
boutell, line 13: Right now we don't look at field here but this is very much subject to change if we find this simple algorithm in adequate to the needs of some field types. Futureproofing.

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -146,6 +149,32 @@ export default {
     saveRelationship() {
       this.$emit('modal-result', this.checkedDocs);
       this.modal.showModal = false;
+    },
+    // Default implementation of isModified is based on whether the
+    // selection has changed, but you can override this and combine
+    // that bit with your own if your manager allows in-context editing
+    // of a piece (i.e. AposMediaManager)
+    isModified() {
```
abea, line 157: I don't see this actually used anywhere. The only manager using `isModified` is the media manager, which has its own implementation. Should this be checked in `saveRelationship`?

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -262,10 +272,9 @@ export default {
         }
       }
       this.editing = this.items.find(item => item._id === id);
-      this.modified = false;
     },
     edited(value) {
```
abea, line 276: It doesn't look like this method is used anymore.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
```diff
@@ -161,16 +165,13 @@ export default {
...
-          if (
-            Object.keys(oldData).length > 0 &&
-            Object.keys(newData).length > 0
-          ) {
-            this.updateDocEdited(true);
+          // If either old or new state are an empty object, it's not "modified."
+          if (!(Object.keys(oldData).length > 0 && Object.keys(newData).length > 0)) {
+            this.$emit('modified', false);
           }
+          this.$emit('modified', detectDocChange(this.schema, this.original, newData));
```
abea, line 174: This should be in an `else` block since we might have just emitted `false`.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -326,9 +325,8 @@ export default {
     },
     async loadNewInstance () {
       this.docReady = false;
-
       const newInstance = await this.getNewInstance();
-
+      this.original = newInstance;
```
abea, line 329: This could stay as `null` for the sake of quicker evaluation. I might suggest that to save an unnecessary cycle or two.

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
```diff
@@ -109,6 +110,7 @@ export default {
       // permanent modifications whether the user
       // clicks save or not
       next: klona(this.items),
+      original: klona(this.items),
```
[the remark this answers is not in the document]
abea: I see it's done everywhere else as well for safety sake. Objection retracted.

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
```diff
@@ -236,6 +235,25 @@ export default {
       const currentIndex = this.next.findIndex(item => item._id === this.currentId);
       this.next[currentIndex] = this.currentDoc.data;
     },
+    isModified() {
```
abea, line 238: I don't see a `return false` case here.

abea (changes requested):

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
boutell:
```
    async confirmAndCancel() {
      let dismiss;
      if (this.isModified()) {
```

In AposModalModifiedMixin.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
boutell: ? updateEditing is still called. The modified flag is consulted.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
boutell: This allows us to determine whether actual effort has been put in to change it from the defaults or not. If we set it null then it is considered never to have changed significantly and the user loses work they may have put in.

modules/@apostrophecms/schema/ui/apos/components/AposArrayEditor.vue
boutell: undefined is falsy, but clarity is good.

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
abea: True. I wasn't seeing them in concert in my head.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
abea: Right. The `edited` method isn't, though.

abea (changes requested):
I think there's just the deprecated `edited` method.

abea (dismissed):

boutell:
Merge changes require review according to github.

abea (approved):

PR #2505, opened 2020-10-23
when using the new fields.group property, the fields should wind up i…

boutell:
…n the order configured in their groups in the schema, even if groups are not actually displayed (media manager)


boutell:
I also put the fields for users into some reasonable groups as part of my testing. But mainly it's about the media manager fields being in the expected order.

abea (approved):

PR #2508, opened 2020-10-26
Basic permissions

boutell:
* Just for 3.0 alpha 1, a logged-in user can do anything. The public of course can only see public content. The full 3.x permissions model is coming in beta 1 if not sooner.
* Published field removed in favor of the visibility field (formerly known as loginRequired), which is sufficient when combined with draft/current/publish workflow, to be standard in 3.x beta.
* `apos.permission.can` Permissions API no longer requires building permissions names as strings and splitting them constantly on every request to separate the type name from the action. Instead an object with action and type properties is passed.
* `apos.permission.criteria` no longer takes a type argument, as it is possible to determine which documents a user can carry out a particular action on purely through appropriate mongo criteria. Limiting the query to a type is the province of the query builder in question and unrelated to permissions themselves.
* Also removed server side notifications on all inserts and updates as this was more of a test of the notifications feature than an actual UX-considered feature.


abea (changes requested):
- It seems like parked pages are still getting `published: true` properties on their docs.
- A few components still have `published` artifacts: the page manager, `AposTreeRows` (page manager adjacent), and `AposDocEditor` in the `filterValues`.

boutell:
* `published: true` on parked pages: this was coming, and is still coming, from a3-demo. That should be removed but it's not an issue with the PR.
* leftover refs to published in the code: thanks, fixed these.

boutell:
I will clean up parked pages in a3-demo after acceptance.

boutell:
Going to add a migration to this PR. We can remove it before shipping since third parties won't have the issue.

boutell:
@abea OK, this is ready for re-review, I added a migration so no more blowing up databases to cross this bridge. You don't have to run the migration task manually. In a dev environment it just runs at startup (we did the same in late A2).

We can remove this particular migration before shipping but writing it revealed bugs so I have no regrets.

abea (approved):

PR #2509, opened 2020-10-26
3/img previews

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-671, fixing the image preview in the image relationship input field slats. Also fixing the thumbnail in the right rail of the media library (image selections).

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -76,8 +76,28 @@ export default {
       return apos.modules[this.field.withType].components.managerModal;
     }
   },
-  mounted() {
+  async mounted() {
     this.validateAndEmit();
+
+    // Check if this is joined to an image.
+    if (this.field.withType === '@apostrophecms/image' && this.value.data && this.value.data.length > 0) {
+      // If it is, go get the URLs and populate them on `value`.
+      let i = 0;
+      for (const image of this.value.data) {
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -76,8 +76,28 @@ export default {
...
-  mounted() {
+  async mounted() {
     this.validateAndEmit();
+
+    // Check if this is joined to an image.
+    if (this.field.withType === '@apostrophecms/image' && this.value.data && this.value.data.length > 0) {
+      // If it is, go get the URLs and populate them on `value`.
+      let i = 0;
+      for (const image of this.value.data) {
+        if (image.attachment._urls) {
```
boutell, line 87: This should always already be present in whatever API response populates the join in the first place. A new fetch should not be necessary, are you sure the URLs are not populating?

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
abea: It isn't present because the image object that's already here is what comes directly from the database. _urls is not populated there.

boutell: It should. Hmm. Let me peep.

boutell (approved):

PR #2510, opened 2020-10-26
3/fix widget control focus

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-683, "Focus is being stolen by other widgets"

In a two column widget, you should now have no trouble accessing widget controls for Widget A with your mouse that happen to appear on top of Widget B. All other widget control operations should work as they have.


abea:
I recommend reviewing the first commit on its own. Those are the meaningful changes. The CSS in this file had repetition and inconsistent organization, so I cleaned up a bit of that in the second commit.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -428,61 +428,69 @@ export default {
...
-      z-index: $z-index-widget-highlight;
       &:before, &:after {
         opacity: 0.4;
       }
     }
     &.apos-focus {
-      z-index: $z-index-widget-focus;
       &:before, &:after {
         opacity: 1;
         border-top: 1px solid var(--a-primary);
```
boutell, line 446: Not an issue to hold this PR for, but have we bogarted the entire "a" namespace for CSS variables? Or is it just a SASS thing?

boutell (approved):
It does feel better.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
abea: we're bogarting it.

PR #2514, opened 2020-10-27
3/fix area menu

abea:
sort of resolves https://apostrophecms.atlassian.net/browse/A30U-672, but really addresses other widget menu issues. I think the original issue is mostly resolved already.

1. There was a gap between the menu and the arrow. If moving the mouse between the two triggered another widget to take focus, you couldn't get to the menu. I added an invisible box in the SVG to help with this.
2. Resolves https://apostrophecms.atlassian.net/browse/A30U-698. The context menu arrows are now always set by the actual v-popover `x-placement` attribute.


modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -131,6 +131,7 @@ export default {
       }
       return {
         menuPlacement: 'bottom',
+        menuOffset: 15,
```
stuartromanek, line 134: Should this offset always be fixed at 15? if so it should come after the instance level options

modules/@apostrophecms/ui/ui/apos/components/AposContextMenuTip.vue
```diff
@@ -52,14 +61,14 @@ export default {
...
   }
 
-  [x-placement^='bottom'] > .apos-context-menu__tip {
-    top: -11px;
+  [x-placement^='bottom'] .apos-context-menu__tip {
+    top: -27px;
     bottom: auto;
   }
 
-  [x-placement^='top'] > .apos-context-menu__tip {
```
stuartromanek, line 60: Did you find the case where these tooltips are created inside one another? Thats why we're using the direct child selector.

abea: jesus. where's that?

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
abea: I'm doing it it here because we're setting pixel styles for it. 15 is already the default, so I'm restating that here to make sure it's not overridden to the point where it no longer matches the offset style.

modules/@apostrophecms/ui/ui/apos/components/AposContextMenuTip.vue
abea: I don't see any cases of that. The rich text linking modal doesn't have it, which is where we suspected it might be. We could maintain this specificity with a few more levels of `> parentEl`, but we don't want to be trying to pass some direction value down to recreate the attribute.

stuartromanek: i think we should put the direct child selectors back for now even though we can't find the nested example .. we never want to reach down the tree to apply these rules, they are only relevant in their immediate context

stuartromanek (changes requested):

modules/@apostrophecms/ui/ui/apos/components/AposContextMenuTip.vue
abea: done

PR #2515, opened 2020-10-27
Apos logo too

localghost443:


stuartromanek (changes requested):
- Get the admin bar logo to 45px wide and kill the right margin it creates (or however you'd rather implement that size change)

- Constrain Confirm logo to 40px wide and give it 10px margin bottom

- Should the logo component take a width prop for these different cases? Fielder's choice

modules/@apostrophecms/modal/ui/apos/components/AposModalConfirm.vue
```diff
@@ -120,9 +120,9 @@ export default {
   padding: 60px;
 }
 
-.apos-confirm__icon {
-  width: 60px;
+.apos-confirm__logo {
   height: 60px;
+  fill: var(--a-black);
```
abea, line 125: This feels... bad since it's assigning a fill meant for a specific SVG to any logo that's passed in on the `img` tag. `fill` won't have any impact there, so I guess it's okay.

abea (approved):

PR #2516, opened 2020-10-28
Drag to trash

boutell:
* Fixed bugs that could cause duplicate page ranks.
* Wrote a migration to fix that in the case of the trash page. (It is unlikely that any other instances exist yet as it has to do specifically with pages marked as "orphans," i.e. excluded from navigation.)
* Dragging to and from the trash works.
* You get a good error message if you try to drag the trash can itself.
* Visit links to pages in the trash are disabled, as they would 404, and those links are never seen, as they are weird and would 404.
* Enforce that the trash can must be the last child of the home page.
* Politely rejigger attempts to drop after (not in) the trash can so they drop just before it, avoiding a frustrating error message.
* Don't throw an error on the link tree cell if the _url is not present (page in trash).
* Disabled appearance for link icon when page is in trash.
* Projection for efficient loading of large page trees.

I thought about adding an icon to pages in the trash and the trash can itself, but decided that is probably not something I should be cowboying and it should be a separate ticket. There is probably already a ticket for page types to have icons in the page tree.


abea:
It looks like we have a failing test here.

boutell:
@abea this is ready for reexamination, see new commit. All tests pass and the UI still works.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -34,6 +34,7 @@
           :href="col.type === 'link' ? row[col.name] : false"
           :target="col.type === 'link' ? '_blank' : false"
           :class="getCellClasses(col, row)"
+          :disabled="getCellDisabled(col, row)"
```
abea, line 37: We should totally hide/remove the trash cells. The edit button is still active as well, for one. They should be gone-gone.

abea: We have a way to disable dragging already as well. See the line: `v-if="options.draggable && index === 0 && !row.parked"`.  Using the `parked` property would also set the `is-parked` class on it (L266), which tells vue-draggable to lock it down as well. So putting `parked: true` on the trash "page" would be the quickest way to prevent dragging. If you don't want to do that, those conditionals could also including looking for the `@apos/trash` type.

abea (changes requested):

boutell:
Well that was more difficult than I anticipated. Back to you. 🏸

abea (changes requested):
One big thing and one smaller thing that could be resolved via the big thing.
- Even with the trash's "edit" button visibly gone and set `disabled`, clicking it still opens the editor modal. That's the event handler there in the AposTreeRows file. We could add a condition to this line to preclude `disabled` cells. `@click="col.action ? $emit(col.action, row._id) : null"` Not necessary for disabled buttons, but this is a span, so there's no such built-in affordance.
- You get the `cursor: not-allowed` cursor on hover at the end of the trash row even though you don't see anything you're hovering over. This is what leads me to think that we shouldn't be putting `disabled` on them at all. The link is already gone: ✔️. The button is gone as well, so once the click handler is removed, there's no need to tell them they are disabled.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
abea: This is still a factor @boutell

abea (changes requested):

abea:
I commented that on the wrong thing. But some of my previous list stands.

abea (approved):
Nevermind. Let's make sure tests pass by merging the main 3.0 branch in.

boutell:
I ran the tests successfully after merging in 3.0.

PR #2517, opened 2020-10-28
Avatar menu works, avatar gradient is chosen via user's _id

boutell:
I'm choosing among the gradients Stu provided via a little math on the _id, which is fun, but of course there's no guarantee two users of interest won't happen to have the same one.

The "My Profile" and "Log Out" buttons here are hooked up now, so I removed "Log Out" from the regular admin bar.

The context menu looks a little weird here, the pointer at top kinda misses the mark, Stu maybe there's a way to adjust that?


boutell:
Closing temporarily because it is inadvertently based on another PR which should be approved soon anyway and then I'll reopen.

PR #2518, opened 2020-10-28
Avatar menu works (correct PR this time with no extraneous guff)

boutell:
I'm choosing among the gradients Stu provided via a little math on the _id, which is fun, but of course there's no guarantee two users of interest won't happen to have the same one.

The "My Profile" and "Log Out" buttons here are hooked up now, so I removed "Log Out" from the regular admin bar.

The context menu looks a little weird here, the pointer at top kinda misses the mark, Stu maybe there's a way to adjust that?

There might be a better way to do the vertical centering of the initials here.

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBarUser.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBarUser.vue
@@ -5,12 +5,12 @@
     :menu="menu"
     menu-placement="bottom-start"
     :menu-offset="2"
+    @item-clicked="emitEvent"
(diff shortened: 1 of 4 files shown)


stuartromanek (changes requested):

stuartromanek:
@boutell Missed your question about the tip not being on target .. we have props that can provide a vertical offset but I think we lack ones that push it horizontally .. we should equip context menus with that .

boutell:
@stuartromanek cleaned this up, am I a CSS expert now?

localghost443 (approved):
love and light. Looks good to me.

PR #2523, opened 2020-10-29
Page picker picks pages passably

boutell:
You can pick pages both via autocomplete in the relationship field and via browsing the page tree.

I also fixed a bug in the implementation of the "Global" button so that you can edit the singular, existing global preferences doc now. previously it was trying to make a new one every time.

See corresponding a3-demo PR, which introduces a simple set of footer links via the global doc.

### DIFF
--- a/modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
+++ b/modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
@@ -6,6 +6,12 @@
   >
     <template #secondaryControls>
       <AposButton
+        v-if="relationshipField"
+        type="default" label="Cancel"
+        @click="confirmAndCancel"
+      />
+      <AposButton
+        v-else
         type="default" label="Finished"
         @click="confirmAndCancel"
       />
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -241,6 +258,27 @@ export default {
         this.checkedDocs.push(doc);
         this.checked.push(doc._id);
       }
+    },
+    updateCheckedDocs() {
+      this.checkedDocs = this.checked.map(_id => find([ this.originalPageTree ], item => {
```
abea, line 263: It looks like we could simply put the unadulterated page object into `pagesFlat` and then use that instead. No need for the additional `find` method.

abea (commented):
I'm hesitant to add another version of the page tree. Check out my pagesFlat suggestion.

abea:
> Page picker picks pages passably
You had fun with this.

abea (approved):

PR #2530, opened 2020-10-30
Fixes for various bugs with media selection

boutell:
* Disable uploads when max is reached. TODO: the button's appearance should change. I messed around with CSS filter but couldn't see an easy way to get a good PoC of this in place. I have opened a separate ticket to properly design this state. For now we just have the functional disabling.
* When switching images in the inline editor, the confirmation prompt should only appear if the editor is modified, the selection is not relevant to that prompt.
* Don't let the media manager save a selection with any relationship errors. Not just min.
* Fix the max error so it is not an "error" if you're just *at* the maximum. Provide a separate `maxReached` mechanism for the cases where that was important. I wanted this to be "computed" but the child components didn't respond to its changes that way, so it's a method.
* If an image is being edited and the user deselects it, confirm before clearing the editor. If they cancel, re-select the item.
* Allow the user to upload the same file twice in a row in the media manager. Note: this wasn't about avoiding duplicates, it was a browser behavior ("same file? I'm not going to emit an event") that provided zero feedback and made it feel broken in a test that users are likely to conduct during evaluation. I was confused myself.

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
@@ -1,6 +1,7 @@
 <template>
   <label
     :class="dropzoneClasses"
+    :disabled="disabled"
     @drop.prevent="uploadMedia"
     @dragover.prevent="dragover = true"
     @dragleave="dragover = false"
@@ -23,9 +24,10 @@
     </div>
     <input
       type="file" class="apos-sr-only"
-      ref="apos-upload-input"
+      ref="upload"
       @input="uploadMedia"
       multiple="true"
+      :disabled="disabled"
     >
   </label>
 </template>
@@ -37,6 +39,10 @@ export default {
     action: {
       type: String,
       required: true
+    },
+    disabled: {
+      type: Boolean,
+      required: false
     }
   },
   emits: [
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
```diff
@@ -183,7 +199,9 @@ export default {
     grid-column: 1 / 3;
     grid-row: 1 / 3;
     @include apos-transition();
-
+    color: inherit;
```
abea, line 202: The disabled version should also give the `cursor: not-allowed` treatment.

abea:
I don't know what you mean by the "inline editor." The steps to reproduce?

> When switching images in the inline editor, the confirmation prompt should only appear if the editor is modified, the selection is not relevant to that prompt.

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -88,6 +88,14 @@ export default {
     }
   },
   methods: {
+    // It would have been nice for this to be computed, however
+    // AposMediaManagerDisplay does not re-render when it is
+    // a computed prop rather than a method call in the template.
+    maxReached() {
+      // Reaching max and exceeding it are different things
+      const result = this.relationshipField.max && this.checked.length >= this.relationshipField.max;
```
abea, line 96: This is odd as a method, I agree. If the issue is the display view not updating, we could have a watch listener in there? But I'm not too offended by this.

abea (changes requested):
Can you provide steps to test this? "Allow the user to upload the same file twice in a row in the media manager. "

boutell:
"I don't know what you mean by the "inline editor." The steps to reproduce?"

1. Start editing an existing media widget (existing selection).
2. Click "Edit" on the image (at right). But don't change anything.
3. Deselect the image. Without this PR, you would be prompted for confirmation, even though you're just closing the in-modal editor without changes, you're not discarding the selection (yet).

"Allow the user to upload the same file twice in a row in the media manager. Steps to reproduce?"

Literally do that. Click the upload button, upload an image, see it appear in the media manager. Then deselect it and immediately upload that same exact file again from the same path on your drive. This should work, because people will test with it and we don't have a proper "hey you already have that" feature, nor is it necessarily good in all cases (that's a rabbit hole, but the point is it should be intentional). But without this PR it does not work because it is the same input element of type file and they don't fire a change event if their value has not changed. That makes sense with the native UI but not with this UI.

"The disabled version should also give the cursor: not-allowed treatment."

I tried that, but it just plain didn't work, and so I backed it out as part of my general handing off of making it actually look right to a new ticket. Alpha 1 time!

PR #2536, opened 2020-10-31
handle field types with no ui (reverse relationships) properly

boutell:
We've been seeing errors that boil down to the fact that `relationshipReverse` fields have no UI.

That's intentional but our frontend needed to know how to deal with it. Now the schemas are filtered appropriately and we don't get console errors and empty tabs.

(The possibility of making relationshipReverse an editable field type has lots of ins and outs and pros and cons and wouldn't be in-scope right now, this PR is just about un-breaking things.)

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -141,7 +141,7 @@ export default {
       return (window.apos.modules[this.moduleName] || {}).action;
     },
     schema() {
-      return this.moduleOptions.schema || [];
+      return (this.moduleOptions.schema || []).filter(field => apos.schema.components.fields[field.type]);
     },
     groups() {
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -141,7 +141,7 @@ export default {
       return (window.apos.modules[this.moduleName] || {}).action;
     },
     schema() {
-      return this.moduleOptions.schema || [];
+      return (this.moduleOptions.schema || []).filter(field => apos.schema.components.fields[field.type]);
```
abea, line 144: Since it's done in so many places, maybe this should be done server-side in the pieces module before browserOptions are set up instead?

abea: okay...

abea (approved):

PR #2542, opened 2020-11-03
3/global busy

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-305, As a user I can see that Apostrophe is busy, and I am blocked from attempting operations that might fail while it is busy

You can trigger by making API calls via the UI, or (even better), copy/paste an admin menu button and have it trigger this method:

```
,
    lookBusy() {
      apos.busy.$emit('apos-busy', {
        active: true,
        name: 'busy'
      });

      setTimeout(function () {
        apos.busy.$emit('apos-busy', {
          active: false,
          name: 'busy'
        }, 5000);
      });
    }`
```

### DIFF
--- a/modules/@apostrophecms/busy/index.js
+++ b/modules/@apostrophecms/busy/index.js
@@ -12,7 +12,7 @@ module.exports = {
       getBrowserData(req) {
         return {
           busy: self.busy,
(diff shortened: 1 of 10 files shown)


boutell (approved):
Thank you for cleaning up so much vestigial nonsense the global busy works well for now

PR #2543, opened 2020-11-03
3/busy media

abea:
The media manager upload is the primary long-running busy operation we have, but as it was the busy state would simply flip on and off. I added a general event for it and implemented a counter.


modules/@apostrophecms/image/ui/apos/components/AposMediaUploader.vue
```diff
@@ -82,6 +82,11 @@ export default {
     },
     async uploadMedia (event) {
       try {
+        apos.bus.$emit('apos-busy', {
```
boutell, line 85: With `busy: true` on the http call I would be surprised if you were spending significant time not already marked busy for you by that.

boutell (approved):

PR #2545, opened 2020-11-04
3/design fixes

stuartromanek:
- Admin bar: buttons use AposButton component
- Admin bar: various size and style adjustments
- Areas: in a singleton scenario, try to put widget's icon in Add button
- Areas: Remove empty state component, use button
- Areas: Remove empty state borders
- Image widget: add default icon for area ui
- Modal Tabs: style adjust
- Notifications: better v align content center
- Pieces manager: add keyboard shortcut for creating a New piece from manager modal
- Area Input: proper wrapper for schema styles
- Time string: hide default browser icon
- Context menus: v-tooltip allows you to express a popovers offset as a comma separated X,Y string, account for it (probably more context menus to zhuzh in this way)
- Context menus: Remove dead modifier
- Slats: better v align content
- Area input: normalize width with other fields


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -269,11 +261,12 @@ $admin-bar-border: 1px solid var(--a-base-9);
...
 
 .apos-admin-bar__create {
-  margin-left: 20px;
+  margin-left: 10px;
   // Adjust button padding and svg size to have a large plus icon while keeping
   // the button size the same.
   /deep/ .apos-context-menu__btn {
-    padding: 4px;
+    padding: 2px;
+    border: 0;
```
abea, line 269: If I find something else, then this would be better as `border-width: 0` imo. If I don't, it's fine to leave.

abea: Reason being that it sets the important property directly rather than re-delcaring all the border property default values.

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -5,6 +5,7 @@
         :is="iconComponent"
         :decorative="true"
         :size="icon ? 16 : 12"
+        class="apos-notification__icon"
```
abea, line 8: You didn't end up using this.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -284,6 +289,25 @@ export default {
       this.currentPage = 1;
 
       this.getPieces();
+    },
+
+    shortcutNew(event) {
+      const interesting = (event.keyCode === 78 || event.keyCode === 67); // C(reate) or N(ew)
```
abea, line 295: We need the keyboard shortcut to show the keyboard shortcuts. I imagine this will be useful to wrap up in shared code later.

modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
```diff
@@ -31,14 +31,16 @@
           :button="more.button"
           :menu="more.menu"
           @item-clicked="$emit('item-clicked', item)"
+          menu-placement="bottom-start"
+          menu-offset="40, 10"
```
abea, line 35: Is this legal in v-tooltip? I'm only seeing reference to using single values.

abea (changes requested):

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
abea: This is the only definite one. Others are suggestion, comment, and question.

localghost443 (dismissed):

modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
stuartromanek: I don't think the documentation specifies exactly but the source treats it that way (and it does works)

abea (approved):

PR #2548, opened 2020-11-05
Fixes bug preventing area widget controls in editor modal

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-706, "Areas in modals's controls are being masked by the inner column of the modal, need to be accessible"

I tested this with the old canvas era style where the controls float outside the area and it works. This was the issue, not that the controls were visually hidden.


boutell (approved):
Good catch.

PR #2550, opened 2020-11-06
for multisite the asset build task must respect the separate npmRootD…

boutell:
…ir option so it can differ from project root


boutell:
apostrophe already supports these with npmRootDir automatically being set to rootDir if it was not passed, but the a3 asset build task didn't yet.

abea (commented):
Looks good.

boutell:
Cool, thanks for checking it.

PR #2552, opened 2020-11-09
3/a30u 733 save button

boutell:
* Your edits do not immediately go live when editing in context on the page.
* The "save" button starts out disabled (these are the current styles for that in our admin bar).
* when you save, it sends up all the patches in a single PATCH operation.
* If you try to close the page prematurely you get the built-in "gee do you want to do that" prompt from the browser. This is not a message that can be adjusted or substituted in modern browsers, we get what we get.
* That message stops appearing if the page has been saved.

We have lots of other pieces that need to come together here, like edit mode itself, but I saw no reason not to start with this.

Something NOT done in this PR: if you try to edit something that's not part of the current "context document" (page or piece that is the best fit for this URL), there is no visual indication yet and the UI is available to you, but it won't save it. There is a separate ticket to address it so the editing UI is not available and you get appropriate affordances to help you figure out where you should go to edit that particular thing. So I'm not worrying about that here, other than making sure I don't send up any patches for the wrong document.

### DIFF
--- a/modules/@apostrophecms/admin-bar/index.js
+++ b/modules/@apostrophecms/admin-bar/index.js
@@ -243,12 +243,15 @@ module.exports = {
           return false;
         }
         const closeDelay = self.options.closeDelay;
+        const context = req.data.piece || req.data.page;
+        const contextId = context && context._id;
         return {
           items: items,
           components: { the: 'TheAposAdminBar' },
           openOnLoad: !!(typeof self.options.openOnLoad === 'undefined' || self.options.openOnLoad),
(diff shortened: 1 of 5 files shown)


boutell:
This PR is not meant to address the optimal appearance of the button as we don't even have edit mode broken out yet.

abea:
@boutell There a bunch of failing tests in the REST API set.

abea (changes requested):
Please resolve the broken tests.

boutell:
Mutter mutter mutter stupid bug took me 30 minutes to figure out because I'm mutter mutter mutter.

Ready for another go.

abea (approved):

PR #2555, opened 2020-11-10
area edits now save properly in piece modals, including nested areas,…

boutell:
… without breaking the widget controls within the area. Comprehensive fixes to ensure docId is not passed where it should not be (in a modal) or when the parent is just another widget that also intentionally does not know docId (metatypes to the rescue). Make sure parent areas know about edits to child areas even inside modals, because even inside a modal a nested area does not have AposInputArea as a direct parent, so the normal changed event does not cover that situation.

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposWidgetControls.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposWidgetControls.vue
@@ -47,10 +47,6 @@ export default {
       default() {
         return {};
       }
-    },
-    item: {
-      type: Object,
-      default: null
     }
   },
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/area/ui/apos/components/AposWidgetControls.vue
```diff
@@ -47,6 +47,10 @@ export default {
       default() {
         return {};
       }
+    },
+    item: {
```
abea, line 51: It doesn't seem this is used. If not, we should remove it here and in `AposAreaWidget.vue` where it's passed in.

abea (dismissed):
It looks good to go once the `item` prop is removed.

PR #2560, opened 2020-11-10
Fixes login event handler and checks for existing user

abea:
resolves: https://apostrophecms.atlassian.net/browse/A30U-418, "At start up, check if there is a user"

This looks like it might have had an outdated event handler syntax. It didn't seem to be triggering in the login module.


boutell (approved):

boutell:
Nice catch

PR #2561, opened 2020-11-11
Disables quickCreate for images, files, and users

abea:
Removes low level piece types from the quick create menu.


boutell (changes requested):
How about file tags and image tags too, those are not really part of the "story of your website" either and I think they are confusing to have without the other two

boutell (approved):

PR #2562, opened 2020-11-11
Draft: link with just a div instead of bubble, demonstrates the bug is resol…

boutell:
…ved, although now you have to click the link icon to dismiss without using ok or cancel

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
@@ -9,6 +9,7 @@
       <AposContextMenuDialog
         menu-placement="top"
         class-list="apos-rich-text-toolbar"
+        :has-tip="false"
         :modifiers="['unpadded']"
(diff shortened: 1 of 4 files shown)


abea:
In all our discussion I lost track of the problem this is solving. Could you link to that or restate here?

boutell:
* Click on the link button
* Link dialog appears
* Dismiss it via cancel or via clicking elsewhere, ish
* Try to bring up the menu bar bubble menu again
* You can't

abea:
Yep, that looks good. I don't know if this was shot down, but using AposContextMenu seems like a fast way to get this implemented in a fairly nice looking way.

abea (commented):
Yep, that looks good. I don't know if this was shot down, but using AposContextMenu seems like a fast way to get this implemented in a fairly nice looking way.

stuartromanek:
Just committed a few UI updates.. the dialog is feeling pretty good.

one last way to break things specific to Link .. if you blur the rich text widget while the link dialog is open both the tiptap editor and link dialog remain on page while the widget itself visually unfocuses.

I can take a look at that tomorrow ..

stuartromanek (approved):

PR #2568, opened 2020-11-12
edit mode toggle, "publish changes" button, auto refresh of main content area on edits, gear icon is contextual to the context page or piece, "context bar" is roughed in (really roughed in)

boutell:
* Functioning edit mode toggle, remembered in session storage so it persists for the lifetime of the current tab
* Main content area refreshes when pieces and page settings change, this mechanism is also used in the implementation of edit mode
* Page always renders normally ("preview") unless apos-edit=1 is present in the query string
* parseQuery method to complement addQueryToUrl, did it this way because we want apos-edit and apos-refresh to be query parameters (because "custom headers should not change the response body") and because we'll want to be able to fully parse filters with nested objects/arrays from the query before too long (thinking ahead to how we'll address what used to be called data-apos-ajax and other things). This implementation understands everything addQueryToUrl can stringify and is much smaller than comparable implementations, might be worth releasing this pair as a module
* Gear icon is contextual to the right context document
* "Save" became "Publish Changes" as that is what it does until we have drafts

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -60,6 +60,20 @@
         </span>
         <span class="apos-admin-bar__context-controls">
           <AposButton
+            v-if="editMode"
+            type="default" label="Preview Mode"
+            icon="eye-icon" class="apos-admin-bar__btn"
+            :icon-only="true"
+            @click="switchToPreviewMode"
+          />
+          <AposButton
+            v-if="!editMode"
+            type="default" label="Edit Mode"
+            icon="pencil-icon" class="apos-admin-bar__btn"
+            :icon-only="true"
+            @click="switchToEditMode"
+          />
+          <AposButton
             v-if="moduleOptions.contextId"
             type="default" label="Page Settings"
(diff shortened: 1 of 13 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -91,7 +115,8 @@ export default {
     return {
       menuItems: [],
       createMenu: [],
-      patches: []
+      patches: [],
+      editMode: window.sessionStorage.getItem('aposEditMode') === 'true'
```
boutell, line 119: sessionStorage used to persist between pages for the lifetime of the tab per discussion with Stu and Ben.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -145,16 +191,46 @@ export default {
       }
     },
     emitEvent: function (name) {
+      console.log(name);
```
abea, line 194: stray log

boutell: Fixed

abea (approved):
One little suggestion, but this is really nice.

PR #2569, opened 2020-11-12
3/tooltips

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-724, "Implement tooltips"

Currently only implemented for buttons. Adding anywhere else only requires using the `v-tooltip` directive with either a string or [configuration object](https://github.com/Akryum/v-tooltip#object-notation).

tooltip design example

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposButton.stories.js
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposButton.stories.js
@@ -46,6 +46,9 @@ export const buttons = () => ({
     disabled: {
       default: boolean('Disabled', false)
     },
+    tooltip: {
+      default: boolean('Activate Tooltip (on hover)', false)
+    },
     busy: {
       default: boolean('Busy', false)
     },
@@ -66,8 +69,10 @@ export const buttons = () => ({
         )
     }
   },
+  // Margin added to make room for demonstrating tooltip.
   template: `
     <AposButton
+      style="margin: 100px;"
       :disabled="disabled"
       :label="label"
       :busy="busy"
@@ -75,6 +80,7 @@ export const buttons = () => ({
       :icon="icon"
       :iconOnly="iconOnly"
       v-bind:modifiers="modifiers"
+      :tooltip="tooltip ? 'Duis mollis, est<br />porttitor ligula.' : null"
     />
   `
 });
--- a/modules/@apostrophecms/ui/ui/apos/lib/tooltip.js
+++ b/modules/@apostrophecms/ui/ui/apos/lib/tooltip.js
@@ -1,0 +1,13 @@
+export default {
+  updateOptions (tooltipObject) {
+    tooltipObject.options.defaultOffset = '11';
+    tooltipObject.options.defaultPlacement = 'bottom-end';
+    tooltipObject.options.defaultTemplate = `
+      <div class="apos-tooltip" role="tooltip">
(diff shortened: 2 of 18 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposButton.stories.js
```diff
@@ -66,15 +69,18 @@ export const buttons = () => ({
...
     <AposButton
+      style="margin: 100px;"
       :disabled="disabled"
       :label="label"
       :busy="busy"
       :type="type"
       :icon="icon"
       :iconOnly="iconOnly"
       v-bind:modifiers="modifiers"
+      :tooltip="tooltip ? 'Duis mollis, est<br />porttitor ligula.' : null"
```
abea, line 83: Looks like we need to use a `br` tag for line breaks.

abea: It looks like we could alternatively set a max-width on the inner bit and v-tooltip will figure out positioning.

modules/@apostrophecms/ui/ui/apos/scss/global/_tooltips.scss
```diff
@@ -0,0 +1,86 @@
+// Styles based on the v-tooltip example styles:
+// https://github.com/Akryum/v-tooltip/blob/83615e394c96ca491a4df04b892ae87e833beb97/demo-src/src/App.vue#L179-L303
+.apos-tooltip {
+  z-index: $z-index-notifications;
+  display: block;
+
+  .apos-tooltip__inner {
+    @include type-large;
```
stuartromanek, line 8: use `type-small`

modules/@apostrophecms/ui/ui/apos/lib/tooltip.js
```diff
@@ -0,0 +1,13 @@
...
+    tooltipObject.options.defaultPlacement = 'bottom-end';
+    tooltipObject.options.defaultTemplate = `
+      <div class="apos-tooltip" role="tooltip">
+        <div class="apos-tooltip__arrow"></div>
+        <div class="apos-tooltip__inner"></div>
+      </div>`;
+    tooltipObject.options.defaultArrowSelector = '.apos-tooltip__arrow';
+    tooltipObject.options.defaultInnerSelector = '.apos-tooltip__inner';
+  }
+};
```
stuartromanek, line 13: Why avoid creating a component out of this? it has configuration, template, CSS .. I imagine if it were a component that was imported by others we wouldn't need the duplicate `Vue.directive` registrations, etc.

naive?

stuartromanek (changes requested):
type fix for but why no component?

modules/@apostrophecms/ui/ui/apos/lib/tooltip.js
abea: Because it's not a component. v-tooltip does this using a custom directive, not a component, so the directive needs to be set up on every Vue app where it might appear. One downside of Apostrophe using multiple apps for various parts is that it needs to be set up on each one individually.

modules/@apostrophecms/ui/ui/apos/scss/global/_tooltips.scss
abea: That looked pretty small versus the design, but I can do that.

stuartromanek (approved):

PR #2570, opened 2020-11-12
Allows clicks anywhere on piece manager table row to open piece

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-441, "Entire row in (piece) manager windows should click through the primary action of editing the item."

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerDisplay.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerDisplay.vue
@@ -11,7 +11,7 @@
           class="apos-table__header" :key="header.label"
         >
           <component
-            :is="getEl(header)" 
+            :is="getEl(header)"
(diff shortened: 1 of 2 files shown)


boutell (approved):

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManagerDisplay.vue
```diff
@@ -48,18 +48,21 @@
...
+            class="apos-table__link"
             :href="item[header.name]"
+            @click.stop
           >
-            <LinkIcon :size="12" />
+            <link-icon :size="14" />
           </a>
           <button
             v-else-if="header.name === 'title'"
-            @click="$emit('open',item._id)"
```
abea, line 62: I'm leaving this button click event here because it's going to do the easy work in keyboard control situations rather than fuss with `tabindex` on the table cell when one also has a link in it.

PR #2571, opened 2020-11-16
page level undo/redo support

boutell:
I added undo/redo buttons to the context bar, as I understand the original design to have been.

This undoes and redoes operations on the context document (the only document that we're going to be editing in context soon, remember, that ticket is open but not done). You can only undo back to the last save operation. 

The strategy is to leverage the existing array of pending patches that we are already building in order to "save all at once" rather than "save as you go." Nothing *really* happens in the database until you save, but we need the server's help to re-render the content with the change undone.

In principle we'd handle that with a GET request containing all the patches we do still want to apply in the query string, as a way of doing a "virtual patch" that gives us back the content rendered with all the pending changes we haven't undone. In practice you can't make giant query strings like that. So instead we use a POST API that stashes the patches in Apostrophe's cache and redirects to the URL we really wanted, with a query parameter referencing that cache. Apostrophe's database query APIs then spot any match with that cache entry and use that patched document instead of the real one.

Some followups I will open if this is incremental step is accepted:

* Hold on to the scroll position so that if you undo while way down the page, you don't get scrolled to the top while the content is rebuilding onscreen.
* To improve performance both here and when loading the page as an editor, make optimizations to require fewer separate render-widget requests. Debounce these and send requests for more than one at a time.
* Wrap a "busy" state around that rendering process, both here and when loading the page as an editor, because otherwise it's a lot of flashing going on.

Those last two aren't specific to undo/redo really, you just notice them more with this feature present.

A point for discussion: how many "undo points" typing in the rich text editor could produce. One per character was really inefficient, so I've debounced it. However if you're furiously typing this could mean a whole paragraph is one undo point. Maybe it should create at least one every 5 seconds or so if the text has changed? Google Docs does something like that.

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -1,4 +1,6 @@
 const _ = require('lodash');
+const cuid = require('cuid');
+
 // This module is responsible for managing all of the documents (apostrophe "docs")
 // in the `aposDocs` mongodb collection.
 //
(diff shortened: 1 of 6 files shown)


boutell:
Another point for discussion: this *could* probably be bound to the undo/redo keystrokes, replacing the undo/redo of tiptap entirely. If we do that, then I would need to add an optimization to be able to quickly undo a text-only change without asking the server for help, which is feasible. I would also need to stash and restore the text selection. Or we could intentionally decide this is the "high level" undo/redo only and that rich text specific undo/redo should still be available.

modules/@apostrophecms/doc/index.js
```diff
@@ -83,6 +85,59 @@ module.exports = {
       }
     };
   },
+  routes(self, options) {
+    return {
+      post: {
+        // POST to /@apostrophecms/doc/api/v1/get-as-patched with
```
abea, line 91: dark magic

modules/@apostrophecms/doc/index.js
```diff
@@ -83,6 +85,59 @@ module.exports = {
...
+              await self.apos.page.applyPatch(req, doc, input);
+            } else {
+              await manager.applyPatch(req, doc, input);
+            }
+          }
+          // As far as applyPatch is concerned, relationships are
+          // real, but as far as mongo is concerned they're not, so
+          // we have to get back to real _ids before we try to substitute
+          // this document in the result of a mongo query
+          manager.prepareRelationshipsForStorage(req, doc);
```
abea, line 128: This is a method on the schema module, but no the doc type modules. It's a handler function, though. Is that how it's declared and referenced here?

abea (approved):
This looks good. I'm just not clear on how the `prepareRelationshipsForStorage` plumbing is working.

boutell:
I think I goofed on that bit of plumbing, so I'll follow up on that before I merge.

boutell:
Actually it's fine: right now, handlers *are* methods. Which, I fully intended to change for A3, because you can do `extendHandlers` instead. But, that's not what this PR is about. So I will open a ticket about it and mention that followup would be needed for this particular case.

PR #2572, opened 2020-11-16
pre-render widgets on the initial apos-edit=1 request for the page

boutell:
Saves a zillion render-widget requests.

Currently if the page request has `apos-edit=1` the server renders areas as just a div with a data attribute and lets Vue call back for renderings of the widgets so it can encapsulate them in components, etc. This is great in principle, slow and visually weird in practice.

So with this PR, in addition to the div with the data attribute, we do render the widgets in a div for each one, with a data attribute containing its _id. The AposAreas app finds these and passes them to the areas as advance renderings, then snips them out of the DOM, allowing Vue to put them back as it sees fit.

Since props don't change but the widgets can be edited, we pass in enough information to figure out if these pre-renderings are still relevant or not; if not we call render-widgets as per normal.

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
+++ b/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
@@ -23,11 +23,30 @@ export default function() {
     const data = JSON.parse(el.getAttribute('data'));
     const fieldId = el.getAttribute('data-field-id');
     const choices = JSON.parse(el.getAttribute('data-choices'));
+    const renderings = {};
+    const _docId = data._docId;
+
+    for (const widgetEl of el.querySelectorAll('[data-apos-widget]')) {
+      const _id = widgetEl.getAttribute('data-apos-widget');
+      const item = data.items.find(item => _id === item._id);
+      if (item) {
+        renderings[_id] = {
+          html: widgetEl.innerHTML,
+          parameters: {
+            _docId,
+            widget: item,
+            areaFieldId: fieldId,
+            type: item.type
+          }
+        };
+        widgetEl.remove();
+      } else {
+        // Nested widget, none of our business, picked up later
+      }
+    }
     el.removeAttribute('data-apos-area-newly-editable');
 
     const component = window.apos.area.components.editor;
-
-    const _docId = data._docId;
 
     return new Vue({
       el: el,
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -23,12 +23,31 @@ export default function() {
...
+          html: widgetEl.innerHTML,
+          parameters: {
+            _docId,
+            widget: item,
+            areaFieldId: fieldId,
+            type: item.type
+          }
+        };
+        widgetEl.remove();
+      } else {
```
abea, line 43: This empty `else` block pattern always grates at me. The comment is useful, but it's a confusing way to present it. Looks like something is missing.

modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -23,12 +23,31 @@ export default function() {
     const data = JSON.parse(el.getAttribute('data'));
     const fieldId = el.getAttribute('data-field-id');
     const choices = JSON.parse(el.getAttribute('data-choices'));
+    const renderings = {};
+    const _docId = data._docId;
+
+    for (const widgetEl of el.querySelectorAll('[data-apos-widget]')) {
```
abea, line 29: Does this work? You removed the only two places this attr appears in AposAreaWidget.

modules/@apostrophecms/area/views/area.html
```diff
@@ -3,10 +3,14 @@
...
-  {%- if not data.canEdit -%}
-    {%- for item in data.area.items -%}
-      {%- set widgetOptions = data.options.widgets[item.type] or {} -%}
-      {% widget item, widgetOptions %}
-    {%- endfor -%}
-  {%- endif -%}
+  {%- for item in data.area.items -%}
+    {%- set widgetOptions = data.options.widgets[item.type] or {} -%}
+    {%- if data.canEdit -%}
+      <div data-apos-widget="{{ item._id }}">
```
abea, line 9: My only hesitation is tying us further to inserting ourselves into the DOM as we consider alternatives, but this seems less of an issue with that than the .apos-area div.

abea (approved):

PR #2574, opened 2020-11-16
3/apos vue

abea:
Packages up a single modified version of Vue for use in Apostrophe UI.


boutell (changes requested):
This is A+, but since you cut the import of portal, you should cut the dependency too.

boutell (approved):

PR #2577, opened 2020-11-17
foreign areas now work approximately per design by Stu but could use …

boutell:
…some CSS followup. Also, apos.area.isEmpty and similar helpers work now. Necessary casualty: apos.area cannot just be a helper that tells you to use the new {% area %} tag, so we may want to lint that another way at some point

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
@@ -260,7 +260,12 @@ export default {
             }
           );
           if (doc._url) {
-            window.location.href = doc._url;
+            if (await apos.confirm({
+              heading: 'Open that document in a new tab?',
+              description: 'That content is part of another document. Would you like to edit that document in another tab?'
+            })) {
+              window.open(doc._url);
+            }
           } else {
             apos.bus.$emit('admin-menu-click', {
               itemName: `${doc.type}:editor`,
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
@@ -1,7 +1,7 @@
 <template>
   <div :data-apos-area="areaId" class="apos-area">
     <div
-      v-if="next.length === 0"
+      v-if="next.length === 0 && !foreign"
       class="apos-empty-area"
     >
       <template v-if="isEmptySingleton">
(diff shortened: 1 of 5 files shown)


boutell:
You can test with the `foreign-areas` branch of `a3-demo`. In that branch, the `blurb` field of articles has gone back to being an area, and it allows both image and rich text widgets to allow for seeing how both of them act.

You can also explore removing your articles page from your site in order to see that if there is no `_url` for a foreign document, it opens the settings modal for it rather than navigating there.

You might have to remove your existing `article` pieces, I'm not sure if that was actually necessary in my case but there is a warning at least because the previously existing `blurb` is not a proper area.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -467,6 +479,13 @@ export default {
     .apos-area-widget-inner &.apos-highlight:before {
       z-index: $z-index-default;
     }
+
+    &.apos-foreign {
+      outline: 1px dashed var(--a-warning);
+      /deep/ .apos-button {
+        background-color: var(--a-warning);
```
abea, line 486: Let's also make the color of the icon white to match the breadcrumbs. `var(--a-white)`

abea (changes requested):

modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -302,6 +334,9 @@ export default {
         }
       }
     },
+    contextualWidgetDefaultData(type) {
```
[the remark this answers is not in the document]
boutell: I was fixing an unrelated bug: it gets called here and in AposAreaMenu but the implementation had been migrated to just AposAreaMenu.

boutell:
I added the confirmation and new tab logic. I wasn't quite sure what you wanted me to do with the colors, and I'm not actually sure I still even see the pencil for these; can you take a look at the visual aspects?

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
abea: This still needs the white icon. That should be done in aposWidgetControls, passing the `fill-color` prop into the icon.

abea (changes requested):

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
abea: Though setting `color: var(--a-white)` here seems to be enough.

abea: I'm talking about this:
Screen Shot 2020-11-23 at 9 22 48 AM

The pencil on yellow not matching the text on yellow.

abea (approved):

PR #2579, opened 2020-11-18
3/clear empty rte

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-682, "If I had empty richtext while in Edit mode, they are removed on publish/save"

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -201,6 +201,7 @@ export default {
         busy: true
       });
       this.patches = [];
+      await this.refresh();
     },
     switchToEditMode() {
       window.sessionStorage.setItem('aposEditMode', 'true');
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -420,7 +420,7 @@ module.exports = {
           }
         }
       },
-      // Called by `docBeforeInsert` to confirm that the user
+      // Called by `beforeInsert` to confirm that the user
       // has the appropriate permissions for the doc's type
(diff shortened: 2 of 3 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -201,6 +201,7 @@ export default {
         busy: true
       });
       this.patches = [];
+      await this.refresh();
```
boutell, line 204: This is going to refresh constantly with autosave.

boutell (changes requested):
Now that I see how this works out, if we're going to do this I think we need to do it on the front end. Otherwise the front end doesn't know what the backend is doing which leads to problems for undo/redo and the integrity of the patching algorithm. Also on the front end we can address what happens with autosave if you add a rich text widget and just don't type right away... blam, it's gone!

You see why I was really leery of going ahead on it, but it's a useful experiment and first step.

boutell:
I would recommend not taking this further until autosave lands for the sake of not having to rework it twice.

abea:
@boutell We can wait, but the idea is that this wouldn't be done until the user's intentional save point. That wouldn't be during auto-save. Do we have those anymore?

boutell:
We don't have them at the *moment*, but we will have them when draft/publish becomes a thing.

At that point we could make sure a refresh happens, that's a reasonable thing to do at such a moment yep.

boutell:
Closing this in its current form as discussed. We talked about various options for the future, including making it easier to spot an empty (or intentional placeholder) rich text widget as an editor, or doing this at specific times in the life cycle that wouldn't cause issues for contextual editing. I also suggested doing it browser-side, but the trouble with that is I don't see any really good time for it. If I just added a rich text widget, I probably do not want it to go away at least half of the time even if I immediately go on to making another.

PR #2580, opened 2020-11-20
3/advisory locking (piece editor / page settings editor only for now)

boutell:
This is the initial implementation of advisory locking. Important caveats for testing:

* It is currently only implemented for AposDocEditor. Editing an article is a good test.
* If the conflict is with another user, and they choose to take control, your dialog closes and you get a notification with the bad news. It's nice.
* If the conflict is with yourself, you automatically take control, because it is impossible to prove for sure that the "conflict" is not just due to the fact that you abandoned the previous page (there is no guarantee of communication on page unload in modern browsers), and telling people about a fake "conflict" caused enormous frustration. However, you do still get notified in the other tab, the impacted one, if it's real. We worked all this out painstakingly in late A2, it's not a new practice.
* In the impacted tab (losing control) I am *temporarily* using a browser `alert` if the conflict is with yourself. This will change when Bea's nice alerts are available as a PR (Bea, I'm thinking you might just submit this part of your "tell the user if they haven't saved in too long" PR, since that is superseded by bringing back autosave at this point).
* The browser `alert` can be a little confusing to test because it pops up over whatever tab you are currently on. But it really is from the affected tab. Please keep in mind this won't be an issue when we switch it to one of Bea's nice alerts.
* I think I should use the nice new alerts in *both* cases, "conflict with self" and "conflict with other", because it is priority information and clearly interrupts your flow. But, we'll worry about that when they exist in the 3.0 branch.


boutell:
Note: this PR also stomps out inappropriate references to "cursor" and fixes an inconsistency between the findOneForEditing methods of pieces and pages, both of which now take a builders argument rather than a one-off options argument, projection argument (only in pages), etc.

boutell:
This incorporates the foreign areas work (I had to build on things to build on things to get things done), so I have marked it draft until I respond to the concerns there, which I'll do ASAP.

abea (commented):
I get the window alert when taking control from myself, though not really consistently. No window alert when taking control from another user, though. The confirmation modal is very nice.

For the notification, I suggest using the `warning` type with the alert icon that some of the other notifications are using already.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -223,6 +226,54 @@ export default {
...
+                      force: true
+                    }
+                  }
+                });
+                this.locked = true;
+              } catch (e) {
+                await apos.notify(e.message, {
+                  type: 'error'
+                });
+                this.modal.showModal = false;
```
abea, line 270: It's also not consistently closing the browser on the unlockee. Not sure why.

boutell:
You are getting the alert from the tab where you lost control, which makes sense, but it is popping up over the tab where you took control, which is confusing but only a temporary problem caused by calling `alert`.

I will test again for the not-consistently-closing thing, are you saying the edit view modal stays open? The manage view should stay open as the conflict is only in the edit view.

boutell:
This is ready for proper review. I brought in your alert code, thank you for that, it's appropriate no matter who took control for you, and is much less confusing than native `alert()` because native `alert()` pops up over foreign tabs.

I also fixed a bug: if "B" steals control from "A", "B" must still watch for the possibility of "A" stealing control back. I think this may be the issue you spotted with inconsistent notification of the jilted party.

boutell:
Ah, I had to put this back to draft because foreign areas are not merged yet. I addressed concerns in that PR but also asked if you might address the visual concern you raised because I'm not sure what is being asked for. Or it could be merged and a followup ticket opened to unblock this. 🙏

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -269,6 +382,11 @@ export default {
         if (this.docId) {
           route = `${this.moduleAction}/${this.docId}`;
           requestMethod = apos.http.put;
+          // Make sure we fail if someone else took the advisory lock
+          body._advisoryLock = {
```
abea, line 386: It seems that I should be failing in my disaster scenario I laid out previously. If I took control, the initial editor should not be able to save. Or should get an error when they do.

abea (changes requested):
- All of the widgets on the home page area are now being mark as foreign to that context document. I get the yellow UI treatment and the confirmation modal when clicking on them. When confirming (to go to that things's context document) it refreshes the page.

boutell:
LOLLLL I named two unrelated things contextId 'cuz I'm dumb and the merge revealed this dumbness. I'll fix it in the morning! Thanks.

boutell:
In your scenario, user A attempts to refresh their lock as part of the PUT, so it should fail.

... Except yeah, it doesn't. So I'll fix that.

Once this works like it's supposed to, polling faster won't be needed.

boutell:
Oops, I have to do this for pages too, hold my beer.

boutell:
OK, now it is truly ready for your attention.

boutell:
(Reminder re: pages that this is still only for AposDocEditor. After this PR is approved and merged I will start work on the contextual case and the media editor case.)

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -123,6 +123,7 @@
 
 <script>
 import klona from 'klona';
+import cuid from 'cuid';
```
abea, line 126: This isn't used anymore.

abea (commented):
It looks good, but tests are now failing.

boutell:
Fixed an unrelated issue breaking the tests.

abea (approved):

PR #2581, opened 2020-11-20
3/context menu btn

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-806, "Style the Undo/Redo Buttons," and finishes https://apostrophecms.atlassian.net/browse/A30U-795,  "Style the context bar"

The undo/redo icons aren't exactly the same as in the design, but they are what's included in Material icons. How do we feel about that?

Also resolves https://apostrophecms.atlassian.net/browse/A30U-735, "Z-index conflict remaining on area buttons"


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -54,7 +54,22 @@
...
+            icon="undo-icon" :icon-only="true"
+            @click="undo"
+          />
+          <AposButton
+            :disabled="undone.length === 0"
+            type="default" label="Redo"
+            class="apos-admin-bar__context-button"
+            icon="redo-icon" :icon-only="true"
+            @click="redo"
+          />
```
stuartromanek, line 71: THese need tooltips

stuartromanek: Also, use the undo/redo variants

image

stuartromanek (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
stuartromanek: Also also, these buttons should use the `outline` and `no-motion` modifiers

stuartromanek (changes requested):
forgot one

localghost443 (approved):

abea:
We're back to think looking junky due to the autosave merge. I'm ignoring that for now since we'll need design direction for that.

stuartromanek (approved):

PR #2582, opened 2020-11-23
Autosave returns, with undo/redo support back to page load time

boutell:


boutell:
The "publish changes" button will be back! It just has no function right now, until draft/live arrives.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -121,6 +114,7 @@
             :icon-only="true"
             @click="redo"
           />
+          <span class="apos-admin-bar__status">{{ status() }}</span>
```
abea, line 117: Unless I'm missing something, this should be a computed property.

abea (changes requested):

abea (approved):

abea:
@boutell The other thing I forgot to mention was that it seems like we're ending up with a lot of editing state logic being stored in here. I'm not sure there's a great alternative. A single `status` property that alternates `editing/retrying/saved` would probably be over-engineering.

I think my best thought regarding this is that after the Vue 3 update we separate the code and use Composition API to keep things cleaner.

boutell:
The admin bar is pretty much responsible for the editing state, but looking forward to Vue 3 ways of factoring things that are unnecessarily complicated.

PR #2584, opened 2020-11-24
3/bearer tokens and api keys

boutell:
It works! Now I have to document it. So I'll do that here and then move it to a docs PR.

Working examples are in `test/pieces.js`.

## API keys

API keys are great for server-to-server communication, because they don't expire.

They are unsuitable for browser-side or headless-app use because it is possible for end users to determine the api key and use it for their own purposes. For these roles see "bearer tokens," below.

 Because 3.x has a simpler permissions model, api keys live in just one place: the `@apostrophecms/express` module.

To set up an api key you write:

```
        '@apostrophecms/express': {
          options: {
            apiKeys: {
              // Use your own api key value, we recommend a strong randomly generated password
              'my api key here': {
                role: 'admin'
              }
            }
          }
        },
```

Right now `admin` is the only supported role value; any other role will get you an anonymous request, which you could get without an api key. But this will change in the next alpha release when the rest of the permissions model comes in.

To use an API key, just add an `authorization` HTTP header to each request, like this:

```
ApiKey ACTUAL-TOKEN-GOES-HERE
```

Alternatively you may pass the api key as the `apikey` or `apiKey` query parameter.

## Bearer tokens

Bearer tokens are more appropriate for browser use and headless apps. That's because they are tied to a single user account. A user who obtains a bearer token via watching the network panel only has access to the resources they could edit via your app anyway.

To obtain a bearer token, make a POST a request to `/api/v1/login/login`, with the following body properties:

```javascript
{
  username: 'username',
  password: 'password'
}
```

You will get back a JSON object with a `token` property:

```javascript
{
  token: 'xyz'
}
```

Now you can carry out all of the REST API actions that require login by passing an `Authorization` header. This includes POST, PUT, PATCH and DELETE operations. You will also get more complete results from GET operations, including access to documents with `visibility` set to `loginRequired`.

Your `authorization` HTTP header should look like:

```
Bearer ACTUAL-TOKEN-GOES-HERE
```

Note the need for the word "Bearer" before the key.

To log out and destroy the token, send a POST request to `/api/v1/login/logout`, with the same `authorization` header. No body properties are required. After logging out, the token is no longer accepted.


boutell:
A lot of docs were missing on the first try, updated.

modules/@apostrophecms/express/index.js
```diff
@@ -65,25 +65,19 @@
 //
 // ```javascript
 // {
-//   // Do not save sessions until something is stored in them.
+//   // Do not save sesions until something is stored in them.
```
falkodev, line 68: `sessions` I think

modules/@apostrophecms/express/index.js
```diff
@@ -158,9 +145,10 @@ const bodyParser = require('body-parser');
 const expressSession = require('express-session');
 const cookieParser = require('cookie-parser');
 const qs = require('qs');
+const expressBearerToken = require('express-bearer-token');
 
 module.exports = {
-  init(self, options) {
+  async init(self, options) {
```
falkodev, line 151: `async` is not necessary here I think (no `await`)

modules/@apostrophecms/express/index.js
```diff
@@ -191,7 +179,7 @@ module.exports = {
       },
       'apostrophe:modulesReady': {
         async addCsrfAndModuleMiddleware() {
```
falkodev, line 181: async is not necessary here I think (no await)

modules/@apostrophecms/express/index.js
```diff
@@ -215,6 +203,78 @@ module.exports = {
...
+          if (info.role === 'admin') {
+            taskReq = self.apos.task.getReq();
+          } else {
+            taskReq = self.apos.task.getAnonReq();
+          }
+          req.user = taskReq.user;
+          req.csrfExempt = true;
+          return next();
+        }
+        return res.status(403).send({ error: 'invalid api key' });
```
falkodev, line 223: Not mandatory, but it would ease reading the code to drop an error `if (key && !_.has(self.options.apiKeys, key))` otherwise `return next()`

modules/@apostrophecms/express/index.js
```diff
@@ -215,6 +203,78 @@ module.exports = {
...
+            return null;
+          }
+          const matches = header.match(/^ApiKey\s+(\S.*)$/i);
+          if (!matches) {
+            return null;
+          }
+          return matches[1];
+        }
+      },
+      beforeMiddleware(req, res, next) {
```
falkodev, line 237: What is the purpose of `beforeMiddleware`? Is it some kind of hook?

modules/@apostrophecms/express/index.js
```diff
@@ -441,7 +504,9 @@ module.exports = {
           // See options.csrfExceptions
           if (!req.cookies[self.apos.csrfCookieName] || req.get('X-XSRF-TOKEN') !== req.cookies[self.apos.csrfCookieName] || req.session['XSRF-TOKEN'] !== req.cookies[self.apos.csrfCookieName]) {
             res.statusCode = 403;
-            return res.send('forbidden');
+            return res.send({
```
falkodev, line 507: I understand you don't want to give too much information, but a message as `crsf exception` would be useful to find where in the code we get a 403 error (I suspect there are many places, at least in A2 that was the case, and it was hard to track where the error was coming from when your request is rejected).

modules/@apostrophecms/login/index.js
```diff
@@ -94,31 +110,52 @@ module.exports = {
...
           }
-          const passportLogin = (user) => {
-            return require('util').promisify(function(user, callback) {
-              return req.login(user, callback);
-            })(user);
-          };
-          await passportLogin(user);
+          if (session) {
+            const passportLogin = (user) => {
+              return require('util').promisify(function(user, callback) {
```
falkodev, line 123: You lost me here 😅 
I see it was already written like that previously, but still, I am not sure of what is going on here.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -102,7 +102,10 @@ export default {
       try {
         await apos.http.post(`${apos.modules['@apostrophecms/login'].action}/login`, {
           busy: true,
-          body: this.doc.data
+          body: {
+            ...this.doc.data,
+            session: true
```
falkodev, line 107: Not sure I understood why this property has been added (still able to log in through the UI)

test/pieces.js
```diff
@@ -804,4 +815,124 @@ describe('Pieces', function() {
...
+      body: {
+        username: 'admin',
+        password: 'admin'
+      }
+    });
+    assert(response.token);
+    token = response.token;
+  });
+
+  it('can POST a product with the bearer token', async () => {
```
falkodev, line 833: I tried to POST and PATCH a product having a join with an existing `article` id (with `"articlesIds": ["id-of-the-article"]`). No errors, but the join is not taken into account. Maybe the syntax has changed, but it would be useful to check that too.

falkodev (changes requested):

modules/@apostrophecms/express/index.js
boutell: Thanks fixing typo

modules/@apostrophecms/express/index.js
boutell: True

modules/@apostrophecms/express/index.js
boutell: I'll put in a check that self.options.apiKeys is actually set so you can't crash it with this, but otherwise keep it the same

modules/@apostrophecms/express/index.js
boutell: Good catch it was testing

modules/@apostrophecms/express/index.js
boutell: You convinced me. Adding this.

modules/@apostrophecms/login/index.js
boutell: We do not use callbacks, but passport's login method takes a callback. So we "promisify" it, which allows us to await it. This is the correct pattern if you just invoke something callback-based in A3.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
boutell: Without this property you get a bearer token instead, which we anticipate will be the more "normal" case for everyone else but us (third party apps logging in).

test/pieces.js
boutell: In A3, articlesIds is an implementation detail the developer should not think about. Instead you should actually pass `_articles`, as an array of objects. However the only mandatory properties of each object are `_id` and, if present, `_fields`. This is what allows us to implement the chooser/relationship field stuff more naturally in A3.

In A3 you patch schema fields, no matter how magical they may seem... not private implementation properties.

PR #2585, opened 2020-11-25
3/indicator

stuartromanek:
- satisfies https://apostrophecms.atlassian.net/browse/A30U-649?atlOrigin=eyJpIjoiMTA2OGQyMzQ2Y2FlNDdhYzgwNDA3YzRhN2RlN2IxZWQiLCJwIjoiaiJ9
- adds icons and tooltips for parked pages and trash inside Page Tree
- adds `<AposIndicator/>`, a wrapper for icons with tooltips
- implements `<AposIndicator/>` inside `<AposButton/>` and single-uses inside `<AposTree />`


abea (commented):
I don't know what I think about this approach. It should probably get some team review before we move ahead.  This makes pretty good sense where we need to put a tooltip on an icon. Where we don't use a tooltip the benefit seems marginal if there at all. I don't see a problem with using the icon components directly.

stuartromanek:
Using `<AposIndicator/>` for all icons would eliminate some amount of unintuitive, component-specific pixel pushing CSS that makes icons align with their peer components just so, plus bake in tooltip capabilities everywhere, however, replacing everything feels like a low priority at present. I am fine with this being an optional enhancement. 

I will follow up and use it in places like the doc switcher and the future clipboard indicator.

abea (approved):
Let's go ahead with this. Is there a reason it shouldn't be `AposIcon`, though? Especially if we're going to use it in most contexts instead of the material component.

boutell:
This is just chillin' in the approved zone, with a clear "let's go ahead with this," so I'm merging it with my eye on Wednesday.

PR #2588, opened 2020-11-25
show only the fields of the first tab at page settings startup. Also:…

boutell:
… fixed console warning from AposInputSlug

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -215,7 +215,7 @@ export default {
       }
     },
     tabs() {
-      if (!this.currentTab) {
+      if ((!this.currentTab) || (!this.tabs.find(tab => tab.name === this.currentTab))) {
         this.currentTab = this.tabs[0].name;
       }
     }
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -213,6 +213,11 @@ export default {
           });
         }
       }
+    },
+    tabs() {
+      if (!this.currentTab) {
+        this.currentTab = this.tabs[0].name;
+      }
     }
   },
   async mounted() {
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
@@ -149,7 +149,7 @@ export default {
       }
       if (this.field.page) {
         const matches = slug.match(/[^/]+$/);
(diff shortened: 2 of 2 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -213,6 +213,11 @@ export default {
           });
         }
       }
+    },
+    tabs() {
+      if (!this.currentTab) {
+        this.currentTab = this.tabs[0].name;
```
boutell, line 219: This isn't known until later in the case of page settings due to the way the type of the page is determined.

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
```diff
@@ -149,7 +149,7 @@ export default {
       }
       if (this.field.page) {
         const matches = slug.match(/[^/]+$/);
-        slug = matches[0] || '';
+        slug = (matches && matches[0]) || '';
```
boutell, line 152: Unrelated console warning fix, came up when editing page settings for the home page.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
abea: Do we not need to do a deep change watch? What if the tabs are the same, but the contents are different? I'm not 100% sure of what is found or not with default watching. We could also tie this to `docType` changing after mount.

abea (commented):
See my note on the tabs watching.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
boutell: `tabs` is a computed property. So what will that actually watch for?

"Computed properties are cached based on their reactive dependencies."

Following the stack of computed properties accessed by the `tabs()` computed function all the way to the bottom, we get to `moduleOptions`, which depends on the data property `this.docType`, which starts out correct for a piece type but as `@apostrophecms/page` for a page type. However it then changes dynamically based on newInstance or a change of `type`.

So since the correct set of tabs is fully determined by `type`, watching for that change of `type` is enough to do the job for us.

However, looking at this I did notice another scenario: the current tab can become invalid due to a switch of page type if the new page type does not have that tab. I added logic for that in a new commit.

boutell: We're watching correctly — default watch behavior for computed properties is to run any time their "reactive dependencies" (other computed and data properties and props) change, and at the bottom of that stack is `moduleOptions`, which depends on `this.docType`, which is what we want.

However I did notice that the current tab can become invalid if you switch to a page type that doesn't have that tab, so I added logic for that too.

abea (approved):

PR #2591, opened 2020-11-25
good error messages for login/logout, and expose error.message and er…

boutell:
…ror.name conveniently on the browser side so you do not have to dig in error.body for them


localghost443 (approved):
these are great

PR #2592, opened 2020-11-25
no undo/redo outside edit mode. Also, make sure video players do thei…

boutell:
…r thing when switching back to preview mode, or in foreign widgets


localghost443 (approved):
beautiful

PR #2594, opened 2020-11-25
fixed undo of adding a widget to a top level area. Fixed patch with @…

boutell:
… reference to top level area id (same thing). Fixed global busy display for HTTP requests. Fixed redundant namespacing of global busy events.

### DIFF
--- a/modules/@apostrophecms/busy/ui/apos/components/TheAposBusy.vue
+++ b/modules/@apostrophecms/busy/ui/apos/components/TheAposBusy.vue
@@ -14,7 +14,7 @@ export default {
     };
   },
   mounted() {
-    apos.bus.$on('apos-busy', state => {
+    apos.bus.$on('busy', state => {
       // TODO: Possibly add a check for `state.name === 'busy'` again if other
       // busy contexts are added.
       if (state.active === false && this.busyCount >= 0) {
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -2190,9 +2190,13 @@ module.exports = {
         }
         _.each(patch, function(val, key) {
           if (key.charAt(0) !== '$') {
-            key = self.apos.util.resolveAtReference(existingPage, key);
+            let atReference = false;
+            if (key.charAt(0) === '@') {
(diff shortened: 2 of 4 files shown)


modules/@apostrophecms/busy/ui/apos/components/TheAposBusy.vue
```diff
@@ -14,7 +14,7 @@ export default {
     };
   },
   mounted() {
-    apos.bus.$on('apos-busy', state => {
+    apos.bus.$on('busy', state => {
```
boutell, line 17: Apostrophe events are specific to apostrophe by definition, we do not namespace any others like this, so I removed the unwanted prefix.

modules/@apostrophecms/schema/index.js
```diff
@@ -2190,9 +2190,13 @@ module.exports = {
         }
         _.each(patch, function(val, key) {
           if (key.charAt(0) !== '$') {
-            key = self.apos.util.resolveAtReference(existingPage, key);
+            let atReference = false;
+            if (key.charAt(0) === '@') {
```
boutell, line 2194: This was the bug that broke undo. Apostrophe has a sweet extension to PATCH to let you patch any sub-object by the _id of that nested object using this syntax: `@id-goes-here-` but it was broken for the edge case where the sub-object you're replacing is a top level property of the page, like the "main" area. Simple to fix.

modules/@apostrophecms/util/ui/public/http.js
```diff
@@ -118,7 +118,17 @@
     if (options.qs) {
       url = apos.http.addQueryToUrl(url, options.qs);
     }
-
+    if (options.busy) {
```
boutell, line 121: This code was in the "load" event, so we didn't start "looking busy" until the request was already done. Oops. Now we start looking busy when the HTTP request begins, and it looks right.

We really should have a nice spinner and a transition so this fades in and doesn't bug you with flashy-flashy if it only takes an instant, and I don't know why we don't have that in 3.x yet (we do in 2.x), but that's another convo entirely.

localghost443 (approved):

PR #2596, opened 2020-11-25
Area-from-text fixes (for the blurb migration)

boutell:
… code in the fromRichText and fromPlaintext methods


boutell:
This PR eliminates duplicate code, and also makes sure areas created from rich text have an _id, which is required in 3.x.

abea (approved):

PR #2605, opened 2020-11-30
Migration to fix duplicate or missing widget ids

boutell:
. I cannot reproduce the issue from scratch anymore (I tested the migration by manually modifying my database).

But it existed on staging in existing content at one point, so others might have this problem with alpha 1 databases and need this migration to have better success in alpha 2.

I don't see a clear point at which the original bug was fixed though. So if we see it creep back in then I will need to know how to reproduce it. Just using the clone button does not reproduce it.

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -31,6 +31,7 @@ module.exports = {
     self.enableBrowserData();
     await self.enableCollection();
     await self.createIndexes();
+    self.addDuplicateOrMissingWidgetIdMigration();
   },
   restApiRoutes(self, options) {
     return {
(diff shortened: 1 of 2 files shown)


abea (commented):
What is the benefit of writing migrations into core (that will live forever) if they are motivated by edge cases we create for ourselves in alpha demoing?

boutell:
Because this isn't just for us. We published alpha 1 and people are hopefully evaluating it, and their evaluation of alpha 2 will be brief and disappointing if this bug is currently tormenting them. Thus the migration. We don't want any friction for people migrating from alpha 1 to alpha 2 and if we can fix something they are experiencing we should do that.

However we could later look at moving the pre-release migrations to a command line task that former alpha/beta testers can choose to run, perhaps.

abea:
Yeah, I'd like to see some sign that these aren't permanent. Either moving into a separate thing at some point or being removed at a certain stage.

boutell:
I opened A30U-867 to cover migrating this later. OK to approve?

abea (approved):

PR #2613, opened 2020-12-02
3/rte defaults

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-707, "Rich text widget default toolbar is not used when there is no project-level config."

Passing: `toolbar: false` as an option will result in no toolbar.


modules/@apostrophecms/area/index.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
             throw self.apos.error('invalid');
           }
           let options;
+          // TODO: Remove `singleton` check if no longer relevant.
```
boutell, line 44: It's no longer relevant, we officially killed that type.

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -127,6 +131,19 @@ export default {
     moduleOptions() {
       return moduleOptionsBody(this.type);
     },
+    editorOptions() {
+      const defaultOptions = moduleOptionsBody(this.type).defaultOptions;
```
boutell, line 135: Seems like you can use moduleOptions here rather than moduleOptionsBody, no?

boutell (changes requested):
Great stuff, minor things.

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
abea: You are correct.

boutell (dismissed):

abea:
@boutell I missed the server side sanitation before. Please review again if you would.

boutell (approved):

PR #2614, opened 2020-12-03
3/fix root path

boutell:


boutell:
Migration to ensure the path property of the home page is equal to its _id, as intended for 3.x. I had an old a3-demo where this was not the case and we increasingly rely on it being true. I do not see this behavior with newly created sites. This would be one of the migrations to review later for possible removal.

modules/@apostrophecms/page/index.js
```diff
@@ -2009,6 +2010,26 @@ database.`);
...
+        self.apos.migration.add('fix-home-page-path', async () => {
+          const home = await self.apos.doc.db.findOne({
+            slug: '/',
+            path: '/',
+            level: 0
+          });
+          if (!home) {
+            return;
+          }
+          home.path = home._id;
```
abea, line 2024: You don't use this reset value (nor do you need to).

abea (changes requested):

abea (approved):

PR #2616, opened 2020-12-04
3/nice remove link

stuartromanek:
- satisfies https://apostrophecms.atlassian.net/browse/A30U-856?atlOrigin=eyJpIjoiMTk0MzZhN2ZjMGVlNDgxMWFlMWU5MmU3MGNjMDVjY2UiLCJwIjoiaiJ9
- Adds a Remove Link button when the link pane is opened and already has a value
- Simplifies link schema
- Adds keyboard accessibility for intuitive link submissions
- keyboard accessible esc
- general 'clicked outside this element' Vue directive

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposTiptapLink.vue
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposTiptapLink.vue
@@ -86,7 +86,6 @@ export default {
     return {
       keepInBounds: true,
       href: null,
-      id: null,
       target: null,
       active: false,
       hasLinkOnOpen: false,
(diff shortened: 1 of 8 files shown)


modules/@apostrophecms/rich-text-widget/ui/apos/components/AposTiptapLink.vue
```diff
@@ -157,9 +164,27 @@ export default {
...
     save() {
+      // cleanup incomplete submissions
+      if (this.value.data.target && !this.value.data.href) {
+        delete this.value.data.target;
+      }
       this.editor.commands[this.name](this.value.data);
       this.active = false;
     },
+    keyboardHandler(e) {
+      if (e.keyCode === 27) {
```
stuartromanek, line 175: close pane on `esc` key

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposTiptapLink.vue
```diff
@@ -157,9 +164,27 @@ export default {
...
+        this.close();
+      }
+      if (e.keyCode === 13) {
+        if (this.value.data.href || e.metaKey) {
```
stuartromanek, line 179: on `return`:
- if we have an href, consider this a submission
- if we held the meta key, consider this a submission

otherwise, prevent the confusing default behavior of closing the pane without considering new field values

modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
```diff
@@ -11,7 +11,6 @@
         :id="getChoiceId(uid, choice.value)"
         :choice="choice"
         :field="field"
-        :status="status"
```
stuartromanek, line 14: unrelated but necessary cleanup to use checkboxes

boutell (approved):
Sweet!

Maybe someday we'll figure out how to write a vue directive for the "escape dance."

At some point we may have to revisit removing anchors from these because I seem to recall an enterprise client considering it important to be able to paste docs with existing named anchor links and named anchors, but we can cross that bridge if we have to.

PR #2617, opened 2020-12-04
3/colorpicker

stuartromanek:
satisfies https://apostrophecms.atlassian.net/browse/A30U-143?atlOrigin=eyJpIjoiNzAyYzhiNDEwNmRhNGM3YzllMmNmMzZlNzNlNmU3MTIiLCJwIjoiaiJ9
- Adds `color` field type
- Adds new color input component
- Adds button configuration for color buttons
- Follow up ticket for cascading field options

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -227,14 +227,6 @@ module.exports = {
           object[field.name] = null;
         }
       },
-      index: function (value, field, texts) {
-        const silent = field.silent === undefined ? true : field.silent;
-        texts.push({
-          weight: field.weight || 15,
-          text: (value || []).join(' '),
-          silent: silent
-        });
-      },
       isEmpty: function (field, value) {
         return !value.length;
       }
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -214,6 +214,33 @@ module.exports = {
     });
 
     self.addFieldType({
+      name: 'color',
+      convert: async function (req, field, data, object) {
+        console.log(data);
+        if (typeof data[field.name] === 'string') {
+          data[field.name] = self.apos.launder.string(data[field.name]).split(',');
+
+          if (!Array.isArray(data[field.name])) {
+            object[field.name] = [];
+            return;
+          }
+
+          object[field.name] = _.filter(data[field.name], function (choice) {
+            return _.includes(_.map(field.choices, 'value'), choice);
+          });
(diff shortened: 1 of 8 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -213,6 +213,33 @@ module.exports = {
...
+        if (field.required && (_.isUndefined(object[field.name]) || !object[field.name].toString().length)) {
+          throw self.apos.error('required');
+        }
+
+        const test = tinycolor(object[field.name]);
+        if (!tinycolor(test).isValid()) {
+          object[field.name] = null;
+        }
+      },
+      index: function (value, field, texts) {
```
boutell, line 230: Remove. We don't want people searching for color values. I assume.

modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
```diff
@@ -145,9 +160,25 @@ export default {
...
+  mounted() {
+    if (this.type === 'color') {
+      this.previewColor();
+    }
+  },
   methods: {
     click($event) {
       this.$emit('click', $event);
+    },
+    previewColor() {
```
boutell, line 172: I think it would be more maintainable to actually chuck a span for this purpose in the component and generate its dynamic styles the more usual vue way, unless I'm missing something.

boutell (changes requested):

modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
stuartromanek: :chefs-kiss-goodbye:

modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
```diff
@@ -110,6 +117,17 @@ export default {
     tabindex() {
       return this.disableFocus ? '-1' : '0';
     },
+    colorStyle() {
+      if (this.type === 'color') {
+        let style = ''
+        style += `background-color: ${this.color};`;
```
boutell, line 123: A lot better — how about:

```
return {
  backgroundColor: this.color,
  border: `2px solid ${tinycolor(this.color).lighten(20).toString()};`
};
```

Which saves you a bunch of of the string building? (Vue will automatically do the right thing if you bind an object with js-style css property names to :style)

boutell (changes requested):

boutell (approved):

PR #2619, opened 2020-12-07
Uses cell components for piece manager display

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-690, "Date columns look bad in the piece manager"

Introduces a manager cell component option to allow project-level introduction of other manager cell formats. This is limited to the piece manager now. It _could_ be introduced in the page manager. That is arguably more fundamental to core (being that there's only one page manager view) and is more finicky with the dragging that I'm holding off on adding it there for now. Plus that's out of scope of this ticket.

### DIFF
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -28,17 +28,21 @@ module.exports = {
     return {
       add: {
         title: {
-          label: 'Title'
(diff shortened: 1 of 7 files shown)


boutell (approved):
Nice. The date component isn't very localization friendly, but a followup ticket to address that would be sufficient.

PR #2621, opened 2020-12-08
last saved timestamp + author displayed in admin UI

stuartromanek:
Satisfies https://apostrophecms.atlassian.net/browse/A30U-623?atlOrigin=eyJpIjoiZGQzNWQ5ZjdmN2RlNGNkYjk2OGYwNGViZDQ1ZWM4ZGIiLCJwIjoiaiJ9

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -106,7 +106,7 @@
               icon="information-outline-icon"
               fill-color="var(--a-primary)"
               :size="15"
-              tooltip="Page Title"
+              :tooltip="docTooltip"
               class="apos-admin-bar__title__indicator"
             />
             {{ moduleOptions.context.title }}
@@ -171,6 +171,7 @@
 
 <script>
 import klona from 'klona';
+import dayjs from 'dayjs';
 
 export default {
   name: 'TheAposAdminBar',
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -312,7 +317,24 @@ export default {
       }
     }
   },
-  mounted() {
+  async mounted() {
+    // Fetch the user data of the person who last edited this doc for display in the admin bar
+    if (this.moduleOptions.context.updatedBy && this.moduleOptions.context.updatedBy !== 'ApostropheCMS') {
+      const editor = await apos.http.get(
+        `${window.apos.user.action}/${this.moduleOptions.context.updatedBy}`,
```
boutell, line 324: Thinking ahead, this won't be possible for non-admin users to fetch. I suggest that updatedBy just be set to an object with all the properties you need for display at the time you set it. This could get out of date but it's a snapshot of what their name was at the time so that's probably OK and as long as you include `_id` among those properties we can better it later.

boutell (changes requested):

modules/@apostrophecms/doc/index.js
```diff
@@ -129,6 +129,9 @@ module.exports = {
             }
           });
           doc.updatedAt = new Date();
+          doc.updatedBy = req.user || {
```
boutell, line 132: You should cherrypick the likely suspects: firstName, lastName, title, and _id. Although passwords are not in req.user restraint in exposing it as a whole is a good practice. Otherwise A-OK

boutell (changes requested):

boutell (changes requested):
Last one I swear to god

Please include `_id` so that if we decide later to fetch up to date info, we can.

boutell (approved):
Go in peace

PR #2625, opened 2020-12-11
radio schema fields

stuartromanek:
satisfies https://apostrophecms.atlassian.net/browse/A30U-869
@boutell this mostly piggybacks radios off of checkboxes .. Is this the right approach?

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -286,6 +286,19 @@ module.exports = {
     });
 
     self.addFieldType({
+      name: 'radio',
+      extend: 'checkboxes',
+      convert: async function (req, field, data, object) {
+        data[field.name] = self.apos.launder.string(data[field.name]);
+
+        if (field.required && !data[field.name]) {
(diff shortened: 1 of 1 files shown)


boutell (changes requested):
There can be only one value yeah? So shouldn't it piggyback off "select" instead?

modules/@apostrophecms/schema/index.js
```diff
@@ -361,6 +361,11 @@ module.exports = {
       }
     });
 
+    self.addFieldType({
+      name: 'radio',
+      extend: 'select'
```
boutell, line 366: I don't see any `vueComponent` property, is this complete for the front end in some way I'm not seeing?

boutell (changes requested):

modules/@apostrophecms/schema/index.js
boutell: Never mind I forgot there's a default naming pattern.

boutell (approved):
OK I forgot there's a default pattern for naming the frontend components

PR #2626, opened 2020-12-11
better empty color pickers, no More menu for New docs

stuartromanek:
Cleaning up some debris from my recent prs
image

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -13,6 +13,7 @@
     </template>
     <template #primaryControls>
       <AposDocMoreMenu
+        v-if="docId"
         :doc-id="docId"
       />
       <AposButton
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -13,6 +13,7 @@
     </template>
     <template #primaryControls>
       <AposDocMoreMenu
+        v-if="docId"
```
boutell, line 16: Will "publish" be on this menu, for one-step save and publish? I guess not because "just publish the fool thing" will be the more common default preference so it'll be a button.

boutell (approved):

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
stuartromanek: Publish will never be in this tertiary menu† (as far as I can see from here). Publish will be the primary button.

Share, Duplicate, and Revert all don't make sense in a New context, so that is fine.

Teasing this out, this means that that button needs to be context-aware of both a non-publisher (no publish rights) and a publisher who has explicitly set Visibility to false during new piece creating. The labels and behaviors need to update in those cases.

PR #2631, opened 2020-12-15
Foreign document editing UX/Design

stuartromanek:


abea:
I'm not getting these audit warnings locally, but they're not going away running the CI tests again.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -620,7 +623,7 @@ export default {
     background-color: var(--a-secondary);
   }
 
-  .apos-show,
+  .is-visible,
   .apos-focus {
```
abea, line 627: Still leaving this `.apos-focus`?

abea (changes requested):
I see some remaining uses of `apos-highlight` and `apos-focus`. Unless I'm missing something those should be replaced as well. Then the audit tests need to pass, of course.

abea (approved):

PR #2634, opened 2020-12-15
Remove potentially confusing changelog line

abea:
This was confusing to me in a release that does have code changes. I think it's referring to that PR or commit, but it made my dog turn her head sideways.


abea:
This would be left in for the next release since the release in question was published already.

boutell (approved):
This was already published, but good to correct the record.

PR #2635, opened 2020-12-15
Image scrolls 2130

### LINKED ISSUE #2130: Scrolling on the Manage Images modal does not work consistantly.
Apos v 2.101.0

**To Reproduce**
1. Go to the images modal with multiple pages of images (I have 200+)
2. Try to scroll
3. It only goes a small part of the way down. There are still images to see below that I can't get to.
4. Keep fiddling with the cursor and the window and eventually it works

**Expected behavior**
If there are images below to scroll to you should be able to scroll down without hesitation.

**Describe the bug**
Scrolling is stuck.

**Version of Node.js**
v12.13.1

**Browsers (only browsers tested - it might happen in others)**
Firefox 72.0.1
Chrome 79.0.3945.117 (Official Build) (64-bit)

**Screenshots**
See the video below. (I can also send you the .MOV file)
ScreenRecording2020-01-21at92717

**Additional context**
This doesn't happen every time and it eventually gets unstuck. It's not clear to me the state of the cursor or the window to cause this, but it is happening pretty consistently to me.

abea:
Resolves #2130 , a double scroll bar bug somehow created by introducing CSS Grid styles.

### DIFF
--- a/lib/modules/apostrophe-images/public/css/grid-item.less
+++ b/lib/modules/apostrophe-images/public/css/grid-item.less
@@ -83,10 +83,6 @@
 
     .apos-transition();
   }
-  input:checked + .apos-field-input-checkbox-indicator,
-  {
-    border-width: 1;
-  }
 
   // border
   &::before
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 
 ## 2.113.1 (2020-11-04)
 * Updates the logo in the README.
-k
 
 ## 2.113.0 (2020-10-21)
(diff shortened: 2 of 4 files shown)


lib/modules/apostrophe-images/public/css/grid-item.less
```diff
@@ -83,10 +83,6 @@
 
     .apos-transition();
   }
-  input:checked + .apos-field-input-checkbox-indicator,
```
abea, line 86: Trailing comma typo, but also the rule doesn't do anything either since there is no unit.

stuartromanek (approved):

PR #2637, opened 2020-12-16
Add orientation and file types to images filters

ValJed:
- Add cursors and filters for orientations and file types
- Add migration to fix all existing attachments orientations
- Fix some modal style with flexbox, to avoid dual scroll and bad sizing
- Remove `resizeContentHeight`  since we don't need it anymore


abea:
This is exciting! Is someone already planning an update to the documentation for this?

ValJed:
@abea @boutell  This is a good question, I don't know if we need to document the existence of two new images filers. 
If yes, I'm not sure of where it should be documented.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -1144,4 +1146,44 @@ module.exports = function(self, options) {
     }
   };
 
+  self.addFixOrientationsMigration = function() {
+    self.apos.migrations.add(self.__meta.name + '.fixOrientations', function(callback) {
```
boutell, line 1150: Do not write new migrations in callback style. Migrations may return a promise.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -1144,4 +1146,44 @@ module.exports = function(self, options) {
...
+      return self.apos.migrations.each(
+        self.db,
+        { group: 'images' },
+        5,
+        function(attachment, callback) {
```
boutell, line 1155: Do not take a callback, just return a promise (mongodb methods return promises if not given a callback, too, so this will be easy to do).

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -1144,4 +1146,44 @@ module.exports = function(self, options) {
...
+            delete attachment.landscape;
+          }
+
+          return self.apos.docs.db.update({
+            _id: attachment._id
+          }, {
+            $set: attachment
+          }, callback);
+        }, function(err) {
+          if (err) {
```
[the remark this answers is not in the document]
boutell: (You should move off the callbacks anyway.)

lib/modules/apostrophe-attachments/lib/api.js
boutell: `each` iterators are permitted to return a promise, in which case they should not take `callback` or invoke it.

lib/modules/apostrophe-images/index.js
```diff
@@ -80,6 +80,50 @@ module.exports = {
...
+            label: 'Square'
+          }
+        ],
+        def: ''
+      },
+      {
+        name: 'fileType',
+        choices: [
+          {
+            value: 'jpg',
```
boutell, line 106: The list of extensions is dynamic, it depends on `self.apos.attachments.fileGroups`, which is an array. One of them will have `name: 'images'`, use the `extensions` property of that to figure out what is supported. Note that svg is not turned on for every site.

See apostrophe-attachments/index.js

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,65 @@ module.exports = {
...
+    self.addFilter('orientation', {
+      finalize: function() {
+        const orientation = self.get('orientation');
+        if (!orientation) {
+          return;
+        }
+
+        self.and(getCriteria(orientation));
+
+        function getCriteria(orient) {
```
boutell, line 52: orient should be orientation

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,65 @@ module.exports = {
...
+          if (orient === 'landscape') {
+            return {'attachment.landscape': true};
+          } else if (orient === 'portrait') {
+            return {'attachment.portrait': true};
+          }
+
+          // Last case is square
```
boutell, line 59: Write an else clause for consistency.

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,65 @@ module.exports = {
...
+        const orientation = self.get('orientation');
+        if (!orientation) {
+          return;
+        }
+
+        self.and(getCriteria(orientation));
+
+        function getCriteria(orient) {
+          if (orient === 'landscape') {
+            return {'attachment.landscape': true};
```
boutell, line 54: Consistent spacing.

boutell: Newlines for objects. (eslint will catch these too I think.)

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,65 @@ module.exports = {
       }
     });
 
+    self.addFilter('orientation', {
+      finalize: function() {
```
boutell, line 44: You should include `def: null` for clarity: if the filter is not explicitly used all orientations are shown.

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,65 @@ module.exports = {
...
+            'attachment.square': true
+          };
+        }
+      },
+      safeFor: 'public',
+      launder: function(s) {
+        return self.apos.launder.string(s);
+      },
+      choices: function(callback) {
+        const {choices} = options.module.filters
```
boutell, line 70: How do we feel about these filters not being responsive to other filters? As written, the list of orientations would not respond to the fact that you've already filtered by tag and no images with this tag are landscape. cc @abea @stuartromanek

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,65 @@ module.exports = {
...
+        };
+
+        self.and(criteria);
+      },
+      safeFor: 'public',
+      launder: function(s) {
+        return self.apos.launder.string(s);
+      },
+      choices: function(callback) {
+        const {choices} = options.module.filters
```
boutell, line 95: How do we feel about these filters not being responsive to other filters? As written, the list of orientations would not respond to the fact that you've already filtered by tag and no images with this tag are landscape. cc @abea @stuartromanek

boutell: We do that for "trash" and, I think, "published" because they are so fundamental, so it's not like we *never* do it.

lib/modules/apostrophe-modal/public/css/components/modal.less
```diff
@@ -74,22 +83,28 @@
...
   width: 100%;
 
-  // background-color: @apos-white;
 	background-color: lighten(@apos-light, 5%);
 }
 
 .apos-modal-body
 {
-	height: 100%;
+  height: calc(100% - 63px);
```
boutell, line 107: Where is this number coming from?

lib/modules/apostrophe-modal/public/js/modal.js
```diff
@@ -535,60 +534,6 @@ apos.define('apostrophe-modal', {
       }
     };
 
-    // Calculates the appropriate modal body height by subtracting
-    // header, breadcrumb, and footer heights and an additional
-    // 50 pixels from the browser window height. Invoked for you
-    // by `self.show()`.
```
boutell, line 541: There is a lot of hard-won logic here. How sure are we that this still works for all the situations it was meant to address?

lib/modules/apostrophe-pieces/public/js/manager-modal.js
```diff
@@ -701,17 +701,6 @@ apos.define('apostrophe-pieces-manager-modal', {
       });
     };
 
-    // shrink and grow make visual reflectments to accommodate the the new Select Everything element
-    self.shrinkGrid = function() {
```
boutell, line 705: Why did this go away?

lib/modules/apostrophe-pieces/public/js/manager-modal.js
```diff
@@ -723,7 +712,6 @@ apos.define('apostrophe-pieces-manager-modal', {
       var $selectEverything = self.getSelectEverything();
       if ($pieces.length === $checked.length) {
         self.getSelectAll().prop('checked', true);
-        self.shrinkGrid();
```
boutell, line 726: We can't just remove methods because third parties may be expecting them (bc break). If they are no longer needed they could become empty methods.

lib/modules/apostrophe-modal/public/js/modal.js
boutell: It sounds like the answer is "we fixed it with flexbox," however A2 supports editing in IE11. Is this tested in IE11?

boutell (changes requested):

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -1144,4 +1146,44 @@ module.exports = function(self, options) {
...
 
+  self.addFixOrientationsMigration = function() {
+    self.apos.migrations.add(self.__meta.name + '.fixOrientations', function(callback) {
+      return self.apos.migrations.each(
+        self.db,
+        { group: 'images' },
+        5,
+        function(attachment, callback) {
+          if (attachment.width > attachment.height) {
+            attachment.landscape = true;
```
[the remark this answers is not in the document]
ValJed: Yes what I wanted to do with the delete which was stupid

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -1144,4 +1146,44 @@ module.exports = function(self, options) {
...
+          } else if (attachment.width < attachment.height) {
+            attachment.portrait = true;
+
+            delete attachment.landscape;
+          } else {
+            attachment.square = true;
+
+            // Normally actually square was portrait, but to be sure we delete landscape too
+            // No risk of error and no cost
+            delete attachment.portrait;
```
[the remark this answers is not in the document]
ValJed: Right, don't know why I did it that way

lib/modules/apostrophe-attachments/lib/api.js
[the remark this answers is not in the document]
ValJed: removed

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -1144,4 +1146,44 @@ module.exports = function(self, options) {
...
+            $set: attachment
+          }, callback);
+        }, function(err) {
+          if (err) {
+            return callback(err);
+          }
+          return callback(null);
+        });
+    }, {
+      safe: true
```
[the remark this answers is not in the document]
ValJed: removed

lib/modules/apostrophe-attachments/lib/api.js
ValJed: callback removed

lib/modules/apostrophe-attachments/lib/api.js
ValJed: removed

lib/modules/apostrophe-images/index.js
ValJed: Thanks for this info.

lib/modules/apostrophe-images/lib/cursor.js
ValJed: That was to avoid same name for param and higher level constant but ok.

lib/modules/apostrophe-images/lib/cursor.js
ValJed: ok

lib/modules/apostrophe-images/lib/cursor.js
ValJed: No eslint error neither for spacing nor for new line. Ok

lib/modules/apostrophe-images/lib/cursor.js
ValJed: ok done

lib/modules/apostrophe-modal/public/css/components/modal.less
ValJed: This is the height of the modal's footer, which changes only on mobiles, and I think we don't really support mobiles, but we could use the resizeHeightContent to make it dynamic.

lib/modules/apostrophe-modal/public/js/modal.js
ValJed: The only purpose of the function is to compute and set height for the modal and its header / contents / footer. Flexboxes allow to make it easily, dynamically with less ressources. You right we sould take car with this type of modifications but I took the liberty to do it because the modal was already buggy on my moderns chrome and firefox, I had two superimposed scroll which was really uncomfortable as a user.
Also I checked on IE11 with BrowserStack, the height of the modal seems ok,  the modal's footer is broken and is out of the modal at the bottom of the window, but this bug is present in the main branch too.

lib/modules/apostrophe-modal/public/css/components/modal.less
ValJed: Ok I found a better solution to remove it, moreover it fixes the modal's footer in IE 11 !

lib/modules/apostrophe-modal/public/js/modal.js
ValJed: Just fixed the modal's footer in IE11, but there are still weirds rendering on the editor modal (which was already existing)

boutell (commented):
I'm good with this, but we need signoff from @stuartromanek on the move to flexbox. Your IE11 testing does sound good though. That is why I did not click "approve" yet.

Re: the filters not taking other filters into account, I took a look and it appears the necessary context to allow you to do this is not readily available in A2 for a filter that is not based on a schema field, so let's not worry about it for this release.

stuartromanek (approved):
I'm good with the overall approach, and testing in IE11 looks good.
The only thing that sticks out to me is this weird partial border artifact
image

This appears in all browsers

lib/modules/apostrophe-images/lib/cursor.js
boutell: Your `addFilter` calls need documentation comments just before the call to `self.addFilter`, like the example below. This allows apostrophe-documentation to pick them up and include them in the documentation site:

```
    // Filter. Sets the MongoDB criteria, discarding
    // criteria previously added using this
    // method or the `and` method. For this reason,
    // `and` is a more common choice. You can also
    // pass a criteria object as the second argument
    // to any `find` method.
```

boutell (changes requested):

lib/modules/apostrophe-images/lib/cursor.js
ValJed: Just added two comments to document filters.

boutell (approved):
Since Stu has approved I'm ready now too.

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,10 @@ module.exports = {
       }
     });
 
+    // Filter. Get images depending of the
```
abea, line 43: Typo: "depending on the"

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,10 @@ module.exports = {
...
+    // orientation of their attachments. This orientation
+    // is computed during import, comparing width and height
```
abea, line 45: Add a period at the end here, please.

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -80,6 +84,10 @@ module.exports = {
       }
     });
 
+    // Filter. Get images depending of their
```
abea, line 87: "depending on their" typo

lib/modules/apostrophe-images/lib/cursor.js
abea: We aren't telling them the options to use in this documentation.

lib/modules/apostrophe-images/lib/cursor.js
abea: We need to provide some examples of extensions to use. Lower case? Upper case? With the leading period? We should be very clear, though we don't need to provide every possible option.

lib/modules/apostrophe-images/lib/cursor.js
```diff
@@ -40,6 +40,79 @@ module.exports = {
...
+        const { choices } = options.module.filters
+          .find((filter) => filter.name === 'orientation');
+
+        return callback(null, choices);
+      }
+    });
+
+    // Filter. Get images depending of their
+    // type, the available images extensions come from
+    // self.fileGroups in apostrophe-attachments
```
abea, line 89: End with a period please. `self.fileGroups` and `apostrophe-attachments` should be in back-ticks for code styling.

abea (changes requested):

lib/modules/apostrophe-images/lib/cursor.js
ValJed: You're right, my bad.

lib/modules/apostrophe-images/lib/cursor.js
ValJed: Done.

lib/modules/apostrophe-images/lib/cursor.js
ValJed: @abea I'm not sure to understand what options you are talking about ? The user has nothing to configure since these filters have been added in `apostrophe-images` directly.

lib/modules/apostrophe-images/lib/cursor.js
ValJed: @abea  Same as above, the user has nothing to configure here, and options are dynamics from `self.fileGroups`.
Basically for now there are only images extensions like `jpg`, `png`, `gif` and `svg` if allowed in `apostrophe-attachments`.

lib/modules/apostrophe-images/lib/cursor.js
abea: Can they not use the filters in API calls?

ValJed: You you're right, I no longer had that in mind. 
Just improved comments.

lib/modules/apostrophe-images/lib/cursor.js
ValJed: Documentation improved.

abea (approved):
Thank you!

boutell:
Merged because all reviewers approved and the test failures are due to an unrelated npm audit issue. I verified a successful `npx mocha` in both node 12 and node 8.

PR #2641, opened 2020-12-18
CKEDITOR: When only one style is available, make it default

fredrikekelund:
The current handling of `defaultStyle` for `apostrophe-rich-text-widgets` isn't really documented, but works as follows:
- firstly, look for a `defaultElement` option and create a new `CKEDITOR.style` from that if available
- secondly, look for a `defaultStyle` option and look up the corresponding `style` object if available

I think a logical third option to add would be:
- if a `styles` option is available and contains a single style object, use it as the `defaultStyle`

This PR makes it so!

Sorry for not opening an issue first, this is sort of a drive by contribution in the middle of other work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.113.4 (2020-12-22)
+* When a `styles` config with a single object is passed to an `apostrophe-rich-text` widget, use that as the defualt style.
+
 ## 2.113.3 (2021-01-13)
 * Fixes image manager double-scroll bar bug.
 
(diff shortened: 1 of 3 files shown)


abea (changes requested):
Thanks, @fredrikekelund. This looks like a good improvement. Please add a changelog entry and bump a minor version number.

fredrikekelund:
Great! I've amended the changelog and bumped to `2.113.4`. 

I noted that `2.113.3` is datestamped 2021-01-13 in the changelog btw. I guess that's incorrect?

abea:
We probably want to bump a minor version, not only a patch version, but I can take care of it. We have a two week release schedule, and with the next one hitting Dec 30, when we're mostly not working, the next scheduled release is Jan 13, hence the date.

abea (approved):

PR #2644, opened 2020-12-22
Make selected Boolean darker

joeinnes:
--a-base-10 for a selected Boolean option is indistinguishable from the background (--a-background-primary). The only other marker is a small dot, which is either grey (unselected) or red/green. 

People who struggle to distinguish shades of red/green often also find it difficult to distinguish these colours from greys, and so may find it difficult to understand which option is selected.

Making the background of the selected Boolean option darker makes it clear which option is selected.

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputBoolean.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputBoolean.vue
@@ -126,7 +126,7 @@ export default {
     }
 
     .apos-boolean__input:checked + & {
-      background-color: var(--a-base-10);
+      background-color: var(--a-base-8);
 
       .apos-boolean__icon {
         color: var(--a-success);


stuartromanek:
Hey @joeinnes thanks for this contribution!

This implementation increases contrast of one part of the UI (the background of the checked item) but decreases the contrast of the accompanying indicator, especially when cycling through some color blindness filters.

What do you think of using [this secondary approach with custom labels](https://ui.apos.dev/?path=/story/inputs-boolean--boolean-input&knob-Is Disabled?=&knob-Is Pre Checked?=true&knob-Custom Labels=true) as the default approach, so as to use the larger UI as the indicator and not muddle it with some smaller ornamentation.

https://ui.apos.dev/?path=/story/inputs-boolean--boolean-input&knob-Is Disabled?=&knob-Is Pre Checked?=true&knob-Custom Labels=true

joeinnes:
Love it - this is very clear, and unmistakable which is selected, regardless of how well you can discriminate between green and red. Thanks for building this out - much better than my solution!

abea:
I'll go ahead and close this.

It looks like @stuartromanek has taken this as feedback.

PR #2649, opened 2020-12-28
media lib: single tag filter

stuartromanek:
only filter a single tag at a time in media lib, remove broken transitions

satisfies https://apostrophecms.atlassian.net/browse/A30U-577?atlOrigin=eyJpIjoiYjE4NmYzZTExMGMyNDBhMDlhYzZiMTRlZThjMGZkMTAiLCJwIjoiaiJ9


abea (changes requested):
The ticket says that the filters themselves should not filter (disappear on selection), but that's happening in my test. The `update` event emitted by `AposTagList` is causing the `AposMediaManager` component to filter the tag list.

abea (approved):

abea (approved):

PR #2651, opened 2020-12-29
Updates pager to allow flexible number of intermediate pgs

abea:
Updates the pager macro to end the tyranny of the 5-item intermediary range. A `shown` option can be included in the macro to change the number of items in the middle of the pager.

See a test PoC for this: https://codepen.io/alexbea/pen/NWRRPyV?editors=0012, changing the variables at the top of the JS section.


stuartromanek (approved):
you are the pager quest hero ⚔️

PR #2656, opened 2021-01-04
Filter by media source

ValJed:


lib/modules/apostrophe-images/index.js
```diff
@@ -80,8 +80,73 @@ module.exports = {
...
+        name: 'fileType',
+        choices: fileImages.extensions.map((ext) => ({
+          value: ext,
+          label: ext.charAt(0).toUpperCase() + ext.slice(1)
+        })),
+        def: null
+      }
+    ];
+
+    self.on('apostrophe:modulesReady', 'getMediaSourcesModules', () => {
```
boutell, line 122: This filter does not belong in the core. It should be added via the `improve` mechanism in the media sources bundle. The others are fine in the core because they do not require an optional module.

boutell (changes requested):

lib/modules/apostrophe-images/index.js
ValJed: Oh ok, yes you right, I remove this PR so.

PR #2657, opened 2021-01-07
onReadyAndRefresh method that can be used to schedule widget players …

boutell:
…and anything else appropriately for "when apostrophe changes stuff".

I changed the video player to just use setTimeout 0 because that's all it was really getting out of the old onReady, which became an inner function in onReadyAndRefresh. Nothing else was calling it. However if there's a chorus of requests we can always factor it back out as a method.


modules/@apostrophecms/util/ui/public/utils.js
```diff
@@ -130,20 +130,28 @@
...
-      setTimeout(fn, 0);
-    } else if (document.addEventListener) {
-      document.addEventListener('DOMContentLoaded', fn);
-    } else {
-      document.attachEvent('onreadystatechange', function() {
-        if (document.readyState !== 'loading') {
-          fn();
-        }
-      });
+  apos.util.onReadyAndRefresh = function(fn, once) {
```
abea, line 140: The `once` argument isn't used.

abea (changes requested):

abea (approved):

PR #2665, opened 2021-01-11
3/secure drafts

boutell:


modules/@apostrophecms/i18n/index.js
```diff
@@ -97,6 +102,9 @@ module.exports = {
           locale = req.locale;
           mode = req.mode;
         }
+        if ((req.mode === 'draft') && (!self.apos.permission.can(req, 'view-draft'))) {
```
abea, line 105: This doesn't seem to be the purpose of this method.

modules/@apostrophecms/i18n/index.js
```diff
@@ -65,6 +65,11 @@ module.exports = {
         }
         req.locale = locale;
         req.mode = mode;
+        if ((req.mode === 'draft') && (!self.apos.permission.can(req, 'view-draft'))) {
```
abea, line 68: Can this be part of permissions or `doc` middleware?

abea (changes requested):

modules/@apostrophecms/i18n/index.js
boutell: We must stop the user from entering draft mode wherever it is possible for them to do that if they do not have permission to do that. This method is one of the two places where `req.mode` can be set. So this is a good place for this check, otherwise it is potentially needed in hundreds of places.

modules/@apostrophecms/i18n/index.js
boutell: Permissions should be checked immediately at the point where the danger begins, and this is one of the two places the mode can potentially be set. I do not want to move this to another handler/middleware function where a possibility exists that something will be inserted between.

abea (approved):
You know this stuff better than me, so I'll go ahead and approve. It seems like this module's methods are getting overstuffed and I worry that it's becoming too much of a catch-all.

PR #2667, opened 2021-01-11
users and other nonlocalized piece types can be edited again

boutell:
The method that sometimes infers the locale from the _id was at fault. Need to check whether it's appropriate to the type before doing that.

### DIFF
--- a/modules/@apostrophecms/i18n/index.js
+++ b/modules/@apostrophecms/i18n/index.js
@@ -76,16 +76,9 @@ module.exports = {
       },
       // Infer `req.locale` and `req.mode` from `_id` if they were
       // not set already by explicit query parameters. Conversely,
-      // if the parameters were set, rewrite `_id` accordingly.
-      // Returns `_id`, after rewriting if appropriate.
-      //
-      // If `localize` is `true`, or the name of a type that is localized,
-      // the method has its effect as described above. Otherwise it has no effect
-      // and returns `_id` unmodified.
-      inferIdLocaleAndMode(req, _id, localize) {
-        if ((localize === false) || ((localize !== true) && (!self.apos.doc.getManager(localize).isLocalized()))) {
-          return _id;
-        }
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -182,7 +182,7 @@ module.exports = {
     },
     async getOne(req, _id) {
       self.publicApiCheck(req);
-      _id = self.apos.i18n.inferIdLocaleAndMode(req, _id);
+      _id = self.apos.i18n.inferIdLocaleAndMode(req, _id, self.name);
```
abea, line 185: Wouldn't it be pretty easy to get the localization option off the piece type to pass in as a boolean here instead?

modules/@apostrophecms/i18n/index.js
```diff
@@ -78,7 +78,14 @@ module.exports = {
       // not set already by explicit query parameters. Conversely,
       // if the parameters were set, rewrite `_id` accordingly.
       // Returns `_id`, after rewriting if appropriate.
-      inferIdLocaleAndMode(req, _id) {
+      //
+      // If `localize` is `true`, or the name of a type that is localized,
+      // the method has its effect as described above. Otherwise it has no effect
+      // and returns `_id` unmodified.
+      inferIdLocaleAndMode(req, _id, localize) {
+        if ((localize === false) || ((localize !== true) && (!self.apos.doc.getManager(localize).isLocalized()))) {
```
abea, line 86: it seems weird to have an argument that is, in part at least, saying whether or not this method should have been run in the first place.

boutell:
@abea Refactored to avoid weird arguments and descriptions that add more cognitive debt than this method was intended to save. Having a wrapper at the piece level and the page level eliminates an argument and just makes this the simple ritual before using req and _id that it was intended to be. Hoping this feels better.

abea (approved):

PR #2668, opened 2021-01-11
publish vs. publish changes

boutell:


localghost443 (dismissed):

boutell:
This needs review from Bea because it got more complicated when I realized "undo published" didn't work at all yet for the first publication. That is now included in this fix because it was tightly married to the needed logic for changing this button's label dynamically to suit the circumstances.

boutell:
I'm npm testing this bad boy.

boutell:
Alas, the unit tests have spotted something iffy, taking it back to Santa's Workshop.

boutell:
Ah, it was the tests, I changed error reporting a little. All good, ready for review now honest.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -677,22 +684,24 @@ export default {
...
+        apos.notify(response.message, {
           type: 'success',
           dismiss: true
         });
         // This handler covers all "undo publish" buttons, so make sure it's
         // for the context document before altering any admin bar state
         // because of it
-        if (doc.aposDocId === (this.context && this.context.aposDocId)) {
+        const aposDocId = data._id.replace(/:.*$/, '');
+        if (data._id.replace(/:.*$/, '') === (this.moduleOptions.context && this.moduleOptions.context._id.replace(/:.*$/, ''))) {
```
abea, line 699: You have `data._id.replace(/:.*$/, '')` set to `aposDocId` now, to replace this.

abea: It's odd that `context` is on `this.moduleOptions` since it isn't, afaik, an option on the module. It's the context in which this particular instance of the component has been loaded. So `this.context` seems better. But this component already uses `this.context`, though I can't figure out where that's being set. Maybe it isn't and could be used instead?

Generally speaking it doesn't seem like `moduleOptions` is a thing that should ever be touched once set.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -677,22 +684,24 @@ export default {
...
           this.draftIsModified = true;
-          this.publishedAt = null;
-          this.moduleOptions.context.lastPublishedAt = doc.lastPublishedAt;
-          this.refreshOrReload(doc._url);
+          this.lastPublishedAt = response.lastPublishedAt;
```
abea, line 701: All these things like `lastPublishedAt` and `draftIsModified` should really be on a `this.context` object instead. The admin bar is never published or modified.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -715,15 +723,47 @@ module.exports = {
...
+      // Revert the doc `published` to its content as of its most recent
+      // previous publication. If it has only been published once ever,
+      // delete the published version of the document completely.
+      // Does not affect the related draft document.
+      //
+      // Returns `false` if it is not possible to revert because
+      // only one previous version is available and it has already
+      // been used in this way.
       async revertPublishedToPrevious(req, published) {
+        if (published.publicationCount === 1) {
```
abea, line 737: Couldn't we look for the `:previous` doc and delete if that doesn't exist? I don't like the idea of adding a `publicationCount` property onto all Apos documents for this one method when there's another option.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -659,7 +664,10 @@ module.exports = {
         }, {
           $set: {
             modified: false,
-            lastPublishedAt: new Date()
+            lastPublishedAt
```
abea, line 667: What's the difference between `updatedAt` on a published document and `lastPublishedAt`?

modules/@apostrophecms/piece-type/index.js
```diff
@@ -268,7 +268,17 @@ module.exports = {
             // Not subject to draft/publish workflow
             throw self.apos.error('invalid');
           }
-          return self.revertPublishedToPrevious(req, published);
+          const result = await self.revertPublishedToPrevious(req, published);
+          if (result) {
+            return {
```
abea, line 273: I guess the response also includes a status code? This seems purpose built for internal API, but the `else` one in particular seems to be less so for project use. Someone would need to look for `!lastPublished` or something. Wasn't it previously returning the reverted doc?

abea (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
boutell: Good catch re: the extra replace call, I removed one.

Re: `context` being on `moduleOptions`, that's because the server gives it to us as module options, we wouldn't have any other way of knowing about it.

`this.context` does not appear in the module at this point, I think you started reviewing before I caught a rogue instance of that.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
boutell: OK, I'm going to pull out `context` from moduleOptions and shallow-clone it as a deep-watched `context` and we can see how that goes. Deep watches have performance concerns and also can be a maintenance burden because people tend to forget you still can't add a new property and expect reactivity unless using Vue.set. But I agree it will be more readable, let's try it that way, will try this and push.

modules/@apostrophecms/doc-type/index.js
boutell: I considered that but I feel it is too risky, as if I'm a third party developer looking to prune the db in various ways I would not assume that deleting the "previous published" could lead to total deletion of published documents any scenario.

modules/@apostrophecms/doc-type/index.js
boutell: While we're not doing it at the moment, I foresee scenarios where direct writes to the published locale could happen, for instance to update programmatically managed data that isn't subject to draft/published workflow. This would update `updatedAt` and I want to make sure we have a separate and accurate `lastPublishedAt`.

modules/@apostrophecms/piece-type/index.js
boutell: ... Yeah.

Looking at this, I think I can go back to returning the reverted published doc, or `false` if there is no longer a published doc. And the front end can go back to being responsible for deciding what message to show. That still provides access to `lastPublishedAt`. I'll try that.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
boutell: (Fundamentally, the reason we need a reactive `this.context` is because while in principle, we're updating the page and we could refresh the page to get a new option with the latest lastPublishedAt, etc. from the server, nobody wants to refresh the page! So we can view all of these scenarios where context changes in the admin bar as an optimization, but a necessary one for a good user experience. I think the cleanest solution is to take all of `context` into reactive land at startup and stay there, yes.)

boutell:
@abea I refactored the things and retested, have a peep.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
abea: Ah, I see. I didn't see `context` on the module options somehow. It still seems like that should be referenced, but not mutated. It seems better to mutate that copy that the component owns.

`this.context` is in several places in this component, starting with L272. It's set in `onDiscardDraft`, but I don't see where it's originally set up.

abea: Nevermind that last bit. I wrote that before I digested your other comments on the other line.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -696,11 +697,41 @@ export default {
...
+        }
+      } catch (e) {
+        await apos.alert({
+          heading: 'An Error Occurred',
+          description: e.message || 'An error occurred while restoring the previously published version.'
+        });
+      }
+    },
+    async onUnpublish(data) {
+      try {
```
abea, line 725: The catch message on this is 'An error occurred while restoring the previously published version.' That should be updated for the `onUnpublish` method.

abea (changes requested):

abea (approved):

PR #2677, opened 2021-01-12
3/fix area menu close after add

stuartromanek:
https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?selectedIssue=A30U-913
- cleans up some unused event listeners
- manually closes context menu when a user adds something to a page/doc

As noted, context menus should really be able to handle this on their own but are currently relying on blurring or using the full `<AposContextMenu/>` implementation, which is being overridden and cherry picked here

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
@@ -4,6 +4,7 @@
       :button="buttonOptions"
       v-bind="extendedContextMenuOptions"
       @open="menuOpen"
+      @close="menuClose"
       ref="contextMenu"
     >
       <ul class="apos-area-menu__wrapper">
@@ -102,7 +103,7 @@ export default {
       type: Boolean
     }
   },
-  emits: [ 'menu-open', 'insert' ],
+  emits: [ 'menu-open', 'menu-close', 'insert' ],
   data() {
     return {
       active: 0,
@@ -160,11 +161,14 @@ export default {
     menuOpen(e) {
       this.$emit('menu-open', e);
     },
+    menuClose(e) {
+      this.$emit('menu-close', e);
+    },
     async add(name) {
       // Potential TODO: If we find ourselves manually flipping these bits in other AposContextMenu overrides
       // we should consider refactoring contextmenus to be able to self close when any click takes place within their el
       // as it is often the logical experience (not always, see tag menus and filters)
-      this.$refs.contextMenu.isOpen = !this.$refs.contextMenu.isOpen;
+      this.$refs.contextMenu.isOpen = false;
       if (this.widgetIsContextual(name)) {
         this.insert({
           _id: cuid(),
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -4,7 +4,7 @@
       :button="buttonOptions"
       v-bind="extendedContextMenuOptions"
       @open="menuOpen"
-      @close="menuClose"
```
abea, line 7: We're not using the close event here at all?

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -102,7 +102,7 @@ export default {
       type: Boolean
     }
   },
-  emits: [ 'menu-close', 'menu-open', 'insert' ],
```
abea, line 105: `AposAreaWidget` is still listening for `@menu-close`

abea (changes requested):

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
stuartromanek: you are correct this is still in play, re implemented

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
stuartromanek: re implemented

modules/@apostrophecms/area/ui/apos/components/AposAreaMenuItem.vue
```diff
@@ -48,42 +48,8 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.apos-tag-list__button {
-  @include apos-button-reset();
-  display: flex;
-  align-items: center;
-  padding: 7.5px 10px;
-  border-radius: 5px;
```
stuartromanek, line 56: unrelated cleanup of copypasta'd template guts

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -164,6 +165,10 @@ export default {
       this.$emit('menu-open', e);
     },
     async add(name) {
+      // Potential TODO: If we find ourselves manually flipping these bits in other AposContextMenu overrides
+      // we should consider refactoring contextmenus to be able to self close when any click takes place within their el
+      // as it is often the logical experience (not always, see tag menus and filters)
+      this.$refs.contextMenu.isOpen = false;
```
stuartromanek, line 171: @abea I also changed this to be explicitly false (closed) because at this stage we always want it closed, truly toggling it felt overly cute and possibly bug prone

abea (approved):

PR #2678, opened 2021-01-12
do not block page when relationship autocomplete is searching

stuartromanek:
https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4

This causing flashing in Safari and is not an intended experience anyway.

requesting Tommy 'lock the page' Boutell for the official blessing

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -108,7 +108,6 @@ export default {
       if (!this.searching) {
         if (this.searchTerm.length) {
           this.searching = true;
-          console.log('search baby');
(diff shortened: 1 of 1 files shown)


boutell:
Sure that's fine, but how about some kind of local busy indicator instead then, in case it is slow?

And why do we have any flashing concerns — shouldn't our global busy indicator fade in slow, like it does in A2, so if it's quick you never see it?

boutell (approved):
Sure that's fine in that there's no huge problem if we don't lock the page for this one. But how about some kind of local busy indicator instead then, in case it is slow?

And why do we have any flashing concerns — shouldn't our global busy indicator fade in slow, like it does in A2, so if it's quick you never see it?

boutell:
Am approving but would like your thoughts on those followup concerns.

stuartromanek:
a local busy indicator would be swell, i can make an issue for that.

To be totally honest I haven't looked at the implementation of our new busy but I've noted it feeling very harsh in other places, I think generally it needs another look

PR #2684, opened 2021-01-13
Fix the headers bug

boutell:
Our default styles were making it to our UI controls, but not to our tiptap extensions, so the styles extension didn't know what to make of the existing markup and reverted to the default p tag.


localghost443 (approved):

localghost443 (approved):

boutell:
This is an ordinary fix among other ordinary fixes crawling toward beta 1, not a hotfix, since we were able to fix the demo to not rely upon it.

PR #2687, opened 2021-01-15
responsive video widget

stuartromanek:
https://apostrophecms.atlassian.net/jira/software/projects/A30U/boards/4?assignee=5d9f805f30387c0dbe48415e&selectedIssue=A30U-852

- add window resize handler when we size video
- move video widget player player to `video-widget` module

### DIFF
--- a/modules/@apostrophecms/video-widget/ui/public/video.js
+++ b/modules/@apostrophecms/video-widget/ui/public/video.js
@@ -54,6 +54,10 @@ apos.util.widgetPlayers['@apostrophecms/video'] = {
         if (result.width && result.height) {
           inner.style.width = '100%';
           inner.style.height = ((result.height / result.width) * inner.offsetWidth) + 'px';
+          // If we need to iniitally size the video, also resize it on window resize.
+          window.addEventListener('resize', function() {
+            inner.style.height = ((result.height / result.width) * inner.offsetWidth) + 'px';
+          });
         } else {
           // No, so assume the oembed HTML code is responsive.
         }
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/video-widget/ui/public/video.js
```diff
@@ -54,6 +54,10 @@ apos.util.widgetPlayers['@apostrophecms/video'] = {
         if (result.width && result.height) {
           inner.style.width = '100%';
           inner.style.height = ((result.height / result.width) * inner.offsetWidth) + 'px';
+          // If we need to iniitally size the video, also resize it on window resize.
+          window.addEventListener('resize', function() {
+            inner.style.height = ((result.height / result.width) * inner.offsetWidth) + 'px';
```
boutell, line 59: This handler should check that `inner` is still in the DOM and if not it should unhook the listener and politely return.

boutell: Also, suggest factoring the logic into a little function that can be called twice rather than repeated.

boutell: Looks like `document.contains` works at least back to IE11.

boutell (changes requested):

abea (changes requested):
What Tom said.

modules/@apostrophecms/video-widget/ui/public/video.js
```diff
@@ -53,13 +56,28 @@ apos.util.widgetPlayers['@apostrophecms/video'] = {
...
         }
       }, 0);
     }
 
+    function resizeVideo(canvasEl) {
+      canvasEl.style.height = ((queryResult.height / queryResult.width) * canvasEl.offsetWidth) + 'px';
+    };
+
+    function resizeHandler() {
+      const canvas = el.querySelector('[data-apos-video-canvas]');
```
boutell, line 73: This tests whether the youtube player is still inside the widget. I don't see how that could change. My concern is that the entire widget may no longer be in the DOM (for instance, because it has been deleted).

boutell (changes requested):

boutell (approved):

PR #2693, opened 2021-01-20
3/resurrect npm bundle functionality

### LINKED ISSUE #2686: 3.0: moogBundle is not a valid top level property
## To Reproduce

1. Add `moogBundle` property to any A3 module
2. Look at the CLI output (app crashed) with message something along the lines `moogBundle is not a valid top level property`

## Expected behavior
Accept moogBundle property as a valid one

## Describe the bug
moogBundle is required for bundled modules (such as Apostrophe Blog). There is a trace in https://github.com/apostrophecms/apostrophe/blob/3.0.0-alpha.3/index.js#L184 of the functionality and indeed mogBundle is required to be a top level property, but it seems the module validator is not accepting this prop.

## Details

**Version of Node.js:**
any

**Server Operating System:**
any

**Apostrophe CMS version**
3.0.0-alpha.3

myovchev:
Fixes #2686

### DIFF
--- a/index.js
+++ b/index.js
@@ -181,11 +181,11 @@ module.exports = async function(options) {
         return;
       }
       const module = require(path);
-      if (module.moogBundle) {
+      if (module.bundle) {
         self.options.bundles = (self.options.bundles || []).concat(name);
-        _.each(module.moogBundle.modules, function(name) {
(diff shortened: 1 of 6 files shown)


boutell (approved):
Thank you!

PR #2695, opened 2021-01-21
Add alt field to apostrophe-images schema

ValJed:


ValJed:
@abea @boutell 
After discussion with @abea and after reading this issue about alt tag. I just want to be sure everything is ok for news and legacy projects.
- It must be possible not to fill alt attribute. (So I guess we cannot use or operation in alt attributes)
- If I understood, some projects use title for alt, some others use description. But actually in images-widget, the title is used for alt attributes. So we cannot run the same migration for everyone.
- Should we let them configure which field they wanna use ?

abea:
Identifying improvements to make alt text work well for new project is probably fairly simple. There are various ways to do that, but ultimate it would involve having a field that's clearly labeled for alt text and using that field in the image widget template.

The challenge is not breaking legacy projects when they simply `npm update apostrophe`. As you point out due to previous confusion and lack of documentation, we cannot assume what field had been used for alt text before. We also can't really depend on them configuring what field to use, since we haven't done that before and that would be a BC break.

We _can_ implement a 🆕 configuration that activates an improved alt text experience. This is what I suggest as of this moment:

- We add an `enableAltField` option to `apostrophe-images`.
- When set to `true`, that adds an optional `alt` field to images, and anywhere in core that uses apos-images-widgets would use `alt` for its alt attributes.
  - It would probably be good to give help text guidance for the other image fields (description, caption, title), but that is a secondary concern.
- There is no migration.
- The `apostrophe-boilerplate` would be updated to enable that option by default.
- Documentation would be updated to cover the legacy behavior, but suggest people switch to the new option
- If a project is not using that option, we use a `warnDev` notification to suggest they add it, linking to the docs.

That's not terribly complicated, but it is several steps of work that haven't been priorities.

ValJed:
@abea Thank you for your feedback.
Your solution sounds painless, and ensuring that legacy projects won't have problem with this.
I'll follow the step you described.

lib/modules/apostrophe-images/index.js
```diff
@@ -104,6 +118,10 @@ module.exports = {
   },
   construct: function(self, options) {
 
+    if (!options.enableAltField) {
+      self.apos.utils.warnDev('⚠️ We recommend to add enableAltField: true in apostrophe-images options, see https://docs.apostrophecms.org/reference/modules/apostrophe-images/#apostrophe-images');
```
abea, line 122: Language suggestion:
> We recommend that you add `enableAltField: true` in apostrophe-images options to use a new `talt text field on images. See https://docs.apostrophecms.org/reference/modules/apostrophe-images/#apostrophe-images

abea (changes requested):
- See my comment on the documentation PR. You'll need to move the feature documentation here. https://github.com/apostrophecms/apostrophe-documentation/pull/338#pullrequestreview-580846719
- Please update the changelog and bump the package version's minor version number.

lib/modules/apostrophe-images/index.js
ValJed: Modified

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.114.1 (2021-02-02)
```
abea, line 3: This adds a new feature, so it should be a minor version increase, to `2.115.0`

abea (changes requested):

abea (approved):

PR #2696, opened 2021-01-21
3/identify unpublished pages in tree

stuartromanek:
https://apostrophecms.atlassian.net/browse/A30U-920?atlOrigin=eyJpIjoiZjZlNmQ2YWRmZjIyNDdkZTk1N2MwNTkwMjNmOTY5MDciLCJwIjoiaiJ9

First part of this ^ .. second part should be its own ticket as it would (probably) impact the pages `getAll` query more significantly than adding an additional projection property.

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
+++ b/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
@@ -5,7 +5,6 @@ import { detectDocChange } from 'Modules/@apostrophecms/schema/lib/detectChange'
 // Vue 3 using the composition API. - AB
 
 import klona from 'klona';
-import _ from 'lodash';
 
 export default {
   data() {
@@ -155,7 +154,8 @@ export default {
           }
         });
 
-        _.uniq(temp).forEach(icon => {
+        // Include only unique in final object
+        Array.from(new Set(temp)).forEach(icon => {
           icons[icon] = `${icon.toLowerCase()}-icon`;
         });
 
--- a/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
+++ b/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
@@ -5,6 +5,7 @@ import { detectDocChange } from 'Modules/@apostrophecms/schema/lib/detectChange'
 // Vue 3 using the composition API. - AB
 
 import klona from 'klona';
+import _ from 'lodash';
 
 export default {
   data() {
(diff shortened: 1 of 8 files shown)


modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -84,6 +84,14 @@ export default {
             label: 'Page Title',
             name: 'title'
           },
+          {
+            label: 'Published',
+            name: 'lastPublishedAt',
+            value: {
+              true: '',
+              false: 'Unpublished'
```
abea, line 92: Maybe this should be `indicator`? `value` is more of a computer-read property. I realize `label` is taken here.

abea (commented):
It is clever. Given that this is the visible label for the value, `value` is striking me as not quite right, but there's not a perfect alternative that comes right to mind.

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
stuartromanek: yeah value didn't seem right to me .. what about `label` being able to be both an object and a string and figuring it out down there at the template?

abea: This is describing the column. `columnHeader` could be appropriate, [based on the related aria-role name](https://w3c.github.io/aria/#columnheader). Then `label` is freed up. Maybe it'd be good to specify both better: `columnHeader` or `header` and then `cellLabel`.

stuartromanek: oo that would not really be that same thing ... maybe `label` should change to `heading` to be more in line with how a table gets constructed?

then i could take `label` for this

abea: Simply making `label` do more work doesn't seem right.

stuartromanek: fine w the explicit versions of `columnHeader` and `cellLabel`

boutell: I like the improved property names.

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -5,6 +5,7 @@ import { detectDocChange } from 'Modules/@apostrophecms/schema/lib/detectChange'
 // Vue 3 using the composition API. - AB
 
 import klona from 'klona';
+import _ from 'lodash';
```
boutell, line 8: Use `Array.from(new Set(array))` to unique-ify an array without hauling in lodash. isObject can be `(typeof foo === 'object')`, isString can be `(typeof foo === 'string')`.

boutell (changes requested):

boutell (approved):

PR #2698, opened 2021-01-21
apos object should be available in macros without "with context", sam…

### LINKED ISSUE #2692: 3.0: nunjucks helpers not found inside of a macro
## To Reproduce

```sh
git clone git@github.com:myovchev/a3-helpers.git && cd $_
```
Detailed instructions and info are available in the [README](https://github.com/myovchev/a3-helpers). I'm providing a copy here as well:
1. Be sure you have running mongodb on default port. Database `a3-helpers` will be auto-created.
1. Run the project in dev mode (`npm i && npm run dev`). Ignore warnings, no need of user/session secret.
2. You should see home page rendered with additional `Helper Test` section.
3. Open `modules/@apostrophecms/home-page/views/page.html` and uncoment around line 23 `<p>{{ helper.render("Same helper called from inside of a macro will fail") }}</p>`
4. You should see error page, look at the CLI output.

Changes made to the original a3 boilerprate to reproduce the issue:
- added `helper.include.html`, `helper.macro.html`, `plain.macro.html` in `modules/@apostrophecms/home-page/views/`
- modified `modules/@apostrophecms/home-page/views/page.html` to include the macros and render additional section reproducing the issue in question
- added `alias` and dummy helper in `modules/@apostrophecms/home-page/index.js`
- `app.js` - shortName renamed to `a3-helpers`
- `package.json`: removed build from `npm run dev` command, added `postinstall` for initial build and better startup time

## Expected behavior
Execute the module nunjucks helper without an error.

## Describe the bug
Registered via a3 module helpers are not available inside of any nunjucks macro.

## Details
An example call `apos.moduleName.someHelper()` raises error:  
`Error: Unable to call `apos["moduleName"]["someHelper"]`, which is undefined or falsey`

**Version of Node.js:**
v12.8.1

**Apostrophe CMS version**
3.0.0-alpha.3

**Server Operating System:**
```sh
 > lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```
No docker involved. Node env provided by `nvm`.

boutell:
…e for getOption and module

closes #2692

### DIFF
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -255,17 +255,6 @@ module.exports = {
         const args = {};
 
         args.data = merged;
-        args.module = self.templateApos.modules[module.__meta.name];
-        args.getOption = function(key, def) {
-          const colonAt = key.indexOf(':');
-          let optionModule = module;
-          if (colonAt !== -1) {
-            const name = key.substring(0, colonAt);
-            key = key.substring(colonAt + 1);
-            optionModule = self.apos.modules[name];
-          }
-          return optionModule.getOption(req, key, def);
-        };
 
         // // Allows templates to render other templates in an independent
         // // nunjucks environment, rather than including them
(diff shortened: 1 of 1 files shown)


boutell:
See #2692 

In A3 we are injecting the template apos object, via which helpers are accessed, as an argument to the original template. But this does not expose it to macro imports, unless "with context" is used.

Since apos is a singleton, there is no reason we can't safely expose it to macro imports by treating it as a nunjucks global rather than an argument. So I made that change here.

I did the same for "getOption" and "module". These are lesser known, but they are still singletons as far as the module calling the template is concerned, because *we create and reuse one nunjucks environment per module name per HTTP request.*

I did not attempt to do this for `data` because that would not make sense. Templates can invoke other templates via async components, and these would have their own `data`, so it cannot be shared in the same per-request, per-module environment. Also `data` is a reasonable and even pretty common name for a parameter in a macro, so we don't want it just shoving in there. If people really want it they can use `with context`.

abea (approved):

PR #2701, opened 2021-01-25
3/range field

stuartromanek:
For testing you can use the schema
```
      myRange: {
        type: 'range',
        label: 'My home on the range',
        help: 'This is a slider range and its alright',
        min: -10,
        max: 100,
        step: 2,
        unit: 'px',
      },
```

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -603,8 +603,8 @@ module.exports = {
         // `min` here does not imply requirement, it is the minimum value the range UI will represent
         if (
           !data[field.name] ||
-          data[field.name] < parseFloat(field.min) ||
-          data[field.name] > parseFloat(field.max)
+          data[field.name] < field.min ||
+          data[field.name] > field.max
         ) {
           object[field.name] = null;
         }
@@ -616,16 +616,16 @@ module.exports = {
         if (!field.max) {
           fail('Property "max" must be set.');
         }
-        if (isNaN(parseFloat(field.max))) {
-          fail('Property "max" must parse to a float');
+        if (typeof field.max !== 'number') {
+          fail('Property "max" must be a number');
         }
-        if (isNaN(parseFloat(field.min))) {
-          fail('Property "min" must parse to a float.');
+        if (typeof field.min !== 'number') {
+          fail('Property "min" must be a number');
         }
-        if (field.step && isNaN(parseFloat(field.step))) {
-          fail('Property "step" must parse to a float.');
+        if (field.step && typeof field.step !== 'number') {
+          fail('Property "step" must be a number.');
         }
-        if (field.unit && (typeof field.unit) !== 'string') {
+        if (field.unit && typeof field.unit !== 'string') {
           fail('Property "unit" must be a string.');
         }
       }
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -601,7 +601,11 @@ module.exports = {
         }
         // Allow for ranges to go unset
         // `min` here does not imply requirement, it is the minimum value the range UI will represent
-        if (!data[field.name] || data[field.name] < parseFloat(field.min)) {
+        if (
+          !data[field.name] ||
+          data[field.name] < parseFloat(field.min) ||
+          data[field.name] > parseFloat(field.max)
+        ) {
           object[field.name] = null;
         }
       },
@@ -613,13 +617,13 @@ module.exports = {
           fail('Property "max" must be set.');
         }
         if (isNaN(parseFloat(field.max))) {
-          fail('Property "max" must be an integer.');
+          fail('Property "max" must parse to a float');
         }
         if (isNaN(parseFloat(field.min))) {
-          fail('Property "min" must be an integer.');
+          fail('Property "min" must parse to a float.');
         }
         if (field.step && isNaN(parseFloat(field.step))) {
-          fail('Property "step" must be an integer.');
+          fail('Property "step" must parse to a float.');
         }
         if (field.unit && (typeof field.unit) !== 'string') {
           fail('Property "unit" must be a string.');
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -592,7 +592,7 @@ module.exports = {
       name: 'range',
       vueComponent: 'AposInputRange',
       convert: async function (req, field, data, object) {
-        object[field.name] = self.apos.launder.integer(data[field.name], field.def, field.min, field.max);
+        object[field.name] = self.apos.launder.float(data[field.name], field.def, field.min, field.max);
         if (field.required && (_.isUndefined(data[field.name]) || !data[field.name].toString().length)) {
           throw self.apos.error('required');
         }
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -239,29 +239,6 @@ module.exports = {
     });
 
     self.addFieldType({
-      name: 'range',
-      convert: async function (req, field, data, object) {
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -595,6 +588,45 @@ module.exports = {
...
+        object[field.name] = self.apos.launder.float(data[field.name], field.def, field.min, field.max);
+        if (field.required && (_.isUndefined(data[field.name]) || !data[field.name].toString().length)) {
+          throw self.apos.error('required');
+        }
+        if (data[field.name] && isNaN(parseFloat(data[field.name]))) {
+          throw self.apos.error('invalid');
+        }
+        // Allow for ranges to go unset
+        // `min` here does not imply requirement, it is the minimum value the range UI will represent
+        if (!data[field.name] || data[field.name] < parseFloat(field.min)) {
```
abea, line 604: Wouldn't it be invalid if the value was less than the minimum? Is this how we're doing it in 2.x?

modules/@apostrophecms/schema/index.js
```diff
@@ -595,6 +588,45 @@ module.exports = {
...
+        if (!field.min) {
+          fail('Property "min" must be set.');
+        }
+        if (!field.max) {
+          fail('Property "max" must be set.');
+        }
+        if (isNaN(parseFloat(field.max))) {
+          fail('Property "max" must be an integer.');
+        }
+        if (isNaN(parseFloat(field.min))) {
```
abea, line 618: The message is that it must be an integer, but this (and above) is testing for float.

abea: HTML range inputs aren't limited to integers, so I don't know why ours should be.

modules/@apostrophecms/schema/index.js
```diff
@@ -595,6 +588,45 @@ module.exports = {
...
+          fail('Property "min" must be an integer.');
+        }
+        if (field.step && isNaN(parseFloat(field.step))) {
+          fail('Property "step" must be an integer.');
```
abea, line 622: You're testing for float, but the message is integer. We should allow float values as well.

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
```diff
@@ -0,0 +1,248 @@
...
+  >
+    <template #body>
+      <div class="apos-input-wrapper">
+        <div class="apos-range-input">
+          <input
+            type="range"
+            :min="field.min"
+            :max="field.max"
+            :step="field.step"
+            class="apos-range"
```
abea, line 14: `apos-range-input__field`? Or `__input`?

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
```diff
@@ -0,0 +1,248 @@
...
+            v-model="next"
+            :id="uid"
+          >
+          <div class="apos-range-input__scale">
+            <span class="apos-range-input__scale-value apos-range-input__scale-value--min">
```
abea, line 19: You don't use the `--min` and `--max` classes. You could also get away with first-child and last-child if you needed to. Let's remove these classes.

abea: Also doesn't look like `apos-range-input__scale-value` are used. I guess the spans are needed for flexing. If they're there we should have some visually hidden labeling (Min: Max:)

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
```diff
@@ -0,0 +1,248 @@
...
+              {{ minLabel }}
+            </span>
+            <span class="apos-range-input__scale-value apos-range-input__scale-value--max">
+              {{ maxLabel }}
+            </span>
+          </div>
+        </div>
+        <div class="apos-range-value" :class="{'is-unset': !isSet}">
+          {{ valueLabel }}
```
abea, line 28: I'm wondering if this should be hidden from AT. The input itself will announce the value and then you don't need more labeling to distinguish from the min/max values.

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
```diff
@@ -0,0 +1,248 @@
...
+  .apos-range:focus::-ms-fill-lower {
+    background: var(--a-base-7);;
+  }
+
+  .apos-range:focus::-ms-fill-upper {
+    background: var(--a-base-7);;
+  }
+
+  /*TODO: Use one of the selectors from https://stackoverflow.com/a/20541859/7077589 and figure out
+  how to remove the virtical space around the range input in IE*/
```
abea, line 240: "vertical"

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
```diff
@@ -0,0 +1,248 @@
...
+
+  .apos-range:focus::-ms-fill-lower {
+    background: var(--a-base-7);;
+  }
+
+  .apos-range:focus::-ms-fill-upper {
+    background: var(--a-base-7);;
+  }
+
+  /*TODO: Use one of the selectors from https://stackoverflow.com/a/20541859/7077589 and figure out
```
abea, line 239: Do we need to fuss with little alignment bits in IE11?

abea (changes requested):

modules/@apostrophecms/schema/index.js
stuartromanek: `parseFloat` won't error if you give it a stringified float .. you want to change the error language to something akin to that? (sounds wordy which is why i didnt bother)

abea: I'm saying if we're requiring a float, the message should say that it must be a float value. This error is going to the dev, who can/should know the difference.

modules/@apostrophecms/schema/index.js
stuartromanek: we are using an out-of-range value as a fake null value, which gets converted here.

There is no UI for the editor to force an out-of-range value so this feels safe to squash and reset back to null, no?

abea: Sounds good.

stuartromanek: i will add a greater-than-max check as well, for completeness

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
stuartromanek: axed

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
stuartromanek: axed

modules/@apostrophecms/schema/ui/apos/components/AposInputRange.vue
stuartromanek: cleaned all this up to start with `apos-range`

modules/@apostrophecms/schema/index.js
boutell: Also, why are we parsing here? It should be an actual float. We don't need to accept strings for numeric values as part of schema field definitions.

boutell: So typeof should be `number` rather than doing any parsing.

abea (approved):

modules/@apostrophecms/schema/index.js
stuartromanek: I was being soft on this because technically range inputs accept strings buuuutt we do not need to support both, clamped these down

abea (approved):

PR #2718, opened 2021-02-08
Webpack 5

boutell:
Webpack 5 and its consequences.

Also: a better error message for nonexistent page type modules, correct merging of options at app.js and index.js level (like 2.x), and no reliance on setImmediate in the browser especially where polyfills might not be present.

We needed all of these (except the nice error message) for this week's priority project.


boutell:
Oh excuse me, and of course the big one, in terms of files touched: no storybook.

modules/@apostrophecms/asset/index.js
```diff
@@ -6,18 +6,6 @@ const globalIcons = require('./lib/globalIcons');
 const path = require('path');
 const express = require('express');
 
-// Wrap webpack to report its errors readably, in
-// preparation to promisify it
-
-function webpack(config, cb) {
```
boutell, line 12: No longer necessary because webpackModule's function signature is now correct to pass directly to promisify.

modules/@apostrophecms/asset/lib/webpack.config.js
```diff
@@ -1,5 +1,5 @@
 const path = require('path');
-const merge = require('webpack-merge');
+const merge = require('webpack-merge').merge;
```
boutell, line 2: New version of webpack-merge offers no "smart" method, suggests this way of requiring and using. Seems OK so far.

modules/@apostrophecms/asset/lib/webpack.scss.js
```diff
@@ -6,17 +6,39 @@ module.exports = (options, apos) => {
...
         {
           test: /\.css$/,
-          use: [ 'vue-style-loader', 'css-loader' ]
+          use: [
+            'vue-style-loader',
+            // https://github.com/vuejs/vue-style-loader/issues/46#issuecomment-670624576
+            {
+              loader: 'css-loader',
+              options: {
+                esModule: false
```
boutell, line 15: Required for vue-style-loader compatibility with webpack 5.

modules/@apostrophecms/asset/lib/webpack.scss.js
```diff
@@ -6,17 +6,39 @@ module.exports = (options, apos) => {
...
+              }
+            },
             {
               loader: 'postcss-loader',
               options: {
-                plugins: () => [ require('autoprefixer') ]
+                postcssOptions: {
+                  plugins: [
+                    [
+                      'autoprefixer',
```
boutell, line 37: Pulls in plugins differently with webpack 5.

lib/moog-require.js
```diff
@@ -121,6 +121,12 @@ module.exports = function(options) {
     };
 
     _.defaults(definition, projectLevelDefinition);
+    // app.js wins. However if an option exists in index.js and not
+    // in app.js, that should be honored, like it was in 2.x when
```
abea, line 125: We shouldn't compare to 2.x in code.

abea (changes requested):
Other than that comment, it looks fine. I haven't tested it in projects with different webpack versions, so I'm trusting you've done that.

boutell:
Rewrote the comment and re-requested review.

To be clear, there is no support for multiple versions of webpack right now. Webpack 5 must be used at project level also, due to the upstream issue with vue-loader. Tickets exist to try to contribute to the resolution of that problem in the future.

abea (approved):

PR #2719, opened 2021-02-08
3/date min max

abea:
Adds date min/max validation. Also allows dates to be unset.

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -519,7 +519,29 @@ module.exports = {
       name: 'date',
       vueComponent: 'AposInputString',
       convert: async function (req, field, data, object) {
+        if (!data[field.name] && object[field.name]) {
+          // Allow date fields to be unset.
+          object[field.name] = null;
+          return;
+        }
+        if (field.min && dayjs(data[field.name]).isBefore(field.min)) {
+          // If the min requirement isn't met, leave as-is.
+          return;
+        }
+        if (field.max && dayjs(data[field.name]).isAfter(field.max)) {
+          // If the max requirement isn't met, leave as-is.
+          return;
+        }
+
         object[field.name] = self.apos.launder.date(data[field.name], field.def);
+      },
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -519,8 +519,30 @@ module.exports = {
       name: 'date',
       vueComponent: 'AposInputString',
       convert: async function (req, field, data, object) {
+        if (!data[field.name] && object[field.name]) {
+          // Allow date fields to be unset.
+          object[field.name] = null;
+          return;
+        }
+        if (field.min && dayjs(data[field.name]).isBefore(field.min)) {
```
boutell, line 527: The nice thing about YYYY-MM-DD strings is that you can compare them directly, there's no need for dayjs for this purpose.

modules/@apostrophecms/schema/index.js
```diff
@@ -519,8 +519,30 @@ module.exports = {
...
+        }
+        if (field.max && dayjs(data[field.name]).isAfter(field.max)) {
+          // If the max requirement isn't met, leave as-is.
+          return;
+        }
+
         object[field.name] = self.apos.launder.date(data[field.name], field.def);
       },
+      validate: function (field, options, warn, fail) {
+        if (field.max && !dayjs(field.max).isValid()) {
```
boutell, line 539: dayjs is nice here though.

boutell (changes requested):

modules/@apostrophecms/schema/index.js
abea: This is true, but unless we enforce `YYYY-MM-DD` for `min/max` values, which we could do, we'd need the help here.

boutell: We should enforce that for min/max I think. We don't want to turn dayjs into our official standard of what inputs are allowed, because then we're on dayjs forever. YYYY-MM-DD is what we store and what we allow for configuration, although we might allow other things in the UI.

boutell (approved):

PR #2724, opened 2021-02-11
Fixes chooser grid hidden footer

### LINKED ISSUE #2722: v2.114.0 piece selection modal pagination not visible
## To Reproduce

> 1. Create 11 pieces, of any kind.
> 2. Create a widget that accepts pieces. (have the ability to select pieces individually)
> 3. Select "Individually" for the piece selector.
> 4. Click on "Browser"
> 5. Look at the bottom left corner.

## Expected behavior

The pagination should be visible.

## Describe the bug

The moment you have a piece selector, at least 11 pieces, you try to select pieces individually and hit browse, you won't be able to see the pagination.
This works in version 2.103.1 but after I updated, directly, to 2.114.0 I've noticed that the pagination is missing.

The problem is that `div.apos-chooser` overflows on top of the `div.apos-modal-footer` thus making the pagination not visible.

## Details

**Version of Node.js:**
v12.0.0

**Server Operating System:**
Windows 10, CentOS

**Screenshots**
Apos-Pagination-Bug

abea:
Resolves #2722

Adds a new modal body class block to better target specific modal cases.

### DIFF
--- a/lib/modules/apostrophe-pages/public/css/user.less
+++ b/lib/modules/apostrophe-pages/public/css/user.less
@@ -44,9 +44,15 @@
   }
 }
 
+.apos-modal-body--reorganize-tree {
+  display: flex;
+  flex-direction: row;
+  overflow-y: hidden;
+}
+
 .apos-modal .apos-reorganize-tree--chooser {
   width: 80%;
-  float: right;
+
   @media @apos-breakpoint-desktop-xl { width: 75%; }
 }
 
--- a/lib/modules/apostrophe-doc-type-manager/public/css/chooser.less
+++ b/lib/modules/apostrophe-doc-type-manager/public/css/chooser.less
@@ -1,16 +1,25 @@
 .apos-ui
 {
+
+  .apos-modal-body--chooser {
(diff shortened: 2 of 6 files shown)


lib/modules/apostrophe-pages/public/css/user.less
```diff
@@ -44,9 +44,15 @@
   }
 }
 
+.apos-modal-body--reorganize-tree {
+  display: flex;
+  flex-direction: row;
+  overflow-y: hidden;
+}
+
 .apos-modal .apos-reorganize-tree--chooser {
```
abea, line 53: This also sits next to `.apos-chooser`.

gregvanbrug (changes requested):
Seeing a gap between the dialog content and the footer. Looks like adjusting the inline style to `height: calc(100% - 60px)` on `apos-modal-body` and `apos-chooser` would fix it. Otherwise looks good.

Screen Shot 2021-02-17 at 1 52 07 PM

boutell (approved):

PR #2727, opened 2021-02-12
Fixes min/max input issues

abea:
A few input bits I found while writing docs.

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
@@ -132,7 +132,8 @@ export default {
       const minMaxFields = [
         ...minMaxNumFields,
         'string',
-        'password'
+        'password',
+        'slug'
       ];
 
       if (this.field.min && minMaxFields.includes(this.field.type)) {
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
@@ -8,7 +8,7 @@
       <div class="apos-input-wrapper">
         <textarea
           :class="classes"
-          v-if="field.textarea" rows="5"
+          v-if="field.textarea && field.type === 'string'" rows="5"
           v-model="next" :placeholder="field.placeholder"
           @keydown.enter="enterEmit"
           :disabled="field.disabled" :required="field.required"
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
@@ -132,8 +132,7 @@ export default {
       const minMaxFields = [
         ...minMaxNumFields,
         'string',
-        'password',
-        'slug'
+        'password'
       ];
 
       if (this.field.min && minMaxFields.includes(this.field.type)) {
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -55,7 +55,7 @@ export default {
         if (this.field.type === 'float' || this.field.type === 'integer') {
           return 'number';
         }
-        if (this.field.type === 'string') {
+        if (this.field.type === 'string' || this.field.type === 'slug') {
```
abea, line 58: This will go into the `type` attribute on the input. There's no `type="slug"` input in HTML.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -121,17 +121,30 @@ export default {
           return 'required';
         }
       }
-      if (this.field.min) {
-        if (this.type === 'number' || this.type === 'date') {
+
+      const minMaxNumFields = [
```
abea, line 125: This makes it clearer and prevents us from needing to jam `range` in there some other way.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -121,17 +121,30 @@ export default {
...
+        'date',
+        'integer',
+        'float',
+        'range'
+      ];
+
+      const minMaxFields = [
+        ...minMaxNumFields,
+        'string'
```
abea, line 134: We should limit `min/max` use more. Emails shouldn't have min/max limitations. Nor URLs. Maybe passwords and slugs? I'd rather start more limited.

boutell (approved):

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
boutell: I see no reason to forbid min on passwords or slugs. Clearly max on passwords is kind of crappy, but the field is legitimate for uses other than our own passwords and they may have constraints and it doesn't serve the user to let them enter something that won't work with system X (that the field is being passed to) if we can help them by rejecting it. So I think passwords should support both min and max, and slugs too.

boutell (approved):

boutell (changes requested):
Requesting a change to allow min/max on slugs and passwords.

boutell (dismissed):

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
abea: Actually, slugs are less obvious. In the field's `convert` function we do a certain amount of mutation to the slug. We run `slugify` as well as adding and removing slashes if it's a page. Do we match the min/max lengths before or after all of that?

It's not likely that min/max would be used with the page option, but even with the basic version you could be converting multiple spaces and punctuation to a single dash.

abea:
See comment on slug fields: https://github.com/apostrophecms/apostrophe/pull/2727#discussion_r576307433

boutell (approved):

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
boutell: Good points. I would apply min/max to slugs at the end.

boutell (changes requested):
I guess I'm requesting a change because I responded that slug length should be enforced after all transformations.

abea:
Right on. That must have got hidden from the commit pushes.

boutell (approved):

PR #2735, opened 2021-02-18
we were using yargs, but not in a way that leveraged it in any useful…

boutell:
… degree, I think it was an oversight. We want the boring parser, with the end option so we can do passthrough to other commands from a task

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 const path = require('path');
 const _ = require('lodash');
-const argv = require('yargs').argv;
+const argv = require('boring')({ end: '--' });
 const fs = require('fs');
 const npmResolve = require('resolve');
 let defaults = require('./defaults.js');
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
     "bluebird": "^3.7.2",
     "body-parser": "^1.18.2",
+    "boring": "^1.0.0",
```
abea, line 37: `^1.1.0`

index.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 const _ = require('lodash');
-const argv = require('yargs').argv;
+const argv = require('boring')({ end: '--' });
```
abea, line 3: Now set to `true`.

abea (changes requested):

abea (approved):

PR #2738, opened 2021-02-18
Updates the klona major version

abea:
Updates klona to 2.x. Fewer files changed than I expected, to be honest.


boutell (changes requested):
We also have a bunch of `import` statements in vue code / frontend js code.

abea:
I knew that was too few.

boutell (approved):

PR #2740, opened 2021-02-19
the discard draft button appeared in error on parked pages because th…

boutell:
…e mechanism to also insert a draft anytime we directly insert a published document did not set lastPublishedAt. Lock the barn door, and hunt down the missing cows.


abea (approved):
Looks good. I look forward to moving all alpha migrations into a module.

PR #2741, opened 2021-02-19
1046 optimization of db queries

ValJed:
Modify `updateDocReferences` method to avoid processing attachments  update when none could exist here. To do that we parse the doc schema recursively to check that no field has an `attachment` type .

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -861,8 +861,7 @@ module.exports = function(self, options) {
   // Or if the document schema doesn't allow attachments.
 
   self.updateDocReferences = function(doc, callback) {
-    // Adding a 's' for plural
-    const docModule = self.apos.modules[doc.type] || self.apos.modules[`${doc.type}s`];
+    const docModule = self.apos.docs.getManager(doc.type);
 
     if (!docModule) {
       return callback();
@@ -963,7 +962,6 @@ module.exports = function(self, options) {
         return self.db.update(criteria, query, callback);
       }, callback);
     }
-
   };
 
   // We check that no attachment is possible for a doc by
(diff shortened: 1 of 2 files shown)


ValJed:
Should I bump the package version ?

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -856,8 +856,23 @@ module.exports = function(self, options) {
   //
   // This method is invoked after any doc is inserted, updated, trashed
   // or rescued.
+  // We don't go further if the document has no module associated to its type
+  // (ex: pages types are not associated to modules, but they don't have attachments)
+  // Or if the document schema doesn't allow attachments.
 
   self.updateDocReferences = function(doc, callback) {
+    // Adding a 's' for plural
+    const docModule = self.apos.modules[doc.type] || self.apos.modules[`${doc.type}s`];
```
boutell, line 865: Instead, use `self.apos.docs.getManager(doc.type)` to get the manager object. Then you can refer to the `schema` property of the manager object.

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -944,13 +959,29 @@ module.exports = function(self, options) {
     ], callback);
 
     function updateCounts(callback) {
-      return async.eachSeries(commands, function(command, callback) {
-        return self.db.update(command[0], command[1], callback);
+      return async.eachSeries(commands, function([criteria, query], callback) {
```
boutell, line 962: Yes time to modernize (:

lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -944,13 +959,29 @@ module.exports = function(self, options) {
...
+  self.findAttachmentsInSchema = function(schema) {
+    if (!schema) {
+      return false;
+    }
+
+    return findAttachment(schema);
+
+    function findAttachment(fields) {
+      return fields.some((field) => field.type === 'attachment' ||
+        (field.schema && findAttachment(field.schema)));
```
boutell, line 981: This works for arrays, but it does not work for widgets which are part of the document too.

If field.type is array or singleton, then you need to iterate the allowed widget types in it, and for each of those, call `self.apos.areas.getManager(widgetTypeName)` to get a manager. Then you can access `.schema` again and make a recursive call to your method.

boutell (changes requested):

lib/modules/apostrophe-attachments/lib/api.js
ValJed: Thanks a lot, this is the piece I missed.

lib/modules/apostrophe-attachments/lib/api.js
ValJed: done

boutell (approved):
This is nice and succinct!

boutell (changes requested):
Should be `return callback(null)`, for consistency with our standard. Otherwise great.

boutell (approved):

boutell (changes requested):
Sorry Val,

Unfortunately I missed something important because I am not used to looking at A2 code as much recently.

In A3, the schema is completely known at startup, and so we can use this technique.

However in A2 templates can unfortunately introduce their own areas, with widgets the possibility of which cannot be known by analyzing the schema of the page.

This means there is always a possibility of a widget that has its own attachments.

I will look for another way to optimize this and I will make a note to bring your original code to A3 where it has a lot of value.

PR #2743, opened 2021-02-19
Addresses wonky toolbar buttons (due to content-box)

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-1035, "Rich Text Toolbar is chunky and busted"

This was another instance of removing normalize.css inflicting content-box on everything. Rather than set these individual things to `border-box`, I removed some padding and used flex powers. We could decide to use border-box for all our things, but if not we should live with what we have set.

Previously:
Screen Shot 2021-02-16 at 5 33 06 PM


abea:
👀

stuartromanek (approved):

PR #2744, opened 2021-02-19
Detect changes in relationships (does not yet support alternate `idsStorage` settings)

boutell:
This fixes the problem where the images widget doesn't reliably save it when you swap out one of the images.


boutell:
I will open a followup ticket about idsStorage.

boutell:
Actually idsStorage is not directly examined, which is the whole point really (because it's not updated at this stage, which is how we got this mystery bug). So we're fine with this PR.

abea (approved):

PR #2749, opened 2021-02-22
3/relationship transition

abea:
Resolves https://apostrophecms.atlassian.net/browse/A30U-1034, "Images: Strange transition when moving between images in the modal/array"


abea:
To test, create an array field with a join field inside. populate at least two array items and use the left sidebar to switch between them. There should no longer be any animation on the relationship field.

boutell (approved):

PR #2754, opened 2021-02-23
3/area context options

abea:
Adds the option to pass context options to an area for its widgets. Context options for widgets not in that area (or that don't exist) are ignored.

### DIFF
--- a/modules/@apostrophecms/area/lib/custom-tags/widget.js
+++ b/modules/@apostrophecms/area/lib/custom-tags/widget.js
@@ -4,7 +4,42 @@
 module.exports = function(self, options) {
 
   return {
-    async run(context, item, options) {
+    // We need a custom parser because of the "with" syntax
+    parse(parser, nodes, lexer) {
+      // get the tag token
+      const token = parser.nextToken();
+
+      const args = new nodes.NodeList(token.lineno, token.colno);
+      let argsCount = 0;
+
+      while (true) {
+        // get the arguments before "with"
+        const object = parser.parseExpression();
+
+        if (argsCount < 2) {
+          args.addChild(object);
+          argsCount++;
+        } else {
+          throw new Error('Too many arguments were passed to the widget tag before the "with" keyword.');
+        }
+
+        const w = parser.peekToken();
+        if (!(w.type === 'comma')) {
+          break;
+        }
+        parser.nextToken();
+      }
+
+      const w = parser.peekToken();
+      if ((w.type === 'symbol') && (w.value === 'with')) {
+        parser.nextToken();
+        const _with = parser.parseExpression();
+        args.addChild(_with);
+      }
+      parser.advanceAfterBlockEnd(token.value);
+      return { args };
+    },
+    async run(context, item, options, _with) {
       const req = context.env.opts.req;
       if (!item) {
         self.apos.util.warn('a null widget was encountered.');
@@ -13,13 +48,19 @@ module.exports = function(self, options) {
       if (!options) {
         options = {};
       }
+
+      let contextOptions = {};
+      if (typeof _with === 'object' && _with[item.type]) {
+        contextOptions = typeof _with[item.type] === 'object' ? _with[item.type] : {};
+      }
+
       const manager = self.getWidgetManager(item.type);
       if (!manager) {
         // Not configured in this project
         self.warnMissingWidgetType(item.type);
         return '';
       }
-      return manager.output(req, item, options);
+      return manager.output(req, item, options, contextOptions);
     }
   };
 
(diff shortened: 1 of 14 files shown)


modules/@apostrophecms/widget-type/index.js
```diff
@@ -184,12 +184,12 @@ module.exports = {
...
       // Apostrophe 3.x.
-
-      async output(req, widget, options) {
+      async output(req, widget, options, _with) {
         return self.render(req, self.template, {
           widget: widget,
           options: options,
-          manager: self
+          manager: self,
+          contextOptions: _with
```
abea, line 192: They do seem like "options," but more specifically belong to the context in which the area was used.

modules/@apostrophecms/area/lib/custom-tags/widget.js
```diff
@@ -4,7 +4,42 @@
 module.exports = function(self, options) {
 
   return {
-    async run(context, item, options) {
+    // We need a custom parser because of the "with" syntax
+    parse(parser, nodes, lexer) {
```
abea, line 8: Reused from the area.js file (`area` tag).

modules/@apostrophecms/area/lib/custom-tags/widget.js
```diff
@@ -13,13 +48,19 @@ module.exports = function(self, options) {
       if (!options) {
         options = {};
       }
+
+      let contextOptions = {};
+      if (typeof _with === 'object' && _with[item.type]) {
```
abea, line 53: There's probably a better way.

test/widgets.js
```diff
@@ -0,0 +1,127 @@
...
+
+    assert(result.indexOf('<h2>Good args page</h2>') !== -1);
+    assert(result.indexOf('<p>You can control what happens when the text reaches the edges of its content area using its attributes.</p>') !== -1);
+    assert(result.indexOf('<li>color: 🟣</li>') !== -1);
+  });
+
+  it('should error while trying to render page template with poorly constructed area tag', async function() {
+    const req = apos.task.getAnonReq();
+
+    const goodPage = await apos.page.find(req, { slug: '/bad-page' }).toObject();
```
boutell, line 117: [squints] correct, but confusing variable name.

test/widgets.js
```diff
@@ -0,0 +1,127 @@
...
+    goodPage.metaType = 'doc';
+
+    try {
+      await apos.modules['args-bad-page'].renderPage(req, 'page', { page: goodPage });
+      assert(false);
+    } catch (error) {
+      assert(error);
```
boutell, line 124: Sounds like it's doing the job, but I don't actually see why the parser fails on this, so it may be worth inspecting the error to make sure it's saying something useful.

boutell (changes requested):
This is very 💯 - thank you for doing it!

I had one tiny quibble and one slightly bigger query about exactly what the bad page error looks like because I can't figure out why it would fail although it apparently does (as planned).

But mainly I'd like us to show the syntax to Ben and hopefully Stu in standup today and get buy-in before we merge.

test/widgets.js
abea: The error is caused in the Bad Page template, passing three arguments to the area tag before `with`. It'd definitely be better if everything was in the test file, but it seems we're already using fake modules and their templates in other tests already.

test/widgets.js
abea: Ah, I forgot to change this.

test/widgets.js
boutell: What I mean is that I'm not clear on where you are *checking* for too many arguments, and so it's not clear to me if the actual error message is a good error message.

boutell (changes requested):

test/widgets.js
abea: I see. I might need a hand with this. I thought the `await apos.modules['args-bad-page']` line would throw an error, but it doesn't (there, at least). The response is simply the error page template.

Is there a way we can capture the template render error, "Too many arguments were passed to the 'baddies' area before the 'with' keyword."?

modules/@apostrophecms/template/index.js
```diff
@@ -657,8 +657,8 @@ module.exports = {
         try {
           content = await module.render(req, template, args);
         } catch (e) {
-          // The page template
-          // threw an exception. Log where it
+          console.info('🍅');
```
boutell, line 660: Leftover tomato

test/widgets.js
```diff
@@ -114,14 +130,26 @@ describe('Widgets', function() {
...
+    };
+
     try {
-      await apos.modules['args-bad-page'].renderPage(req, 'page', { page: badPageDoc });
+      await apos.modules['args-bad-page'].render(req, 'page', args);
+
       assert(false);
     } catch (error) {
-      assert(error);
+      assert(error.toString().indexOf('Too many arguments were passed'));
```
boutell, line 152: indexOf returns -1 on no match, which is truthy.

boutell (changes requested):

boutell (approved):

PR #2758, opened 2021-02-25
improve attachment performance and provide an option to do an insert …

boutell:
…or update that skips all attachment related work

### DIFF
--- a/lib/modules/apostrophe-attachments/lib/api.js
+++ b/lib/modules/apostrophe-attachments/lib/api.js
@@ -874,16 +874,14 @@ module.exports = function(self, options) {
     // entry in the array is a 2-element array. Element 0
     // is the criteria, element 1 is the command
 
-    var commands = [];
-
-    if (!doc.trash) {
-      if (ids.length > 0) {
-        commands.push([
+    const commands = [
+      ...ids.length ? [
+        [
           {
             _id: { $in: ids }
           },
           {
-            $addToSet: { docIds: doc._id }
+            $addToSet: { [!doc.trash ? 'docIds' : 'trashDocIds']: doc._id }
           }
         ],
         [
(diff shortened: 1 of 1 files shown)


lib/modules/apostrophe-attachments/lib/api.js
```diff
@@ -874,70 +877,97 @@ module.exports = function(self, options) {
     var commands = [];
 
     if (!doc.trash) {
+      if (ids.length > 0) {
```
ValJed, line 880: [Code proposal]
To avoid code redundancy you can do something like that:
(it replaces all the if else blocks of this function)
```javascript
    const commands = [
      ...ids.length ? [
        [
          {
            _id: { $in: ids }
          },
          {
            $addToSet: { [!doc.trash ? 'docIds' : 'trashDocIds']: doc._id }
          }
        ],
        [
          {
            _id: { $in: ids }
          },
          {
            $pull: { [!doc.trash ? 'trashDocIds' : 'docIds']: doc._id }
          }
        ],
        [
          {
            $or: [
              {
                trashDocIds: { $in: [ doc._id ] }
              },
              {
                docIds: { $in: [ doc._id ] }
              }
            ],
            _id: { $nin: ids }
          },
          {
            $pull: {
              trashDocIds: doc._id,
              docIds: doc._id
            }
          }
        ],
        [
          {
            _id: { $in: ids }
          },
          {
            $set: {
              utilized: true
            }
          }
        ]
      ] : [
        // This is the only command that must be run when
        // the document now has no attachments
        [
          {
            $or: [
              {
                trashDocIds: { $in: [ doc._id ] }
              },
              {
                docIds: { $in: [ doc._id ] }
              }
            ]
          },
          {
            $pull: {
              trashDocIds: doc._id,
              docIds: doc._id
            }
          }
        ]
      ]
    ];
```

ValJed (commented):
Just a little concern about workflow, I tried to insert pieces with the option `skipAttachments` but it seems that only the first insert receive the option.
I think we should modify the `insert` function inside `insertInMissingLocales` which is run in the `self.docAfterSave`, to passe this option. So maybe creating a PR in `apostrophe-workflow` ?
Otherwise it works great.

ValJed (approved):
Seems ok, just need to create a PR in workflow to pass the `skipAttachments` option to all inserts.

boutell:
I used your version. Much better.

PR #2763, opened 2021-02-25
3/no options components

abea:
Removes references to, and support for, an `options.components` config object. Everywhere a module's Vue components are set, they are set in an override or extension of `getBrowserData`.

### DIFF
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -554,17 +554,6 @@ module.exports = {
         return self.apos.modules['@apostrophecms/email'].emailForModule(req, templateName, data, options, self);
       },
 
-      // Given a Vue component name, such as AposPiecesManager,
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/module/index.js
```diff
@@ -554,17 +554,6 @@ module.exports = {
...
       },
 
-      // Given a Vue component name, such as AposPiecesManager,
-      // return that name unless `options.components[name]` has been set to
-      // an alternate name. Overriding keys in the `components` option
-      // allows modules to provide alternative functionality for standard
-      // components while maintaining readabile Vue code via the
-      // <component :is="..."> syntax.
-
-      getVueComponentName(name) {
```
abea, line 564: I don't see this used anywhere.

abea:
Closing this. I think this is better to make available than `options.browser.components`. See #2766

PR #2766, opened 2021-02-26
3/no browser

abea:
Deprecates the `browser` module option and related method, `getComponentName`.

### DIFF
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -507,14 +507,6 @@ module.exports = {
         return _.get(self.options, dotPathOrArray, def);
       },
 
-      // If `name` is `manager` and `options.browser.components.manager` is set,
-      // return that string, otherwise return `def`. This is used to decide what
-      // Vue component to instantiate on the browser side.
-
-      getComponentName(name, def) {
-        return _.get(self.options, `browser.components.${name}`, def);
-      },
-
       // Send email. Renders an HTML email message using the template
       // specified in `templateName`, which receives `data` as its
       // `data` object (literally called `data` in your templates,
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -426,8 +426,7 @@ module.exports = {
 
       // Extend this method to return the appropriate browser data for
       // your module. If you want browser data for the given req, return
-      // an object. That object is merged with the `browser` option passed
-      // to your module, then assigned to `apos.modules['your-module-name']`
+      // an object. That object is assigned to `apos.modules['your-module-name']`
       // in the browser. Do not return huge data structures, as this will impact
       // page load time and performance.
       //
@@ -443,8 +442,7 @@ module.exports = {
       // you must explicitly opt in.
 
       getBrowserData(req) {
-        self.options.browser = self.options.browser || {};
-        return self.options.browser;
+        return {};
       },
 
       // Transform a route name into a route URL. If the name begins with `/` it is understood to
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -507,6 +507,13 @@ module.exports = {
         return _.get(self.options, dotPathOrArray, def);
       },
 
+      // If `name` is `manager` and `options.components.manager` is set,
+      // return that string, otherwise return `def`. This is used to decide what
+      // Vue component to instantiate on the browser side.
+      getComponentName(name, def) {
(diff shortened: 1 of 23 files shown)


modules/@apostrophecms/module/index.js
```diff
@@ -509,14 +507,6 @@ module.exports = {
         return _.get(self.options, dotPathOrArray, def);
       },
 
-      // If `name` is `manager` and `options.browser.components.manager` is set,
-      // return that string, otherwise return `def`. This is used to decide what
-      // Vue component to instantiate on the browser side.
-
-      getComponentName(name, def) {
-        return _.get(self.options, `browser.components.${name}`, def);
```
abea, line 517: An alternative would be to update `getComponentName` to use `options.components` instead. No modules are currently using that option, but some, like the modal module, do open the door:

```
  getBrowserData(req) {
    return {
      modals: self.modals,
      components: {
        the: options.components.the || 'TheAposModals',
        confirm: options.components.confirm || 'AposModalConfirm'
      }
    };
  }
```

abea: ... I'll do that.

modules/@apostrophecms/module/index.js
```diff
@@ -509,12 +507,11 @@ module.exports = {
...
       },
 
-      // If `name` is `manager` and `options.browser.components.manager` is set,
+      // If `name` is `manager` and `options.components.manager` is set,
       // return that string, otherwise return `def`. This is used to decide what
       // Vue component to instantiate on the browser side.
-
       getComponentName(name, def) {
-        return _.get(self.options, `browser.components.${name}`, def);
+        return _.get(self.options, `components.${name}`, def);
```
boutell, line 514: This is the same ol' problem: you have to reset the whole thing, including new components you might not know about that are added to that module's configurable components object later, just to change one. It's askin' for trouble.

So I would recommend:

```
return self.options[`${name}Component`] || def
```

Which can be overridden with no surprises.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -324,12 +250,88 @@ module.exports = {
       },
       // Add on the core default options to use, if needed.
       getBrowserData(_super, req) {
-        const result = _super(req);
+        const initialData = _super(req);
 
-        _.defaults(result, {
+        const rteData = {
+          components: {
```
boutell, line 256: If we go with the `${name}Component` pattern we can set the original defaults as options which is more self-documenting and can potentially be extracted for documentation.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -324,12 +250,88 @@ module.exports = {
...
+            widgetEditor: 'AposRichTextWidgetEditor',
+            widget: 'AposRichTextWidget'
+          },
+          tools: {
```
boutell, line 260: Same comment for tools.

boutell (changes requested):

modules/@apostrophecms/module/index.js
abea: That's a new bucket of worms: creating a component name structure. Maybe we should actually remove this for now and create a ticket to come up with a better way for project-level component override.

boutell: Rich text widgets already need and use it, so we can't punt. We could discuss in a meeting.

abea: Rich text widgets don't use this method currently. It hasn't really been doing anything lately, I don't think. Not other than returning the default values we pass in.

abea: The goal of this PR isn't to establish a new and better way to set module components. It's to get rid of the barely- and inconsistently-used `browser` option. If we switch to `components` option instead, for now, we've not gained or lost anything there other than not having to dig deeper in a dot path.

modules/@apostrophecms/rich-text-widget/index.js
abea: [See above](https://github.com/apostrophecms/apostrophe/pull/2766/files?file-filters%5B%5D=.js#r584841341). I don't think we should make this call in this PR. I can return these to the options object in a `components` option.

modules/@apostrophecms/rich-text-widget/index.js
abea: As above, I can move this back into the options in its own `tools` option.

boutell (approved):
That works for now.

PR #2773, opened 2021-03-01
Fix relationship change detection so the global doc does not always offer "discard changes"

boutell:
Related changes:

* A relationship that has no ids or no findable ids should still populate as an empty array, so it can be saved again as empty, rather than falling back to a nonempty ids array that wasn't part of what was copied during "discard" because relationships are writable in 3.x.
* In revert methods, documents should be loaded as if they went through the full query process so their relationships can be compared; this includes calling finalize to make sure relationships defaults to true.
* The relationship comparison logic should only care about _id and _fields, even for a top level relationship field.

Unrelated minor change:

* Removed jsdiff and dead code that used it, to quiet npm audit.

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -1,5 +1,4 @@
 const _ = require('lodash');
-const { klona } = require('klona');
 
 module.exports = {
   options: {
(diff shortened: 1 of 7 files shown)


abea:
Some eslint errors to fix.

abea (dismissed):

abea:
Good to merge once the ESLint jawns are fixed.

PR #2777, opened 2021-03-02
previously the auto-default to first option behavior for a required s…

boutell:
…elect field with no def setting only took effect visually, when you hit save you would be told the field was still blank.


modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
```diff
@@ -44,6 +44,8 @@ export default {
   },
   data() {
     return {
+      next: (this.value && this.value.data != null)
```
boutell, line 47: Overriding similar logic in the mixin.

abea (changes requested):
I think the issue is more that it appears that the input has a value when it does not. Even if not required, someone could look at it, see a label, and assume that value was already chosen for them as the default (or previously selected).

boutell:
The convention going back to A2 and before is that if a select element is required, the first option serves as the default value.

If we don't want that, then we need to have a design discussion about changing it. Whereas this PR fixes 3.x to behave as previous releases have.

So I would suggest that we accept this for now, as a bug fix, and that we open a ticket for a design discussion about changing the behavior if desired.

abea:
That makes sense. As long as there is no null option, this seems fine. If there is a null option (not a required field) and has no configured default, it should not show any value and default to null, which I think is what we have already.

abea (dismissed):

abea (dismissed):

boutell:
@abea I did dead code removal and purged the diff module here because it was blocking npm audit and thus the testing and the merge.

abea (approved):

PR #2782, opened 2021-03-03
Advisory lock factored out to mixin, works in context

boutell:
The advisory locking feature now works in the on-page contextual editor, including both regular on-page editing and the palette.

The palette is a separate document; we let go of the lock on the page when we take the lock on the palette.

I shortened the advisory lock lifetime to 15 seconds because the refresh time (every 5 seconds) is more than sufficient and we want to minimize the unavoidable reality that when a user moves from page to page in edit mode, we can't really unlock, so another user could be briefly irritated by an incorrect message about a lingering lock. (There may be a viable scheme involving the navigator beacon API and timestamps, but that would be a subject for a whole new ticket.)


boutell:
There is a corresponding PR on palette which is required for the palette to correctly close itself if a lock is lost or declined.

boutell:
Improving this just a lil bit for the original use case.

abea (dismissed):

PR #2792, opened 2021-03-04
changes piecesModuleName to pieceModuleName

abea:
Renames the `piecesModuleName` option to `pieceModuleName` to bring it in sync with the singular naming scheme for 3.x


abea:
The build is breaking due to the rich text widget lodash dependency. Should we be fixing that in every PR? It'll be merged momentarily.

boutell (dismissed):

PR #2795, opened 2021-03-05
Conditional fields

boutell:
Also: input fields should strive to emit values of the correct type, so that conditional fields and other browser-side applications don't have problems testing them for equality, etc. I upgraded the string input component to do this when dealing with the various number types. While I was doing that I also fixed min/max comparison for number types.


modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -166,6 +154,15 @@ export default {
       if (this.type === 'number') {
         this.step = this.field.type === 'float' ? 'any' : 1;
       }
+    },
+    convert(s) {
+      if ([ 'integer', 'range' ].includes(this.field.type)) {
```
stuartromanek, line 159: same Q here

stuartromanek (changes requested):

boutell:
Good catch re: the range field. I removed vestigial range support from the string input and added a convert method to the range input. I also updated the conditional-fields-demo branch with a range example, although it is a silly one since we only do strict equality for now. Select "yes" for rock, set "volume" to 11 and an additional field appears.

PR #2796, opened 2021-03-05
removed the options parameter from all of the module section function…

boutell:
…s in our core modules. Did not stop passing it from moog yet.

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -591,13 +591,10 @@ module.exports = {
       // If `builders` is an object its properties are invoked as
       // query builders, for instance `{ attachments: true }`.
       async findOneForEditing(req, criteria, builders) {
-        const query = await self.findForEditing(req, criteria, builders);
-        const doc = query.toObject();
-        if (self.options.annotate) {
-          self.apos.attachment.all(doc, { annotate: true });
-        }
-        return doc;
+        return self.findForEditing(req, criteria, builders).toObject();
       },
+      // Identical to findOneForEditing by default, but could be
+      // overridden usefully in subclasses.
       async findOneForCopying(req, criteria) {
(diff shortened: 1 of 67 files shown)


abea (commented):
I see some comments and code in `lib/moog.js` that still have the `self, options` arguments.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -593,7 +593,7 @@ module.exports = {
       async findOneForEditing(req, criteria, builders) {
         const query = await self.findForEditing(req, criteria, builders);
         const doc = query.toObject();
-        if (options.annotate) {
+        if (self.options.annotate) {
```
abea, line 596: What does this option do?

boutell: It never worked and was never documented, plus we have the `attachments` query builder now as well as REST APIs that automatically take care of annotating attachments, so I'm removing it. Good catch.

modules/@apostrophecms/launder/index.js
```diff
@@ -3,11 +3,11 @@
 // Apostrophe to sanitize user input.
 
 module.exports = {
-  init(self, options) {
+  init(self) {
     self.apos.launder = require('launder')({
       // A3 _ids may contain :-separated components
       idRegExp: /^[A-Za-z0-9_-]+(:[A-Za-z0-9_-]+)*$/,
-      ...options
+      ...self.options
```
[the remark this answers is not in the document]
boutell: No, because launder doesn't do anything fancy or expensive with its options object like try to sanitize everything recursively and omitting the apos object would thus itself be overhead. In theory we could maintain logic here to pass only options to launder that are valid, but that would add to our maintenance burden, so I would say not today at least.

abea (approved):

PR #2797, opened 2021-03-08
tweaked lock language

boutell:
This language isn't quite what you suggested because:

* 1. In the case of another person taking the lock, they actively, affirmatively did that, so the language should make this clear to discourage slapfights between users.
* 2. The suggested language didn't quite address the issue when you take control from yourself. The new language is accurate in either scenario: you can only edit a particular document in one place (tab or window) at a time.

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
+++ b/modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
@@ -93,10 +93,6 @@ export default {
     // detect lock errors with `isLockedError` and call this method. The rest of the time,
     // it is called for you.
     async showLockedError(e) {
-      // We use an alert because it is a clear interruption of their
-      // work, and because a notification would appear in both windows
-      // if control was taken by the same user in another window,
-      // which would be confusing.
       await apos.alert({
         heading: 'Multiple Editors',
         description: `${e.body.data.me ? 'You' : e.body.data.title} took control of this document in another tab or window. A document can only be edited in one place at a time.`
--- a/modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
+++ b/modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
@@ -93,21 +93,14 @@ export default {
     // detect lock errors with `isLockedError` and call this method. The rest of the time,
     // it is called for you.
     async showLockedError(e) {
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
```diff
@@ -93,21 +93,14 @@ export default {
...
-          description: 'Another user took control of the document.'
-        });
-      }
+      // We use an alert because it is a clear interruption of their
+      // work, and because a notification would appear in both windows
+      // if control was taken by the same user in another window,
+      // which would be confusing.
+      await apos.alert({
+        heading: 'Multiple Editors',
+        description: `${e.body.data.me ? 'You' : e.body.data.title} took control of this document in another tab or window. A document can only be edited in one place at a time.`
```
localghost443, line 102: What if the person who took control is not "You", this feels very confusing to me. This could lead someone to think they'r doing something wrong. I don't think pointing out that a specific **other** user is doing it will create a 'slap fight', its just transparency. Are we able to identify the user who took control?

localghost443 (changes requested):

modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
```diff
@@ -93,21 +93,14 @@ export default {
...
-          heading: 'You Took Control in Another Window',
-          description: 'You took control of this document in another tab or window.'
-        });
-      } else {
-        await apos.alert({
-          heading: 'Another User Took Control',
-          description: 'Another user took control of the document.'
-        });
-      }
+      // We use an alert because it is a clear interruption of their
```
localghost443, line 96: I don't think we need this comment, or rather, it should just be "Alert user that the document is being edited in another window. Appears in all windows...." for example. We don't need to justify that its not a notification.

localghost443 (changes requested):

modules/@apostrophecms/ui/ui/apos/mixins/AposAdvisoryLockMixin.js
boutell: This notification now reads:

"Bob Smith took control of this document in another tab or window. A document can only be edited in one place at a time."

This seems like what you're asking for. I think you might be responding to the word "title," but that's the `title` property of the user who took control (their name).

localghost443 (approved):

PR #2798, opened 2021-03-09
36 show suggested terms from tags

ValJed:
Resolve: https://apostrophecms.atlassian.net/browse/MAE-36

Adds suggestions list when typing on images search bar, based on autocomplete cursor, it shows words from `highSearchWords` that march the search.
takes filters into account.

Suggestions are word related, since we perform multi words search, so we can have independent  suggestions for each word in the search bar.

You can easily test it by using the related branch in your project, going to the images modal and trying to search for images.


boutell (changes requested):
See notes on the API response. Also, I think it would make sense to do this for the pieces manager in general, not just images, if Stuart has no objection.

lib/modules/apostrophe-images/lib/routes.js
```diff
@@ -0,0 +1,22 @@
+module.exports = (self) => {
+  self.route('post', 'suggestions', self.apos.docs.requireEditorOfSomething, async (req, res) => {
```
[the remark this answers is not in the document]
ValJed: I moved this into `apostrophe-pieces` routes so I followed the way it's done here, I modified the response using `self.apiResponse()` to stay coherent indeed.

lib/modules/apostrophe-images/public/js/manager-modal.js
```diff
@@ -65,5 +64,125 @@ apos.define('apostrophe-images-manager-modal', {
       return $checkbox;
     };
 
+    self.enableSearch = function() {
+      var isListening = false;
+      var inputSelector = '[name="search-' + self.options.name + '"]';
+
+      self.$filters.on('keyup', inputSelector, function(e) {
+        var $input = $(e.target);
+        var $search = self.$filters.find('.apos-modal-filters-search');
```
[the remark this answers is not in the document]
ValJed: This one had no data attribute, I  added one.

ValJed:
Since the search is a multi words search I improved the suggestions:
* Only returns words matching from `highSearchWords` fields, no more entire title since it was very confusing...
* Big improvement: the suggestions are words dependents, so we watch the cursor position to know on which word we are currently and to replace only this one when choosing a suggestion.

Suggestions seem really more relevant now, matching the way we perform searches.

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -124,6 +124,45 @@ module.exports = function(self, options) {
     return next();
   };
 
+  // This method returns list of suggestions based on highSearchWords array,
+  // it uses the autocomplete cursor to request documents, it also checks for filters.
+  self.getSuggestions = async function(req, term, filters) {
+    const search = self.apos.launder.string(term).toLowerCase();
+    const cursor = getCursorWithFilters(filters);
+    const images = await cursor.autocomplete(search).limit(15).toArray();
```
boutell, line 132: should be `pieces`

lib/modules/apostrophe-pieces/lib/api.js
```diff
@@ -124,6 +124,45 @@ module.exports = function(self, options) {
...
+      return [
+        ...acc,
+        ...matchingWords
+      ];
+    }, []);
+
+    return suggestions;
+
+    // adding filter to cursor only if this one exist for the current piece
+    function getCursorWithFilters (filters) {
```
boutell, line 157: This will crash on made-up query strings.

Just call:

`self.find(req, {}).queryToFilters(filters, 'manage')`

To apply those that really exist, are safe for the manage view, and actually have a launder method.

boutell (changes requested):

lib/modules/apostrophe-pieces/lib/api.js
ValJed: Yes, forgot this one.

lib/modules/apostrophe-pieces/lib/api.js
ValJed: Thanks for this one, way cleaner indeed. 👍🏼

lib/modules/apostrophe-pieces/lib/routes.js
```diff
@@ -34,6 +35,27 @@ module.exports = function(self, options) {
...
+  self.routes.suggestions = async function(req, res) {
+    try {
+      const {term, filters} = req.body;
+
+      if (!term || typeof filters !== 'object') {
+        const errorMsg = !term
+          ? 'A search term must be provided.'
+          : 'Filters must be an object containing search filters.';
+
+        return self.apiResponse(res, errorMsg);
```
boutell, line 47: This is fine. You could have used `self.apiRoute` to remove a little more code I think.

boutell (approved):

lib/modules/apostrophe-pieces/lib/routes.js
ValJed: I didn't use it initially because it wasn't used in this file.
Just modified it.

PR #2800, opened 2021-03-09
lint for incorrectly configured modules (ported from late A2), throw …

boutell:
…an error on a page type without module to back it, allow publishing of a page type change, remove parked fields from the page settings UI

All of these issues added up to the solution to the "can't switch page type and publish that change" ticket.

I'm aware we might prefer a read-only treatment for parked fields. Removing them is simpler and consistent with 2.x, so I'd like to do that just for this ticket — which is intended as a bug fix, not as the final be-all end-all of how we handle parked fields.

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,7 +106,7 @@ module.exports = async function(options) {
     defineModules();
 
     await instantiateModules();
-    lintOrphanModules();
+    lintModules();
     await self.emit('modulesReady');
     await self.emit('afterInit');
     if (self.taskRan) {
@@ -329,7 +329,7 @@ module.exports = async function(options) {
     });
   }
 
-  function lintOrphanModules() {
+  function lintModules() {
     const validSteps = [];
     for (const module of Object.values(self.modules)) {
       for (const step of module.__meta.chain) {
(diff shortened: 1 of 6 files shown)


abea:
@boutell It looks like the "Pages Public API" test is failing at the moment.

boutell:
@abea fixed the tests.

index.js
```diff
@@ -369,6 +369,64 @@ module.exports = async function(options) {
...
+      if (module.options.singletonWarningIfNot && (name !== module.options.singletonWarningIfNot)) {
+        lint('The module ' + name + ' extends ' + module.options.singletonWarningIfNot + ', which is normally\na singleton (Apostrophe creates only one instance of it). Two competing\ninstances will lead to problems. If you are adding project-level code to it,\njust use lib/modules/' + module.options.singletonWarningIfNot + '/index.js and do not use "extend".\nIf you are improving it via an npm module, use "improve" rather than "extend".\nIf neither situation applies you should probably just make a new module that does\nnot extend anything.\n\nIf you are sure you know what you are doing, you can set the\nsingletonWarningIfNot: false option for this module.');
+      }
+      if (name.match(/-widget$/) && (!extending(module)) && (!module.options.ignoreNoExtendWarning)) {
+        lint('The module ' + name + ' does not extend anything.\n\nA -widget module usually extends @apostrophecms/widget or another widget type.\nOr possibly you forgot to npm install something.\n\nIf you are sure you are doing the right thing, set the\nignoreNoExtendWarning option to true for this module.');
```
abea, line 381: Should be `@apostrophecms/widget-type`.

index.js
```diff
@@ -369,6 +369,64 @@ module.exports = async function(options) {
...
+      }
+      // If we got to the base class of all modules, the module
+      // has no construct of its own
+      if (self.synth.myToOriginal(d.__meta.name) === '@apostrophecms/module') {
+        return false;
+      }
+      return doesWork(d);
+    }
+    function doesWork(d) {
+      const code = d.routes || d.apiRoutes || d.renderRoutes || d.init || d.methods || d.beforeSuperClass || d.handlers || d.helpers || d.restApiRoutes || d.middleware || d.customTags || d.components || d.tasks;
```
abea, line 405: It seems fragile to have all these hard-coded here. Maybe at least in an array? Can we connect it to the module error when an unapproved key is used in a module on the top level?

modules/@apostrophecms/page/index.js
```diff
@@ -481,6 +481,23 @@ module.exports = {
...
+            if (!choice.name) {
+              throw new Error('One of the page types specified for your types option has no name property.');
+            }
+            if (!choice.label) {
+              throw new Error('One of the page types specified for your types option has no label property.');
+            }
+            if (!self.apos.modules[choice.name]) {
+              let error = `There is no module named ${choice.name}, but it is configured as a page type\nin your types option.`;
+              if (choice.name === 'home-page') {
+                error += '\n\nYou probably meant @apostrophecms/home-page.';
```
abea, line 495: I don't think we need this. If they're coming directly from 2.x the type would be `home`, so it's not helping that case.

modules/@apostrophecms/page/index.js
```diff
@@ -2148,6 +2156,13 @@ database.`);
         // For pages we currently always do this. For pieces it's conditional
         // on whether the type is localized.
         return self.apos.i18n.inferIdLocaleAndMode(req, _id);
+      },
+      // Copy any parked properties of `page` back into `input` to
+      // prevent any attempt to alter them via the PUT or PATCH APIs
+      enforceParkedProperties(req, page, input) {
+        for (const parked of (page.parked || [])) {
```
abea, line 2163: `parked` for the field name is confusing. Maybe `field`.

abea (changes requested):

index.js
boutell: I'll make it an array. Not all approved keys count as "doing work" in this sense (options don't, fields don't, etc.), so the distinction makes more sense here than it does in moog's core.

modules/@apostrophecms/page/index.js
boutell: This was the bug in a3-assembly-boilerplate. I think it's an easier mistake for an A3 native to make because all the other page types look like `foo-type`. So I'd like to keep it.

boutell: We'll want to deal with A2 migration in the A2 migration tool one day.

abea (approved):

PR #2811, opened 2021-03-10
discard changes should reset undo/redo. (You could never undo discard…

boutell:
… changes which is why it has a confirm prompt. Previously we just had buggy behavior where it looked like you could undo it, and then more stuff broke.)

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -833,6 +833,23 @@ export default {
         ...this.context,
         modified: false
       };
+      if (this.contextStack.length) {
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -833,6 +833,23 @@ export default {
         ...this.context,
         modified: false
       };
+      if (this.contextStack.length) {
+        // Pushed contexts might edit any property of the doc
+        this.original = klona(result.doc);
+      } else {
+        // On-page we only edit areas
```
boutell, line 840: This logic exists for parity with our behavior when first loading the context document and populating this.original based on the areas actually in the page.

abea (approved):

PR #2817, opened 2021-03-12
choices for a filter should ignore the setting of that filter

boutell:
This is related to PRO-985 but that might not be the whole story, see ticket.


abea (approved):

abea:
I think there's a separate front end bug that inspired this, so I wouldn't close the ticket out yet.

PR #2819, opened 2021-03-14
3/cut copy paste

boutell:
Works; follows the design. Needs design input on the paste experience and surfaces an issue with the grouped widgets experience styles which seem to have drifted off target a bit.

Re: the paste experience, the design is based on the idea of widget groups, but when widgets aren't grouped the presence of the clipboard kinda forces the use of the "ungrouped widgets" section, which also isn't toggled open by default because it isn't the first section. I didn't want to cowboy this bit so I'm sharing it in its unvarnished glory to get input.

Right now the icons crowd each other.

Draft for now while we figure out these issues.

See screenshots.

Per previous discussion there is just one widget on the clipboard at a time.

Screen Shot 2021-03-14 at 4 33 15 PM
Screen Shot 2021-03-14 at 4 33 48 PM
Screen Shot 2021-03-14 at 4 39 17 PM

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
+++ b/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
@@ -1,6 +1,9 @@
 import Vue from 'Modules/@apostrophecms/ui/lib/vue';
+import { klona } from 'klona';
 
 export default function() {
+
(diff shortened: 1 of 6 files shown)


abea:
I imagine this will be closed and replaced anyway?

PR #2821, opened 2021-03-15
Implied default of select field should work with conditionalFields

boutell:
The logic to select the first choice when required was failing to actually emit that choice as a change, now it is compatible with conditional fields.


boutell:
This goes with https://github.com/apostrophecms/basics/pull/22

stuartromanek (approved):

PR #2825, opened 2021-03-16
PRO-980 update secondary UI selector to match new button structure 

stuartromanek:
https://linear.app/apostrophecms/issue/PRO-980/nested-area-breadcrumbs-seems-to-be-bleeding-into-the-button

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
@@ -583,7 +583,7 @@ export default {
     transform: translate(-50%, 0);
   }
 
-  .apos-area-widget-inner .apos-area-widget-inner /deep/ .apos-context-menu__btn {
+  .apos-area-widget-inner .apos-area-widget-inner /deep/ .apos-context-menu__btn .apos-button {
     background-color: var(--a-secondary);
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -583,7 +583,7 @@ export default {
     transform: translate(-50%, 0);
   }
 
-  .apos-area-widget-inner .apos-area-widget-inner /deep/ .apos-context-menu__btn {
+  .apos-area-widget-inner .apos-area-widget-inner /deep/ .apos-context-menu__btn .apos-button {
```
abea, line 586: That's a deep boi. Doesn't look like there's an easy alternative. Unless we pass a class into the button?

stuartromanek: does the Vue component receive any of the 'normal' element attributes that get tacked onto the `<Component>`? .. if so, we could generate a child class based on the one first attached, that might shortcut some of this

abea: There are probably ways, but it'd be some decent refactoring. Not a fight for this day.

abea (approved):

PR #2827, opened 2021-03-16
PRO-94 ensure disabled fields

stuartromanek:
- `readOnly`'ify everything
- correct disabled states and slat behavior for Relationships
- Separate `disabled` from `limitReached` concepts in Attachment fields to preserve correct edit states
- provide `field` to areas so they can be disabled in editors
- Ability to disable `AposContextMenu`s from the top level props
- Provide generic `disabled` tooltip message as computed property from the inputMixin
- be able to disable `range`, `color`, `select`, `radio`
- Various style updates to slats, buttons

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -1828,6 +1828,7 @@ module.exports = {
               await self.apos.util.recursionGuard(req, `${_relationship.type}:${_relationship.withType}`, () => {
                 // Allow options to the getter to be specified in the schema,
                 // notably editable: true
+                // TODO editable going away in favor of disabled for consistency
                 return self.fieldTypes[_relationship.type].relate(req, _relationship, _objects, options);
               });
               _.each(_objects, function (object) {
@@ -1875,6 +1876,7 @@ module.exports = {
 
           // Allow options to the getter to be specified in the schema,
           // notably editable: true
+          // TODO editable going away in favor of disabled for consistency
           await self.apos.util.recursionGuard(req, `${relationship.type}:${relationship.withType}`, () => {
             return self.fieldTypes[relationship.type].relate(req, relationship, _objects, options);
           });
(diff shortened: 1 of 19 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -1828,6 +1828,7 @@ module.exports = {
               await self.apos.util.recursionGuard(req, `${_relationship.type}:${_relationship.withType}`, () => {
                 // Allow options to the getter to be specified in the schema,
                 // notably editable: true
+                // TODO editable going away in favor of disabled for consistency
```
stuartromanek, line 1831: @boutell I found this note while normalizing relationship field's disabled states, I don't think we use this `editable` property anymore, is there anything to be done here?

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
```diff
@@ -76,6 +76,14 @@ export default {
         return 20;
       }
     },
+    tooltip () {
+      let msg = false;
+      if (this.field.readOnly) {
+        // TODO For editor-facing tooltips is this the correct language?
```
stuartromanek, line 82: Not sure if there should be any more nuance to the generic tooltip message. Disabled? Read Only?

modules/@apostrophecms/schema/index.js
boutell: It's dead, you can drop the note.

modules/@apostrophecms/schema/index.js
```diff
@@ -1875,6 +1876,7 @@ module.exports = {
 
           // Allow options to the getter to be specified in the schema,
           // notably editable: true
+          // TODO editable going away in favor of disabled for consistency
```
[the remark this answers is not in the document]
boutell: You can drop the TODO.

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
boutell: 🤷‍♂️

boutell (dismissed):
I answered your Q's about those TODOs but they aren't what you came to do so I approved as well. This was clearly a Jorb, so thanks.

stuartromanek:
if you could ring the bell again, post-comment cleanup

boutell (approved):

PR #2829, opened 2021-03-17
PRO-1008: trash is an action, with autopublish

boutell:
Moving pieces to the trash is now an action, and that action autopublishes. This fixes the loophole where it is impossible to publish the fact that you've moved something to the trash if it's just a boolean field, you don't publish on the first try, and you can't edit the trash.

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
@@ -74,13 +74,15 @@
         <AposButton
           v-if="activeMedia._id && !restoreOnly"
           @click="trash"
+          icon="trash-can-icon"
+          :icon-only="true"
           class="apos-media-editor__trash"
-          label="Move to Trash"
+          label="Trash"
         />
         <AposButton
           @click="save" class="apos-media-editor__save"
           :disabled="docFields.hasErrors"
-          :label="restoreOnly ? 'Restore from Trash' : 'Save'" type="primary"
+          :label="restoreOnly ? 'Restore' : 'Save'" type="primary"
         />
       </div>
     </AposModalLip>
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
@@ -72,6 +72,12 @@
           label="Cancel"
         />
         <AposButton
+          v-if="activeMedia._id && !restoreOnly"
+          @click="trash"
+          class="apos-media-editor__trash"
+          label="Move to Trash"
+        />
+        <AposButton
           @click="save" class="apos-media-editor__save"
           :disabled="docFields.hasErrors"
           :label="restoreOnly ? 'Restore from Trash' : 'Save'" type="primary"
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -577,13 +577,6 @@ module.exports = {
         });
       },
 
-      // Implementation of the PATCH route. Factored as a method to allow
```
boutell, line 580: I moved this back up to the route because there is no longer a generic PATCH route so it was carrying water for no one.

localghost443:
@boutell When a doc is removed from the trash, does it become a live draft? or is it auto published in reverse? I believe the former is the designed intent.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
```diff
@@ -71,6 +71,12 @@
           class="apos-media-editor__back" type="outline"
           label="Cancel"
         />
+        <AposButton
```
abea, line 74: I made a note in the ticket. We should split this into a separate branch to hold until we have proper designs for the image editor. This doesn't seem right with our design patterns.

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -36,6 +36,8 @@ export default {
           field.readOnly = true;
         }
       }
+      // Trash UI is handled via action buttons
+      schema = schema.filter(field => field.name !== 'trash');
```
abea, line 40: Should trash still be a field in the doc-type schema? 🤔

modules/@apostrophecms/piece-type/index.js
```diff
@@ -607,6 +600,9 @@ module.exports = {
       // within a single lock and without redundant network operations. This greatly
       // improves the performance of saving all changes to a document at once after
       // accumulating a number of changes in patch form on the front end.
+      //
+      // If `input._publish` launders to a boolean and the type is subject to draft/publish
```
abea, line 604: I got confused here. It seems to suggest that it'll pass if it's simply registered as a boolean. It also has to be true.

abea (changes requested):
When I open the piece editor for something _already_ in the trash, I still get the "Move to Trash" button in the more menu.

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
boutell: Yes, for pieces it should, because it is still handled that way on the back end, e.g. via PATCH of a normally converted schema field. So the front end presenting it in a particular way is the front end's business but it's still a schema field.

modules/@apostrophecms/piece-type/index.js
boutell: Thanks for the opportunity to say "launders to a truthy boolean" with a straight face.

abea: "Truthy" is one of programming's greatest gifts to the world.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
boutell: I switched to the trash icon per team discussion.

boutell:
I have fixed the errata and taken the notes from the ticket.

Among these:

* Use trash can icon only for media
* Button is "Restore" not "Restore from Trash"
* Fixed incorrect visibility of "move to trash" and "discard draft" buttons
* Do not auto-publish when restoring, let them publish the draft again when they are ready

abea (approved):

abea (approved):

PR #2830, opened 2021-03-17
Drag and drop of media fixed per various stackoverflow hints PRO-987

boutell:


stuartromanek:
@boutell can you explain what the heck is happening, why it needs to be happening, etc? 😅

boutell:
What we want: to know when the drag pointer is "over" the element.

What the dragover event actually does: fire constantly as you move around over the element. Which is a lot.

So we use dragenter and dragleave.

But: dragleave also bubbles up from sub-elements. Which is what was causing the more obvious symptoms before this PR. Except: fortunately, dragenter also bubbles. So per various stackoverflow threads, the solution is to track the depth.

The only alternative I'm aware of would be a CSS hack that would break the use of buttons as sub-elements, so for us it's not a good solution.

stuartromanek (approved):

PR #2837, opened 2021-03-18
allow apos.doc.insert in init of any module initialized after the db

boutell:
allow apos.doc.insert in init of any module initialized after the db module. Divorces uploadfs from the attachment module, it is now initialized by its own module and made available as apos.uploadfs.

This fixes https://github.com/apostrophecms/apostrophe/issues/2729. See also PRO-1055 and PRO-1056.


defaults.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     '@apostrophecms/util': {},
     '@apostrophecms/task': {},
     '@apostrophecms/schema': {},
-    '@apostrophecms/attachment': {},
+    '@apostrophecms/uploadfs': {},
```
abea, line 7: Is there sense in making this '@apostrophecms/uploader'? Might people want to use an alternative or something like that? Or is uploadfs so baked in that we are one?

modules/@apostrophecms/attachment/index.js
```diff
@@ -2,44 +2,13 @@ const _ = require('lodash');
...
-        width: 190,
-        height: 350
-      }
-    }
-  },
 
   async init(self) {
     self.name = 'attachment';
-
+    self.imageSizes = self.apos.modules['@apostrophecms/uploadfs'].imageSizes;
```
abea, line 11: This seems like a deal breaker for an `uploadfs` module. Let's call it `uploader`.

modules/@apostrophecms/attachment/index.js
```diff
@@ -212,22 +155,9 @@ module.exports = {
...
-          await self.db.createIndex({ trashDocIds: 1 });
-        },
-        addFixLengthPropertyMigration() {
-          self.addFixLengthPropertyMigration();
-        }
-      },
       'apostrophe:destroy': {
         async destroyUploadfs() {
-          await Promise.promisify(self.uploadfs.destroy)();
+          await Promise.promisify(self.apos.uploadfs.destroy)();
```
abea, line 160: Should this handler go in `@apostrophecms/uploader`?

abea (changes requested):
The module name is breaking my brain. Otherwise this seems great.

modules/@apostrophecms/attachment/index.js
abea: Thinking of documentation, I don't want people to have to think about what `uploadfs` is when they're adding image sizes.

defaults.js
boutell: This module is similar to `@apostrophecms/express` and `@apostrophecms/db` in that the module is not interesting to outside code, the actual service is. I don't think the name should change to suggest it's anything more than "the thing that fires up uploadfs and then you have uploadfs."

modules/@apostrophecms/attachment/index.js
boutell: Since uploadfs exports its options, I can change this to `self.apos.uploadfs.options.imageSizes` to address this concern. Then there will be zero references to the module outside itself.

modules/@apostrophecms/attachment/index.js
boutell: It should move yes, good catch.

defaults.js
abea: The issue I have is that it is something people will configure. If you want to add your own image size, you  have to configure it. Express and "database" are relatively generic or well-known. uploadfs is a really deep cut package with a funny name. If we're asking people to configure it to do a pretty normal thing, let's give them a pretty recognizable label to reference.

modules/@apostrophecms/attachment/index.js
abea: I don't understand. So image sizes would be added to `attachments`?

boutell:
I understand your point that `uploadfs` is not a brand as obvious as mongodb or express.

But the reality is that people should take a minute to understand it, especially if they might do a cloud deployment at some point. Including understanding that the attachment module is not the only thing that uses it. The imageSizes configuration is general to all use of uploadfs and its `copyImageIn` method in this instance of apostrophe, not just by attachments.

We cannot move that option to the attachments module because that would defeat the purpose by requiring the attachments module to initialize too early, causing problems for assets or for doc inserts in init, take your pick.

boutell:
(Yes attachments are the only use of copyImageIn in totally stock apostrophe, but they don't have to be. OK I Have to go to the 11am now)

modules/@apostrophecms/uploadfs/index.js
```diff
@@ -0,0 +1,101 @@
+const _ = require('lodash');
```
abea, line 1: We can only grab `merge` out of this here.

abea (changes requested):

abea:
If it didn't have normal-use configuration then I'd be fine with it. Most devs aren't going to be spending enough time with this to make it worth digesting the name. We can make it simpler for devs, so why shouldn't we?

boutell:
We cannot put `imageSizes` in the configuration of the `attachment` module because the `attachment` module cannot start up before assets, which may use uploadfs, and must start up before most other modules, which may use the database and insert documents, something that will not work properly if the attachment module isn't ready to run its handlers.

Also, attachments are not guaranteed to be the only place the `copyImageIn` feature of uploadfs gets used, although we don't currently use it elsewhere. We *do* use uploadfs in a second place (asset deployment), which is why it's been in need of liberation from the attachment module for a while.

We could move `imageSizes` up to be a top-level option, but I don't think that's a good idea. It's confusing for people who are involved in ops and *do* know uploadfs, because all other uploadfs options will not be there. And it's a decision we could make for convenience in too many other places.

So I still think imageSizes belongs in the uploadfs module, because that is what processes images to the required sizes, The attachment module documentation can cross-reference that, and the example can be given in a suitable HOWTO.

Learning about uploadfs is a good idea anyway. Developers need to know how to manipulate or modify files in Apostrophe in a way that will still work when they deploy to the cloud.

abea:
I'm not suggesting that `imageSizes` go in `attachments`. Only renaming the new module.

The new module has its primary option as `options.uploadfs`, so that's not hidden away. It also become redundant since the `uploadfs` module has an `uploadfs` option. I just don't agree that developers gain anything by us not making this simpler (imo) for them to learn.

boutell:
Good point! `options.uploadfs` does not make sense. So what if the new module is still called `uploadfs`, because that's what you're configuring, but the options are moved up a level, i.e. imageSizes is  not a sub-option of `uploadfs` anymore? This also makes overrides easier, since options don't merge recursively.

There is one more step we could take. If I make a small change to uploadfs to allow specifying `sizes` on individual calls to `copyImageIn`, then `imageSizes` can remain an option of the `attachment` module.

Which means that the only people who would have to learn about the `uploadfs` module are those who are truly keen on reconfiguring uploadfs in terms of how it stores media or chooses processing backends.

That is a small, bc change that could be done in a minor uploadfs release without too much creep in our scope.

boutell:
This is really better-er now

abea (changes requested):
It looks like there are a few failing tests.

boutell:
All tests now passing. https://app.circleci.com/pipelines/github/apostrophecms/apostrophe/2363/workflows/7c72cb7d-a79b-4cfc-8c55-fddba9f4cf5c/jobs/2966

abea (approved):

PR #2841, opened 2021-03-22
PRO-1075 3/tab id replaces html page id

boutell:
Advisory locking based on an id that remains constant for the lifetime of the tab. This eliminates nuisance logging about locks but more importantly it gives us the ability to detect when we really are taking control from another tab that belongs to ourselves and ask if we want to go through with that. This prevents lost work without nuisance prompts.

Also removed some dead code to parse a header we no longer use to transmit the lock identifier.

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -525,7 +525,7 @@ export default {
       this.refresh();
     });
 
-    // sessionStorage because it is deliberately tab-specific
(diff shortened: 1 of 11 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
```diff
@@ -524,6 +525,14 @@ export default {
       this.refresh();
     });
 
+    // sessionStorage because it is deliberately tab-specific
```
abea, line 528: Can we specify "browser tab" since we have tabs in our UI? I was confused for a minute there.

abea (approved):

abea (approved):

PR #2842, opened 2021-03-22
PRO-1074 skip migrations on new sites

boutell:
There is no need to run migrations on a new site. We must detect new-ness early enough that there's no possibility of being faked out by inserts in another module.

This won't detect a site that was constructed in advance in mongo by some custom script, but that advanced developer is probably not going to care, and migrations have a documented obligation to do no harm.

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -30,7 +30,7 @@ module.exports = {
     self.managers = {};
     self.enableBrowserData();
     await self.enableCollection();
-    self.apos.isNew = self.detectNew();
+    self.apos.isNew = await self.detectNew();
     await self.createIndexes();
     self.addDuplicateOrMissingWidgetIdMigration();
     self.addDraftPublishedMigration();
@@ -257,8 +257,8 @@ module.exports = {
       // This can't be done later because after this point init()
       // functions are permitted to insert documents
       async detectNew() {
-        const existing = await self.db.find({}).project({ _id: 1 }).limit(1).toArray();
-        self.apos.isNew = existing.length === 0;
+        const existing = await self.db.countDocuments();
+        return !existing;
       },
       async createSlugIndex() {
         const params = self.getSlugIndexParams();
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -30,6 +30,7 @@ module.exports = {
     self.managers = {};
     self.enableBrowserData();
     await self.enableCollection();
+    self.apos.isNew = self.detectNew();
     await self.createIndexes();
     self.addDuplicateOrMissingWidgetIdMigration();
     self.addDraftPublishedMigration();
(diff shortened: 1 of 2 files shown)


abea:
Little lint error

modules/@apostrophecms/doc/index.js
```diff
@@ -252,6 +253,13 @@ module.exports = {
       async enableCollection() {
         self.db = await self.apos.db.collection('aposDocs');
       },
+      // Detect whether the database is brand new (zero documents).
+      // This can't be done later because after this point init()
+      // functions are permitted to insert documents
+      async detectNew() {
+        const existing = await self.db.find({}).project({ _id: 1 }).limit(1).toArray();
```
abea, line 260: Couldn't you `count`? Or `toCount`?

modules/@apostrophecms/migration/index.js
```diff
@@ -248,7 +260,7 @@ module.exports = {
           await migration.fn();
           await self.db.insertOne({
             _id: migration.name,
-            at: new Date()
+            at: new Date(),
```
abea, line 263: there's your dangling comma.

modules/@apostrophecms/migration/index.js
```diff
@@ -230,8 +230,20 @@ module.exports = {
       async migrate(options) {
         await self.apos.lock.lock(self.__meta.name);
         try {
-          for (const migration of self.migrations) {
-            await self.runOne(migration);
+          if (self.apos.isNew) {
+            // Since the site is brand new (zero documents), we may assume
+            // it requires no migrations. Mark them all as "done" but note
+            // that they were skipped, just in case we decide that's an issue later
```
abea, line 236: Do we have guidance for this that needs to be documented? Should all migrations be looking for skipped docs?

abea (changes requested):

modules/@apostrophecms/migration/index.js
boutell: No guidance is required. What's being skipped, and marked as such, is the entire migration, not an apostrophe document. The migration won't be run at all because we know it doesn't have to be if the site is brand new. We mark it just in case we change our minds about this.

modules/@apostrophecms/migration/index.js
boutell: Fixed.

modules/@apostrophecms/doc/index.js
boutell: I could have used [looks at latest recommended mongo APIs] `countDocuments`, sure. Works fine though; how about we open a ticket to refine this as it's well tested and can make it into this release?

modules/@apostrophecms/migration/index.js
abea: Right, it's that "just in case we change our minds" bit that I think should be documented somewhere. It could be a documentation ticket to be done later, but the feature is useless if it's not documented. If it's useless, why have it?

modules/@apostrophecms/doc/index.js
abea: I don't think that ticket would ever get done. Would that be a big lift to change? It seems like you're doing extra work to get to the same place as a method that already exists to do this.

abea (changes requested):
Definitely the documentation one. The count thing would be good to get in as well if it's not time-intensive.

boutell:
Turns out that code was shaggy in other ways, so thanks for the check.

modules/@apostrophecms/doc/index.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
     self.managers = {};
     self.enableBrowserData();
     await self.enableCollection();
-    self.apos.isNew = self.detectNew();
+    self.apos.isNew = await self.detectNew();
```
abea, line 33: good catch.

abea (approved):

PR #2844, opened 2021-03-22
PRO-1020 the page chooser should not allow choosing trash.

boutell:
The "all" option of the pages rest API needed an upgrade to respect a "trash" query parameter and the component needed to send it.

### DIFF
--- a/modules/@apostrophecms/page/index.js
+++ b/modules/@apostrophecms/page/index.js
@@ -74,6 +74,7 @@ module.exports = {
       async getAll(req) {
         self.publicApiCheck(req);
         const all = self.apos.launder.boolean(req.query.all);
+        const trash = self.apos.launder.booleanOrNull(req.query.trash);
         const flat = self.apos.launder.boolean(req.query.flat);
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/page/index.js
```diff
@@ -94,7 +95,7 @@ module.exports = {
           }
           const page = await self.getRestQuery(req).and({ level: 0 }).children({
             depth: 1000,
-            trash: null,
+            trash: trash,
```
stuartromanek, line 98: just `trash` ?

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -210,8 +210,8 @@ export default {
         '/api/v1/@apostrophecms/page', {
           busy: true,
           qs: {
-            all: 1,
-            trash: null
+            all: '1',
+            trash: this.relationshipField ? '0' : 'any'
```
stuartromanek, line 214: had to read `launder`'s source but i see that this is legal 👍

stuartromanek (approved):
optional

PR #2845, opened 2021-03-22
both piece and page choosers lacked support for max in the expanded m…

boutell:
…odal, and also page chooser modals completely lacked the relationship editor, necessary for ordering and sometimes relationship fields

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
+++ b/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
@@ -14,7 +14,7 @@ export default {
       // as initially checked.
       checked: Array.isArray(this.chosen) ? this.chosen.map(item => item._id)
         : [],
-      checkedDocs: Array.isArray(this.chosen) ? klona(this.chosen) : false
+      checkedDocs: Array.isArray(this.chosen) ? klona(this.chosen) : []
     };
   },
   props: {
@@ -88,9 +88,6 @@ export default {
       }
     },
     checked () {
-      if (!this.checkedDocs) {
-        return;
-      }
       this.updateCheckedDocs();
     }
   },
--- a/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
+++ b/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
@@ -33,11 +33,13 @@ export default {
       if (!this.relationshipField) {
         return false;
       }
-
+      if (this.relationshipField.required && !this.checked.length) {
+        // Treated as min for consistency with AposMinMaxCount
+        return 'min';
+      }
       if (this.relationshipField.min && this.checked.length < this.relationshipField.min) {
         return 'min';
       }
-
       if (this.relationshipField.max && this.checked.length > this.relationshipField.max) {
         return 'max';
       }
--- a/modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
+++ b/modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
@@ -31,10 +31,25 @@
       <AposButton
         v-if="relationshipField"
         type="primary"
-        label="Select Pages"
-        :disabled="relationshipErrors === 'min'"
+        :label="saveRelationshipLabel"
+        :disabled="(relationshipErrors === 'min') || (relationshipErrors === 'max')"
         @click="saveRelationship"
       />
+    </template>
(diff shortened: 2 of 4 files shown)


modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -14,7 +14,7 @@ export default {
       // as initially checked.
       checked: Array.isArray(this.chosen) ? this.chosen.map(item => item._id)
         : [],
-      checkedDocs: Array.isArray(this.chosen) ? klona(this.chosen) : false
+      checkedDocs: Array.isArray(this.chosen) ? klona(this.chosen) : []
```
boutell, line 17: There was only one place this was tested in a boolean way, and it wasn't necessary, so a consistent type is less confusing

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -88,9 +88,6 @@ export default {
       }
     },
     checked () {
-      if (!this.checkedDocs) {
```
boutell, line 91: Always truthy now

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -298,6 +320,9 @@ export default {
         this.checked.push(doc._id);
       }
     },
+    setCheckedDocs(checkedDocs) {
+      this.checked = checkedDocs.map(doc => doc._id);
```
boutell, line 324: This will cause checkedDocs to be reactively refilled, this dance is present because of the mixin, it makes more intuitive sense with pieces

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -306,4 +331,13 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+  .apos-pages-manager__relationship__rail {
```
boutell, line 334: Borrowed from the pieces manager, looks good?

modules/@apostrophecms/schema/ui/apos/components/AposInputAttachment.vue
```diff
@@ -106,7 +106,6 @@ export default {
   methods: {
     watchNext () {
       this.validateAndEmit();
-      this.limitReached = !!this.next;
```
boutell, line 109: Fixed unrelated console error. This is a computed properly, shouldn't be set.

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -31,11 +31,26 @@
       <AposButton
         v-if="relationshipField"
         type="primary"
-        label="Select Pages"
-        :disabled="relationshipErrors === 'min'"
+        :label="saveRelationshipLabel"
+        :disabled="(relationshipErrors === 'min') || (relationshipErrors === 'max')"
```
stuartromanek, line 35: what errors would not warrant disabling save? Should this just be `:disabled ="!!(relationshipErrors)"` ?

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
stuartromanek: pieces manager is also wrong, see `AposArrayEditor` and extend the rail component, lose the height/bg color styles

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -28,8 +28,8 @@
       <AposButton
         v-if="relationshipField"
         type="primary"
-        :label="`Select ${moduleLabels.pluralLabel || ''}`"
-        :disabled="relationshipErrors === 'min'"
+        :label="saveRelationshipLabel"
+        :disabled="(relationshipErrors === 'min') || (relationshipErrors === 'max')"
```
stuartromanek, line 32: same Q from `AposPagesManager`

stuartromanek (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
boutell: Oops, caught this in new commit.

stuartromanek (approved):

PR #2850, opened 2021-03-23
deleted unnecesary assignation of choices

ecb34:
field.idField always should be empty for a joinByArray

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
-## UNDEPLOYED 
-* Deleted unnecesary assignation of choices in addFieldType joinByArray
+## Unreleased 
+* Deleted unnecesary assignment of choices in addFieldType joinByArray
 
 ## 2.117.0 (2021-03-24)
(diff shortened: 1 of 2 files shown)


abea (commented):
The more obvious change there would be to use `idsField` instead of `idField`. Can you explain why this isn't needed?

I'm not super familiar with this UI code myself, so I'm looking to understand where this might end up. @boutell Do you have an insight there?

ecb34:
because you set the choices afterwards in the set method, I think that chooser.set([]) can also be removed, but Iḿ not 100% sure
image

abea:
👍 Ah, that looks right. Tagging @boutell for an additional review to be sure.

boutell (changes requested):
The fix is correct, thank you @ecb34! I appreciate you tracking this down.

We should change "assignation" to "assignment." An assignation is something much sneakier 😂 But that is a minor concern.

ecb34:
so chooser.set([]) should be also removed or is necessary for something?

boutell:
Your code is completely correct. Just need to fix the word "assignation" to "assignment" in the changelog. Thanks!

abea (approved):

boutell (approved):

PR #2856, opened 2021-03-26
PRO-1105 save draft of new thing should always be permitted

boutell:


abea (changes requested):
Please update the changelog

abea (approved):

PR #2857, opened 2021-03-26
Adds CI test for node 14 with mongo 4

abea:
Adds a CI test to start checking Node 14 on the latest version of MongoDB and npm as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,30 @@
 version: 2
 jobs:
-  build-node14:
+  build-node14-mongo44:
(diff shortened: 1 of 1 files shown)


boutell (changes requested):
You should add entries for both "node 14, mongo 4.2" and "node 14, mongo 4.4". 4.4 is latest stable, but 4.2 is still in heavy circulation. I like that the node 12 test checks mongo 3.6. That is probably enough coverage for now.

boutell (approved):

PR #2859, opened 2021-03-26
PRO-117 in-context editing of areas inside array items

boutell:


abea:
We should add to the changelog for this.

boutell:
I added it to the changelog.

abea (approved):

PR #2861, opened 2021-03-29
PRO-1108 Detect a2

boutell:
A3 uses the database similarly enough that if a developer tries pointing A3 at an existing A2 database, it'll make a mess, but just subtly enough that they might not figure out what happened. Ouch. So prevent that.

Content migration tools from A2 to A3 are planned of course.

### DIFF
--- a/modules/@apostrophecms/db/index.js
+++ b/modules/@apostrophecms/db/index.js
@@ -24,6 +24,13 @@
 // connection instance is created that reuses the same sockets,
 // and `uri`, `host`, `connect`, etc. are ignored.
 //
+// ### `versionCheck`
+//
+// If `true`, check to make sure the database does not belong to an
+// older, incompatible major release release of Apostrophe and exit if it does.
+// Defaults to `true`. You can set this to `false` to avoid an extra query at startup
+// time.
+//
 // ## Command line tasks
 //
 // ### `@apostrophecms/db:reset`
@@ -46,8 +53,12 @@
 const mongo = require('mongodb');
 
 module.exports = {
+  options: {
+    versionCheck: true
+  },
   async init(self) {
     await self.connectToMongo();
+    await self.versionCheck();
     // TODO: Remove this conditional and `self.trace` if not necessary or add
     // documentation explaining utility and usage.
     if (process.env.APOS_TRACE_DB) {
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/db/index.js
```diff
@@ -24,6 +24,13 @@
 // connection instance is created that reuses the same sockets,
 // and `uri`, `host`, `connect`, etc. are ignored.
 //
+// ### `versionCheck`
+//
+// If `true`, check to make sure the database does not belong to an
+// older, incompatible major release release of Apostrophe and exit if it does.
+// Defaults to `true`. You can set this to `false` to avoid an extra query at startup
```
abea, line 31: If the check is only at start up, what do we gain by ever turning this off?

modules/@apostrophecms/db/index.js
```diff
@@ -124,9 +135,23 @@ module.exports = {
...
+          return;
+        }
+        const oldGlobal = await self.apos.db.collection('aposDocs').findOne({
+          type: 'apostrophe-global'
+        });
+        if (oldGlobal) {
+          throw new Error(`There is a problem with the database: ${self.uri ? (self.uri + ' ') : ''}
 
+This database already contains an Apostrophe 2.x website. Exiting to avoid
+causing content loss. Content migration tools from 2.x to 3.x are planned.`);
```
abea, line 152: Let's not include this last sentence. It dates the error message and code like this isn't where we should be discussing plans.

boutell: @localghost443 wanted to sign off on the language too so I'm requesting review from him so we can change this once rather than twice.

modules/@apostrophecms/db/index.js
boutell: Apostrophe does a lot at startup. It's time-consuming, and enterprise clients have historically asked for options to tone it down where possible when you know what you're doing.

abea: 👍

modules/@apostrophecms/db/index.js
localghost443: "There is a problem with the database: mongodb://localhost:27017/whatever

This database contains an Apostrophe 2.x website. Exiting to avoid content loss."

Removed the last sentence and tweaked slightly for brevity.

localghost443 (changes requested):

abea (approved):

PR #2868, opened 2021-03-30
PRO-1083 when a page is moved to a trash, it should only exist as a draft. Thi…

boutell:
…s avoids conflict with the mechanisms that autopublish page tree movements for previously published pages to prevent UX and permissions nightmares previously discussed and resolved

- [ ] Link to the related issue with requirements and/or clearly describe 1) what problem this solves and 2) the requirements to review it against.
- [ ] Update the `CHANGELOG.md` file with the added feature or fix. If there is not yet a heading for an upcoming release, add one following the established pattern.
- [ ] Keep the pull request minimal! Break large updates into separate pull requests for individual features/fixes whenever possible. Small requests get reviewed more quickly.
- [ ] All tests must pass, including the linters. Run `npm run lint` to test code linting independently.

Thanks for contributing!

### DIFF
--- a/modules/@apostrophecms/page-type/index.js
+++ b/modules/@apostrophecms/page-type/index.js
@@ -118,6 +118,11 @@ module.exports = {
           if (!doc._id.includes(':draft')) {
             return;
           }
+          if (doc.parkedId === 'trash') {
+            // The root trash can exists in both draft and published to
+            // avoid overcomplicating parked pages
+            return;
+          }
           return self.apos.doc.db.removeMany({
             _id: {
               $in: [
--- a/modules/@apostrophecms/page-type/index.js
+++ b/modules/@apostrophecms/page-type/index.js
@@ -123,6 +123,13 @@ module.exports = {
             // avoid overcomplicating parked pages
             return;
           }
+          await self.apos.doc.db.updateOne({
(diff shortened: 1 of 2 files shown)


stuartromanek:
Still working on testing the actual case of trashing a page, but this came up consistently and is **not** the case when doing the same thing on the `3.0` branch

image

To get this all i have to do i `node app.js @apostrophecms/db:reset && node app.js @apostrophecms/user:add stuart admin` and login to the page manager

stuartromanek:
I also get this generic error when creating a new page from this branch

image

stuartromanek (changes requested):
See previous comments, seems like some major blockers in this branch

boutell:
I was able to reproduce the mess and I think I've addressed it based on my tests.

stuartromanek (changes requested):
Trashing a page with children results in the parent page's `published` doc being removed from the db but not the children, they remain

to reproduce
1. create page
2. create child page of step one
3. go to page manager, drag page in step 1 to the trash
4. poke around in mongo for the child page

I assume this case should follow suit, although its not outlined

boutell:
Addressed this subpage situation.

stuartromanek (approved):

PR #2872, opened 2021-04-01
PRO-164 refactor of admin bar, also includes PRO-1002

boutell:
Also the tray (for palette toggle) needs to be styled.

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -4,247 +4,20 @@
     <nav class="apos-admin-bar" ref="adminBar">
       <div class="apos-admin-bar__row">
         <AposLogoPadless class="apos-admin-bar__logo" />
-        <ul class="apos-admin-bar__items">
-          <li class="apos-admin-bar__item" v-if="pageTree">
-            <AposButton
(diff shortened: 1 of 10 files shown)


boutell:
This will also resolve PRO-1002 when complete.

stuartromanek:
Page title no longer shows in context doc bar 

Screen Shot 2021-04-02 at 8 47 34 AM

stuartromanek (changes requested):
context doc title and button styles

stuartromanek:
lot of green and red in this diff, @boutell is there anything I should be looking at extra hard? Noted the surface level differences but if there are more subtle items could you call them out?

stuartromanek:
@boutell when testing this with the assembly boilerplate + palette, there are some strange reactivity cases changing values of palette. Sometimes the Publish Changes button lights up after a change, sometimes not (usually after you've already hit Publish once that page load) .. sometimes it re enables itself seemingly without prompt .. Not sure if you'd consider that bug part of this work

boutell:
I fixed the two bugs you mentioned. I was not able to reproduce a problem with the "publish changes" button so far. I suspect that it would be a preexisting issue, because in general with this refactoring so far, either it works like before or it's just not hooked up at all. Please do open a separate ticket if you can figure out how to reproduce that one.

stuartromanek (approved):

PR #2873, opened 2021-04-02
PRO-1115 switch edit mode if destination doc + mode will cause a 404 from a manager

stuartromanek:


stuartromanek:
this work is based on the new organization of `3/pro-164-refactor-admin-bar`

modules/@apostrophecms/page-type/index.js
```diff
@@ -376,6 +376,9 @@ module.exports = {
   },
   extendMethods(self) {
     return {
+      enableAction() {
```
stuartromanek, line 379: @boutell as far as I could tell a custom page's `MODULE.action` was not actually useful, we often hard code in an action of `@apostrophecms/page` to do things like get documents.

In a consult w @abea some version of this seemed reasonable

modules/@apostrophecms/page/ui/apos/components/AposPagesManager.vue
```diff
@@ -138,7 +138,12 @@ export default {
           {
             columnHeader: 'Link',
             property: '_url',
-            type: 'link',
+            action: {
+              name: 'set-context-if-needed-and-redirect',
+              component: 'bus',
+              params: [ '_id', 'type', '_url' ]
+            },
```
stuartromanek, line 145: expand `action`'s scope to be able to emit events from the bus

stuartromanek:
As per our last scoping meeting this issue should be handled by a more global mechanism that detects whether _any_ requests should result in an on-the-fly mode switch that avoids a 404.

I still think [these changes](https://github.com/apostrophecms/apostrophe/pull/2873#discussion_r606366384) might be worth rolling in, possibly even the expanded `action` logic, although nothing would make use of it at the moment.

modules/@apostrophecms/page-type/index.js
abea: Later we could probably clean up some other code with this available.

boutell:
Closing this PR to avoid confusion here and in linear but I don't have anything against the idea of rolling in some desirable code in a separate PR.

PR #2874, opened 2021-04-02
find the alt attribute as we pass by in image.first and friends

boutell:
Turns out we were very close to doing this already, so I just updated the list of attributes that make sense to capture.

I simplified the logic a bit by always cloning the attachment object, which is not complex.


modules/@apostrophecms/image-widget/views/widget.html
```diff
@@ -3,7 +3,7 @@
   <img{% if data.manager.options.className %} class="{{ data.manager.options.className }}"{% endif %}
     srcset="{{ apos.image.srcset(attachment) }}"
     src="{{ apos.attachment.url(attachment, { size: data.options.size or 'full' }) }}"
-    alt="{{ data.widget._image[0].alt or '' }}"
+    alt="{{ attachment._alt or '' }}"
```
boutell, line 6: I tested this.

abea (approved):

PR #2877, opened 2021-04-05
PRO-1121 search in media

boutell:
This will throw a bunch of console errors as you type, that is an unrelated bug which I just fixed in a different PR, the search will work.

I also cleaned up a bunch of inaccurate `emits` sections.

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
@@ -373,9 +373,9 @@ export default {
       this.$emit('trash', this.checked);
     },
 
-    search(query) {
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -373,9 +373,9 @@ export default {
       this.$emit('trash', this.checked);
     },
 
-    search(query) {
-      // TODO stub
-      this.$emit('search', query);
+    async search(query) {
+      console.log(`query is ${query}`);
```
abea, line 377: stray log

abea (approved):

PR #2878, opened 2021-04-05
PRO-383 remove profiles

boutell:
I'm done with the current sprint, so I'm working on the next one and targeting the beta 1 branch here to avoid any conflict with landing the last alpha smoothly this week.


abea (approved):

abea (changes requested):
Is 3.0.0-beta.1 the right branch to merge into?

PR #2879, opened 2021-04-05
WIP. You can edit the role but I am still building out the grid that …

boutell:
…explains the role. Roles have no actual impact yet, that's in other tickets.


abea:
We can close this, right @boutell ? It looks like it was merged into `3.0` already. #2899

PR #2880, opened 2021-04-05
PRO-1130 PRO-1083 trash is really truly unpublished now

boutell:


modules/@apostrophecms/doc-type/index.js
```diff
@@ -139,6 +139,39 @@ module.exports = {
         }
       },
       afterTrash: {
+        // Mark draft only after moving to trash, to reactivate UI associated with
+        // things never published before
```
abea, line 143:
```suggestion
        // Mark draft only after moving to trash, to reactivate UI
        // associated with things never published before
```

abea: Aligns to 80 chars

abea (changes requested):
Only a small formatting improvement

abea (approved):

PR #2882, opened 2021-04-05
navigating to another page always puts you back in published mode, th…

boutell:
…e mode is no longer recorded in the session

This is a draft because we need to talk about the UX consequences. Simply refreshing the page after your edits have saved currently appears to wipe them away, and longtime apostrophe developers tend to refresh for reassurance that autosave worked. If necessary it is technically possible to save some state browser side after all and still keep the architectural improvement of not using the session.

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
@@ -40,17 +40,19 @@ export default {
   name: 'TheAposContextBar',
   mixins: [ AposPublishMixin, AposAdvisoryLockMixin ],
   data() {
-    // Since cookie-based login sessions and sessionStorage are not precisely the same
-    // thing, correct any forbidden combination at page load time
-    if (window.apos.mode === 'published') {
-      window.sessionStorage.setItem('aposEditMode', JSON.stringify(false));
+    const query = apos.http.parseQuery(location.search);
+    // If the URL references a draft, go into draft mode but then clean up the URL
+    const draftMode = query['apos-mode'] || 'published';
+    if (draftMode === 'draft') {
+      delete query['apos-mode'];
+      history.replaceState(null, '', apos.http.addQueryToUrl(location.href, query));
     }
     return {
       patchesSinceLoaded: [],
       undone: [],
       patchesSinceSave: [],
-      editMode: window.sessionStorage.getItem('aposEditMode') === 'true',
-      draftMode: window.apos.mode,
+      editMode: false,
+      draftMode,
       original: null,
       saving: false,
       editing: false,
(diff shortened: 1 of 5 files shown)


boutell:
We agreed to go ahead and land this PR, however there will be a followup ticket to address not losing edit mode merely because you refreshed the page, using browser side session storage rather than an old-fashioned req.session thing.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -40,17 +40,19 @@ export default {
   name: 'TheAposContextBar',
   mixins: [ AposPublishMixin, AposAdvisoryLockMixin ],
   data() {
-    // Since cookie-based login sessions and sessionStorage are not precisely the same
-    // thing, correct any forbidden combination at page load time
-    if (window.apos.mode === 'published') {
-      window.sessionStorage.setItem('aposEditMode', JSON.stringify(false));
+    const query = apos.http.parseQuery(location.search);
+    // If the URL references a draft, go into draft mode but then clean up the URL
+    const draftMode = query['apos-mode'] || 'published';
```
stuartromanek, line 45: the `draftMode` var can be `'published'`? should this variable just be `mode` ?

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -308,26 +310,32 @@ export default {
...
+            'apos-locale': locale
           }
         });
+        if (navigate && (!modeDoc._url)) {
+          await apos.alert({
+            heading: 'Page Does Not Exist Yet',
+            description: `The page that provides a listing for this type of piece is not yet available as ${mode} in the ${locale} locale.`
+          });
+          return;
+        }
```
stuartromanek, line 326: this will get axed when we land second chance 404 draft mode, yes? Should we add that comment?

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
stuartromanek: or, isn't this really the version we're referencing? should it be `version` ?

modules/@apostrophecms/login/index.js
```diff
@@ -158,36 +158,6 @@ module.exports = {
             await destroySession();
           }
         },
-        async setContext(req) {
-          if (!(self.apos.i18n.isValidLocale(req.body.locale) && [ 'draft', 'published' ].includes(req.body.mode))) {
-            throw self.apos.error('invalid');
-          }
-          const locale = req.body.locale;
-          const mode = req.body.mode;
```
stuartromanek, line 166: Are we going to need to bring this right back when we have second chance context switch? or will that be handled elsewhere

stuartromanek (approved):
i left some questions but none of them are dealbreakers

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
boutell: This is a previously established and released convention in the component, so I'd like to not reopen the subject of the name of this variable, at least in this ticket.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
boutell: This PR contains second chance 404 draft mode, which is working so far as I know. But, this alert is a good fallback because the setContext API can also be called during the time you're on the page, and because of potential situations with locales in the future. The second chance 404 mode is only for one case: (1) you arrived on the page "naturally" via a page transition, and (2) it's about to 404.

modules/@apostrophecms/login/index.js
boutell: No, the second chance switch is in this PR, see below.

modules/@apostrophecms/page/index.js
```diff
@@ -1395,6 +1395,29 @@ database.`);
         if (self.isFound(req)) {
           return;
         }
+        if (req.user && (req.mode === 'published')) {
```
boutell, line 1398: This is how the second chance switch works. Server side, when it's about to 404, we look for whether the same URL would succeed in draft mode. if it would, we redirect with a query parameter as a hint that gets removed later via the history API so people don't share ugly URLs.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -40,17 +40,19 @@ export default {
...
-    // Since cookie-based login sessions and sessionStorage are not precisely the same
-    // thing, correct any forbidden combination at page load time
-    if (window.apos.mode === 'published') {
-      window.sessionStorage.setItem('aposEditMode', JSON.stringify(false));
+    const query = apos.http.parseQuery(location.search);
+    // If the URL references a draft, go into draft mode but then clean up the URL
+    const draftMode = query['apos-mode'] || 'published';
+    if (draftMode === 'draft') {
+      delete query['apos-mode'];
+      history.replaceState(null, '', apos.http.addQueryToUrl(location.href, query));
```
boutell, line 48: This query parameter got here via the second chance redirect in the page module, we needed it to load the URL with draft content, but now we can remove it from the address bar so folks don't get stuck sharing / linking to URLs with an inappropriate query parameter for most audiences.

stuartromanek (approved):

PR #2890, opened 2021-04-07
PRO-1009 3/disable unpublished relationship choices

stuartromanek:
https://linear.app/apostrophecms/issue/PRO-1009/drafts-and-relationships

This PR only disables unpublished choices in a relationship manager.

As I understand it, the way the above ticket describes _how_ managers display pieces (a combination of live and draft-only) will become the default across all managers and is not specific to relationships.

@boutell @localghost443 If i'm wrong about the division of work, feel free to reject. If there are more implementation tickets to make let's make sure they get made.


boutell (approved):
Covers this part just fine, will help with opening other tickets as you nmentioned.

localghost443 (approved):

PR #2896, opened 2021-04-08
3/chrome source map crash

stuartromanek:
blindly following suggestions from https://github.com/webpack/webpack/issues/7199

### DIFF
--- a/modules/@apostrophecms/asset/lib/webpack.scss.js
+++ b/modules/@apostrophecms/asset/lib/webpack.scss.js
@@ -12,7 +12,8 @@ module.exports = (options, apos) => {
             {
               loader: 'css-loader',
               options: {
-                esModule: false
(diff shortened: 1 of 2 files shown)


boutell (changes requested):
This kills the ability to debug the SASS, so it's not a sufficient solution yet.

We have a few components that contain two "style" tags. There's a reason for that: one is scoped, the other is not. However, I came across a report that two "style" tags in the same Vue single-file component can cause the behavior we're seeing. Would you please try temporarily removing that to see if it fixes the issue?

AposInputColor
TheAposLogin

I think that's all

stuartromanek:
@boutell please read https://linear.app/apostrophecms/issue/PRO-1084/chrome-crashes-whenever-you-inspect-an-apostrophe-dom-node

PR #2898, opened 2021-04-09
Adds warning for malformed $or conditionals

abea:
Do we need this? I ran into it trying an object and thought this error handling could help.


boutell (changes requested):
Good idea, but it belongs in the schema module which already does a lot of one time startup validation, rather than all the time browser console stuffs.

boutell (approved):
Sounds good. You can do `JSON.stringify(thing, null, '  ')` to get readable formatting and line breaks, but I'm not sure if you even wanted a multiline response here.

PR #2899, opened 2021-04-10
3/pro 382 edit role

boutell:
This PR implements PRO-382, "as an admin, I can edit a user's role." However I haven't matched the visual design entirely, specifically there is just one column and I think I have at least one of the icons wrong. I think it would be a good idea for me to hand off that part.


stuartromanek (changes requested):
- On a fresh reset + new user creation (in the normal style `node app.js @apostrophecms/user:add stuart admin`), my user's permissions are reflected as Guest
- I assume I'm supposed to see some kind of UI here, but i do not
image
- Related: We should have a follow up ticket for updating @apostrophecms/palette with the new `singleton: true` setting for it's piece type.

stuartromanek (approved):

PR #2904, opened 2021-04-12
PRO-1104 3/preview draft

stuartromanek:
https://linear.app/apostrophecms/issue/PRO-1104/preview-button-on-pagepiece-creation

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -31,6 +31,10 @@
         @copy="onCopy"
       />
       <AposButton
+        v-if="canPreviewDraft" type="secondary"
+        @click="saveDraftAndPreview" label="Preview Draft"
+      />
+      <AposButton
         type="primary" :label="saveLabel"
         :disabled="saveDisabled"
         @click="submit"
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -525,6 +540,16 @@ export default {
         }
         this.$emit('modal-result', doc);
         this.modal.showModal = false;
+        if (navigate) {
+          if (doc._url) {
+            window.location = doc._url;
+          } else {
+            await apos.notify(`Draft saved but could not navigate to a preview. Try creating a ${(this.moduleOptions.label || '')} index page`, {
```
stuartromanek, line 547: this language stinks

stuartromanek:
@boutell Do I need to do anything before [kicking the user out to the `_url`](https://github.com/apostrophecms/apostrophe/pull/2904/files#diff-deaa2014c6fa12403ae9a4304464a5fa140efe1aa78ea9502e562fde73782b3eR545) ? .. Currently this gives me a 404 even when the index page is ready to go.

boutell (approved):
This logic is correct. If it 404s even though _url is set, then that's a bug for me, please open a ticket for me. I think it's as simple as make an articles page in a3-demo and create a draft article, hit preview draft, 404, right?

PR #2908, opened 2021-04-14
PRO-1177 username following

abea:
Reduces user creation steps by telling the username field to follow the title.


boutell (approved):
This is not a blocker, but in addition, should we do any transformations here? i.e. do we want to encourage mixed case and punctuation and spaces in usernames? I don't think we've ever forbidden them, so maybe yes? But it might not be what users find convenient to remember and type later.

PR #2909, opened 2021-04-14
Changes users title field to Display Name

abea:
It may not always be a full human name or they  may not want to give their full name. We're requiring it, so we can reduce mental friction by calling it Display Name instead.


stuartromanek (approved):
can't be mad

PR #2910, opened 2021-04-14
PRO-1161 restrict permissions by role

boutell:
Permissions restrictions by role are now fully enforced.

This is *not* a PR to adjust the UI so we don't show inappropriate functionality, etc. I'll be addressing that in other tickets. However it's actually pretty convenient for evaluation purposes, because you can try things and get appropriate error messages. (Bearing in mind that most of those would never be seen under normal circumstances once we do add the UI affordances, so we don't need to go nuts making every "forbidden" pretty.)

### DIFF
--- a/modules/@apostrophecms/permission/index.js
+++ b/modules/@apostrophecms/permission/index.js
@@ -41,7 +41,7 @@ module.exports = {
       // the action on that doc type generally.
       //
       // If a permission is not specific to particular documents,
-      // for instance `view-draft`, `docOrType` may be omitted.
+      // `docOrType` may be omitted.
       //
       // The doc passed need not already exist in the database.
       //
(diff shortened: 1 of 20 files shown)


boutell:
Also, the grid of roles is now populated with the truth (although there are separate tickets to adjust the grid on various points).

boutell:
There is a separate ticket for the page tree rules, I implemented the most obvious one here (contributors can't drag published pages) but not the stuff about the trash. Would like to land this incremental step first.

boutell:
If there's anything I *didn't do yet* in this ticket, I would appreciate a followup ticket, as long as this implementation is good in what it does do and constitutes an iterative improvement. We need to land this soon so I can start to put other tickets truly "in review."

modules/@apostrophecms/permission/index.js
```diff
@@ -67,28 +103,82 @@ module.exports = {
...
+            };
+          } else if (role === 'editor') {
+            return {
+              type: {
+                $nin: restrictedTypes
+              }
+            };
+          } else {
+            return {
+              _id: 'thisIdWillNeverMatch'
```
abea, line 161: Can't we throw an error? Somehow they've tried to initiate an editing action without being a contrib, editor, or admin. Something is actually wrong there.

modules/@apostrophecms/permission/index.js
```diff
@@ -67,28 +103,82 @@ module.exports = {
...
+        } else if (action === 'publish') {
+          if (role === 'editor') {
+            return {
+              type: {
+                $nin: restrictedTypes
+              }
+            };
+          } else {
+            return {
+              _id: 'thisIdWillNeverMatch'
```
abea, line 173: Same here and below

modules/@apostrophecms/permission/ui/apos/components/AposInputRole.vue
```diff
@@ -109,8 +114,13 @@ export default {
       // Allows expression of non-string values
       this.next = this.choices.find(choice => choice.value === JSON.parse(value)).value;
     },
-    async getTypes() {
-      return (await apos.http.get(`${apos.permission.action}/grid`, {})).types;
+    async getTypes(role) {
```
abea, line 117: It's really `getPermissions`, right? Or `getPermissionSets`? `getTypes` seems like it's populating `choices`.

modules/@apostrophecms/task/index.js
```diff
@@ -203,35 +206,56 @@ module.exports = {
...
+      // object simulating a user with the editor role.
+      getEditorReq(options) {
+        return self.getReq({
+          role: 'editor',
+          ...options
+        });
+      },
+
+      // Convenience wrapper for `getReq`. Returns a request
+      // object simulating a user with the editor role.
```
abea, line 253: Typo here: "simulating a user with the editor role"

modules/@apostrophecms/task/index.js
```diff
@@ -203,35 +206,56 @@ module.exports = {
...
+      // and no `req.user`.
+      getAnonReq(options) {
+        return self.getReq({
+          role: 'anon',
+          ...options
+        });
+      },
+
+      // Convenience wrapper for `getReq`. Returns a request
+      // object simulating a user with the editor role.
```
abea, line 226: Typo here: "simulating a user with the editor role"

test/users.js
```diff
@@ -31,6 +31,7 @@ describe('Users', function() {
     user.username = 'JaneD';
     user.password = '123password';
     user.email = 'jane@aol.com';
+    user.role = 'admin';
```
abea, line 34: Maybe add a test or tests showing other user roles can't insert or edit users. A big CYA there.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -178,8 +178,8 @@ module.exports = {
       return result;
     },
     async getOne(req, _id) {
-      self.publicApiCheck(req);
       _id = self.inferIdLocaleAndMode(req, _id);
+      self.publicApiCheck(req);
```
abea, line 182: Why move these?

abea (changes requested):
Contributor:
- Moving an unpublished page is unpredictable. I had two published pages created by the admin, then added a page as a Contrib. As the Contrib, I tried to move that page about. I could move it to be the first child of Home, but not between the two others... unless I was dragging it out of the archive. It's not clear what should be working there.
- The inability to discard a page draft is murky territory. I guess it makes sense to not be able to do that as a Contrib.

modules/@apostrophecms/permission/index.js
boutell: I wanted to make this a little less brittle than that, in case someone is using the super pattern on it; exception throwing is a little expensive if it has to happen every time and then get caught and turned into something else. Whereas changing the return value would be lightweight and speedy.

modules/@apostrophecms/task/index.js
boutell: I don't see the typo?

modules/@apostrophecms/task/index.js
boutell: This is intentional.

test/users.js
boutell: Agreed will do that.

modules/@apostrophecms/piece-type/index.js
boutell: Inferring the mode alters the result of the permissions check made in publicApiCheck, allowing contributors to do things they should be able to do without forcing us to change our REST API design to always require ?apos-mode even when it can be inferred from the _id.

boutell:
* "Moving an unpublished page is unpredictable:" this was actually a bug in "move." Fixed.
* "Discard draft": this now works for contributors, as long as it's a never-published document. I hadn't implemented the "delete" permission.

abea (approved):

abea (changes requested):
still think we throw errors

abea (approved):
Looks good. We could do the same `_id: null` on L184 instead of the error now, but also would be good to merge as-is

PR #2913, opened 2021-04-14
PRO-366 admin and context bar permissions

boutell:
This is ready for review after PRO-1161, on which it is based. Please go there first.

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -35,6 +35,7 @@ module.exports = {
     self.addDuplicateOrMissingWidgetIdMigration();
     self.addDraftPublishedMigration();
     self.addLastPublishedToAllDraftsMigration();
+    self.addAposModeMigration();
   },
   restApiRoutes(self) {
     return {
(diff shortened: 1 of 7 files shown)


abea (commented):
This should be just about good to go. I'm seeing a CI failure, though. I'd say merge the other branch into this and see if it clears up?

modules/@apostrophecms/doc/index.js
```diff
@@ -75,10 +76,11 @@ module.exports = {
   handlers(self) {
     return {
       '@apostrophecms/doc-type:beforeInsert': {
-        setLocale(req, doc, options) {
+        setLocaleAndMode(req, doc, options) {
           const manager = self.getManager(doc.type);
           if (manager.isLocalized()) {
             doc.aposLocale = doc.aposLocale || `${req.locale}:${req.mode}`;
+            doc.aposMode = req.mode;
```
abea, line 83: How does this update from draft to published? Publishing isn't creating a new document, is it?

boutell: This is a generic listener for "insert" of any document. When any document that is of a localizable type (not users) gets inserted, its `aposMode` property needs to get set, to facilitate queries without writing a lot of regexes in mongo based on aposLocale or _id.

This isn't part of the mechanism of publishing things, that's at a higher level of the stack. This is just guaranteeing that certain queryable properties exist for documents in certain modes.

abea: This is throwing me. Does `aposLocale` need to include the mode, then? I get that it's easier to not have to use regex to parse those properties in many circumstances, so why don't we just have these things as separate properties?

abea (approved):

PR #2917, opened 2021-04-16
Propose Changes, Manage Submitted Drafts: basic implementation

boutell:
This doesn't cover every planned feature. Want to land this and keep iterating on more tickets. This PR covers:

* Change "Publish" to "Propose Changes" when appropriate
* This carries out a submit draft operation
* Submitted drafts can be browsed via a virtual piece type (TODO: show how many in the admin bar per design, I am opening a ticket)

This is already useful. Currently your only option when browsing submitted drafts is to edit them and click "publish changes" or "discard draft" via the more menu of the individual page or piece's settings modal, or do the same after clicking its link, but that takes you quite far.


modules/@apostrophecms/admin-bar/index.js
```diff
@@ -251,14 +251,11 @@ module.exports = {
           req.res.setHeader('Cache-Control', 'no-cache');
         }
         let contextEditorName;
-        let contextAction;
```
boutell, line 254: We get this dynamically now, see getContextModule.

modules/@apostrophecms/page/index.js
```diff
@@ -597,6 +643,15 @@ database.`);
       async patch(req, _id) {
         return self.withLock(req, async () => {
           const input = req.body;
+          const keys = Object.keys(input);
+          let possiblePatchedFields;
```
boutell, line 647: Avoids updating updatedAt just because of a lock. Already had this fix in pieces.

modules/@apostrophecms/page/index.js
```diff
@@ -676,14 +735,15 @@ database.`);
           components: {}
         });
         _.defaults(browserOptions.components, {
-          insertModal: 'AposDocEditor',
+          editorModal: 'AposDocEditor',
```
boutell, line 738: More accurate, used for all cases.

modules/@apostrophecms/submitted-draft/index.js
```diff
@@ -0,0 +1,151 @@
...
+            $exists: 1
+          }
+        });
+      }
+    };
+  },
+  queries(self, query) {
+    return {
+      builders: {
+        _type: {
```
boutell, line 124: The "submitted drafts" manager has a working "type" filter.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
```diff
@@ -276,23 +279,37 @@ export default {
...
-        moduleName: this.moduleName,
-        docId: pieceId,
+    async edit(piece) {
+      let moduleName;
+      // Don't assume the piece has the type of the module,
+      // this could be a virtual piece type such as "submitted-draft"
+      // that manages docs of many types
+      if (piece) {
+        if (piece.slug.startsWith('/')) {
+          moduleName = '@apostrophecms/page';
```
abea, line 289: It seems this should now be `AposDocsManager.vue`.

modules/@apostrophecms/ui/ui/apos/components/AposCellType.vue
```diff
@@ -0,0 +1,29 @@
...
+    class="apos-table__cell-field"
+    :class="`apos-table__cell-field--${header.name}`"
+  >
+    {{ label(item[header.name]) }}
+  </p>
+</template>
+
+<script>
+export default {
+  name: 'AposCellBasic',
```
abea, line 12: Name needs updating

modules/@apostrophecms/ui/ui/apos/components/AposCellType.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <p
+    class="apos-table__cell-field"
+    :class="`apos-table__cell-field--${header.name}`"
+  >
+    {{ label(item[header.name]) }}
```
abea, line 6: we can just say `item.type`, right? This is a cell only for doc types.

abea (changes requested):
- As a contrib I keep getting logged out unwillingly. I'm doing some basic piece editing.
- As a contrib I no longer get the page tree menu item. Is that from this?
- As a contrib I switch a page (home) to edit mode. I make edits. I cannot go back to published mode as it says the page has never been published. But if I refresh I'm back in the published mode. Same thing happens if I switch to draft again.
- As a contrib I make edits to the home page. The admin user opens the home page change submission and takes control. As a contrib, I get the warning modal and can't edit, but the context menu still says I'm in Draft mode.
- As an admin when I reject changes the doc stays in the submitted drafts manager. I no longer have "reject changes" as an option.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
boutell: I don't think I understand what you're suggesting here. I'm determining the name of the module with the API suited to the document.

boutell:
Thanks for the thorough review!

I believe I've addressed all of the above concerns with the possible exception of:

"As a contrib I keep getting logged out unwillingly. I'm doing some basic piece editing." I have not seen this. However this could be a reference to something I have seen where I have to refresh and try the "edit" button again. And it looks like I fixed that in these latest commits when I addressed the bug where publicApiCheck was too strict. If not please let me know how to reproduce what you are seeing.

"As a contrib I make edits to the home page. The admin user opens the home page change submission and takes control. As a contrib, I get the warning modal and can't edit, but the context menu still says I'm in Draft mode."

This is not a bug - this is preview mode. Maybe that's not the right place to land after being booted, but it's not new behavior in this sprint, so changing it would be a conversation for a new design ticket.

modules/@apostrophecms/piece-type/ui/apos/components/AposPiecesManager.vue
abea: The component is `AposPiecesManager`, but with this change it is no longer specific to pieces. I don't want that to be misleading or confusing.

abea (approved):

PR #2924, opened 2021-04-19
PRO-1183 "before" support for routes

boutell:
Also middleware may be specified with a `url`, otherwise defaults to global as before.

Notifications are no longer implemented as a wacky middleware function, they now have a clean `getAll` route that uses `before` to skip the performance impact of loading the global doc (which, in many projects, is too large for polling).

This is prelude to a clean implementation of PRO-1181, polling for the count of submitted docs.

Targets the submissions branch since that has not yet landed.


abea (changes requested):
Looks good. We should have tests for this, no? I don't see any for adding middleware.

boutell:
Unit test added.

abea (approved):

PR #2926, opened 2021-04-19
3/pro 1181 count submitted drafts

boutell:
This is marked as a draft to prevent it from landing before PRO-1183, as it would otherwise prematurely land that code too. I suggest waiting on review to avoid confusion with that code.

This one needs CSS love after acceptance although it's in the neighborhood of right.


boutell:
This is no longer a draft.

stuartromanek (commented):
@boutell are there tests to be written/updated for the new middleware capabilities?

PR #2927, opened 2021-04-19
PRO-1177 Removes first and last name, has username follow display name

abea:


modules/@apostrophecms/submitted-draft/index.js
```diff
@@ -23,12 +23,12 @@ module.exports = {
...
-        label: 'Submitted',
-        component: 'AposCellDate'
+        label: 'Submitted'
+        // component: 'AposCellDate'
       },
       'submitted.by': {
-        label: 'Proposed By',
-        component: 'AposCellDate'
+        label: 'Proposed By'
+        // component: 'AposCellDate'
```
boutell, line 31: This one should never have used AposCellDate. The other one should, not sure why that's not an option at the moment.

boutell (approved):

modules/@apostrophecms/submitted-draft/index.js
abea: Oh, I didn't mean to leave these commented out.

abea: There's a separate ticket to work on those. It's because they aren't digging into the data property with the dot path.

boutell (approved):

PR #2928, opened 2021-04-19
sets custom page types to default their action to @apostrophecms/page's

stuartromanek:
custom page's `action` route doesn't do anything as far as i know .. this would be helpful for generic functions that work with both pages and pieces

### DIFF
--- a/modules/@apostrophecms/page-type/index.js
+++ b/modules/@apostrophecms/page-type/index.js
@@ -392,6 +392,9 @@ module.exports = {
   },
   extendMethods(self) {
     return {
+      enableAction() {
+        self.action = self.apos.modules['@apostrophecms/page'].action;
+      },
       copyForPublication(_super, req, from, to) {
         _super(req, from, to);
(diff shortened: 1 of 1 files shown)


boutell (approved):

boutell (changes requested):
... I was wringing my hands a little bit because a developer might decide to write some apiRoutes etc. that make sense for a particular page type. I can see how that could arise. But if they really want to do that they can still construct that URL themselves in browserland, or set another property besides `action` in `getBrowserData`.

And with your PR we could discard a lot of frontend logic currently in use to determine the right action URL.

However that does bring me to a better reason to request changes: how about replacing that logic in this PR?

stuartromanek:
@boutell would a better approach be that page-type inherits `@apostrophecms-page`'s apiMethods? That way the constructed action would work as expected and the developer could make per page-type's enhancements and all that just works as expected?

Or is that messier

boutell:
I think it's definitely messier, those methods use `self` extensively, and so on. Not a good candidate to become "mixins."

Better to point the action to the page module.

PR #2929, opened 2021-04-20
PRO-1185 fragments calling fragments

boutell:
Make sure fragments can:

* render other fragments declared in the same source file, like macros can
* render other fragments imported by their source file, like macros can

It took me forever to figure out these six lines of code on my own time tonight. Just couldn't let this go unfixed. Needed to know if we were going to have to pivot to twigjs or even write something new. The answer is "not today."

Nunjucks makes the proper context available in the run() function that handles the "fragment" tag itself, so I needed to return that context as part of the node so it would be available later in the run() fragment of the "render" tag.

### DIFF
--- a/modules/@apostrophecms/template/lib/custom-tags/render.js
+++ b/modules/@apostrophecms/template/lib/custom-tags/render.js
@@ -5,11 +5,6 @@ module.exports = (self) => {
         // get the tag token
         const token = parser.nextToken();
         const args = new nodes.NodeList(token.lineno, token.colno);
-        // TODO this only allows a simple name, we need
-        // to permit at least a dotted expression to
-        // accommodate imports. We can't parse it as an
-        // expression because that would invoke it
-        // as a function with the args
         const invocation = parser.parsePrimary();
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/template/lib/custom-tags/render.js
```diff
@@ -5,11 +5,6 @@ module.exports = (self) => {
         // get the tag token
         const token = parser.nextToken();
         const args = new nodes.NodeList(token.lineno, token.colno);
-        // TODO this only allows a simple name, we need
```
boutell, line 8: This comment was out of date, you can successfully render an imported fragment.

boutell:
For verification see the home page template in this branch of a3-boilerplate:

fragments-calling-fragments

Without the PR it'll throw an error. With the PR you'll get two areas at the top of the page.

abea (approved):

abea (changes requested):
Looks good. Please update the changelog.

abea (approved):

PR #2933, opened 2021-04-20
Pro 1176 following no longer triggers required error

boutell:


modules/@apostrophecms/schema/index.js
```diff
@@ -130,7 +130,8 @@ module.exports = {
             return self.sortedDistinct(field.name, query);
           }
         });
-      }
+      },
+      def: ''
```
boutell, line 134: With this change the required mechanism still triggers when it should (empty string is still empty as we tab off a field or attempt to save), and we don't get a false positive due to a transition from null to empty string.

abea (approved):

PR #2935, opened 2021-04-20
PRO-1184 contributors can see their submissions

boutell:
Contributors can see their submissions. Piggybacks on the PR for showing the submission count to editors/admins.

Contributors can only see their own submissions. A button to display the manage view appears in the tray area, like the palette icon. I used a custom icon component. The button appears only when you actually have submissions. No count is displayed because that sounds like an unnecessary distraction, it's just a way to get back to your own submissions.

The ability to dismiss a submission is a separate ticket already and when implemented it will be supported for contributors too as long as it was their own.


modules/@apostrophecms/submitted-draft/index.js
```diff
@@ -105,24 +105,49 @@ module.exports = {
...
+          self.pluralLabel,
+          {
+            action: 'edit',
+            type: self.name
+          },
+          {
+            icon: 'AposSubmittedDraftAdminBarIcon',
+            contextUtility: true,
+            tooltip: 'Submitted Drafts',
+            when(req) {
```
boutell, line 126: This admin bar item is to be displayed only when the other one can't be, requiring custom logic. So I added a "when" option to admin bar items, which accepts a function that makes that determination on the fly based on `req`.

modules/@apostrophecms/submitted-draft/index.js
```diff
@@ -105,24 +105,49 @@ module.exports = {
...
         );
+        self.apos.adminBar.add(
+          `${self.__meta.name}:manager`,
+          self.pluralLabel,
+          {
+            action: 'edit',
+            type: self.name
+          },
+          {
+            icon: 'AposSubmittedDraftAdminBarIcon',
```
boutell, line 123: I wrote a component to wrap the icon in a v-if test, using the same REST API call as the fancier one for the counter.

modules/@apostrophecms/submitted-draft/index.js
```diff
@@ -105,24 +105,49 @@ module.exports = {
...
       find(_super, req, criteria, options) {
-        return _super(req, criteria, options).type(null).and({
+        const query = _super(req, criteria, options).type(null).and({
           submitted: {
             $exists: 1
           }
         });
+        if (!self.apos.permission.can(req, 'publish', self.name)) {
+          query.and({
+            'submitted.byId': req.user && req.user._id
```
boutell, line 147: This will only match for new submissions made after this PR, so if you have test contributors with test submissions now, you'll need to submit more.

modules/@apostrophecms/submitted-draft/ui/apos/components/AposSubmittedDraftAdminBarIcon.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <arrow-expand-right-icon
```
stuartromanek, line 2: This should be an `<AposIndicator />`, use the tray-full icon https://materialdesignicons.com/icon/tray-full

stuartromanek (changes requested):

stuartromanek:
@boutell after a quick chat w @localghost443 we should use the icon/right-side trigger UI for all users, not just non-admin. This PR feels like the place to make that change, no?

boutell:
@stuartromanek I switched to just the one icon component for both editors and contributors as we discussed, using AposIndicator, however with a number present (editor/admin) the indicator looks funny, and without a number present it's actually not even visible although I can tell it's there.

Could you help out with the remaining frontend on this one?

stuartromanek:
@boutell i will after merge, things are already out of hand and being broken pull to pull

stuartromanek (approved):

PR #2936, opened 2021-04-21
dismiss submissions

boutell:
* "Dismiss Submission" on the "more" menu for both pieces and pages
* Contributors can do it too, as long as it was their own submission
* Also refactored a little to reduce code duplication in the submit/dismiss routes
* /dismiss route is now /dismiss-submission to match the language elsewhere


boutell:
I added a separate ticket to track adding this choice in the manage view as well.

stuartromanek (changes requested):
very cool, add the `:can-dismiss-submission` prop to `modules/@apostrophecms/ui/ui/apos/components/AposCellContextMenu.vue` as well

boutell:
I mentioned there's a separate ticket for that, but yeah let's do it here and be done will get to that today

stuartromanek (approved):

PR #2940, opened 2021-04-21
PRO-1170 contributors cannot touch files and images

boutell:
But you still have to select things for relationships. So there is a basic read only experience in the manage view when used as a chooser. Did not attempt to open the editors in read only mode, I just don't let you launch editors for now when using the manage view as a chooser without permission to edit. That could be a follow-on at some point.

One unrelated change: "this.pieces" is now "this.items" in the docs manager because that is what some of the mixins were expecting and I ran into showstoppers that wouldn't let me continue.

### DIFF
--- a/modules/@apostrophecms/page/index.js
+++ b/modules/@apostrophecms/page/index.js
@@ -77,8 +77,7 @@ module.exports = {
       // array of pages.
 
       async getAll(req) {
-        // Edit access to draft is sufficient to fetch either
-        self.publicApiCheck(req, 'draft');
+        self.publicApiCheck(req);
         const all = self.apos.launder.boolean(req.query.all);
         const archived = self.apos.launder.booleanOrNull(req.query.archived);
         const flat = self.apos.launder.boolean(req.query.flat);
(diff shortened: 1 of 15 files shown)


abea (changes requested):
I'll try to start reviewing the new code, but it looks like this branch and the branch it is targeting are not updated to the same point. I see this has the `firstname/lastname` removal commit since that's not in the downstream branch.

modules/@apostrophecms/page/index.js
```diff
@@ -77,8 +77,7 @@ module.exports = {
       // array of pages.
 
       async getAll(req) {
-        // Edit access to draft is sufficient to fetch either
-        self.publicApiCheck(req, 'draft');
+        self.publicApiCheck(req);
```
abea, line 80: Didn't you just recently do the opposite?

boutell: No, before it was checking whether you could edit, now it's enough just to be able to view drafts in general on the site, which is a better test that doesn't run afoul of six reasons why it's not permissive enough.

boutell: Because of this change I no longer need to pass a mode override, view-draft is itself a permission.

abea (approved):

PR #2945, opened 2021-04-22
3/page title

abea:
Updates the base layout template's title tag to include the piece or page title by default. This can be overridden in project templates.


boutell (approved):
This is a great improvement. I wonder if the template module should accept an option for a fallback title as well (not a blocker, just a thought).

abea:
That would be through a variable, or a hard-coded value? I'd be interested in understanding some real world cases where that's used. 

It would break some separation of concerns that makes sense. Template stuff going in templates.

boutell:
I guess a default title is inherently bad SEO practice, so never mind.

Issue #2930, opened 2021-04-20
PRO-69 Fragments should support the call / caller syntax like macros do


boutell:
Macros support this:

```
{% macro render(options) %}
<section
data-` 'accordion' | prefix `='` options | json `'
class="` 'accordion' | prefix `"
>
{% if caller %}
{{ caller() }}
{% endif %}
</section>
{% endmacro %}
```

You call that like:

```
{% call render({ ... }) %}
Body that gets inserted by `caller()` goes here
{% endcall %}
```

It is highly likely that areas could be used with a feature like that, so fragments should support it.

We can do it with rendercall and rendercaller keywords, for parallelism with the "render" keyword already used to render fragments without a body.

 So to get the equivalent of `{% call macroName ... %}` you would use `{% rendercall fragmentName... %}`.

Similarly `rendercaller` (since we probably can't just recycle caller without breaking macros).

boutell:
In the meantime the workaround for generating custom bodies is passing fragments to other fragments as parameters, which does work.

myovchev:
I can give it a try. I can use `3/pro-1185-fragments-calling-fragments` as source branch. Does it sound OK @boutell?

boutell:
Yes that makes sense. Thanks for giving it a go!

myovchev:
Actually it's quite easy, I have a working prototype on my test case repo already. However there are few ways to go, so I'll create a draft PR against the `3/pro-1185-fragments-calling-fragments` and we can discuss it.

myovchev:
@boutell I forgot to add the `closes` magic in the PR, please close.

PR #2946, opened 2021-04-22
Fragment `rendercall` and keyword arguments

myovchev:
## Feature `rendercall`
Allow similar to macro `call`/`caller` feature:
```njk
{% rendercall fragment.test() %}
  BODY
{% endrendercall %}
```
where `BODY` can be any valid sync and async nunjucks source.
The fragment definition is able to access the `BODY` via `rendercall` local function:
```njk
{% fragment test() %}
  ...
  {{ rendercaller() }}
  ...
{% endfragment %}
```
ref #2930

## Feature [keyword arguments](https://mozilla.github.io/nunjucks/templating.html#keyword-arguments)
Allows a mix of positional and keyword arguments (see the docs from the link above)
```njk
{% fragment test(pos1, pos2, kw1 = 'default value', kw2 = 'default value') %}
  ...
{% endfragment %}
```
This would be called in various ways, just some examples (some weird too!):
```njk
{% render test('pos1', 'pos2') %}
{% render test('pos1', 'pos2', kw2 = 'value') %}
{% render test(kw1 = 'value', 'pos1', 'pos2') %}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@
 
 * There was a bug that allowed parked properties, such as the slug of the home page, to be edited. Note that if you don't want a property of a parked page to be locked down forever you can use the `_defaults` feature of parked pages.
 * Fragments can now call other fragments, both those declared in the same file and those imported, just like macros calling other macros. Thanks to Miro Yovchev for reporting the issue.
+* Fragments support `rendercall`/`rendercaller` syntax similar to the `call`/`caller` macro syntax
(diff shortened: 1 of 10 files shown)


boutell (approved):
Heck yes! Thank you very much Miro.

myovchev:
My pleasure.
I'm gonna need to provide a patch tomorrow  - I think I've already found a bug :)
Need to look closer and if this is the case the next PR should come tomorrow.

PR #2947, opened 2021-04-22
changelog entry for fragments

boutell:
Miro contributed great work here, just mopping up comments and the changelog entry.


abea (dismissed):

boutell:
Apparently needs re-review due to merge, even if it looks the same.

abea (approved):

PR #2949, opened 2021-04-23
Dependencies not complaining, all tests passing

boutell:


abea (commented):
It doesn't seem like our tests are failing now. Where's this coming from?

abea:
From Miro's note in Discord it sounds like this dep might not be needed at all. Added here, which seems unnecessary: https://github.com/apostrophecms/apostrophe/pull/2948/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R131

boutell:
He said it would fail without it, but one of our other dependencies has a peer dep on the 4.x version of it, which led to trouble.

I guess I can try removing it entirely and running through the npm install / npm test task from the top, yep.

abea (approved):
I see that peer dep warning. Looks good.

boutell:
You're right, it's completely superfluous. This PR is now to remove the dependency.

abea (changes requested):
😬 I think you were right. Let's keep 4 in.

abea (approved):

abea (approved):

PR #2951, opened 2021-04-23
Fixed pager links when logged in

boutell:
Also cleaned up superfluous append=1 parameter.

I know replacing the query parameter by hand is a lot, but I'm doing it rather than fully parsing the URL because this is needed on every page load.

### DIFF
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -644,7 +644,7 @@ module.exports = {
           // Make the query available to templates for easy access to
           // filter settings etc.
           query: req.query,
-          url: req.url
+          url: unrefreshed(req.url)
         };
 
         _.extend(args, data);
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/template/index.js
```diff
@@ -673,6 +673,22 @@ module.exports = {
           req.statusCode = 500;
           return self.render(req, 'templateError');
         }
+
+        function unrefreshed(url) {
+          // Including apos-refresh=1 in data.url leads to busted pages in
+          // navigation links, so strip that out. However this is invoked on
+          // every page load so do it as quickly as we can to avoid the
+          // overhead of a full parse and rebuild
+          if (!url.includes('apos-refresh=1')) {
```
boutell, line 682: I tested this solution with this:

```
console.log(unrefreshed('/wiggle/wiggly?apos-refresh=1'));
console.log(unrefreshed('/wiggle/wiggly?apos-refresh=1&coolness=true'));
console.log(unrefreshed('/wiggle/wiggly?coolness=true&apos-refresh=1'));
console.log(unrefreshed('/wiggle/wiggly?coolness=true&apos-refresh=1&badness=false'));
        function unrefreshed(url) {
          // Including apos-refresh=1 in data.url leads to busted pages in
          // navigation links, so strip that out. However this is invoked on
          // every page load so do it as quickly as we can to avoid the
          // overhead of a full parse and rebuild
          if (url.endsWith('?apos-refresh=1')) {
            return url.replace('?apos-refresh=1', '');
          } else if (url.includes('?apos-refresh=1')) {
            return url.replace('?apos-refresh=1&', '?');
          } else {
            return url.replace('&apos-refresh=1', '');
          }
        }
```

abea (approved):

PR #2956, opened 2021-04-24
PRO-1031 unpublished schema changes are flagged

stuartromanek:
https://linear.app/apostrophecms/issue/PRO-1031/approvals-as-an-admineditor-i-can-see-fields-that-have-changed-in-the

The design of this changed a bit during development. 
The original highlighted the entire field in a warning yellow, but not all fields have that kind of UI to style. I changed this to be a small Changed label appended after the field title, with a fuller tooltip explaining the state. The yellow input also felt a lil scary in practice.

original design:
image

implemented design
image


modules/@apostrophecms/util/ui/public/2-http.js
```diff
@@ -127,7 +129,7 @@
         } else {
           query = {};
         }
-        query['apos-mode'] = 'draft';
+        query['apos-mode'] = options.draft ? 'draft' : 'published';
```
boutell, line 132: You need an `undefined` check on `options.draft` here, otherwise something is passed all the time, which is not always appropriate.

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
```diff
@@ -1,13 +1,24 @@
...
-  });
+export function detectDocChange(schema, v1, v2, options = {}) {
+  if (options.differences) {
+    const differences = [];
+    schema.forEach(field => {
+      if (detectFieldChange(field, v1[field.name], v2[field.name])) {
+        differences.push(field.name);
+      }
+    });
+    return differences;
```
boutell, line 16: 💯

boutell (changes requested):
This is very 💯, just one thing: don't pass `apos-draft` for requests that have no value for `draft` at all.

modules/@apostrophecms/util/ui/public/2-http.js
stuartromanek: @boutell isn't [this check](https://github.com/apostrophecms/apostrophe/pull/2956/files#diff-de132b0a468c668b0f24c2aaa7cba4687c855728c2897417b445377ef9f9bcd4R117) going to safely skip this block in the case of `options.draft` being `undefined` ?

https://github.com/apostrophecms/apostrophe/pull/2956/files#diff-de132b0a468c668b0f24c2aaa7cba4687c855728c2897417b445377ef9f9bcd4R117

boutell: There are two forks of this if statement, the one for options.qs being truthy is correct, but the one for options.qs not being truthy contains the bad assumption that options.draft should always do something (line 132).

boutell (changes requested):

modules/@apostrophecms/util/ui/public/2-http.js
stuartromanek: @boutell is this check enough https://github.com/apostrophecms/apostrophe/pull/2956/files#diff-de132b0a468c668b0f24c2aaa7cba4687c855728c2897417b445377ef9f9bcd4R132-R134 ?

modules/@apostrophecms/util/ui/public/2-http.js
```diff
@@ -127,7 +129,9 @@
         } else {
           query = {};
         }
-        query['apos-mode'] = 'draft';
+        if (typeof options.draft !== 'undefined') {
```
boutell, line 132: Use `options.draft != null` instead (use of only one = is intentional). That's consistent with our "null means don't care" policy elsewhere and it also matches undefined and saves code. Sorry

boutell (changes requested):

modules/@apostrophecms/util/ui/public/2-http.js
stuartromanek: this entire block it wrapped in a `if (options.draft != null) {` ... I must be really missing something here

modules/@apostrophecms/oembed-field/ui/apos/components/AposInputOembed.vue
```diff
@@ -19,15 +19,16 @@
...
       </div>
     </template>
-    <template #secondary>
-      <div
-        v-if="!error && oembedResult.html" v-html="oembedResult.html"
-        class="apos-input__embed" :class="{ 'is-dynamic': !!dynamicRatio }"
-        :style="{ paddingTop: dynamicRatio && `${(dynamicRatio * 100)}%` }"
-      />
-    </template>
+    <!-- <template #secondary>
```
abea, line 29: comments.

abea (changes requested):

boutell (approved):
I missed the outer if statement all good

abea (approved):

modules/@apostrophecms/ui/ui/apos/scss/global/_theme.scss
```diff
@@ -65,6 +65,7 @@
   --a-type-tiny: 9px;
   --a-type-small: 10px;
   --a-type-base: 12px;
+  --a-type-medium: 13px;
```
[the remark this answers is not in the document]
stuartromanek: renamed to `type-label`

PR #2958, opened 2021-04-26
PRO-1204 Archive/Published switcher for Page Manager

stuartromanek:
https://linear.app/apostrophecms/issue/PRO-1204/archive-filter

image

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
@@ -81,16 +81,16 @@ export default {
     draftMenu() {
       return [
         {
-          label: (this.draftMode === 'draft') ? '✓ Draft' : 'Draft',
+          label: 'Draft',
           name: 'draft',
           action: 'draft',
-          modifiers: (this.draftMode === 'draft') ? [ 'disabled' ] : null
+          modifiers: (this.draftMode === 'draft') ? [ 'disabled', 'selected' ] : null
         },
         {
-          label: (this.draftMode === 'published') ? '✓ Published' : 'Published',
+          label: 'Published',
           name: 'published',
           action: 'published',
-          modifiers: (this.draftMode === 'published') ? [ 'disabled' ] : null
+          modifiers: (this.draftMode === 'published') ? [ 'disabled', 'selected' ] : null
         }
       ];
     }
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
```diff
@@ -81,16 +81,16 @@ export default {
...
-          modifiers: (this.draftMode === 'draft') ? [ 'disabled' ] : null
+          modifiers: (this.draftMode === 'draft') ? [ 'disabled', 'selected' ] : null
         },
         {
-          label: (this.draftMode === 'published') ? '✓ Published' : 'Published',
+          label: 'Published',
           name: 'published',
           action: 'published',
-          modifiers: (this.draftMode === 'published') ? [ 'disabled' ] : null
+          modifiers: (this.draftMode === 'published') ? [ 'disabled', 'selected' ] : null
```
stuartromanek, line 93: unrelated to main PR, side effect of new context menu convenience

boutell (changes requested):
* A newly created, unpublished draft page is not archived or published. So "published" is not the opposite of "archived." We need another word encompassing all "not archived" pages. "Live" possibly.
* Not sure why you get everything ("any") in the archived case, rather than getting just things that are archived? But maybe there's an issue getting the tree when the home page is not in the result set, in which case that's cool, I don't think it's a serious issue we need to go rewrite code in the backend for.

stuartromanek:
> * A newly created, unpublished draft page is not archived or published. So "published" is not the opposite of "archived." We need another word encompassing all "not archived" pages. "Live" possibly.
> * Not sure why you get everything ("any") in the archived case, rather than getting just things that are archived? But maybe there's an issue getting the tree when the home page is not in the result set, in which case that's cool, I don't think it's a serious issue we need to go rewrite code in the backend for.

As for getting just the archived node, how would I do that? Would i need to make a request first to find out the archived page's id, then make a request for the node?

localghost443:
**"A newly created, unpublished draft page is not archived or published. So "published" is not the opposite of "archived." We need another word encompassing all "not archived" pages. "Live" possibly."**

Language Proposal:

1. Archived -> **Archived**
2. Unpublished content that has been proposed by a contributor -> **Draft**, **Pending**
5. Published page with active changes -> **Pending Updates**
3. Published with no active draft -> **Published**
4. Newly created, unpublished draft -> **Draft**

localghost443:
https://linear.app/apostrophecms/issue/PRO-1147/language-reckoning

stuartromanek:
@boutell @localghost443 language updated Published -> Live

boutell (approved):

PR #2961, opened 2021-04-26
PRO-1186 unpublished label should only appear for relevant content types

boutell:


abea (changes requested):
This is failing some tests.

boutell:
Tests passing, also switched to computed

abea (approved):

PR #2963, opened 2021-04-26
Override the default redirection status code when req.statusCode is set

jogarijo:
Currently, when redirecting a page using `req.redirect = '/somewhere'`, it is not possible to set the status code to be used for the redirection. In fact, the redirection will always use the status code 302 Found.

However, some modules like `apostrophe-personas` makes the assumption that setting `req.statusCode` will override the status code ([check index.js:302-305 on the latest commit to this day](https://github.com/apostrophecms/apostrophe-personas/blob/31c84c86bd26f42c901b4c91c2d1fd2464da8622/index.js#L302-L305)).

I found this assumption very reasonable in the idea of using ["req as a notepad of things we'd like to happen in res"](https://github.com/apostrophecms/apostrophe/blob/3360af419d73a80a7828472078da4d9b9242edfb/lib/modules/apostrophe-pages/lib/api.js#L1125-L1126), so this PR is my take on it. :slightly_smiling_face: 

Here is the updated behavior when `req.redirect` is set:
- If `req.statusCode` is set, use it for the redirection;
- If it isn't, try to use the option `statusCode` from `apostrophe-soft-redirects` ([similar to what `apostrophe-personas` does](https://github.com/apostrophecms/apostrophe-personas/blob/31c84c86bd26f42c901b4c91c2d1fd2464da8622/index.js#L304));
- Otherwise fallback to using `302` like before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Changes
+* It's now possible to set the status code when redirecting using `req.redirect` by setting `req.statusCode`. When it it is not set, `apostrophe-pages` will attempt to use the `statusCode` option of `apostrophe-soft-redirects` if it exists. It will fallback to 302 (Found) otherwise, which was the existing behavior.
+
 ## 2.117.1 (2021-04-07)
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 ### Changes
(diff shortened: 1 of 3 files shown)


jogarijo:
When I started I was thinking to use `req.statusCode` for the redirection only when it's in the range `3xx`. But as much as it makes sense to me, some users might have really good reasons to have redirections with the status code 419. The restriction would seem arbitrary to them, so I ended up not keeping it.

boutell:
Thank you!

However it looks like the soft redirects module actually does use `req.res.redirec()` directly, so it is not involved in this situation, and you should be able to remove the extra logic for it.

boutell (changes requested):
Thank you!

However it looks like the soft redirects module actually does use `req.res.redirect()` directly, so it is not involved in this situation, and you should be able to remove the extra logic for it.

jogarijo:
Good point, I made the change and removed the related test.

boutell (approved):
Looks good, thank you

PR #2966, opened 2021-04-26
PRO-1206 startup build

abea:
On startup:
- [x] If there's no UI build, **run the UI build**
- [x] If the UI build timestamp is older than the package.json timestamp, **run the UI build**
- [x] If the core-dev flag is on, **run the UI build**
- [x] **Run the public assets build**
- [x] **Merge the UI build and the public assets build**


abea:
We discussed using chokidar to watch project level public assets to trigger a rebuild of those. I looked into it and held off for a couple of reasons.

1. chokidar handlers aren't run async, so there's not a straightforward way to use our existing task. Pieces of the build task can run syncronously so we could break it out. I didn't do that because...
2. Responsibility gets confusing and we would need a clear vision for that. In the demo we are using nodemon to watch stuff, which I think we're going to continue recommending. This chokidar task would be something that nodemon would or could also be doing. It's not clear that we gain a lot with chokidar if we're already recommending a file watcher.

modules/@apostrophecms/asset/index.js
```diff
@@ -85,9 +131,11 @@ module.exports = {
             }
           }
 
-          function buildPublicCssBundle() {
+          async function buildPublicCssBundle() {
             const publicImports = getImports('public', '*.css', { });
-            fs.writeFileSync(`${bundleDir}/public-bundle.css`,
+            await fs.ensureFile(`${bundleDir}/${PUBLIC_BUNDLE_CSS}`);
```
abea, line 136: This was for the test suite. For some reason when it ran without this it choked, even though `writeFileSync` should create the file. The same process is done for the JS bundle, but it isn't coking on that. Not sure why that is.

@boutell Any ideas?

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
 * Admin bar menu items can now specify a custom Vue component to be used in place of `AposButton`.
 * Sets `username` fields to follow the user `title` field to remove an extra step in user creation.
 * Adds default data to the `outerLayoutBase.html` `<title>` tag: `data.piece.title or data.page.title`.
+* Moves the core UI build task into the start up process. The UI build runs when:
+  1. The build folder does not yet exist.
+  2. The package.json file is newer than the existing UI build.
```
boutell, line 33: package-lock.json would make more sense.

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
...
+  3. You explicitly tell it to by setting the environment variable `CORE_ENV=true`
```
boutell, line 34: `CORE_DEV`

modules/@apostrophecms/asset/index.js
```diff
@@ -22,12 +23,21 @@ module.exports = {
       ...globalIcons
     };
   },
+  handlers (self) {
+    return {
+      'apostrophe:modulesReady': {
+        async runUiBuildTask() {
+          await self.apos.task.invoke('@apostrophecms/asset:build');
```
boutell, line 30: This would run it all the time...?

modules/@apostrophecms/asset/index.js
```diff
@@ -22,12 +23,21 @@ module.exports = {
...
+      }
+    };
+  },
   tasks(self) {
     return {
       build: {
         usage: 'Build Apostrophe frontend javascript master import files',
         afterModuleInit: true,
-        async task(argv) {
+        async task() {
```
boutell, line 40: If someone invokes the task as a task, it should do all the things by default. I think people expect that.

Whereas if the handler invokes it, it should pass an option specifying the "only if you think it's necessary" behavior.

modules/@apostrophecms/asset/index.js
boutell: OK I see, but it should pass an option (which will arrive in argv) signifying that the "only if you think it's necessary" behavior is wanted.

modules/@apostrophecms/asset/index.js
boutell: Also modulesReady should not invoke the task if `self.apos.isTask()` is true. We don't want asset builds happening during, say, migrations. And we don't want a conditional asset build happening before a full asset build in the same invocation either if the command line task requested is the actual asset build task.

modules/@apostrophecms/asset/index.js
```diff
@@ -37,12 +47,48 @@ module.exports = {
...
+          if (!fs.exists(bundleDir)) {
+            rebuildAposUi = true;
+            await fs.mkdirp(bundleDir);
+          } else {
+            await fs.remove(`${bundleDir}/${APOS_MERGED_BUNDLE}`);
+            await fs.remove(`${bundleDir}/${PUBLIC_BUNDLE_CSS}`);
+            await fs.remove(`${bundleDir}/${PUBLIC_BUNDLE_JS}`);
+          }
+
+          if (process.env.CORE_DEV !== 'true') {
```
boutell, line 68: Environment variable convention is to allow `CORE_DEV=1`, or really anything, but `1` should definitely work.

modules/@apostrophecms/asset/index.js
```diff
@@ -37,12 +47,48 @@ module.exports = {
...
+
+          let rebuildAposUi = false;
+          const APOS_MERGED_BUNDLE = 'apos-bundle.js';
+          const APOS_ONLY_BUNDLE = 'apos-only-bundle.js';
+          const APOS_ONLY_TS = '.apos-only-timestamp.txt';
+          const PUBLIC_BUNDLE_CSS = 'public-bundle.css';
+          const PUBLIC_BUNDLE_JS = 'public-bundle.js';
+          let checkTimestamp = false;
+
+          if (!fs.exists(bundleDir)) {
```
boutell, line 59: You did not use `existsSync` or `await`. `exists` returns a promise, which is always truthy, so the mkdirp call never happens and the flag is never set.

modules/@apostrophecms/asset/index.js
boutell: See my note earlier about the lack of await on a mkdirp call. May not have happened yet.

modules/@apostrophecms/asset/index.js
```diff
@@ -130,28 +178,28 @@ module.exports = {
...
-    ${componentImports.importCode}
-    ${tiptapExtensionImports.importCode}
-    ${appImports.importCode}
-    ${iconImports.registerCode}
-    ${componentImports.registerCode}
-    ${tiptapExtensionImports.registerCode}
-    setTimeout(() => {
-    ${appImports.invokeCode}
-    }, 0);
+            fs.writeFileSync(importFile, stripIndent`
```
boutell, line 181: Nice

boutell (changes requested):
We should log something if we start one of these auto builds. The extra time could look like a hung state to a frustrated developer if it's not clear to them what is happening. Bear in mind it could be a lot more than 13 seconds if they are not on a baller Mac.

boutell:
Regarding leaving autodetection that it's time to do this up to nodemon, yes... we could configure nodemon to watch package-lock, if it doesn't already.

CHANGELOG.md
abea: I thought about that, but if they use yarn it there wouldn't be one. We could look for yarn.lock as well, but then it feels like we're chasing technologies.

Though you're right that a lock file update could indicate a need to rebuild the UI....

boutell: Interesting point about chasing technologies. You could investigate whether all updates bump the semver ranges in package.json with a new minimum version, but I suspect they do not. So we might have to chase these two. I don't think a third major npm client is likely.

modules/@apostrophecms/asset/index.js
abea: So nice.

abea:
> Regarding leaving autodetection that it's time to do this up to nodemon, yes... we could configure nodemon to watch package-lock, if it doesn't already.

I'm only suggesting having nodemon check if the `ui/public/*` files are updated. The lock file will be checked on normal start up.

modules/@apostrophecms/asset/index.js
```diff
@@ -27,7 +27,13 @@ module.exports = {
     return {
       'apostrophe:modulesReady': {
         async runUiBuildTask() {
-          await self.apos.task.invoke('@apostrophecms/asset:build');
+          if (!self.apos.isTask()) {
+            // If starting up normally, run the build task, checking if we
+            // really need to update the core UI build.
+            await self.apos.task.invoke('@apostrophecms/asset:build', {
+              checkUiBuild: true
```
boutell, line 34: This should be `check-ui-build`, as it is a command line option too.

boutell (changes requested):

boutell:
Sorry for the holdup, no other concerns

boutell (approved):
Approving. `check-ui-build` makes me a little nervous because it doesn't leave the namespace terribly open to having command line options to check the other aspects of the build or not later, but (1) we aren't likely to need that, and (2) `check-ui-apos` is correct but very opaque-sounding.

boutell (changes requested):
Sorry, I missed something:

This should never happen when `NODE_ENV` is `production`, which is always the case on staging and production servers. In staging/production environments it's not good to do this kind of thing spontaneously as it's often part of a build process where the artifact must be copied somewhere to be shared, etc. 🙏

boutell (dismissed):

boutell (approved):

PR #2967, opened 2021-04-27
PRO-1225 routes specified as an array with middleware and a route work again (such as the upload route for attachments)

boutell:


abea (commented):
can we get a test for this since it burned us recently?

abea (dismissed):

PR #2972, opened 2021-04-28
PRO-1155 Show published titles, etc. in page and piece managers, when possible (please update docs with query parameter naming convention change)

boutell:
* The page and piece manager views now display the title, etc. of the published version of a document, unless that document only exists in draft form. However a label is also provided indicating if a newer draft is in progress.
* The new `._id(_idOrArrayOfIds)` query builder replaces `explicitOrder` and accepts an array of document `_id`s or a single one. `_id` can be used as a multivalued query parameter. Documents are returned in the order you specify (unlike an ordinary $in query), and just like with single-document REST GET requests, the locale of the `_id`s is overridden by the `apos-mode` query parameter if present. This replaces the old `explicitOrder` query builder, which took a second argument (not really a supported pattern for query builders) and had a less intuitive name when your goal is to filter by _id.
* The `.published(true)` query builder adds a `_publishedDoc` property to each returned draft document that has a published equivalent. `published=1` can be used as a query parameter. The "Doc" suffix distinguishes `_publishedDoc` from `_published`, which is the published *permission* for the document for the current user, a preexisting convention.
* `inferIdLocaleAndMode` was factored up a level to `doc-type` to accommodate calling it from query builders.
* Cells use a `displayValue` mixin to hide the complexity of fetching a property from publishedDoc if available, otherwise the draft doc.


modules/@apostrophecms/doc-type/index.js
```diff
@@ -1337,56 +1351,62 @@ module.exports = {
...
+              const parts = values[0].split(':');
+              if (parts.length > 1) {
+                values = values.map(value => self.inferIdLocaleAndMode(query.req, `${value.split(':')[0]}:${parts[1]}:${parts[2]}`));
+              }
+            }
+            criteria._id = { $in: values };
             query.and(criteria);
-            query.set('explicitOrderSkip', query.get('skip'));
-            query.set('explicitOrderLimit', query.get('limit'));
+            query.set('_idSkip', query.get('skip'));
```
abea, line 1392: If we're specifying IDs, why are we inviting skipping?

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@
 * Admin bar menu items can now specify a custom Vue component to be used in place of `AposButton`.
 * Sets `username` fields to follow the user `title` field to remove an extra step in user creation.
 * Adds default data to the `outerLayoutBase.html` `<title>` tag: `data.piece.title or data.page.title`.
+* The new `._id(_idOrArrayOfIds)` query builder replaces `explicitOrder` and accepts an array of document `_id`s or a single one. `_id` can be used as a multivalued query parameter. Documents are returned in the order you specify, and just like with single-document REST GET requests, the locale of the `_id`s is overridden by the `apos-mode` query parameter if present.
+* The `.published(true)` query builder adds a `_publishedDoc` property to each returned draft document that has a published equivalent. `published=1` can be used as a query parameter. Note this is not the way to fetch only published documents. For that, use `.locale('en:published')` or similar.
```
abea, line 34: The naming of this is confusing. It's more like `includePublished` or better, `withPublished`. Something like that. `addPublished`? As-is it seems like it'd return the published doc.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -191,6 +191,9 @@ export default {
       return (this.moduleOptions.columns || []).filter(column => {
         return (column.name !== '_url') || this.items.find(item => item._url);
       });
+    },
+    manuallyPublished() {
```
abea, line 195: This is a duplicate.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -258,17 +263,20 @@ export default {
...
-      ));
+      );
 
       this.currentPage = getResponse.currentPage;
       this.totalPages = getResponse.pages;
-      this.items = getResponse.results;
+      // Don't set this.items too soon, it won't detect
+      // the addition of the _publishedDoc sub-properties
+      const items = getResponse.results;
+      this.items = items;
```
abea, line 279: What is this additional step doing?

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  methods: {
+    // Fetch a field from the published version of the item, if available and
+    // defined, otherwise from the draft version of the item
+    displayValue(fieldName) {
+      const publishedDoc = this.item._publishedDoc;
+      return (publishedDoc && (publishedDoc[fieldName] !== undefined) && publishedDoc[fieldName]) || this.item[fieldName];
```
abea, line 7: Why do you need the `(publishedDoc[fieldName] !== undefined)` step?

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  methods: {
+    // Fetch a field from the published version of the item, if available and
+    // defined, otherwise from the draft version of the item
+    displayValue(fieldName) {
```
abea, line 5: I don't think this logic should be in the cell. If the published doc has an empty field that has a value in the draft, we could end up showing the published title with draft values in other columns. We should be feeding all cells the same doc and doing this work on the manager level.

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -373,10 +373,11 @@ export default {
       }
 
       // Original default of just printing the row property value
-      if (row[col.cellValue]) {
-        return row[col.cellValue];
+      const publishedDoc = row._publishedDoc;
+      const value = (publishedDoc && (publishedDoc[col.cellValue] !== undefined) && publishedDoc[col.cellValue]) || row[col.cellValue];
```
abea, line 377: Again, we should be only showing one doc across the row.

abea (changes requested):

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
boutell: We can't do that because some cells make different choices intentionally. Examples being the submitted properties (which are only available on the draft) and the cell context menu (which looks at lots of things). So I put the smarts in one place to be shared by the cells that made typical choices, but I didn't dare oversimplify for them.

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
boutell: I'm falling back to draft properties when the published doc truly doesn't have them at all, which increases the set of use cases that can be covered without one-off logic. (There is a separate bug currently breaking submitted.by and submitted.at, but I'll be addressing that elsewhere after this merges)

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
boutell: Also the submitted at / submitted by cells (those have a different problem right now, unrelated to this ticket but easier to fix when it's done).

boutell: Those are more relevant because the context menu cell is separately called.

modules/@apostrophecms/doc-type/index.js
boutell: Just because you know you are interested in 200 ids in a specific order doesn't mean you don't still want pagination to avoid exhausting memory by carrying out all the relationship nested queries for 200 docs when you're only showing "page two." More generally, if a query builder can avoid breaking other query builders, it should do that.

abea: makes sense.

CHANGELOG.md
abea: This note needs to be updated.

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
abea: How does this solve the issue of values showing from different doc versions?

modules/@apostrophecms/template/index.js
```diff
@@ -629,7 +629,7 @@ module.exports = {
         // Waits for DOMready to give other
         // things maximum opportunity to happen.
 
-        const decorate = (req.query['apos-refresh'] !== '1');
+        const decorate = (req.query['aposRefresh'] !== '1');
```
abea, line 632: ESLint is tripping on this since it can use dot path notation.

abea (changes requested):

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
boutell: We are, not sure what this comment refers to.

CHANGELOG.md
abea: This changelog entry is still outdated. `.published` to `.withPublished`

modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
abea: We're potentially showing data from both the published and the draft documents.

abea:
@boutell Following up on this: https://github.com/apostrophecms/apostrophe/pull/2972#discussion_r622363106

We're opening up to stuff like the row below. That "Gyro - Pub'd" is the published version title (it's a sandwich thing). The custom column at the right with the clown header label has a value from the draft document because the published doc field is empty. So we're not presenting the published version of this document in the row. We're presenting both versions depending on what the data is.

Screen Shot 2021-04-28 at 1 53 44 PM

abea (changes requested):
Also see the most recent comment.

boutell:
Uncle. I will change this over to pass the items a pre-digested item to display, but also pass them the draft and published items as such as otherwise even registering custom components can't help you out of a jam where you need to do something with that information.

abea:
aita

boutell:
You are not. I think there are better arguments for your take than mine and the workaround I have in mind will still allow the better points I was making to be addressed.

modules/@apostrophecms/ui/ui/apos/components/AposCellButton.vue
```diff
@@ -3,7 +3,7 @@
     class="apos-table__cell-field" type="button"
     :class="`apos-table__cell-field--${header.name}`"
   >
-    {{ displayValue(header.name) }}
+    {{ get(header.name) }}
```
abea, line 6: This one still has the item and header props, which can now be removed.

abea (changes requested):
Looks great. There's only the one cell type you missed removing the props. No other notes.

abea (approved):

PR #2973, opened 2021-04-28
PRO-1215 do not allow the primary save button for a piece or page editor modal…

boutell:
… if it is a preexisting document and has not been modified in the modal


boutell:
I have to amend it to allow it if it will result in publication of a draft, because that is a meaningful change even if you don't touch any fields.

stuartromanek (approved):

PR #2978, opened 2021-04-28
PRO-1229 contributors can delete drafts via the kebab menu

boutell:
No changelog because contribs are new in this release.


abea (approved):
Are we calling this the kebab menu?

boutell:
Seems like it

stuartromanek:
_They weren't_

abea:
I guess it is a word people use. https://ux.stackexchange.com/questions/115468/what-the-difference-between-the-2-menu-icons-3-dots-kebab-and-3-lines-hambur

stuartromanek:
it is a modern phrase to describe the icon itself. Context menu is more holistically correct and has been part of OS interface guideline vernacular since the 90s

PR #2979, opened 2021-04-28
submission tray style, tray items can provide their own components

stuartromanek:


stuartromanek:
@boutell if the flag rename looks good i'll update the palette module as well.

boutell (changes requested):
👮 `contextUtility` was a change made at @abea's request in code review, to make it less opinionated about the exact frontend styling in the backend config. I originally had `tray`.

stuartromanek:
hm . I can appreciate the sentiment, but ultimately some flag is going to have to be used to determine where the UI goes, no? If `contextUtility` is a way to say 'not in the main admin bar', then wouldn't we need another flag to split UI between the tray and some future home?

stuartromanek:
@abea ^

abea:
> wouldn't we need another flag to split UI between the tray and some future home?

I don't really understand. The `contextUtility` name is meant to identify a class of UI, not the location. The UI components then translate that class to a location. Is there another class of UI you are thinking of or might anticipate? If there was then it could be placed similarly by identifying its class (e.g., `userUtility`) then placing that class of UI in the components.

If there's some specific piece of UI (like the locale switcher) that is unique, important, and transcends it's general UI class, then it would be placed directly rather than through a flag like this.

stuartromanek:
buut right now that is only the tray.. Say there becomes another position _somewhere_ where controls specific to the context are needed. This need not be a place in the admin bar, imagine some kind of fixed floating button in the bottom right of the screen. These are still context controls and they need to be plucked by some name to get them where they're going. 

I'm playing devil's advocate now, but this came up for me because the flag `contextUtility` just doesn't mean much.

Not going to die on this hill, just a thing that could be abstracted when we have real use cases.

abea:
In that hypothetical we might use another non-contextual name for it. `contextEditor` or something. The point is to separate the kind of thing from what the UI looks like.

boutell (approved):

boutell (approved):

PR #2982, opened 2021-04-29
PRO-1219 PRO-1257 Contributors can "discard draft" properly via the on-page kebab menu

boutell:
Fetch draft properly when refreshing context bar after discarding draft; remove beforeMove handlers made obsolete and problematic by the decision to just autopublish moves (those scenarios no longer apply and the handlers caused the original bug of this ticket); remove redundant logic for discard draft messaging; return doc correctly from discard draft; consistent messaging

### DIFF
--- a/modules/@apostrophecms/page-type/index.js
+++ b/modules/@apostrophecms/page-type/index.js
@@ -154,23 +154,6 @@ module.exports = {
           }
         }
       },
-      afterPublish: {
-        async replayMoveAfterPublish(req, { published, firstTime }) {
-          if (!firstTime) {
-            // We already do this after every move of the draft, so
-            // if there was already a published version it will have
-            // already been moved
-            return;
-          }
-          // Home page does not move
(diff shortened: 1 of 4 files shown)


boutell:
(Nothing here is changelog-worthy because contributors are a new beast.)

modules/@apostrophecms/page-type/index.js
```diff
@@ -154,23 +154,6 @@ module.exports = {
           }
         }
       },
-      afterPublish: {
```
abea, line 157: Is this related to the in-context discard draft change? If not, it's a confusing inclusion.

boutell: I believe I needed this fix to get the other to work.

This code was removed because moving a page now automatically reflects in the other mode, so this "replay it on publish" code is redundant, and also harmful.

abea (approved):

PR #2984, opened 2021-04-29
PRO-1255 fix for edge cases re: when you're allowed to click Submit Changes and when you're not

boutell:
I believe this resolves it for all cases. Keep in mind though that detecting changes in real time — changes made by another user, while this page was open for this user— is out of scope, so a page refresh is needed before checking what a contributor can do after an admin in another window takes some action.


stuartromanek (changes requested):
Still one broken in-context experience I can see:
- As a contributor open a page with no draft. add some junk, see the Submit button activate (draft gets saved)
- open the page manager, find the page your on and Discard Draft it. Close all modals
- The in-context content correctly reflects the drop in draft, but the Submit button is still activated. It should be disabled

stuartromanek (approved):

PR #3000, opened 2021-05-03
Workflow auto commit errors

ValJed:
Linked to this PR: https://github.com/apostrophecms/apostrophe-workflow/pull/335 (this one should be merged after the 335).
* Blocks user from moving if `autoCommitPagesMoves` is true in workflow and user can only edit pages.
* Adds specific error in this case.


lib/modules/apostrophe-pages/lib/api.js
```diff
@@ -651,9 +664,22 @@ module.exports = function(self, options) {
...
+      (data.parent.type !== 'trash') &&
+      (!data.parent._edit)
+    ) {
       return callback('forbidden');
     }
+
+    if (
+      workflow &&
+      workflow.options.autoCommitPageMoves &&
+      !self.apos.permissions.can(req, 'admin-apostrophe-page', moved)
```
boutell, line 678: Are you sure this verifies the user can commit in the locale?

boutell (changes requested):

lib/modules/apostrophe-pages/public/js/reorganize.js
```diff
@@ -504,8 +504,12 @@ apos.define('apostrophe-pages-reorganize', {
       self.hide();
     };
 
-    self.errorOnMove = function() {
-      apos.notify('You may only move pages you are allowed to publish. If you move a page to a new parent, you must be allowed to edit the new parent.', { type: 'error' });
+    self.errorOnMove = function(error) {
+      if (error === 'auto-commit-forbidden') {
+        apos.notify('Since autoCommitPageMoves option has been activated for workflow, you can move pages only if you can commit them too.', { type: 'error' });
```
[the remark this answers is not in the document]
ValJed: Yes, modified.

lib/modules/apostrophe-pages/lib/api.js
ValJed: I tested by setting different permissions between locales, was working.

boutell (approved):

boutell (approved):

lib/modules/apostrophe-pages/lib/api.js
boutell: A user can have only `edit-apostrophe-page` and still be granted commit capability, so I don't think this is right.

Test with `edit-apostrophe-page`, but with commit rights for the locale, and via an admin account specifically grant edit access to the page in question via its page settings (that's what edit-apostrophe-page requires before the user can actually edit a page).

boutell (changes requested):

ValJed:
Modified and tested with a specific edit group.

boutell (approved):

boutell (changes requested):
Sorry, one last thing —

Did you test the scenario where the user's group can edit the draft but cannot commit? They should be blocked from moving the page.

boutell (approved):

PR #3002, opened 2021-05-03
PRO-1213 remove off-topic things from the submissions manager

boutell:


modules/@apostrophecms/ui/ui/apos/components/AposCellContextMenu.vue
```diff
@@ -68,6 +68,22 @@ export default {
         return true;
       }
     },
+    canDiscardDraft() {
+      let initial = true;
```
abea, line 72: "initial" is weird since you reset it, fwiw.

abea (approved):
The code is good. `initial` as a variable name is misleading since it changes.

PR #3006, opened 2021-05-03
Allows more time for soft-redirects test

abea:
This test never passes when run locally for me. I'm giving it some more time to let it pass with others.


abea:
This was included in #3019

PR #3013, opened 2021-05-04
PRO-1277: newly inserted images and files should still be visible to the public

boutell:
There is no migration to go with this because I did not want to add that weight for something that was never publicly released. So if you try to use an EXISTING image in an image widget to test this you will not be able to see it in an incognito window, but if you upload a new image to an image widget you will see it.

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -646,6 +646,11 @@ module.exports = {
           // different from the published, which won't exist yet
           doc.modified = true;
         }
+        if (!doc.visibility) {
+          // If the visibility property has been removed from the schema
+          // (images and files), make sure public queries can still match this type
+          doc.visibility = 'public';
+        }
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/doc/index.js
```diff
@@ -646,6 +646,11 @@ module.exports = {
           // different from the published, which won't exist yet
           doc.modified = true;
         }
+        if (!doc.visibility) {
+          // If the visibility property has been removed from the schema
+          // (images and files), make sure public queries can still match this type
+          doc.visibility = 'public';
```
boutell, line 652: None of the proper values for `visibility` are falsy, so this is a reasonable test and saves performance and lines of code over grabbing the manager and checking the schema etc.

I did not use a promise event handler because each `await` does impact performance slightly and we don't want to introduce a new delay on every insert for this.

abea (approved):

PR #3016, opened 2021-05-04
PRO-1279 new user appears modified

boutell:
WIP


modules/@apostrophecms/user/index.js
```diff
@@ -101,6 +101,7 @@ module.exports = {
               value: 'admin'
             }
           ],
+          def: 'guest',
```
boutell, line 104: Permissions type is not designed with a "none chosen yet" option right now and guest is a reasonable default. Without this setting we have an additional reason for the same perceived bug.

stuartromanek (dismissed):

abea (approved):

PR #3017, opened 2021-05-04
PRO-1282 check for drafts modified prop

stuartromanek:
https://linear.app/apostrophecms/issue/PRO-1282/pieces-not-getting-the-pending-update-label

@boutell as a side note, its confusing we're passing `item`, `draft`, and `published` into all these components, `item` being a moving target. Should we just normalize all the cells to expect `draft` and `published`?

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
@@ -3,7 +3,7 @@
     class="apos-table__cell-field apos-table__cell-field--labels"
     :class="`apos-table__cell-field--${header.name}`"
   >
-    <span v-if="manuallyPublished && item.modified && !item.submitted && item.lastPublishedAt">
+    <span v-if="manuallyPublished && (item.modified || draft.modified) && !item.submitted && item.lastPublishedAt">
       <AposLabel
         label="Pending Updates" class="apos-table__cell-field__label"
         tooltip="There are active changes to this document."
@@ -38,6 +38,10 @@ export default {
   name: 'AposCellLabels',
   props: {
     item: {
+      type: Object,
+      required: true
+    },
+    draft: {
       type: Object,
       required: true
     },
--- a/modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
@@ -3,28 +3,28 @@
     class="apos-table__cell-field apos-table__cell-field--labels"
     :class="`apos-table__cell-field--${header.name}`"
   >
-    <span v-if="manuallyPublished && (item.modified || draft.modified) && !item.submitted && item.lastPublishedAt">
+    <span v-if="manuallyPublished && draft.modified && draft.submitted && draft.lastPublishedAt">
       <AposLabel
         label="Pending Updates" class="apos-table__cell-field__label"
         tooltip="There are active changes to this document."
         :modifiers="[ 'is-success', 'is-filled' ]"
       />
     </span>
-    <span v-else-if="item.submitted">
+    <span v-else-if="draft.submitted">
       <AposLabel
         label="Pending" class="apos-table__cell-field__label"
         tooltip="Changes to this document are awaiting approval by an admin or editor."
         :modifiers="[ 'is-filled' ]"
       />
     </span>
-    <span v-if="manuallyPublished && !item.lastPublishedAt">
+    <span v-if="manuallyPublished && !draft.lastPublishedAt">
       <AposLabel
         label="Draft" class="apos-table__cell-field__label"
         :modifiers="[ 'is-warning', 'is-filled' ]"
         tooltip="This document hasn't been published yet."
       />
     </span>
-    <span v-if="item.archived">
+    <span v-if="draft.archived">
       <AposLabel
         label="Archived" class="apos-table__cell-field__label"
         :modifiers="[ 'is-filled', 'is-danger' ]"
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -3,7 +3,7 @@
     class="apos-table__cell-field apos-table__cell-field--labels"
     :class="`apos-table__cell-field--${header.name}`"
   >
-    <span v-if="manuallyPublished && item.modified && !item.submitted && item.lastPublishedAt">
+    <span v-if="manuallyPublished && (item.modified || draft.modified) && !item.submitted && item.lastPublishedAt">
```
boutell: Use `draft && draft.modified`. `item` is for display purposes and isn't guaranteed to know about draft/live matters, it exists so you don't have to worry about that if all you want to do is display "what we think you should display."

boutell (commented):
@stuartromanek re: your side note it was agreed in a previous round of code review changes that the "item" passed to cells should be "what we want you to show people," abstracting that decision up a level and out of these cell components, but that we also need "draft" for cases where what we want to show is literally related to draft publication progress.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -3,28 +3,28 @@
     class="apos-table__cell-field apos-table__cell-field--labels"
     :class="`apos-table__cell-field--${header.name}`"
   >
-    <span v-if="manuallyPublished && item.modified && !item.submitted && item.lastPublishedAt">
+    <span v-if="manuallyPublished && draft.modified && draft.submitted && draft.lastPublishedAt">
```
abea, line 6: Shouldn't the last part still be `item.lastPublishedAt`? It's indicating a pending draft on a published thing.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -3,28 +3,28 @@
...
       <AposLabel
         label="Pending Updates" class="apos-table__cell-field__label"
         tooltip="There are active changes to this document."
         :modifiers="[ 'is-success', 'is-filled' ]"
       />
     </span>
-    <span v-else-if="item.submitted">
+    <span v-else-if="draft.submitted">
```
abea, line 13: Why do we have this if the previous one also checks for `draft.submitted`?

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -3,28 +3,28 @@
...
-    <span v-if="manuallyPublished && !item.lastPublishedAt">
+    <span v-if="manuallyPublished && !draft.lastPublishedAt">
       <AposLabel
         label="Draft" class="apos-table__cell-field__label"
         :modifiers="[ 'is-warning', 'is-filled' ]"
         tooltip="This document hasn't been published yet."
       />
     </span>
-    <span v-if="item.archived">
+    <span v-if="draft.archived">
```
abea, line 27: I think this should still be `item`. Is there even a draft to pass in if it's archived?

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -3,28 +3,28 @@
...
-    <span v-else-if="item.submitted">
+    <span v-else-if="draft.submitted">
       <AposLabel
         label="Pending" class="apos-table__cell-field__label"
         tooltip="Changes to this document are awaiting approval by an admin or editor."
         :modifiers="[ 'is-filled' ]"
       />
     </span>
-    <span v-if="manuallyPublished && !item.lastPublishedAt">
+    <span v-if="manuallyPublished && !draft.lastPublishedAt">
```
abea, line 20: Seems like an unnecessary change.

abea (changes requested):
`draft` defaults to `null` in the mixin, so should be update there or overridden here.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
boutell: `lastPublishedAt` is an aspect of the publication status, which apostrophe always records on the draft for certain, so it should come from `draft`.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
boutell: Just because it is submitted doesn't mean it's all those other things in that "and" statement. If it isn't, we have a fallback that applies to the case where it is merely submitted.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
boutell: There is always a draft. Archived things only exist as drafts.

boutell: (Which means I could check item.archived, technically, but the simplifying principle I'm pushing for here is *if you are checking something relating to the lifecycle status of the document, look at draft.*)

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
boutell: We do eventually copy this to `published`, but I believe it is best to get anything relating to publication status from the draft, which is the primary place Apostrophe stores it.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
abea: oh right.

modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
```diff
@@ -12,9 +12,7 @@ export default {
     },
     draft: {
       type: Object,
-      default() {
-        return null;
-      }
+      required: true
```
abea, line 15: Since `draft` is now required, we have no need for `item` as a prop. The manager is doing logic that could be done in the mixin instead. Contain that and let the manager only pass in two clear things.

abea (changes requested):

boutell (approved):

PR #3019, opened 2021-05-04
PRO-1287 now there is a "right time" to ddo things only after database migrations have definitely been performed. Also fixed a timeout issue with tests that interfered with validation.

boutell:


test/soft-redirects.js
```diff
@@ -67,7 +67,9 @@ describe('Soft Redirects', function() {
 
 });
 
-describe('Soft Redirects - with `statusCode` option', async() => {
+describe('Soft Redirects - with `statusCode` option', async function() {
```
boutell, line 70: Arrow functions break the way mocha uses "this" which prevents calls to "this.timeout".

boutell: It fails silently, too. Fun.

test/soft-redirects.js
```diff
@@ -67,7 +67,9 @@ describe('Soft Redirects', function() {
...
+
+  this.timeout(t.timeout);
```
boutell, line 72: Default timeout was too short every time today, versus the preferred timeout that we set in all our other test cases (in the same way I'm doing it here).

I think our startups are getting slower, but that's not a problem for this morning.

modules/@apostrophecms/migration/index.js
```diff
@@ -245,6 +245,15 @@ module.exports = {
...
           }
+          // In production, this event is emitted only at the end of the migrate command line task.
+          // In dev it is emitted at every startup after the automatic migration.
+          //
+          // Intentionally emitted regardless of whether the site is new or not.
+          //
+          // This is the right time to park pages, for instance, because the
+          // database is guaranteed to be in a sane state, whether because the
+          // site is new or because migrations ran successfully.
+          await self.emit('after');
```
boutell, line 256: In late A2, we introduced a top-level `apostrophe:migrate` event for the same reason. However I think `@apostrophecms/migrate:after` is better because migrations are managed by that module and things that can be managed by modules in Apostrophe should be. The other top level events are legitimately top level because they are things like `modulesReady` and `afterInit` that are defined by *all* modules having reached a certain point.

abea (dismissed):

abea (approved):

PR #3022, opened 2021-05-04
bring back missing column headings especially in the page tree

boutell:
* The page manager needs to pass data to the cell components the same way the docs manager does.
* Don't use lodash `.get`, it's overkill and it doesn't seem to like digging into computed properties.


modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -34,7 +34,9 @@
           v-for="(col, index) in headers"
           :key="`${col.property}-${index}`"
           :is="getEffectiveType(col, row)"
-          :item="row"
+          :item="row._publishedDoc || row"
```
abea, line 37: `item` shouldn't be passed in anymore.

abea (changes requested):

abea (approved):

abea (changes requested):
item shouldn't be passed in anymore.

abea (approved):

PR #3028, opened 2021-05-05
PRO-1298 PRO-1300 make sure parked pages get inserted if the database is brand new, even if in production

boutell:
I tested this using:

```
NODE_ENV=production npx mocha test/pages
```

Which fails without this patch, passes with it.


abea (changes requested):
This conflicts a bit with this nugget, which suggests the opposite: https://github.com/apostrophecms/apostrophe/blob/3.0/modules/@apostrophecms/migration/index.js#L233-L236

boutell:
It's not a contradiction — even if the migration code just says "since this is new I will mark all the migrations as skipped," that is work it needs to do on a new database on first time run, as well as emitting the "after" event so other code responsible for database sanity (like page parking) can decide what it needs to do.

abea (approved):
okle dokle

PR #3039, opened 2021-05-07
there is no addHelper

abea:
Fixes a doc comment referencing a non-existent method.

### DIFF
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -272,7 +272,7 @@ module.exports = {
       // be used very sparingly, and pretty much never in npm modules. The
       // only exceptions in apostrophe core are `apos.area` and `apos.singleton`.
       //
-      // The helper must be added first with `addHelpers` or `addHelper`.
+      // The helper must be added first with `addHelpers`.
 
       addHelperShortcut(name) {
         self.apos.template.addHelperShortcutForModule(self, name);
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/module/index.js
```diff
@@ -272,7 +272,7 @@ module.exports = {
       // be used very sparingly, and pretty much never in npm modules. The
       // only exceptions in apostrophe core are `apos.area` and `apos.singleton`.
       //
-      // The helper must be added first with `addHelpers` or `addHelper`.
+      // The helper must be added first with `addHelpers`.
```
boutell, line 275: The helper really should be added by putting it in the helpers section, although addHelpers still exists for implementation purposes, so how about documenting that instead.

boutell (changes requested):

modules/@apostrophecms/module/index.js
abea: There's no method by that name.

modules/@apostrophecms/module/index.js
```diff
@@ -272,7 +272,7 @@ module.exports = {
       // be used very sparingly, and pretty much never in npm modules. The
       // only exceptions in apostrophe core are `apos.area` and `apos.singleton`.
       //
-      // The helper must be added first with `addHelpers` or `addHelper`.
+      // The template helper must be added first. [See the module configuration reference](https://a3.docs.apostrophecms.org/reference/module-api/module-overview.html#customization-functions) for more.
```
abea, line 275: Note that we're not definitely going to be generating documentation from this. I'm hedging by using markdown syntax and using the full URL.

boutell (approved):

PR #3040, opened 2021-05-08
1381 empty image placeholder

### LINKED ISSUE #1381: When deleting media, widgets that use that media do not know they are now empty
This creates broken UI when there both no placeholder and no image. Perhaps some kind of 'this was recently deleted' message when revisiting the affected field?

zindarato:
This PR fixes #1381 

When an apostrophe-images widget is created but no images are selected, an empty space shows up with no information about the lack of content unless you hover over the area. This can be confusing for editors.

Added a message ("No images are currently selected.") when the widget exists but no image pieces have been selected. This message will only show up if the user has editing permissions on the area.

This branch also contains a past PR (I didn't branch the last one), but I assume these will get merged together if/when the other PR gets merged. Thanks!


abea:
@zindarato I don't think that last commit applies to this PR. That's for #3026 , right?

zindarato:
> @zindarato I don't think that last commit applies to this PR. That's for #3026 , right?

Yes, you're correct - it seems I accidentally committed to the wrong branch. Should be equivalent to 3026 - will this one be okay if 3026 is merged first, or should I cancel and resubmit this change under a different branch? Still getting used to GitHub's merging/PRs.

abea:
@zindarato Don't worry about it. We appreciate your work here. I'll get this reviewed this week.

abea (changes requested):
Thanks for taking this on! A helpful improvement.

[There is a style already for when rich text areas are empty.](https://github.com/apostrophecms/apostrophe/blob/1260f2cba260f88c3b44dfd32c43854707483753/lib/modules/apostrophe-rich-text-widgets/public/css/components/rich-text-widget.less#L25-L31) It might be good to apply a class like that to use that min-height and background color. It could center the text with flexbox. Additionally, adding the `.apos-ui` class would apply text styles to make it more readable.

Finally, please do update the changelog with an entry in the "unreleased" section.

lib/modules/apostrophe-images-widgets/views/widgetBase.html
```diff
@@ -1,49 +1,53 @@
...
+            {%- endblock -%}
+            {%- block description -%}
+              {%- if image.description -%}
+                <div>{{ image.description | nlbr }}</div>
+              {%- endif -%}
+            {%- endblock -%}
+          </div>
+        {%- endfor -%}
+      {%- elseif data.widget._edit -%}
+        <h4>No images are currently selected.</h4>
```
abea, line 49: This should be a paragraph tag.

abea (changes requested):

zindarato:
@abea I've set up the text to center and wrapped in apos-ui to apply text styling, as well as changing to a <p> tag. I was looking at setting up apos-empty for the height and background, but that's looking much more difficult. It appears that the apos-empty class is normally applied via JS in the schema editor. If I attempt to add it in the nunjucks template, it is removed by JS as soon as a user hovers over it. Do you think its still worth it to either set up a new class with similar behavior, or start changing the JS code around apos-empty to stop removing it?

Thanks!

zindarato:
Alright, thanks for the guidance. I'll split the relevant CSS into a separate mixin that both classes can inherit from - I agree on not refactoring the client side JS, especially since this template won't be rendered client-side anyway. I'll get a jump on this tomorrow. Thanks!

lib/modules/apostrophe-ui/public/css/components/fields.less
```diff
@@ -33,6 +33,12 @@
   padding: @apos-padding-1 @apos-padding-2;
 }
 
+.apos-empty-widget(@minHeight: 100px) {
```
abea, line 36: This should move to `lib/modules/apostrophe-ui/public/css/mixins/utils.less`. The areas module user.less file isn't finding it.

lib/modules/apostrophe-ui/public/css/components/fields.less
```diff
@@ -33,6 +33,12 @@
...
+    min-height: @min-height;
```
abea, line 37: The argument is `@minHeight` and this is `@min-height`.

lib/modules/apostrophe-images-widgets/public/css/user.less
```diff
@@ -20,5 +20,8 @@
       background-size: contain;
       background-repeat: no-repeat;
     }
+    .apos-empty-image {
```
abea, line 23: Right now this only affects image widgets in modals. It should apply to those in-context as well.

lib/modules/apostrophe-ui/public/css/components/fields.less
abea: The style is missing text centering. It should be fine to make it `display: flex` and use flex properties to center.

abea (changes requested):
Thanks @zindarato. A few technical fixes left in here. Please also add a changelog entry.

lib/modules/apostrophe-ui/public/css/components/fields.less
zindarato: Ah, thank you! I was testing last night trying to find out why the areas file wasn't able to find it. Good to know there's a dedicated mixins file.

The default styling for the existing class doesn't center the text - is it okay to change the rich-text empty area to be centered as well? I guess it should be fine as there isn't usually text shown there.

abea: Yeah, exactly. If we ever did add placeholder text in there we'd want it centered as well.

zindarato: Great, thank you. I just discovered NPM link, which should help in the future - I was pushing my code just so I could test it in a separate project, not realizing there was another way to pull in another NPM package locally. Next time I'll be able to test prior to pushing, which would probably help for things like this 😛. Thanks again, I'll have all of these items fixed shortly.

abea: Oh yes! `npm link` is crucial for this kind of thing. It'll be a game changer.

zindarato:
@abea Okay, think I've got it ready to go now. Sorry for all the back and forth, and thanks for the help! I learned a ton about changing NPM modules here as well as Apostrophe standards - looking forward to more in the future, during which hopefully I'll need less hand-holding! I got things set up using NPM link, which made testing much easier. Hopefully I caught everything, let mw know if you see any other glaring issues.

zindarato:
@abea Hold off on merging this one in - it seems that the centering of items did end up affecting other areas. The apos-empty class is being applied to all empty areas, not just rich text items like I believed before. The display: flex and justify attributes are causing the '+' icon to add a widget to be aligned on the right-hand side, as they are normally centered already.

Not sure what the correct fix for this is - I'm going to look at turning off and on the centering as another param in the mixin. If you think there's a better way, let me know! Here's an example of the bug that ends up happening.
Apos-Add-Widget

abea:
I didn't realize the `.apos-empty` class was at that higher level. I suggest you only add the flex centering to the empty image class, not on all apos-empty. Then you should be good.

zindarato:
@abea Alright, I went ahead and did as you suggested. The centering styles only get applied to empty image widgets. I ran through another round of testing, and things look correct now. Thanks!

abea:
@stuartromanek Do you have thoughts on this?

lib/modules/apostrophe-images-widgets/views/widgetBase.html
```diff
@@ -1,49 +1,55 @@
...
+            {%- block description -%}
+              {%- if image.description -%}
+                <div>{{ image.description | nlbr }}</div>
+              {%- endif -%}
+            {%- endblock -%}
+          </div>
+        {%- endfor -%}
+      {%- elseif data.widget._edit -%}
+        <div class="apos-ui apos-empty-image">
+          <p>No images are currently selected.</p>
```
abea, line 50: Oh, this will need to be in a localization helper:

```
<p>{{ __ns('apostrophe', 'No images are currently selected.') }}</p>
```

I forgot about that.

abea (changes requested):

stuartromanek (approved):

stuartromanek:
the appearance and behavior look good to me, thanks @zindarato !

zindarato:
@abea Thank you, I haven't worked with Apostrophe localization before, completely forgot about that. Just fixed it!

abea (approved):

PR #3054, opened 2021-05-11
PRO-1310 home page and other parked pages should not immediately show as "pending changes"

boutell:


modules/@apostrophecms/doc/index.js
```diff
@@ -644,8 +644,12 @@ module.exports = {
         const manager = self.apos.doc.getManager(doc.type);
         if (manager.isLocalized(doc.type) && doc.aposLocale.endsWith(':draft')) {
           // We are inserting the draft for the first time so it is always
-          // different from the published, which won't exist yet
-          doc.modified = true;
+          // different from the published, which won't exist yet. An exception
+          // is when the published doc is inserted first (like a parked page)
+          // in which case updateModified: false will be passed in
+          if (options.updateModified !== false) {
```
abea, line 650: I'm bumping on `updateModified`. It makes sense to have this as an option, but if it's about preventing setting `modified: true` because it's published already, how about `publishedOnInsert: true`? It's less about whether to change (or set) the `modified` property and more about reacting to the existence of the published doc.

boutell: updateModified already exists as an option and is supported in update operations, so I think it makes sense to continue this through, but I'll change it to `setModified` to avoid the implication it is only about updates. I think a more general option ("leave modified the heck alone, I know what I'm doing") is better than maintaining two options as long as it meets the need.

abea (dismissed):

abea (dismissed):

abea (dismissed):

PR #3055, opened 2021-05-11
PRO-1290 PRO-1162: permissions explorer no longer shows "submitted drafts," options to control that at the module level, any-page-type replaces page as a type name for relationships

boutell:


modules/@apostrophecms/doc/index.js
```diff
@@ -1025,6 +1025,16 @@ module.exports = {
       },
       isDraft(doc) {
         return doc.aposLocale.endsWith(':draft');
+      },
+      // Given a type name, normalize for any backwards compatibility
+      // provisions such as accepting @apostrophecms/page for
+      // @apostrophecms/any-page-type
+      normalizeType(type) {
```
boutell, line 1032: Now that we have reached beta, we should not introduce bc breaks without a compelling need.

modules/@apostrophecms/file/index.js
```diff
@@ -18,7 +18,8 @@ module.exports = {
     slugPrefix: 'file-',
     autopublish: true,
     editRole: 'editor',
-    publishRole: 'editor'
+    publishRole: 'editor',
+    showPermissions: true
```
boutell, line 22: Replaces `interestingTypes`, allows module developers to say what should happen with their type. In this case, opting in to always being called out separately in the permissions explorer.

abea (commented):
For what it's worth, it seems like this should have been two PRs.

modules/@apostrophecms/any-page-type/index.js
```diff
@@ -10,8 +10,6 @@ module.exports = {
   extend: '@apostrophecms/doc-type',
```
abea, line 10: The comment above needs updating, right?
> special doc type manager for the `@apostrophecms/page` type

modules/@apostrophecms/permission/index.js
```diff
@@ -288,7 +282,15 @@ module.exports = {
       },
       matchTypicalPieceType(permissionSet) {
```
abea, line 283: What is a typical piece type?

abea: I don't think this is requiring all custom doc types to include `showPermissions: true`, but I also am not fully understanding how a normal dev should think about that. And what purpose it's serving other than excluding when we say `false`. Right now it sounds like it's a binary choice, but it doesn't seem like it actually is.

abea (changes requested):

modules/@apostrophecms/permission/index.js
boutell: The design for the permissions explorer calls for folding all "typical, ordinary" pieces into one section, unless they (a) have permissions that actually would not match what is shown there, (b) explicitly opt into having their own entry in the permissions explorer with `showPermissions: true`, or (c) explicitly opt out of being mentioned at all in the permissions explorer, even in the "typical" section's tooltip, with `showPermissions: false`.

This is not really new. All that is new is that I removed the `interestingPieces` option, which would have been difficult to extend and maintain, in favor of a setting that can be made on a per-type level so that new types can specify their treatment in the permissions explorer.

abea: Got it. I think the method naming could be clearer, but I think the actual option makes sense.

abea (approved):

abea (approved):

PR #3064, opened 2021-05-12
PRO-1341 preserve the hash when updating query strings

boutell:
Also:

* eslint fixes.
* Don't manually chop off old query strings before calling, `addQueryToUrl` is smart enough to do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 ### Fixes
 
 - The `name` option to widget modules, which never worked in 3.x, has been officially removed. The name of the widget type is always the name of the module, with the `-widget` suffix removed.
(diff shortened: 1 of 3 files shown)


boutell:
More importantly I guess, the existing logic failed to add the query string properly because it added it after the # and it never reached the server.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 ### Fixes
 
 - The `name` option to widget modules, which never worked in 3.x, has been officially removed. The name of the widget type is always the name of the module, with the `-widget` suffix removed.
+- In-context editing works properly when the current browser URL has a `#` part, enabling the use of the `#` part for project-specific work. Thanks to [https://stepanjakl.com/](Štěpán Jákl) for reporting the issue.
```
abea, line 12: I think we can call it a hash.

abea (changes requested):
One changelog note. The term "hash" is part of the location API.

abea (changes requested):
Still the changelog bit.

abea (approved):

PR #3069, opened 2021-05-14
Sets the default fallback views directory

abea:
Rather than include a confusing core module configuration in our boilerplate, we can provide a default value and let devs override it if it they want. Our documentation has long discussed the root views directory as if it was a core feature.


modules/@apostrophecms/template/index.js
```diff
@@ -327,8 +328,12 @@ module.exports = {
         });
         // Final class should win
         dirs.reverse();
-        if (self.options.viewsFolderFallback) {
-          dirs.push(self.options.viewsFolderFallback);
+
+        const viewsFolderFallback = self.options.viewsFolderFallback ||
+          path.join(self.apos.rootDir, 'views');
+
+        if (viewsFolderFallback) {
```
boutell, line 335: This will always be truthy, based on line 332. May as well make clear it can't be entirely shut off by dropping the `if`.

boutell (changes requested):

boutell (approved):

stuartromanek:
🥂 here's to never seeing that glop in `app.js` again

PR #3070, opened 2021-05-14
Copies page browser data into any-page-type

abea:
This is an imperfect fix for a bug where a relationship chooser cannot currently be opened for an any-page-type relationship since there is no manager modal registered.

This does significantly duplicate methods from the `@apostrophecms/page` module, however. I started doing more significant refactoring to move all this out of that module and update UI components. Eventually it was clear it was a bigger thread to pull than I should without checking in.

Either we should get this in as a temporary fix for the next release or go ahead with the refactoring to move no longer reference the page module in the UI components other than in API routes.


boutell:
Oof. This is clearly my oversight and a big one. That change was more expensive than I thought.

Thanks for digging into this. I don't think using the docs/pieces manager is the way to go however. We allowed users to choose from the page tree before. Page titles aren't always chosen for global uniqueness so that can be important.

I think a better strategy would be to use the page module's UI for these, with the minimum glue needed to make that happen.

The main goal of this change was to get to "module name equals type name" for developers trying to figure out what to do in relationship config. Not to force us to move a whole lot of UI code. Hrrm.  Should we pair on this tomorrow (Tuesday)?

boutell:
Excuse me, I see that you did use the pages manager.

boutell:
Yeah, actually this approach ain't bad. We do need to do a little work to fix what's going on with getBrowserData, possibly call apos.page.getBrowserData and amend it rather than reimplementing it, that would eliminate most of the duplication.

boutell (changes requested):
See my comments — on reflection this is rather good, but I think we should directly invoke `apos.page.getBrowserData` and then amend it, particularly `browserOptions.name`, before returning the result. This will eliminate most of the duplicate code.

modules/@apostrophecms/any-page-type/index.js
```diff
@@ -12,18 +12,41 @@ module.exports = {
...
   methods(self) {
     return {
-      // Returns a string to represent the given `doc` in an
-      // autocomplete menu. `doc` will contain only the fields returned
-      // by `getAutocompleteProjection`. `query.field` will contain
-      // the schema field definition for the relationship the user is attempting
-      // to match titles from. The default behavior is to return
-      // the `title` property, but since this is a page we are including
-      // the slug as well.
+      // Returns a string to represent the given `doc` in an autocomplete menu.
```
abea, line 22: i'm triggered when they're not using the 80 char line limit.

boutell (approved):

PR #3073, opened 2021-05-17
PRO-409: rename the rest of the convert method object params to destination

boutell:


boutell:
Brought to you by vscode's "rename symbol" aka the F2 key

abea (changes requested):
Love it. It looks like string, slug, boolean, and email still use `object`. That should be it.

abea (approved):

PR #3074, opened 2021-05-18
PRO-1372 dethrone the P tag, impostor king

boutell:
The rich text editor now supports specifying a `styles` array with no `p` tags properly. A newly added rich text widget initially contains an element with the first style, rather than always a paragraph. If no styles are configured, a `p` tag is assumed. Thanks to Stepan Jakl for reporting the issue.

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Styles.js
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Styles.js
@@ -16,7 +16,7 @@ export default class Styles extends Node {
     return {
       attrs: {
         tag: {
-          default: defaultStyle.p
+          default: defaultStyle.tag
         },
         class: {
           default: defaultStyle.class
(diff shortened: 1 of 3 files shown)


boutell:
Closes #3063

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Styles.js
```diff
@@ -10,13 +10,16 @@ export default class Styles extends Node {
...
 
   get schema() {
+    const defaultStyle = (this.options.styles && this.options.styles[0]) || {
+      tag: 'p'
+    };
     return {
       attrs: {
         tag: {
-          default: 'p'
+          default: defaultStyle.p
```
abea, line 19: should be defaultStyle.tag, methinks.

abea (changes requested):

abea (approved):

boutell:
Good catch. Makes me wonder if setting that default there matters, but I don't dare leave out any of the copypasta magic when it comes to tiptap 1.

PR #3083, opened 2021-05-19
PRO-1371 Allows an empty radio input

abea:
Also gets the select and radio validation in sync since they work the same way.

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
@@ -70,7 +70,7 @@ export default {
   },
   methods: {
     validate(value) {
-      if (this.field.required && (!value || !value.length)) {
+      if (this.field.required && (value === null)) {
         return 'required';
       }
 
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
```diff
@@ -70,7 +70,7 @@ export default {
   },
   methods: {
     validate(value) {
-      if (this.field.required && !value.length) {
+      if (this.field.required && (!value || !value.length)) {
```
boutell, line 73: Select elements have logic to support preserving non-string values as such (that's why JSON is used), but it looks like this got broken at some point along the way by checking for `value.length`. Your code fixes the crash on that but doesn't allow the number zero to be chosen as a value. So I think it would be better to just check if `(value == null)`, unless that fails to do the job.

boutell: If that proves problematic though we could go with this fix as-is for now as it would not be a new regression.

boutell (changes requested):

boutell (approved):

PR #3085, opened 2021-05-19
PRO-1374 Removes duplicated _url from new doc

abea:


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -400,7 +401,10 @@ export default {
         newInstance.slug = this.copyOf.slug.replace(/([^/]+)$/, 'copy-of-$1');
       }
       delete newInstance._id;
+      delete newInstance._url;
```
abea, line 404: That's the ticket. Should be fine for everything since saving will generate the new URL.

boutell (approved):

PR #3086, opened 2021-05-19
PRO-1379 no relationship

abea:
Makes sure there are actual relationship fields there before we show the context menu.

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
@@ -27,7 +27,7 @@
           :size="13"
         />
         <AposContextMenu
-          v-if="item._fields"
+          v-if="hasFields"
           :button="more.button"
           :menu="more.menu"
           @item-clicked="$emit('item-clicked', item)"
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposSlat.vue
```diff
@@ -140,6 +140,21 @@ export default {
...
       }
+    },
+    hasFields() {
+      if (!this.item._fields) {
+        return false;
+      }
+
+      let ruling = false;
+
+      for (const key in this.item._fields) {
```
abea, line 151: I stole this like an artist.

modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
```diff
@@ -34,10 +34,6 @@
         />
       </transition-group>
     </draggable>
-
-    <div class="apos-slat-status">
-      {{ message }}
```
abea, line 39: There is no message.

modules/@apostrophecms/ui/ui/apos/components/AposSlatList.vue
```diff
@@ -65,12 +61,6 @@ export default {
     selected: {
       type: String,
       default: null
-    },
-    field: {
```
abea, line 69: There is no field. This prop isn't used anymore, only serves to show the message, the message has been replaced in design, and the field is not passed in in any use of the thing.

boutell (changes requested):
Hmm. I think consulting `_fields` was the original mistake here. This code would be shorter and more correct if the slat did receive the actual schema field, and could simply look at `field.schema` and check it for existence and length to determine if it is *supposed to* have a relationship. It's kind of immaterial whether it currently has populated fields or not; what matters is whether it's supposed to.

abea:
I went with passing the boolean through because the relationship input is the relevant context and we don't need to make the slats do work looking for stuff.

boutell (approved):

PR #3089, opened 2021-05-19
PRO-1385 PRO-1245 Fixes slug troubles.

abea:
Fixes slug troubles.

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
@@ -164,7 +164,7 @@ export default {
       return menu;
     },
     moduleName() {
-      if (this.context.slug.startsWith('/')) {
+      if (apos.modules[this.context.type].action === apos.modules['@apostrophecms/page'].action) {
         return '@apostrophecms/page';
       } else {
         return this.context.type;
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -696,7 +696,7 @@ export default {
         // automatically change the empty string to the prefix, so to
         // prevent a false positive for this being considered a change,
         // do it earlier when creating a new user
-        this.original.slug = slugField.def || slugField.prefix;
+        this.original.slug = this.original.slug || slugField.def || slugField.prefix || '';
```
abea, line 699: If slug.def was a string but prefix was undefined, it would end up undefined. It was also ditching the original slug set up for a page. Let the kids play, ref.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
```diff
@@ -164,7 +164,7 @@ export default {
       return menu;
     },
     moduleName() {
-      if (this.context.slug.startsWith('/')) {
+      if (apos.modules[this.context.type].action === apos.modules['@apostrophecms/page'].action) {
```
abea, line 167: The other fix might make this unnecessary, but it's also a bit of the tail wagging the dog as-is.

boutell (approved):
And the band wagged on.

PR #3092, opened 2021-05-20
PRO-1398 dismiss submission should default to false because it is really only for the manage submissions view

boutell:
Added missing plumbing to get these props where they need to go. Finished the job of renaming "can" to "show" when it denotes which functionality we want to surface in this location, not what the user can actually accomplish.

Note that "showCreate" mapping to "showCopy" is correct because "showCreate" is the more general tip coming down from the module level and is honored in two places, the "new" button and the "copy" button.


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextModeAndSettings.vue
```diff
@@ -50,6 +50,7 @@
         :doc="context"
         :show-preview="false"
         :show-edit="false"
+        :show-dismiss-submission="false"
```
abea, line 53: not necessary since the default is false.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -19,6 +19,7 @@
         :current="docFields.data"
         :published="published"
         :show-edit="false"
+        :show-dismiss-submission="false"
```
abea, line 22: not necessary since the default is false.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -7,8 +7,8 @@ module.exports = {
     perPage: 10,
     quickCreate: true,
     previewDraft: true,
-    canCreate: true,
-    canDismissSubmission: false
+    showCreate: true,
+    showDismissSubmission: false
```
abea, line 11: not necessary since the default is false.

abea (changes requested):
Some code cleanup bits. In contexts that are unrelated to submissions the extraneous props clutter and distract from the important stuff.

abea (approved):

PR #3093, opened 2021-05-24
PRO-1328 Updated what can be updated and wrote a notice for what cannot be

boutell:


boutell:
Bea, I did not see a changelog notice for the dependencies you eliminated (maybe I missed it) so I added it to the "unreleased" section of the changelog here.

abea (approved):

PR #3097, opened 2021-05-24
Replaces uses of whitelist

abea:
Replaces the last uses of whitelist in core code documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+* Updates uses of "whitelist" to "allowlist" to follow project practices.
```
abea, line 7: does this need a changelog entry? It'll mostly be seen in docs.

boutell: Can't hurt. I like that we're putting it out there.

boutell (approved):

PR #3098, opened 2021-05-24
PRO-1189 if the user clicks refresh, return to the same draft/published and edit/preview mode

boutell:


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -142,24 +142,17 @@ export default {
       sessionStorage.setItem('aposTabId', tabId);
     }
     window.apos.adminBar.tabId = tabId;
-
-    if (this.editMode) {
-      // Watch out for legacy situations where edit mode is active
```
boutell, line 147: This is no longer a concern.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -142,24 +142,17 @@ export default {
...
-      // but we are not in draft
-      if (!await this.lock(`${this.action}/${this.context._id}`)) {
-        this.lockNotAvailable();
-        return;
+    const lastBaseContext = JSON.parse(sessionStorage.getItem('aposLastBaseContext') || '{}');
+    if (lastBaseContext.aposDocId === this.context.aposDocId) {
+      if (lastBaseContext.draftMode !== this.draftMode) {
+        await this.setContext({ mode: lastBaseContext.draftMode });
       }
-      if (this.draftMode !== 'draft') {
```
boutell, line 153: More of the dead code for a no longer extant edge case that predates the change away from storing the editing state in the session.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -142,24 +142,17 @@ export default {
...
-
-    if (this.editMode) {
-      // Watch out for legacy situations where edit mode is active
-      // but we are not in draft
-      if (!await this.lock(`${this.action}/${this.context._id}`)) {
-        this.lockNotAvailable();
-        return;
+    const lastBaseContext = JSON.parse(sessionStorage.getItem('aposLastBaseContext') || '{}');
+    if (lastBaseContext.aposDocId === this.context.aposDocId) {
+      if (lastBaseContext.draftMode !== this.draftMode) {
```
abea, line 147: Any sense in checking if `lastBaseContext.draftMode` exists? And `lastBaseContext.editMode` below?

boutell: If lastBaseContext exists it'll have these properties. We don't always sanity-check every object we get from aposDocs (we trust our inserts and updates to insert well formed stuff) so I think this is OK here as well. Worst case scenario would be an unnecessary request on a one-time basis.

abea (dismissed):

PR #3099, opened 2021-05-25
PRO-1420 fix CSRF risk in nlbr and nlp filters

boutell:


abea:
@boutell FYI - it looks like you named the branch `3/` out of habit. Something to watch for to avoid confusion.

abea (changes requested):
I think this should include a test using the safe filter as well. I created one while testing, which you could use: https://github.com/apostrophecms/apostrophe/pull/3100

abea (approved):

PR #3104, opened 2021-05-26
PRO-1403 3/no flash login

stuartromanek:
Project level styles get loaded as part of outerlayout while the login vue app renders. Do our best to shut those visuals off while the login app fades in.

### DIFF
--- a/modules/@apostrophecms/login/views/login.html
+++ b/modules/@apostrophecms/login/views/login.html
@@ -3,6 +3,14 @@
 
 {% block bodyClass %}apos-login-page{% endblock %}
 
+{% block beforeMain %}
+  {# Hush up project level styles while login loads #}
+  <style>
+    body {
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/login/views/login.html
```diff
@@ -3,6 +3,14 @@
 
 {% block bodyClass %}apos-login-page{% endblock %}
 
+{% block beforeMain %}
+  {# Hush up project level styles while login loads #}
+  <style>
+    body {
+      background-color: var(--a-background-primary) !important;
```
abea, line 10: 😱

stuartromanek: ELAINE: "So, this is over, not over? I'm betting not over."

abea (approved):
🤷

PR #3110, opened 2021-05-27
Unit tests showing that "fields" is already implemented for relationships, also singularize module names in this test file for consistency

boutell:


test/pieces.js
```diff
@@ -464,6 +475,34 @@ describe('Pieces', function() {
...
+      {
+        ...thing1,
+        _fields: {
+          skillLevel: 5
+        }
+      }
+    ];
+    await apos.http.put('/api/v1/person/person1:en:published', {
+      body: person1,
+      jar
```
boutell, line 493: Without using the jar here, I would have to add _tools to the `publicApiProjection` for the `person` module. Not fetching fields not in that projection is a feature rather than a bug. Threw me for a second.

abea (approved):

PR #3112, opened 2021-05-27
The data retained for "Undo Publish" no longer causes slug conflicts in certain situations

boutell:


modules/@apostrophecms/doc-type/index.js
```diff
@@ -186,99 +186,20 @@ module.exports = {
             self.emit('afterDelete', req, previous, { checkForChildren: false });
           }
         },
-        deduplicateArchive(req, doc) {
-          const deduplicateKey = doc.aposDocId;
-          if (doc.parkedId === 'archive') {
-            // The primary archive itself should not deduplicate
-            return;
+        async deduplicateArchive(req, doc) {
+          const $set = await self.deduplicate(req, doc);
```
boutell, line 190: More functional approach: the shared "deduplicate" method has no side effects.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -931,6 +854,108 @@ module.exports = {
...
+      // Returns an object containing the properties of doc that
+      // require deduplication when archived, stored as
+      // "previous published" or any other scenario where the slug
+      // and similar properties should never be treated as "in conflict"
+      // with content that is in play on the site. The returned object
+      // contains values for those properties that have been deduplicated, and
+      // can be passed to $set or Object.assign or both, depending on your
+      // situation. `doc` is not changed.
+
+      async deduplicate(req, doc) {
```
boutell, line 868: Also no side effects in a sense, although it does have to do read operations on the database to do its job.

modules/@apostrophecms/doc-type/index.js
abea: Since this isn't doing the duduplication, I suggest `getDeduplicationSet` or `deduplicationSet`

modules/@apostrophecms/doc-type/index.js
```diff
@@ -186,99 +186,20 @@ module.exports = {
...
-          async function update() {
-            const action = { $set: $set };
-            // So methods called later, or extending this method, see the change in docs
-            _.assign(doc, $set);
-            if (_.isEmpty($set)) {
-              // Nothing to do
-              return;
-            }
-            return self.apos.doc.db.updateOne({ _id: doc._id }, action);
+        async reduplicateRescue(req, doc) {
```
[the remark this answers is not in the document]
abea: Or `restoreDeduplications`

modules/@apostrophecms/doc-type/index.js
abea: I don't see any `await` in here.

abea (changes requested):

boutell:
Back to you. I went with these names because "restore" is too general a concept for what is happening here and I don't want to steal it. We're specifically deduplicating and reverting deduplication, we're not addressing other concerns around restoration (and there are).

boutell:
I kept "async" for `getDeduplicationSet` because I don't want to slam the door too hard on the possibility that we'll decide to be more subtle in our deduplicated slugs, etc., at some cost in additional logic.

abea (approved):

boutell:
(Thank you for the nicely functional method name suggestion.)

PR #3118, opened 2021-05-28
PRO-1473 PRO-1405 dynamically determine whether a "save and view" button etc. is a good idea

boutell:
This way we don't hit the wall on the first custom type we encounter that has these properties.


abea (approved):

boutell:
Actually, manual publication isn't the determining factor here. The relevance and existence of piece pages is. Which is 1405. Going to come back with that ticket incorporated because this is really one jawn.

PR #3122, opened 2021-06-01
PRO-1378 move color format option out of UI options

stuartromanek:


abea (commented):
We're no longer passing the format to the `Picker` component. Is that okay?

stuartromanek:
> We're no longer passing the format to the `Picker` component. Is that okay?

it is. in fact, it is the purpose of this very PR.

abea (approved):

PR #3140, opened 2021-06-03
PRO-1441 PRO-1524 PRO-1538 attachment reference issues

boutell:


boutell:
QA steps are in linear.

modules/@apostrophecms/attachment/index.js
```diff
@@ -122,6 +123,10 @@ module.exports = {
       'migrate-from-disabled-file-key': {
         usage: 'Usage: node app @apostrophecms/attachment:migrate-from-disabled-file-key\n\nThis task should be run after removing the disabledFileKey option from uploadfs.\nIt should only be relevant for storage backends where\n' + 'that option is not mandatory, i.e. only local storage as of this writing.',
         task: self.migrateFromDisabledFileKeyTask
+      },
+      'recompute-all-doc-references': {
+        usage: 'Recompute mapping between attachments and docs,\nshould only be needed for rare repair situations',
```
abea, line 128: Why would this need to stay around as a task?

modules/@apostrophecms/attachment/index.js
```diff
@@ -1037,6 +1042,66 @@ module.exports = {
...
+      addDocReferencesContainedMigration() {
+        self.apos.migration.add(self.__meta.name + '.docReferencesContained', self.recomputeAllDocReferences);
+      },
+      // Recompute the `docIds` and `archivedDocIds` arrays
+      // from scratch. Should only be needed by the
+      // one-time migration that fixes these for older
+      // databases, but can be run at any time via the
+      // `apostrophe-attachments:recompute-doc-references`
+      // task, just in case the need arises or your site
+      // was affected by the very brief availability of 2.77.0
```
abea, line 1062: Is this `2.77.0` version number right? This is for 3.x

modules/@apostrophecms/attachment/index.js
```diff
@@ -1037,6 +1042,66 @@ module.exports = {
...
+      // which effectively marked all attachments as
+      // not in use.
+      async recomputeAllDocReferences() {
+        const attachmentUpdates = {};
+        await self.apos.migration.eachDoc({}, 5, addAttachmentUpdates);
```
abea, line 1067: This works, but I'm confused. They're in aposAttachments, but my understanding is that eachDoc only hits aposDocs.

abea (changes requested):
In my test of 1524 the page doesn't seem to realize that the file has been removed from the widget's attachment field. If I completely delete the widget teh attachment is removed from the filesystem.

1441 tests well.

modules/@apostrophecms/attachment/index.js
boutell: People have been known to manipulate mongo directly and could miss a nuance like the attachment reference tracking.

modules/@apostrophecms/attachment/index.js
boutell: It was accurate but a reference to a very old incident not really relevant in 3.x source comments, so removed.

modules/@apostrophecms/attachment/index.js
boutell: Correct, but we use what we learn from scanning those documents to build an array of mongodb bulk operations that are then executed against aposAttachments (self.db, in the attachment module).

boutell:
Re: your test of 1524, my suspicion is that what really happened is that your additional step, purging the widget and again publishing, completely removed any memory of the attachment from draft, published and previous, which is the point where the deletion takes place. This probably would have happened too with a different kind of edit unrelated to the widget, what mattered was one more publish to push out the old "previous published." I suspect this test passed, in other words.

abea (changes requested):
When I test 1524, I add an attachment to a page with a new attachment widget (not an image widget). When I edit that to remove the attachment (not removing the widget itself) the page doesn't have a highlighted "Update" button as if it doesn't know anything changed.

I'm using the directions:
- create a widget with an optional attachment field of its own (NOT a relationship to the image piece type). Populate it in draft on a page that is NOT the home page. Publish it. It's still there on the hard drive.
- Now edit the widget and REMOVE the attachment. Still on the hard drive though.
- Publish that change, then immediately UNDO publish. Still on the hard drive.

abea:
I'm also getting this error in the terminal after removing the attachment from the field:
```
PATCH /api/v1/@apostrophecms/page/ckphgp3m1000f963r8h9p44cg:en:draft:

TypeError: Cannot read property 'title' of null
    at Object.index (/Users/abea/Apos/apostrophe/modules/@apostrophecms/attachment/index.js:281:23)
    at /Users/abea/Apos/apostrophe/modules/@apostrophecms/schema/index.js:1526:21
    at arrayEach (/Users/abea/Apos/apostrophe/node_modules/lodash/lodash.js:530:11)
    at Function.forEach (/Users/abea/Apos/apostrophe/node_modules/lodash/lodash.js:9410:14)
```

abea:
I can't tell if I'm testing this wrong. When I undo on the last step the attachment is still gone from the filesystem.
> Publish that change, then immediately UNDO publish. Still on the hard drive.

boutell:
OK, that sounds like a bug I'll have to reproduce.

abea (dismissed):

PR #3146, opened 2021-06-04
PRO-1525 SVGs by default

stuartromanek:


modules/@apostrophecms/attachment/index.js
```diff
@@ -51,8 +51,9 @@ module.exports = {
         extensions: [
           'gif',
           'jpg',
-          'png'
-        ].concat(self.options.svgImages ? [ 'svg' ] : []),
+          'png',
+          'svg'
```
boutell, line 55: Not having an option is bold, but let's try it man, see if anybody has real concerns; can always add options, can't take 'em away

boutell (approved):

PR #3151, opened 2021-06-08
PRO-1535: relationship subfields working in expanded chooser (manager view), also fixed initialization with support for defaults

boutell:
* Relationship subfields can now be edited when selecting in the full "manage view" browser, as well as in the compact relatio
nship field view which worked previously.
* A console warning when editing subfields for a new relationship was fixed.
* Relationship subfields now respect the `def` property (there was no real previous implementation of initializing the object, with or without defaults, so no reason not to do it right the first time).

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
+++ b/modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
@@ -6,6 +6,15 @@
 import { isEqual } from 'lodash';
 
 export function detectDocChange(schema, v1, v2, options = {}) {
+  // Handle null docs
+  if (!v1) {
+    // If there is no v1 then it's a change if there is a v2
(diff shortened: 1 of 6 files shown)


abea (changes requested):
Change in functionality was updated in the Linear ticket.

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
```diff
@@ -6,6 +6,15 @@
 import { isEqual } from 'lodash';
 
 export function detectDocChange(schema, v1, v2, options = {}) {
+  // Handle null docs
```
boutell, line 9: This can come up when _fields isn't set yet.

abea (approved):

PR #3152, opened 2021-06-08
Adds docs on publishing users

abea:
Clears up the sentence about being able to reintroduce the publish field for users. This came up in SO recently.


lib/modules/apostrophe-users/index.js
```diff
@@ -26,13 +26,20 @@
...
+// allowed to log in**.
 //
 // So for a staff directory, we suggest you create a separate `employee` module
 // or similar, extending `apostrophe-pieces`, unless it's true that basically
 // everyone should be allowed to log in.
 //
+// ::: tip
+// To bring the `published` field back for the `apostrophe-users` piece type,
+// set the `defaultRemoveFields`, `defaultRemoveFilters`, and
+// `defaultRemoveColumns` properties on that module to an empty array.
```
boutell, line 40: Have you verified that you can still log in and perform other user actions like editing as a user who is not "published" once that field has been brought back? Note that when you make this transition all users will be unpublished as of that moment.

boutell (changes requested):

PR #3157, opened 2021-06-10
ui/src implementation

boutell:


modules/@apostrophecms/asset/lib/refresh-on-restart.js
```diff
@@ -7,11 +7,15 @@
...
+    if (!window.apos) {
+      // Wait for the js bundle to evaluate
+      return setTimeout(check, 100);
+    }
     if (document.visibilityState === 'hidden') {
       // No requests when tab is not active
       setTimeout(check, 5000);
     } else {
-      apos.http.get(document.querySelector('[data-apos-refresh-on-restart]').getAttribute('data-apos-refresh-on-restart'), {
+      apos.http.post(document.querySelector('[data-apos-refresh-on-restart]').getAttribute('data-apos-refresh-on-restart'), {
```
boutell, line 18: This became a `post` because of the risk of caching the GET response, which defeats the purpose.

boutell: (It happens to be something IE11 does, but IE11 isn't wrong to cache GET.)

boutell:
In draft while I figure out why production builds aren't working.

modules/@apostrophecms/asset/lib/refresh-on-restart.js
```diff
@@ -7,11 +7,15 @@
   var fast = '1';
   check();
   function check() {
+    if (!window.apos) {
+      // Wait for the js bundle to evaluate
+      return setTimeout(check, 100);
```
boutell, line 12: We are using the "module/nomodule" pattern to avoid the overhead of loading the IE11 build in modern browsers. Since we are using the "module" attribute to the script tag in modern browsers, the load is always deferred. This is a good thing, however it means our refresh-on-restart script must be savvy and wait for apos to pop up. It's the only piece of javascript that has any business being outside the bundle.

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -25,7 +25,9 @@ module.exports = ({
   const config = {
     entry: importFile,
     mode: process.env.NODE_ENV || 'development',
-    optimization: { minimize: false },
+    optimization: {
+      minimize: process.env.NODE_ENV === 'production'
```
boutell, line 29: Approx 1/3 size reduction in my tests.

modules/@apostrophecms/asset/lib/webpack/src/webpack.es5.js
```diff
@@ -0,0 +1,33 @@
+module.exports = (options, apos) => {
+  return {
+    target: 'es5',
+    module: {
+      rules: [
+        {
+          test: /\.js$/,
+          exclude: [
+            /\/core-js($|\/)/,
```
boutell, line 9: babel must not be applied to its own polyfills.

modules/@apostrophecms/asset/lib/webpack/src/webpack.es5.js
```diff
@@ -0,0 +1,33 @@
...
+          use: [
+            {
+              loader: 'babel-loader',
+              options: {
+                presets: [
+                  [
+                    '@babel/preset-env',
+                    {
+                      targets: {
+                        browsers: '> 1%, IE 11, not dead'
```
boutell, line 21: A lot of stackoverflow research led to this working config.

modules/@apostrophecms/asset/index.js
```diff
@@ -11,17 +11,103 @@ module.exports = {
...
+    es5: false,
     // If this option is true and process.env.NODE_ENV is not `production`,
     // the browser will refresh when the Apostrophe application
     // restarts. A useful companion to `nodemon`.
-    refreshOnRestart: false
+    refreshOnRestart: false,
+    // Frontend builds. Changing these settings is not recommended
+    // except as part of development of the Apostrophe core. To enable
+    // the src-es5 build just set the `es5: true` option of this module.
+    builds: {
```
boutell, line 26: The goal here isn't perfect configurability for future, unanticipated builds of every possible stripe right now. It's more to ensure we're taking an approach that can grow in that direction if needed, as requested.

modules/@apostrophecms/asset/index.js
```diff
@@ -37,9 +123,9 @@ module.exports = {
             self.apos.options.autoBuild !== false
           ) {
             // If starting up normally, run the build task, checking if we
-            // really need to update the core UI build.
+            // really need to update each build
             await self.apos.task.invoke('@apostrophecms/asset:build', {
-              'check-ui-build': true
+              'check-frontend-build': true
```
boutell, line 128: It's not just the UI build anymore.

modules/@apostrophecms/asset/index.js
```diff
@@ -56,65 +142,56 @@ module.exports = {
...
           await fs.remove(buildDir);
           await fs.mkdirp(buildDir);
-          await fs.remove(modulesDir);
-          await fs.mkdirp(modulesDir);
 
-          let rebuildAposUi = argv && !argv['check-ui-build'];
+          for (const [ name, options ] of Object.entries(self.options.builds)) {
+            let rebuild = argv && !argv['check-ui-build'];
 
-          const APOS_MERGED_BUNDLE = 'apos-bundle.js';
```
boutell, line 74: All coming from the "builds" config now.

modules/@apostrophecms/asset/index.js
```diff
@@ -56,65 +142,56 @@ module.exports = {
...
-          const APOS_ONLY_BUNDLE = 'apos-only-bundle.js';
-          const APOS_ONLY_TS = '.apos-only-timestamp.txt';
-          const PUBLIC_BUNDLE_CSS = 'public-bundle.css';
-          const PUBLIC_BUNDLE_JS = 'public-bundle.js';
-          let checkTimestamp = false;
+            let checkTimestamp = false;
 
-          const bundleExists = await fs.pathExists(bundleDir);
+            if (options.core) {
```
boutell, line 160: Only the apos build currently has the core flag, others will rebuild all the time, like ui/public always did. The expectation is that they're small.

modules/@apostrophecms/asset/index.js
```diff
@@ -56,65 +142,56 @@ module.exports = {
...
-            await fs.remove(`${bundleDir}/${APOS_ONLY_BUNDLE}`);
           }
 
-          await moduleOverrides();
-          await buildPublicCssBundle();
-          buildPublicJsBundle();
-
-          if (rebuildAposUi) {
-            await buildAposBundle();
+          const scenes = [...new Set(Object.values(self.options.builds).map(options => options.scenes).flat()) ];
```
boutell, line 188: Identify the unique scenes (just "public" and "apos" for now and for the foreseeable future, but we're not hardcoding that).

modules/@apostrophecms/asset/index.js
```diff
@@ -123,7 +200,9 @@ module.exports = {
             });
           }
 
-          async function moduleOverrides() {
+          async function moduleOverrides(modulesDir, source) {
```
boutell, line 203: Each build needs its own working folder where the assets for that module name get copied in "subclass last" order, so that webpack can resolve things simply against that folder. This isn't new but it's per webpack build now, not just the apos build.

modules/@apostrophecms/asset/index.js
```diff
@@ -286,15 +373,43 @@ module.exports = {
...
+            const jsNoModules = `${scene}-nomodule-bundle.js`;
+            const css = `${scene}-bundle.css`;
+            fs.writeFileSync(
+              `${bundleDir}/${jsModules}`,
+              Object.entries(self.options.builds).filter(
+                ([ name, options ]) => options.scenes.includes(scene) &&
+                options.outputs.includes('js') &&
+                (!options.condition || options.condition === 'module')
+              ).map(([ name, options ]) => {
+                return `// BUILD: ${name}\n` + fs.readFileSync(`${bundleDir}/${name}-build.js`, 'utf8');
```
boutell, line 387: Adding the build name comment was a debugging thing, but I'd like to keep it, at least for now. Just a few bytes and it shoudl help us understand what we're looking at.

modules/@apostrophecms/asset/index.js
```diff
@@ -286,15 +373,43 @@ module.exports = {
...
+              }).join('\n')
+            );
+            fs.writeFileSync(
+              `${bundleDir}/${jsNoModules}`,
```
boutell, line 391: The jsNoModules bundle contains code from builds that have "condition: nomodule" (the src-es5 build) and those that don't care. Thus everything IE11 can handle.

modules/@apostrophecms/asset/index.js
```diff
@@ -286,15 +373,43 @@ module.exports = {
             return output;
           }
 
-          function merge() {
+          function merge(scene) {
+            const jsModules = `${scene}-module-bundle.js`;
+            const jsNoModules = `${scene}-nomodule-bundle.js`;
+            const css = `${scene}-bundle.css`;
+            fs.writeFileSync(
+              `${bundleDir}/${jsModules}`,
```
boutell, line 381: The modern browsers bundle.

modules/@apostrophecms/asset/index.js
```diff
@@ -286,15 +373,43 @@ module.exports = {
...
             fs.writeFileSync(
-              `${bundleDir}/${APOS_MERGED_BUNDLE}`,
-              fs.readFileSync(`${bundleDir}/public-bundle.js`) +
-                fs.readFileSync(`${bundleDir}/apos-only-bundle.js`)
+              `${bundleDir}/${css}`,
+              Object.entries(self.options.builds).filter(
+                ([ name, options ]) => options.scenes.includes(scene) &&
+                options.outputs.includes('css')
+              ).map(([ name, options ]) => {
+                return `/* BUILD: ${name} */\n` + fs.readFileSync(`${bundleDir}/${name}-build.css`, 'utf8');
```
boutell, line 406: Being able to figure out where CSS came from (ui/public vs ui/src) can be nice too.

boutell:
Ready for review.

IE11 note: IE11 isn't smart enough to *not* try to load the modern browsers bundle; it loads both, can't understand one, and then the ES5 ("nomodule") bundle actually works. So you will see one syntax error in the console as it gives up on the modern browsers bundle. This is just how it is with this technique, which is weighted in favor of eliminating any performance impact on the modern browsers, which are smart enough to load only the js file with "script module" and ignore the one with "script nomodule".

boutell:
Aha, correction: no more syntax error in IE11, I had written `script module`, the correct syntax is `script type="module"`. IE11 is smart enough to ignore a script tag with a "type" it does not understand.

modules/@apostrophecms/asset/index.js
```diff
@@ -11,17 +11,103 @@ module.exports = {
...
+        `
+      },
+      apos: {
+        scenes: [ 'apos' ],
+        outputs: [ 'js' ],
+        webpack: true,
+        label: 'Apostrophe admin UI',
+        // Only rebuilt on npm updates unless CORE_DEV is set in the environment
+        core: true,
+        icons: true,
```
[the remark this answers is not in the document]
abea: I'm imagining some future installed module to add to the UI. It might have apps and components but no tiptap extensions. That step would simply continue quietly once it saw no tiptaps, right?

modules/@apostrophecms/asset/index.js
```diff
@@ -152,29 +72,42 @@ module.exports = {
...
-              }
+                if (!process.env.APOS_DEV) {
+                  checkTimestamp = await fs.pathExists(`${bundleDir}/${name}-only-timestamp.txt`);
+                }
 
-              if (!rebuild && checkTimestamp) {
-                // If we have a UI build timestamp file compare against the app's
-                // package.json modified time.
-                if (await lockFileIsNewer(name)) {
+                if (!rebuild && checkTimestamp) {
```
abea, line 97: we don't need to check !rebuild again since it's inside a conditional for that already.

modules/@apostrophecms/asset/index.js
```diff
@@ -607,6 +539,92 @@ module.exports = {
...
+          apos: {
+            scenes: [ 'apos' ],
+            outputs: [ 'js' ],
+            webpack: true,
+            label: 'Apostrophe admin UI',
+            // Only rebuilt on npm updates unless APOS_DEV is set in the environment
+            // to indicate that the dev writes project level or npm linked admin UI
+            // code of their own which might be newer than package-lock.json
+            apos: true,
+            icons: true,
```
abea, line 603: No longer needs icons, components, tiptapExts, and apps.

modules/@apostrophecms/asset/lib/webpack/src/webpack.scss.js
```diff
@@ -37,7 +37,7 @@ module.exports = (options, apos) => {
     plugins: [
       new MiniCssExtractPlugin({
         // Should be automatic but we wind up with main.css if we try to go with that
-        filename: options.outputFilename.replace('.js', '.css')
+        filename: options.outputFilename.replace('.js', '.css'),
```
abea, line 40: hanging comma

abea (changes requested):
A few small bits

abea (approved):

PR #3162, opened 2021-06-11
PRO-1443: global doc respects "def" of its schema fields at site creation time

boutell:


stuartromanek (approved):
COol and nice. This wouldn't cover newly added fields with a `def` property though, would it? Having your global config locked and loaded on first run is not the norm for development.

Probably a different enhancement

PR #3167, opened 2021-06-14
PRO-1510: fix number field and min/max problems

boutell:
* `min` and `max` work properly for both string-like and number-like fields.
* Negative numbers, leading minus and plus signs, and trailing periods are accepted in the right ways by appropriate field types.


stuartromanek (changes requested):
This kind of thing should have tests

boutell:
I have added test notes to the ticket.

stuartromanek (dismissed):

PR #3168, opened 2021-06-14
PRO-1555 data.page and data.contextOptions available when widgets are first inserted on the page or edited in the widget editor dialog and saved back to the page

boutell:


CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ myColorField: {
 }
 ```
 * Restored Vue dependency to using semantic versioning now that Vue 2.6.14 has been released with a fix for the bug that required us to pin 2.6.12.
+* `data.page` and `data.contextOptions` are now available in `widget.html` templates even when the widget has just been inserted or has just been edited in a dialog box and saved back to the page. Note however that they are never available when a widget is rendered directly inside a dialog box. Your widget templates should still be written to tolerate the absence of context information from the page template.
```
stuartromanek, line 54: Can you make this clearer? "or has just been edited in a dialog box and saved back to the page" vs "never available when a widget is rendered directly inside a dialog box" is pretty muddy

stuartromanek (approved):
nervously approving this change to a bedrock feature this close to publish

CHANGELOG.md
boutell: I tried, but short of screenshots in the changelog it's pretty tough.

PR #3178, opened 2021-06-17
PRO-1617: hotfix to support multiple ui/src folders, including a robust test in…

boutell:
… the form of migrating our own code to that model, which required me to address several timing issues to ensure everything works right both logged in and logged out


modules/@apostrophecms/asset/index.js
```diff
@@ -554,6 +554,16 @@ module.exports = {
       // If you are trying to enable IE11 support for ui/src, use the
       // `es5: true` option (es5 builds are disabled by default).
       configureBuilds() {
+        const srcPrologue = stripIndent`
```
boutell, line 557: Moved this prologue from public to src because src loads first (as it must, if custom webpack builds are going to use apos frontend utilities) and src now contains our core frontend js

modules/@apostrophecms/util/ui/src/http.js
```diff
@@ -1,11 +1,11 @@
...
+export default () => {
+  // Adds the apos.http client, which has the same API
+  // as the server-side apos.http client, although it may
+  // not have exactly the same features available.
+  // This is a lean, IE11-friendly implementation.
 
-(function() {
-  var busyActive = {};
-  var apos = window.apos;
+  const busyActive = {};
```
boutell, line 7: eslint

abea (approved):

PR #3187, opened 2021-06-24
Adds totp rate limit

ValJed:
[IN-22](https://apostrophecms.atlassian.net/browse/IN-22)

## Purpose
Extracting the throttle system from the classic login logic to reuse it for the login totp logic.
Also avoid that servers breaks if we call the `/login-totp` route without any user in **req**.

We use the throttle system already used for classic login.
I extracted this logic into three methods:
* checkThrottleAttempts: Verify before trying to log the user that he didn't already exceed the throttle limit.
* addThrottleAttempt: During login fails, it adds an attempt to the user and checks if he should be locked or not.
* cleanUpThrottleAttempts: Clean old attempts if we are in a login fail, if the login succeed we remove all attempts (to avoid any conflicts between classic login attempts and TOTP attempts).

@boutell for the redirection issue, for now I just added the `/modules/apostrophe-areas/editor` path to the safeList. It fixed the problem even if definitely not ideal.
Not sure about your `404 things matching that pattern`, if you want me to implement it this way let me know.


boutell (changes requested):
Please see the existing `throttle` option to the login module. I would suggest that this should apply equally to regular login attempts and TOTP attempts, rather than introducing a separate mechanism and a separate syntax for configuration for TOTP throttling only.

lib/modules/apostrophe-login/index.js
```diff
@@ -428,40 +404,102 @@ module.exports = {
...
-        if (userSafe.attempts && userSafe.attempts.filter(attempt => (attempt >= perSince)).length >= allowedAttempts) {
-          await safe.update({
-            _id: user._id
-          }, {
-            $set: {
-              lockout: new Date()
-            }
-          });
-          return true;
+        const unallowedAttemps = userSafe.attempts && userSafe.attempts
```
boutell, line 409: Typo: should be `unallowedAttempts` and `disallowedAttempts` would be better grammar. Actally let's just change it to `disallowed`, no need for the Attempts part of this variable name.

boutell (changes requested):
One minor change.

Has this been thoroughly tested for both regular login throttling and TOTP throttling?

lib/modules/apostrophe-login/index.js
ValJed: Yes it's clearer!

ValJed:
@boutell Yes I did for both, and I made sure that the attempts are independent between `login` and `login-totp` by cleaning them when login.
You can test it easily in a boilerplate if you want to.

boutell (approved):

boutell:
Tested successfully. Nice work!

ValJed:
I just removed the `express-rate-limit` package and merged 2.0 updates.

boutell (approved):

PR #3188, opened 2021-06-24
PRO-1582 normalize widget chrome

stuartromanek:
- widgets at all depths have the same visual look
- breadcrumbs are now theme inverted along with widget controls
- secondary theme color removed

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
@@ -531,11 +531,6 @@ export default {
     &.apos-is-focused:before {
       opacity: 0.15;
     }
-
-    &.apos-is-highlighted,
-    &.apos-is-focused {
-      outline-color: var(--a-secondary);
-    }
   }
 
   .apos-area-widget-controls {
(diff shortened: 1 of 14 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -615,11 +605,8 @@ export default {
     align-items: center;
     margin: 0;
     padding: 2px;
-    background-color: var(--a-primary);
-  }
-
-  .apos-area-widget-inner .apos-area-widget-inner .apos-area-widget__breadcrumbs {
-    background-color: var(--a-secondary);
+    background-color: var(--a-background-primary);
+    border: 1px solid var(--a-primary-button-hover);
```
stuartromanek, line 609: @abea i know this un-semantic use of variables will grind your gears, I suggest a ticket that converts them to something more agnostic of use, more like the base vars

abea: Is this not a find all and replace quick thing? Why not do it in this change?

stuartromanek: because it seemed like a different thing that would grind your gears

stuartromanek: i can do it that way though, sure

abea: Yeah, it'd be better to get it done. No idea when the ticket would happen.

abea (commented):
See comment on the var names.

abea (approved):

PR #3203, opened 2021-06-29
PRO-1679 lint for missing export default

boutell:
Since `ui/src/index.js` files must export a function to avoid a browser error in production which breaks the website experience, we now detect this at startup and throw a more helpful error to prevent a last-minute discovery in production.

This is not perfect, in particular we could look more specifically for the export being a function, but I don't want to mess this up by failing to think about cases that are valid but use different syntax. Making sure "export default" exists should cover 99% of likely developer errors on this point.

### DIFF
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -188,7 +188,8 @@ module.exports = {
             if (options.index) {
               indexJsImports = getImports(source, 'index.js', {
                 invokeApps: true,
-                importSuffix: 'App'
+                importSuffix: 'App',
+                requireDefaultExport: true
               });
               indexSassImports = getImports(source, 'index.scss', {
                 importSuffix: 'Stylesheet'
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/asset/index.js
```diff
@@ -396,6 +397,16 @@ module.exports = {
             };
 
             components.forEach((component, i) => {
+              if (options.requireDefaultExport) {
+                if (!fs.readFileSync(component, 'utf8').match(/export[\s\n]+default/)) {
```
abea, line 401: Would there be no way to use a named export in a front end file?

boutell: Not in a `ui/src/index.js` no, those are for Apostrophe to import, if you want to provide code for reuse by various modules you can, you just wouldn't put it in `ui/src/index.js` and then you can import to your heart's content, use named exports, etc.

abea (approved):

PR #3208, opened 2021-06-30
oembed field fixes

stuartromanek:
Fixes two bugs in oembed fields
- Detecting existing field values was getting tripped up between `null` and `undefined`. If our content has an existing oembed value we know it will be an object so we don't have to be so careful as to what to expect
- the field's validation didn't account for filling in a field and then manually clearing it. A newly-empty oembed string field always produced an error and blocked save.

### DIFF
--- a/modules/@apostrophecms/oembed-field/ui/apos/components/AposInputOembed.vue
+++ b/modules/@apostrophecms/oembed-field/ui/apos/components/AposInputOembed.vue
@@ -136,9 +136,6 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-  .apos-input--oembed {
-    margin-bottom: $spacing-double;
-  }
   .apos-input__embed {
     ::v-deep iframe {
       max-width: 100%;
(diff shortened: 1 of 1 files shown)


stuartromanek:
In a round-about way this is also fixing https://linear.app/apostrophecms/issue/PRO-1397/on-error-user-is-often-knocked-out-of-all-tabs-in-editor-modals

boutell (approved):

abea (changes requested):
I found some alignment issues testing. 

Screen Shot 2021-07-01 at 8 13 25 AM

It's also not clearing that warning when the field is empty. I had a real YouTube url that I backspaced over utnil it wasn't a real thing, then empty.

I added the oembed field back to the testbed for reviewing: https://github.com/apostrophecms/testbed/pull/3

abea:
That is with the field *not* required. Same error when required as well.

stuartromanek:
@abea you're def looking at the feature branch? My experience looks like this

https://user-images.githubusercontent.com/1889830/124139079-433bcd00-da55-11eb-8f5f-3cce0818b8b0.mov

I also did a test where i (painstakingly) backspaced the entire url and got the same error clear

stuartromanek:
The alignment issue persists though, i didn't try to address that, but can

stuartromanek:
@abea fix for alignment is in

abea:
You're right. Good after a fresh install.

abea (approved):

PR #3211, opened 2021-06-30
Release 2.220.2

boutell:


abea (changes requested):
Those should all be 120. That was a typo fixed last week or so. The previous version is 119.

abea:
Nevermind. Ridic. Can we put a note in there explaining the bonkers jump?

boutell:
I did do that.

abea (approved):

abea:
so you did.

PR #3214, opened 2021-07-01
3.0: fix field label typo

myovchev:
There is a field typo in the `user` module, reported by @kimsmouter 
My most significant yet apos contribution :)

I suppose it doesn't need a special CHANGELOG entry? If it needs indeed, let me know.


abea (approved):
Thanks, @myovchev

PR #3217, opened 2021-07-02
Removes a defunct jqtree method; tidies code

abea:
- `mapMongoIdToJqtreeId` isn't used anymore because we don't use jqTree anymore.
- An eslint fix in `AposDocsManager.vue` since the return isn't doing anything.
- Some line break stuff since 80 character line lengths are the meaning of life.

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
@@ -240,14 +240,12 @@ export default {
       } else {
         moduleName = this.moduleName;
       }
-      const doc = await apos.modal.execute(apos.modules[moduleName].components.editorModal, {
+
+      await apos.modal.execute(apos.modules[moduleName].components.editorModal, {
         moduleName,
         docId: piece && piece._id,
         filterValues: this.filterValues
       });
-      if (!doc) {
-        // Cancel clicked
-      }
     },
     async finishSaved() {
       await this.getPieces();
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
@@ -247,7 +247,6 @@ export default {
       });
       if (!doc) {
         // Cancel clicked
-        return;
       }
     },
     async finishSaved() {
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -247,7 +247,6 @@ export default {
       });
       if (!doc) {
         // Cancel clicked
-        return;
```
boutell, line 250: It looks to me like assigning to doc and having this if statement at all is pointless now, no? I imagine at one point there must have been more logic in the method for the case where doc exists, but there is none now. Probably because we do whatever it was in an event listener instead.

modules/@apostrophecms/video-widget/ui/src/index.js
```diff
@@ -57,7 +57,8 @@ export default () => {
           if (result.width && result.height) {
             inner.style.width = '100%';
             resizeVideo(inner);
-            // If we need to iniitally size the video, also resize it on window resize.
+            // If we need to initially size the video, also resize it on window
+            // resize.
```
boutell, line 61: I got so much grief for sticking to 80 column line lengths in our PHP days. I So I hate to give you grief for sticking to 80 column line lengths, and I'm going to happily accept this PR, but if we find it requires bigger changes to other files you touch, maybe we should rethink the requirement or do it all in one swell foop.

boutell (changes requested):
Question about the if statement and assignment and whether they still matter at all.

modules/@apostrophecms/video-widget/ui/src/index.js
abea: One weekend I tried to work out a utility to take blocks of text and return them at 80 characters lines. Punctuation and indentation complicates things. So I approach this a bit at a time.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
abea: Ah, you are correct. I'll excise more.

abea: Yeah it used to continue to relationship stuff https://github.com/apostrophecms/apostrophe/blame/c72f42b0ea209de5400ea0b405b8a8e841a3c9ea/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue#L345-L352

boutell (approved):

PR #3219, opened 2021-07-02
start of moving legacy migrations, is this approach OK to follow from here

boutell:


modules/@apostrophecms/attachment/index.js
```diff
@@ -1100,7 +1070,8 @@ module.exports = {
           }
           return bulk.execute();
         }
-      }
+      },
+      ...require('./lib/legacy-migrations')(self)
```
boutell, line 1074: The migrations are methods, and there are other methods, so we use the spread syntax. Short of adding a new module format section just for migrations this seems like the way to go. `self` is passed in so we have access to the same scope as before.

boutell:
Marked as draft because it's not finished, but I'm looking for approval on this PoC before I do the rest.

modules/@apostrophecms/attachment/index.js
abea: Sure, this looks good.

abea: Sure, this looks good. Thanks.

modules/@apostrophecms/doc/lib/legacy-migrations.js
```diff
@@ -0,0 +1,181 @@
...
+          self.apos.area.walk(doc, (area, dotPath) => {
+            if (!area.items) {
+              return;
+            }
+            for (let i = 0; (i < area.items.length); i++) {
+              const item = area.items[i];
+              if (!item) {
+                continue;
+              }
+              if ((!item._id) || _.has(widgetIds, item._id)) {
```
abea, line 40: This seems like an easy replacement to get rid of the lodash import.

abea (changes requested):
One bit where we can get rid of the lodash import pretty easily. Also needs a changelog entry.

modules/@apostrophecms/doc/lib/legacy-migrations.js
boutell: I added a changelog entry.

`_.has` is omitted from most "you don't need lodash anymore" articles because it's actually still a bit goofy to replace.  I guess `Object.keys(foo).includes('bar')` would do it, but my goal in this work was to relocate this code with no risk of functional changes at minimum cost and the simplest way to ensure that is not to make unnecessary changes to the logic, I think.

abea: I'm speaking of this particular case. It's only one level deep, so it could be replaced with `widgetIds && widgetIds[item._id]`

abea: I'm speaking of this particular case. It's only one level deep, so it could be replaced with `widgetIds && widgetIds[item._id]`. Am I missing something there?

abea (changes requested):
I added a comment, Github. Get out of my todos.

modules/@apostrophecms/doc/lib/legacy-migrations.js
boutell: I removed it. We made widgetIds so we don't have to test whether it exists. The ids came from Apostrophe so we don't have to worry about a sneaky one called `hasOwnProperty`, therefore we can do the lazy test with `widgetIds[item._id]`.

abea (approved):

PR #3230, opened 2021-07-12
PRO-1729 PRO-1708 PRO-1695 PRO-1706 PRO-1727

boutell:
We have static localization on the browser side.

Some phrases are still missing and plurals don't seem to pluralize. I opened separate followup tickets in this cycle for those issues to keep the PRs moving along.


modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -3,31 +3,31 @@
     <span class="apos-notification__indicator">
       <AposIndicator
         :icon="iconComponent" class="apos-notification__indicator__icon"
-        :icon-size="icon ? 16 : 12"
+        :icon-size="notification.icon ? 16 : 12"
```
boutell, line 6: The forest of props was becoming unmaintainable and mirrored the notification object 1-1, so I changed to a single notification prop.

modules/@apostrophecms/notification/ui/apos/components/TheAposNotifications.vue
```diff
@@ -29,31 +25,7 @@ export default {
     };
   },
   async mounted() {
-    apos.notify = async function(message, options) {
```
boutell, line 32: We had support for %s placeholders, but haven't documented it in the 3.x era, and i18next approaches interpolation in a different way. Decided to gut this, simplifying the code. Worst case a few people with custom 2.x style notifications using interpolation (which is pretty rare) will see some %s placeholders and switch them over to `{{ }}`. I think this early in 3.x we can do this and avoid a lot of redundant code that would require maintenance.

boutell: apos.notify, apos.confirm, and apos.alert all support an "interpolate" option which may contain an object of key-value pairs to be inserted using `{{ key }}`, using `$t`. They also support a `localize: false` option for situations where the message has already been localized, or is in some cases, in which case `$t` is used directly for some of the inputs.

While i18next's `$t` handles interpolation keys just by chucking them in the same options object with options that have a special impact on the function, and I can do nothing about that, I didn't want to continue that madness into our own APIs in an unmaintainable way.

modules/@apostrophecms/ui/ui/apos/components/AposCellType.vue
```diff
@@ -15,7 +15,7 @@ export default {
   mixins: [ AposCellMixin ],
   methods: {
     label(typeName) {
-      return (apos.modules[typeName] && apos.modules[typeName].label) || typeName;
+      return this.$t((apos.modules[typeName] && apos.modules[typeName].label) || typeName);
```
abea, line 18: this seems inconsistent with the more common pattern of localizing in the template block. the button, label, and context menu item components here use that, for a few.

modules/@apostrophecms/ui/ui/apos/mixins/AposArchiveMixin.js
```diff
@@ -30,55 +30,65 @@ export default {
...
         }
 
-        description += '.';
-
         // Confirm archiving
         const confirm = await apos.confirm({
-          heading: `Archive ${plainType}`,
-          description,
-          affirmativeLabel: `Yes, archive ${plainType}`,
+          heading: this.$t('apostrophe:archiveType'),
```
abea, line 74: This needs the interpolation object.

test/templates.js
```diff
@@ -182,9 +182,6 @@ describe('Templates', function() {
   it('should render fragments containing async components correctly', async () => {
     const req = apos.task.getReq();
     const result = await apos.modules['fragment-page'].renderPage(req, 'page');
-    if (result.match(/error/)) {
```
abea, line 185: Why remove all these?

abea (changes requested):

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerEditor.vue
```diff
@@ -239,8 +239,8 @@ export default {
     },
     async archive() {
       if (!await apos.confirm({
-        heading: 'Are You Sure?',
-        description: 'This will move the image to the archive.'
+        heading: 'apostrophe:areYouSure',
```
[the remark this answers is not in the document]
abea: Nevermind this for now.

test/templates.js
boutell: Now that we are pushing phrases to the browser, there are far too many with the word "error" in them present as data for a simplistic test like this to work. But the test is also unnecessary because none of the expected outputs will be present if the page is a 500 or 404 error.

boutell: (See the rest of the asserts — they are sufficient.)

boutell:
I pushed up a commit to localize the `AposCellType` label in the template as requested, and provided a response on the query about the "error" test.

modules/@apostrophecms/ui/ui/apos/mixins/AposArchiveMixin.js
boutell: Fixed.

abea (approved):

PR #3232, opened 2021-07-13
PRO-1753 i18n debugging tools, more complete settings so the debugger does not complain

boutell:
…  so we can distinguish what went through it from what did not go through at all

Also fixed one unrelated missing namespace.


modules/@apostrophecms/i18n/index.js
```diff
@@ -9,6 +9,16 @@ const i18next = require('i18next');
 const i18nextHttpMiddleware = require('i18next-http-middleware');
 const fs = require('fs');
 
+const apostropheI18nDebugPlugin = {
+  type: 'postProcessor',
+  name: 'apostrophei18nDebugPlugin',
+  process(value, key, options, translator) {
+    // For ease of tracking down which phrases were
+    // actually passed through i18next
+    return `* ${value}`;
```
abea, line 18: This is marking strings that _have_ been localized? The idea being that one would look through the UI and find anything without the marker, look for the code it's coming from and add the `t` there? That sound helpful. I'm wondering we should separate i18next debugging and this since i18next debugging is its own thing. `APOS_SHOW_L10N`?

And I'm always for making things more obvious if they should be obvious. Maybe the `🌍` for this instead of an asterisk.

abea (changes requested):

modules/@apostrophecms/i18n/index.js
abea: Also needs a changelog update.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -592,7 +592,7 @@ export default {
     },
     async redo() {
       this.patchesSinceLoaded.push(this.undone.pop());
-      await this.refreshAfterHistoryChange('redoFailed');
+      await this.refreshAfterHistoryChange('apostrophe:redoFailed');
```
boutell, line 595: Unrelated missing namespace fix

CHANGELOG.md
```diff
@@ -6,8 +6,10 @@
 
 * Removes an unused method, `mapMongoIdToJqtreeId`, that was used in A2 but is no longer relevant.
 * Removes deprecated and non-functional steps from the `edit` method in the `AposDocsManager.vue` component.
-* The `apostrophe-i18n` module now uses `i18next` to implement static localization.
-* The `i18next-http-middleware` locale guesser is installed.
+* The `@apostrophecms/i18n` module now uses `i18next` to implement static localization. All phrases in the Vue-based admin UI are passed through `i18next` via `this.$t`, and `i18next` is also available via `req.t()` in routes and `__t()` in templates. Apostrophe's own admin UI phrases are in the `apostrophe` namespace for a clean separation. An array of locale codes, such as `en` or `fr` or `en-au`, can be specified using the `locales` option to the `@apostrophecms/i18n` module. The first locale is the default, unless the `defaultLocale` option is set. If no locales are set, the locale defaults to `en`. The `i18next-http-middleware` locale guesser is installed and will select an available locale if possible, otherwise it will fall back to the default.
```
abea, line 9: This is nested in today's release changelog section. Should be in a new Unreleased section, probably.

abea (changes requested):
bookkeeping bit, then good.

boutell:
Merged latest main and resolved the changelog ish.

abea (approved):

PR #3240, opened 2021-07-15
PRO-1773: support for localizing tooltips, plus some new translations

boutell:
I was going to just localize stuff before calling `v-tooltip`, but realized this did not make sense. It would be very fiddly logic to impose in each of those places.

So instead I applied the decorator pattern to `v-tooltip`, resulting in a directive that natively supports i18next.

Right now it does not have `interpolate` and `localize: false`, but these can easily be added if the use cases turn out to include them when all tooltips have been converted: call `v-tooltip` using its object syntax, in which the tooltip key is in the `content` property, and add support for `interpolate` and `localize` properties of that object in the new wrapper.


modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
```diff
@@ -33,7 +33,7 @@
         />
         <slot name="label">
           <span class="apos-button__label" :class="{ 'apos-sr-only' : (iconOnly || type === 'color') }">
-            {{ $t(label) }}
+            {{ $t(label, interpolate) }}
```
boutell, line 36: I also introduced interpolate support for AposButton as that is sometimes needed.

modules/@apostrophecms/ui/ui/apos/lib/vue.js
```diff
@@ -1,11 +1,10 @@
 import Vue from 'vue';
 import VueClickOutsideElement from 'vue-click-outside-element';
-import { VTooltip } from 'v-tooltip';
-import tooltipConfig from './tooltip';
+import LocalizedVTooltip from './localized-v-tooltip';
```
abea, line 3: Let's go with `apos-v-tooltip` in case we ever want to customize anything else.

abea (changes requested):
Please add a changelog entry.

@agilbert We should probably account for this in QA tests.

boutell:
Made those changes. It's `v-apos-tooltip` because the `v-` prefix is imposed by Vue.

abea (approved):

PR #3243, opened 2021-07-16
adds clear color ui to color picker, subs dep in for one that accepts…

stuartromanek:
… null vals

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "v-tooltip": "^2.0.3",
     "vue": "^2.6.14",
     "vue-click-outside-element": "^1.0.13",
-    "vue-color": "^2.8.1",
+    "vue-color": "https://github.com/apostrophecms/vue-color.git#allow-initial-null-vals",
     "vue-loader": "^15.9.6",
     "vue-material-design-icons": "^4.9.0",
     "vue-style-loader": "^4.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -106,7 +106,7 @@
     "v-tooltip": "^2.0.3",
     "vue": "^2.6.14",
     "vue-click-outside-element": "^1.0.13",
-    "vue-color": "^2.8.1",
+    "vue-color": "https://github.com/apostrophecms/vue-color.git#allow-initial-null-vals",
```
boutell, line 109: All dependencies, need to be published to npm before this is actually released. We can't have github dependencies in package.json of a shipping apostrophe release. That's because some of our clients don't allow use of git for module fetches inside their CI environment, as it's one more thing to keep secure and so on.

This concern can be cleared by publishing your version of vue-color as `vue-color-nullable` or similar, feel free to have a README that says it's hopefully only temporary and points to the upstream if you think best.

boutell (changes requested):

package.json
abea: Maybe `@apostrophecms/vue-color`? Definitely say it's temporary as possible.

abea (changes requested):
what tom said.

package.json
boutell: `@apostrophecms/vue-color` is better, yes. The README just needs to say it's intended as a temporary fork until the maintainers resume accepting fixes.

abea (approved):

boutell (approved):

PR #3244, opened 2021-07-16
Removes singleton tag refs

abea:
More cleanup of the singleton area from A2.


boutell (approved):
Removing the shortcut helper capability is an API change, but since the docs say it's reserved for core modules I think we can get away with it.

abea:
A lesson to remove or replace all code comments before a major version launches. And check all methods for removal as well.

PR #3263, opened 2021-07-29
PRO-1845 locale switcher roughed in, needed for testing of other things

boutell:
This is a total work in progress right now, getting it on our radar that a branch exists.


boutell:
See the `static-i18n-demo` branch of `a3-demo` for a live example where `fr-CA` has a translation of the "Global Content" admin bar button. Yes you'll be on a 404 page when you switch locales. Still see the button!

gregvanbrug (approved):

PR #3267, opened 2021-07-31
PRO-1790 accomodations for inline micro range

stuartromanek:
sister PR in `@apostrophecms-pro/palette` completes the lewk
https://github.com/apostrophecms/palette/pull/26


abea (approved):

abea (changes requested):
Just needs a changelog update.

abea (approved):

PR #3269, opened 2021-08-02
PRO-1865: working, I'll open up a followup ticket about published mode

boutell:


boutell:
A draft until I get the tests in.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -243,6 +243,27 @@ module.exports = {
           }
           return self.publish(req, draft);
         },
+        ':_id/localize': async (req) => {
```
gregvanbrug, line 246: Should this be rolled into a utility localize api route function? Seems like a copy of what's done in the `page` api route:  https://github.com/apostrophecms/apostrophe/pull/3269/files#diff-5e6c997e317e3d9d022272dd5b800a9ddd9965d767280831bd1148f202c1dc1dR343.

gregvanbrug (changes requested):

modules/@apostrophecms/piece-type/index.js
boutell: Yes this is an existing pattern for a few other methods as well, they are similar or identical but my feeling is that they may diverge and I don't want to write an awkward meta-method that takes the module as an argument to avoid duplicating these and then just wind up changing one of them for whatever reason anyway.

gregvanbrug (approved):

PR #3274, opened 2021-08-04
PRO-1870: /locales route for discovery of localized versions of a document

boutell:
Includes hilariously crude frontend rough-in demo on top of the existing rough-in, which I'm sure you'll gut Greg, but it's there so you can see the plumbing hooked up.


boutell:
Not to be merged until after PRO-1865 (review that first). This builds on it.

gregvanbrug (approved):

PR #3276, opened 2021-08-05
cut copy paste refinement

stuartromanek:


stuartromanek:
@boutell not sure why it itemized all the merges from `main` .. doing this in `3/cut-copy-paste` squashed them all as one merge ..

modules/@apostrophecms/area/ui/apos/components/AposAreaMenu.vue
```diff
@@ -12,7 +12,8 @@
       <ul class="apos-area-menu__wrapper">
         <li
           class="apos-area-menu__item"
-          v-for="(item, itemIndex) in myMenu" :key="item.label"
+          v-for="(item, itemIndex) in myMenu"
+          :key="(item.type + '_' || '') + item.label"
```
boutell, line 16: I don't think this can ever be falsy due to the `+` with `_`, so I don't see how the empty string case would ever happen?

boutell (changes requested):

boutell (approved):

PR #3280, opened 2021-08-06
styles and filtering for the locale switcher

gregvanbrug:
Resloves: 
- https://linear.app/apostrophecms/issue/PRO-1862/update-the-locale-switcher-to-match-the-design

Locale Switcher UI and filtering. Needs integration with API to complete https://linear.app/apostrophecms/issue/PRO-1845/as-an-editor-i-can-toggle-between-viewing-and-editing-different


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBarLocale.vue
```diff
@@ -41,11 +86,30 @@ export default {
...
+        );
+      });
     }
   },
   methods: {
+    isActive(locale) {
+      return window.apos.i18n.locale === locale.name;
+    },
+    isLocalized(locale) {
+      // TODO: Not sure we have the proper data to make this work properly yet.
```
boutell, line 103: We do. Please see this PR from last week:

https://github.com/apostrophecms/apostrophe/pull/3274

I reckon it shouldn't be tough to integrate that with what you're doing.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBarLocale.vue
```diff
@@ -41,11 +86,30 @@ export default {
         modifiers: [ 'icon-right', 'no-motion' ],
         type: 'quiet'
       };
+    },
+    filteredLocales(input) {
+      return this.locales.filter(locale => {
+        return (
+          locale.label.toLowerCase().indexOf(this.search.toLowerCase()) > -1
```
boutell, line 93: Oh cool, search was going to be pushed back til later, but it never hurts to get ahead.

ES2015 note: could use `.includes` in place of `indexOf` and the `-1` check,  but that's just a nitpick.

boutell: More importantly though, please check the locale name as well as the locale label, in case they don't put the locale's official name in the label but the user searches for it.

boutell (changes requested):

boutell (approved):

PR #3283, opened 2021-08-09
PRO-1754 PRO-1857: determine locale via prefix and domain, with tests and necessary surrounding fixes

boutell:


boutell:
Hitting some complications relating to the consequences of changing the hostname in the URL. Buckle yer somethins.

boutell:
Completes PRO-1857 as a fringe benefit because it eliminates the concern. However I am opening a separate ticket about the need to transfer logins across hostnames when explicitly losing the locale switcher, a stunt we could pull off in 2.x.

abea:
This looks very similar to #3287

boutell:
There's a comment on #3287 saying it shouldn't be reviewed and merged until after this one, which it unavoidably is built upon.

boutell:
Fine with me that you reviewed that one first, but please review this one and we'll sort that first before trying to land all the concerns on the other.

PR #3285, opened 2021-08-10
PRO-1863 call i18n one thing in configuration

boutell:
The per-module option is i18n, the per-module subdirectory is i18n, one thing to remember when coding per discussion.


abea:
We have some failing tests here.

boutell:
Pushed update

abea (approved):

PR #3287, opened 2021-08-10
PRO-1754 PRO-1857 PRO-1868 parked properties need to work in all locales, determine locale via prefix and domain, various fixes

boutell:
This PR should not be reviewed and approved until PRO-1754 is approved.


modules/@apostrophecms/i18n/index.js
```diff
@@ -227,6 +324,15 @@ module.exports = {
           self.getComponentName('localizeModal', 'AposI18nLocalize'),
           { moduleName: self.__meta.name }
         );
+      },
+      setPrefixUrls(req) {
+        // For bc, req.baseUrl is always set, to a best guess if baseUrl is not configured.
+        // When falling back, req.hostname is used if trustProxy is active, otherwise the
+        // Host header to allow port numbers in dev
+        req.baseUrl = self.apos.page.getBaseUrl(req) || (req.protocol + '://' + (self.options.trustProxy && req.get('X-Forwarded-Host')) ? req.hostname : req.get('Host'));
```
abea, line 332: I'd prefer you get the hostname as a step before to make this easier to parse. Then it could be cleanly written with a template literal.

modules/@apostrophecms/page-type/index.js
```diff
@@ -323,7 +323,14 @@ module.exports = {
           return self.apos.page.insert(_req, doc.aposLastTargetId.replace(':draft', ':published'), doc.aposLastPosition, published, options);
         } else if (!doc.level) {
           // Insert the home page
-          return self.apos.doc.db.insertOne(_req, published, options);
+          Object.assign(published, {
+            path: doc.path,
```
abea, line 327: Why this change?

abea (changes requested):

modules/@apostrophecms/express/index.js
```diff
@@ -630,10 +630,6 @@ module.exports = {
       // set the `baseUrl` option for Apostrophe.
 
       addAbsoluteUrlsToReq(req) {
```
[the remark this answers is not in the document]
boutell: This internal API has been made obsolete by the new `req.prefix` feature, to which its logic was moved. I've removed it properly.

modules/@apostrophecms/page-type/index.js
boutell: `published` at this point only has schema properties. If we don't bring over this information the home page will not replicate properly. It's different for non-home pages because we can use `page.insert` for them.

modules/@apostrophecms/i18n/index.js
```diff
@@ -329,7 +329,8 @@ module.exports = {
         // For bc, req.baseUrl is always set, to a best guess if baseUrl is not configured.
         // When falling back, req.hostname is used if trustProxy is active, otherwise the
         // Host header to allow port numbers in dev
-        req.baseUrl = self.apos.page.getBaseUrl(req) || (req.protocol + '://' + (self.options.trustProxy && req.get('X-Forwarded-Host')) ? req.hostname : req.get('Host'));
+        const host = (self.options.trustProxy && req.get('X-Forwarded-Host')) ? req.hostname : req.get('Host');
```
abea, line 332: Blessed.

abea (changes requested):

PR #3288, opened 2021-08-10
PRO-1877 fix duplicte bug, clear clipboard on paste

stuartromanek:
## Bug
- There were conflicting versions of the method `regenerateIds` which was causing a bug in both Duplicate and clipboard paste.

## Improvement
- In demo prep and testing with Alex we decided to clear the clipboard contents on paste, as it felt strange to have the very specific duplication follow you around forever in areas.
- Hide Copy button for now. Between Duplicate and Cut/Paste, Copy felt like a confusing use case. Something to revisit, potentially with some improved widget control bar UI

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
@@ -286,6 +286,7 @@ export default {
     },
     async copy(i) {
       apos.area.widgetClipboard.set(this.next[i]);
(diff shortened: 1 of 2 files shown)


boutell (approved):

boutell:
I don't agree with dropping "copy" yet keeping "duplicate," personally I think it throws off whatever ease of adoption we're getting from following the well-known "cut/copy/paste" triad, but that's not my area of expertise. Excited to see this coming to life

PR #3292, opened 2021-08-10
PRO-1889: ported support for taking sessions with you when you switch locales that have different hostnames

boutell:
This was ported closely from apostrophe-workflow, saving us from reinventing a lot of hard wins.


boutell:
This is a draft until the main PR lands.

boutell:
This is reviewable now.

modules/@apostrophecms/i18n/ui/apos/apps/AposI18nCrossDomainSession.js
```diff
@@ -0,0 +1,23 @@
+export default () => {
```
abea, line 1: This looks like normal `ui/src/*.js` client side JS, but it's in an `ui/apos/apps` directory. I'm guessing that's to keep it in the apos scene. Don't we have `ui/apos/js` or something like that?

abea (changes requested):

modules/@apostrophecms/express/index.js
```diff
@@ -225,6 +225,12 @@ module.exports = {
         url: '/api/v1',
         middleware: cors()
       },
+      createRawRedirect(req, res, next) {
+        // We apply the super pattern variously to res.redirect,
+        // make sure the original version is always available
+        res.rawRedirect = res.redirect;
```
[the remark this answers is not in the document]
boutell: Yes. If we tried to use `res.redirect` to `someDoc._url` on a site with no baseUrl, it would fail due to the prefixing. So we have always offered `rawRedirect` for such cases. At this very moment we are not using it in the core because I decided to handle locale switch URL changes in browser land javascript, but it wouldn't make sense not to offer this.

modules/@apostrophecms/i18n/ui/apos/apps/AposI18nCrossDomainSession.js
boutell: Apostrophe loads admin UI "apps" from `ui/apos/apps`, as opposed to Vue components which are loaded and registered by name from `ui/apos/components`. Apps are automatically invoked, while component .vue files are just components for use by apps and other components.

So admin UI "apps" work similarly to `ui/src/index.js` files but are not loaded for logged-out users. Not a new thing in this PR.

abea: I realize it's not a new thing. It seems like we might have had a blind spot, however. This isn't an app like the Vue apps are, but the file path suggests that it would be. If we don't have a place for logged-in client JS that's not instantiating a Vue app that seems to be a gap.

modules/@apostrophecms/express/index.js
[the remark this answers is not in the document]
abea: It sounds like this is a loose thread for now, then. We should have a ticket to complete the loop.

abea (approved):
This code update is fine. There are two issues it raises that I can create tickets for:
- res.rawRedirect is a dangling feature that we should track else it might be forgotten in the imaginary Boutell bus incident.
- If `/ui/apos/apps` is our only place for logged-in-only client JS then we should give some thought to that. It works, but could be misleading for developers new to the system.

boutell:
It's a documentation gap for sure — we need to document the logged-in javascript world — but there's no functional gap re: the apps. They don't have to be Vue apps; they just frequently are.

PR #3293, opened 2021-08-11
busy screen

gregvanbrug:
Resolves:
- https://linear.app/apostrophecms/issue/PRO-1849

Busy state UI using current components. New ticket added for updating the spinner to the new visual.


gregvanbrug:
Moved the busy state into `AposModal` with the ability to trigger the screen with a `busy` property and the ability to update the text with a `busyTitle` property in the `modal` object.

boutell (changes requested):
The last step here is to integrate it with `TheAposBusy` and `TheAposModals` so that the standard busy mechanism, which you get "for free" with no bookkeeping just by passing the `busy: true` option to an `apos.http.*` call and awaiting it, recognizes when the topmost modal offers a `busyTitle` and engages that rather than the standard global busy display. Then `submit` will not have to set and clear the `busy` property. It'll just work.

boutell (approved):

boutell:
@gregvanbrug planning to merge?

PR #3294, opened 2021-08-11
req.clone is available, and used where mission critical, for localeReq. A separate ticket is open to use req.clone consistently

boutell:
Cloning `req` naively yields an object with no `get` method, which is an issue for code that calls `req.get`. Opening separate ticket to replace *all* naive `{ ...req, ...}` clones of req with calls to this method.

### DIFF
--- a/modules/@apostrophecms/util/index.js
+++ b/modules/@apostrophecms/util/index.js
@@ -776,6 +776,19 @@ module.exports = {
         const result = await fn();
         req.aposStack.pop();
         return result;
+      },
+      // Returns a new `req` object with the properties of the original plus any
+      // in the optional `properties` parameter. Used when a request object
+      // with one change is desired, such as `mode: 'published'`. Avoids
+      // the need to push and pop properties of the original `req`.
+      // Also available as `req.clone(properties = {})`.
+      cloneReq(req, properties = {}) {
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/util/index.js
```diff
@@ -776,6 +776,24 @@ module.exports = {
...
+      // with one change is desired, such as `mode: 'published'`. Avoids
+      // the need to push and pop properties of the original `req`.
+      // Also available as `req.clone(properties = {})`.
+      cloneReq(req, properties = {}) {
+        // Express and Node.js offer no official constructor for a req object.
+        // But a plain object "clone" only takes us so far because we need
+        // access to things like `req.get`. Solution: ask `Object.create` to
+        // create a new object with the same prototype as `req`, then copy the
+        // own properties of `req` into the new object.
+        const result = Object.assign(Object.create(Object.getPrototypeOf(req)), req, properties);
```
abea, line 791: Yo dawg. I hear you like `Object`.

abea (approved):

modules/@apostrophecms/util/index.js
boutell: I certainly have no object-ion.

PR #3301, opened 2021-08-16
PRO-1906: localization links for the public

boutell:
See the `static-i18n-demo` branch of `a3-demo` to try this out.

Assigning to Bea because it touches on stuff I think he's well qualified to give feedback on.


modules/@apostrophecms/i18n/index.js
```diff
@@ -407,6 +459,30 @@ module.exports = {
         req.baseUrlWithPrefix = `${self.apos.page.getBaseUrl(req)}${self.apos.prefix}`;
         req.absoluteUrl = req.baseUrlWithPrefix + req.url;
         req.prefix = `${req.baseUrlWithPrefix}${self.locales[req.locale].prefix || ''}`;
+      },
+      // Returns an Express route suitable for use in a module
+      // like a piece type or the page module. The returned route will
+      // expect req.params._id and req.params.toLocale and redirect,
+      // if possible, to the corresponding version in toLocale.
+      toLocaleRouteFactory(module) {
```
boutell, line 467: Trying out the factory pattern as a response to the problem of many routes being duplicated between piece-type and the page module, which don't and shouldn't share a lineage but do have some cases of routes that aren't going to differ between them.

modules/@apostrophecms/login/index.js
```diff
@@ -91,12 +91,17 @@ module.exports = {
     return {
       get: {
         // Login page
-        [self.login()]: async (req) => {
+        [self.login()]: async (req, res) => {
```
boutell, line 94: Unrelated cleanup to the idiom.

modules/@apostrophecms/login/index.js
```diff
@@ -91,12 +91,17 @@ module.exports = {
...
           if (req.user) {
-            return req.res.redirect('/');
+            return res.redirect('/');
           }
           req.scene = 'apos';
-          await self.sendPage(req, 'login', {});
+          try {
+            await self.sendPage(req, 'login', {});
+          } catch (e) {
+            self.apos.util.error(e);
```
boutell, line 102: Unrelated: noticed this was a request leak before. Error handling could be better, but this is an incremental improvement over leaking requests.

modules/@apostrophecms/page/index.js
```diff
@@ -1391,10 +1401,13 @@ database.`);
           await self.emit('serve', req);
           await self.serveNotFound(req);
         } catch (err) {
-          await self.serveDeliver(req, err);
-          return;
+          return await self.serve500Error(req, err);
```
boutell, line 1404: I ran into this while trying to debug my work on `data.localizations`: we were running 500 error pages through the full stack, including sending out `beforeSend` events, which was not our intent because it is too likely to produce the very same error. The actual templateError template is designed to be self-sufficient, so I altered the logic to take advantage of that.

modules/@apostrophecms/template/index.js
```diff
@@ -666,25 +666,21 @@ module.exports = {
...
           // occurred for easier debugging
-          return error(e, 'template');
+          return error(e);
         }
 
         return content;
 
-        function error(e, type) {
-          let now = Date.now();
-          now = dayjs(now).format('YYYY-MM-DDTHH:mm:ssZZ');
```
boutell, line 684: Factored this out to a method so it can also be used to log 500 errors with other causes.

test/content-i18n.js
```diff
@@ -164,14 +167,12 @@ describe('Locales', function() {
...
 
-    // Confirm login, seems necessary for the session cookie in the jar to work
-    // on the next call
     const page = await apos.http.get('/', {
       jar
     });
 
     assert(page.match(/logged in/));
-
+    assert(page.includes(`<a href="http://localhost:3000/api/v1/@apostrophecms/page/${home._id}/locale/en-CA">Canadian English (en-CA)</a>`));
```
boutell, line 175: Good localization link

abea (changes requested):

modules/@apostrophecms/i18n/index.js
```diff
@@ -62,6 +62,59 @@ module.exports = {
...
+          req.data.localizations = [];
+          for (const name of Object.keys(self.locales)) {
+            const localeReq = self.apos.util.cloneReq(req, {
+              locale: name
+            });
+            self.setPrefixUrls(localeReq);
+            const doc = localizations.find(doc => doc.aposLocale.split(':')[0] === name);
+            if (doc && self.apos.permission.can(req, 'view', doc)) {
+              doc.available = true;
+              doc._url = `${req.prefix}${manager.action}/${context._id}/locale/${name}`;
```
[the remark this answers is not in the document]
boutell: With 90 locales, like some of our enterprise clients, and the need to potentially drag querying for locale-specific piece pages into the calculation, this query would bring the site to its knees. Which is why we also defer the calculation until needed this way in A2. The links we're talking about will more typically be in a dropdown menu, possibly not even hoverable links in the usual way.

We could use slugs rather than ids in this API, although Apostrophe document _ids are not in any way meant to be used as secrets (they are often publicly exposed in headless APIs).

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
...
 * In the admin UI, `v-tooltip` has been extended as `v-apos-tooltip`, which passes phrases through `i18next`.
+* Developers can link to alternate locales by iterating over `data.localizations` in any page template. Each element always has `locale`, `label` and `homePageUrl` properties. Each element also has an `available` property; if true, the current context document is available in that locale, `title` and a small number of other document properties are populated, and `_url` redirects to the context document in that locale. The current locale is marked with `current: true`.
```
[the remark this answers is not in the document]
boutell: Definitely. Addressed this and pushed.

abea (approved):

PR #3306, opened 2021-08-17
PRO-1911 adds color to base type var, all type styles extend this var to receieve benefits

stuartromanek:
this prevents unintended project-level style leaking for our type styles


abea (approved):

abea (changes requested):
Needs the changelog update

abea (dismissed):

abea (approved):

PR #3309, opened 2021-08-18
#1920 bc for old __ helpers

boutell:
Very basic fallback to just outputting the key given (which in 2.x practice is probably the English text) and warning the developer. These were never official in 3.x but from some recent correspondence it's clear they could be encountered, so let's not crash.

### DIFF
--- a/modules/@apostrophecms/template/lib/custom-tags/render.js
+++ b/modules/@apostrophecms/template/lib/custom-tags/render.js
@@ -1,4 +1,5 @@
 const util = require('util');
+
 module.exports = (self) => {
   // Create the render input object, used to parse the fragment source
   function createRenderInput(info) {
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/template/lib/custom-tags/render.js
```diff
@@ -100,7 +101,7 @@ module.exports = (self) => {
       const req = context.env.opts.req;
       const env = self.getEnv(req, context.env.opts.module);
       input.apos = self.templateApos;
-      input.__ = req.res.__;
```
boutell, line 103: No longer a thing anyway.

boutell:
I added invocations of these to the the `static-i18n-demo` branch of a3-demo.

abea (approved):

PR #3311, opened 2021-08-18
PRO-1903 locale switcher refreshes its knowledge when opened

boutell:


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBarLocale.vue
```diff
@@ -96,22 +97,22 @@ export default {
...
-      );
-      this.localized = Object.fromEntries(
-        docs.results
-          .filter(doc => doc.aposLocale.endsWith(':draft'))
-          .map(doc => [ doc.aposLocale.split(':')[0], doc ])
-      );
-    }
-  },
   methods: {
+    async open() {
```
boutell, line 101: Just moved this code.

modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -661,7 +661,7 @@ export default {
           }
         }
         // New documents only
-        newRelatedDocs = relatedDocs.filter(doc => unlocalizedIds.has(doc._id));
+        relatedDocs = relatedDocs.filter(doc => unlocalizedIds.has(doc._id));
```
boutell, line 664: Unrelated fix; this was producing console turds and breaking the filtering for the "new related documents only" case.

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
```diff
@@ -33,6 +33,10 @@ export default {
     // multiple layers of code, as a single `label`
     // property for instance
     Vue.prototype.$t = (phrase, options) => {
+      if (phrase == null) {
```
boutell, line 36: Unrelated fix: don't log console turds if $t is passed null, as can happen if a button has no label, for instance.

gregvanbrug, line 36: Should this check for strict equality? Maybe needs to check undefined and null?

gregvanbrug (changes requested):

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
gregvanbrug: Alright. I thought this was against our linter rules.

gregvanbrug (approved):

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
boutell: It's a common practice for a really useful test (nullishness), so the standard linter rules permit this specific one. Not == in general.

PR #3313, opened 2021-08-18
3.0: fix TypeError while parking missing parked page

myovchev:
This patch fixes a migrate task crash when new parked page is inserted.

## The problem
1. not yet existing in the DB page is available in the parked list
2. migrate task is executed
3. park sync logic is triggered on after:migrate (more than once)
4. the `implementParkOne` calls internally `insert`
5. the internal `insert` method is mutating the original parked configuration
6. on subsequent `implementParkOne` invocation, the script is crashing with error message `TypeError: Cannot read property 'slug' of undefined`

## The fix
The parked configuration is shallow cloned before processed in the internal `insert` method. No other logic is changed. 

A follow up from the PR against a stale branch #3308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v3 Unreleased 
+
+### Fixes
+* Fixes migration task crash when parking new page
+
 ## 3.2.0 - 2021-08-13
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


myovchev:
I'm still lost with the Changelog. There was no "unreleased changes" section.

boutell (approved):
Thanks!

boutell:
@myovchev this will be released in Wednesday's regular release. Thanks for the contribution!

PR #3317, opened 2021-08-19
3.0: [UI] fix AposCellDate typo

myovchev:
AposCellDate component should increment the month value when formatting because the value is zero-based.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## Unreleased
+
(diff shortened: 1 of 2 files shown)


boutell (approved):
Thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+### Fixes
```
abea, line 5: Where did this entry go?

PR #3321, opened 2021-08-19
PRO-1752 static i18n for the vast majority of remaining cases

boutell:
I have opened a followup ticket for the dimensions etc. in the media editor.

I will respond to any omissions pointed out (feel free to do so) by opening followup tickets, because this PR is already more than large enough. Of course regressions and bugs will be fixed before this PR is approved.

Developer QA: to test, use the `static-i18n-demo` branch of a3-demo.

To visualize which phrases have been through the i18n layer, use:

```
APOS_DEBUG_I18N=1 APOS_SHOW_I18N=1 npm run dev
```

**Project-level phrases are going to be shown as "holes" because the project level doesn't have static i18n configured for them. That is not a bug** in the localization of the admin UI.

You should see a 🌎 next to every static text phrase that comes from Apostrophe core. Sometimes two if there are nested phrases being built, that's OK.


modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -105,48 +142,75 @@
...
+  "insertAndReturn": "{{ saveLabel }} and return to the {{ typeLabel }} listing.",
+  "insertAndNew": "{{ saveLabel }} {{ typeLabel }} and create a new one.",
   "italic": "Italic",
   "lastEdited": "Last Edited",
   "leavePageDescription": "The content you're trying to edit belongs to another document and must be edited there.\nChanges made to {{ oldTitle }} are saved automatically.",
   "leavePageHeading": "Leave {{ oldTitle }} to edit {{ newTitle }}?",
   "link": "Link",
   "listJoiner": ", ",
   "live": "Live",
+  "lastUpdatedBy": "Last saved on {{ updatedAt }} <br /> by {{ updatedBy }}",
```
gregvanbrug, line 157: Should this have markup? Might cause issues in responsive layouts.

modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -188,29 +260,50 @@
...
+  "saveDraftAndCreateNew": "Save Draft and Create New",
+  "saveDraftAndPreview": "Save Draft and Preview",
+  "savingDocument": "Saving document...",
+  "searchDocType": "Search {{ type }}",
   "searchLocales": "Search Locales",
+  "searchLocalesPlaceholder": "Search locales...",
+  "select": "Select",
   "selectAll": "Select All",
   "selectContent": "Select Content",
+  "selectedMenuItem": "✓ {{ label }}",
```
gregvanbrug, line 280: Should the check really be part of this? Could be done with CSS.

modules/@apostrophecms/ui/ui/apos/components/AposCellDate.vue
```diff
@@ -22,24 +23,31 @@ export default {
...
+        const period = hour > 11 ? 'pm' : 'am';
+        if (hour > 12) {
+          hour = hour - 12;
+        } else if (hour === 0) {
+          hour = 12;
+        }
+        const minute = d.getMinutes();
 
-      return `${toTwoChars(month)}/${toTwoChars(date)}/${toTwoChars(year)} at ${hour}:${minute} ${period}`;
+        return `${toTwoChars(month)}/${toTwoChars(date)}/${toTwoChars(year)} at ${hour}:${minute} ${period}`;
```
gregvanbrug, line 47: Couldn't most of this be replaced with:
```
const d = new Date(value);
if (Number.isNaN(d.getDate())) {
  return value;
}
dayjs(d).format('MM/DD/YY at h:mm a');

```

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
```diff
@@ -31,16 +31,43 @@ export default {
...
+      const result = i18next.t(phrase, {
         lng: i18n.locale,
         ...options
       });
+      if (i18n.show) {
+        if (result === phrase) {
+          if (phrase.match(/^\S+:/)) {
+            return `❌ ${result}`;
+          } else {
+            return `🕳 ${result}`;
```
gregvanbrug, line 63: Kinda wonder if this should be something else. It's a little confusing in the UI.

gregvanbrug (changes requested):

modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -48,25 +68,39 @@
...
+  "editDocType": "Edit {{ type }}",
   "editingInAnotherTab": "Editing in Another Tab or Window",
   "editor": "Editor",
+  "editRelationshipFor": "Edit Relationship for {{ title }}",
+  "editWidget": "Edit Widget",
   "email": "Email",
+  "emptyRichTextWidget": "Empty Rich Text Widget",
   "error": "An error occurred",
+  "errorCount": "{{ count }} error remaining",
+  "errorCount_plural": "{{ count }} errors remaining",
```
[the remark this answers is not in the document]
boutell: I would, but it's specific magic that i18next looks for when `count` is not equal to `1`.

modules/@apostrophecms/i18n/i18n/en.json
boutell: Feel free to open a ticket on that, but it's not new to this PR (it's just been migrated here so it can be localized).

modules/@apostrophecms/i18n/i18n/en.json
boutell: Changing this is not part of the PR, it's just being made localizable. However I would note that in a right-to-left language it might be good that we made it part of the string so it can be repositioned when localizing the text.

modules/@apostrophecms/ui/ui/apos/components/AposCellDate.vue
boutell: Probably, but again not new in the PR and I don't want to introduce work or potential bugs this close to shipping localiation, so I'm going to suggest you open a followup ticket for that.

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
boutell: Yes, I'm open to suggestions! Only seen when an environment variable is set for debugging though.

modules/@apostrophecms/i18n/i18n/en.json
gregvanbrug: Probably could still be handled in CSS for rtl languages but not sure if we have support for that in core yet. There's probably much more in the UI we'll need to adjust for true rtl support.

gregvanbrug (approved):
Created tickets for the date formatting and the markup in translated strings. Rest is good to go.

PR #3322, opened 2021-08-20
corrected changelog entry for Miro

boutell:
I botched the changelog entry in the merge, here's a followup to fix that.


CHANGELOG.md
```diff
@@ -2,15 +2,14 @@
 
 ## Unreleased
 
-### Fixes
-
 ### Changes
 
 * Bolsters the CSS that backs Apostrophe UI's typography to help prevent unintended style leaks at project-level code.
 
 ### Fixes
```
abea, line 9: I think fixes is typically first. I also made a PR for this: #3323

abea, line 9: We've been putting Fixes first, I think to highlight those. Maybe we don't need to do that?

abea (approved):

PR #3326, opened 2021-08-20
Lints locale names to start with alpha character

abea:
Lints locale names to start with an alphabetic character. This is primarily to further ensure that the `Object.entries` loops maintain order. Apparently numbers as object keys can take precedence in some environments. https://2ality.com/2015/10/property-traversal-order-es6.html

### DIFF
--- a/modules/@apostrophecms/i18n/index.js
+++ b/modules/@apostrophecms/i18n/index.js
@@ -35,6 +35,12 @@ module.exports = {
     self.show = process.env.APOS_SHOW_I18N ? true : self.options.show;
     self.locales = self.getLocales();
     self.defaultLocale = self.options.defaultLocale || Object.keys(self.locales)[0];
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/i18n/index.js
```diff
@@ -365,8 +371,11 @@ module.exports = {
         const hostname = req.hostname;
         let best = false;
         for (const [ name, options ] of Object.entries(self.locales)) {
-          const matchedHostname = options.hostname ? (hostname === options.hostname.split(':')[0]) : null;
-          const matchedPrefix = options.prefix ? ((req.path === options.prefix) || req.path.startsWith(options.prefix + '/')) : null;
+          const matchedHostname = options.hostname
```
abea, line 374: No change, only line break for readability.

PR #3329, opened 2021-08-23
Walks back AposButton label change

abea:
A change was made to AposButton.vue in #3321 to accept an object with `phrase` and `localize` properties. Those properties were not actually used, so the two components using this new prop option did not have visible labels.


gregvanbrug:
This should be fixed in #3330

PR #3332, opened 2021-08-24
display the busy indicator while the docs are being localized

gregvanbrug:
Resolves:
- https://linear.app/apostrophecms/issue/PRO-1849

To test, go through the localize wizard. After clicking "Localize Content" on the third step, the busy screen should show.


modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -587,7 +588,12 @@ export default {
           }
         }
       }
-      this.close();
+
+      // Prevent flashing of the UI if the request returns quickly
+      setTimeout(() => {
```
gregvanbrug, line 593: Without this when the docs are localized quickly, the screen "flashes" and you're not able to see the busy screen. While it's nice we're super fast, it's not so nice that there's a blip after clicking and before being directed back to the page.

boutell (approved):

PR #3337, opened 2021-08-24
PRO-1751: fixed pluralization for notifications

boutell:
People were wrong on the Internet, so this took longer than it should've. The real issue is that `count` must be a number, not a string, for i18next to do the pluralization dance. Fixed the way we sanitize this when it comes over the wire. Other substitutions should be fine to treat as strings but `count` has a privileged status already in `i18next`.


package.json
```diff
@@ -66,7 +66,6 @@
     "glob": "^5.0.15",
     "he": "^0.5.0",
     "html-to-text": "^5.1.1",
-    "i18n": "^0.8.6",
```
boutell, line 69: Unrelated cleanup: dead dependency.

abea (approved):

PR #3340, opened 2021-08-24
PRO-1952 Fix/i18n noti color

abea:


gregvanbrug (approved):

abea:
we could run into a few more of these text style things since Stu normalized text styles in the main branch recently.

PR #3362, opened 2021-08-27
Pro 1981 onready alias

stuartromanek:


stuartromanek:
Added a note about using `onReady` going forward

@abea if you have stronger word choices or other recommendation please flesh them out here.

abea (approved):

abea (approved):

PR #3365, opened 2021-08-30
PRO-2001 fix and test previous and next filters (bumm, bumm, BUMMM, BWAH-NAAA)

### LINKED ISSUE #3364: Option `Previous` doesn't work
When I enable option {previous: true} in piece then the page thought error `TypeError: sort is not iterable`

## To Reproduce

My piece page 

```
module.exports = {
  extend: '@apostrophecms/piece-page-type',
  options: {
    label: 'Blog page',
    pieceModuleName: 'article',
    next: true,
    previous: true,
  },
  fields: {
    add: {},
    group: {},
  },
};
```
I saw in code doc-type/index.js in block 
```
if (direction === -1) {
            // Flip the sort, we need to look backwards
            for (const [ key, val ] of sort) {
              if (typeof (val) === 'number') {
                sort[key] = -val;
              }
            }
          }
```
the sort is an object, so I change to 

```
if (direction === -1) {
            // Flip the sort, we need to look backwards
            _.each(sort, function (key, val) {
              if (typeof (val) === 'number') {
                sort[key] = -val;
              }
            })
          }
```
It works!

So, if that's a bug, please fix, thanks

boutell:
Fixes #3364

This is a draft because we have a release today that this isn't QA'd for. Will add release notes before making it a final PR.

"Previous" and "next" have funny semantics when the default sort order happens to be negative, i.e. reverse chronological by updatedAt, and this could also be an issue for a blog post, but in that situation it would be up to the frontend developer to decide what to actually label these on the front end. From the back end's standpoint "previous" and "next" take you to the previous or next thing according to the sort order in effect, whatever that is. "Next" means "one step further down the list."


abea (changes requested):
Changelog please. Then is good.

abea (changes requested):
it looks like this is going to run into merge issues in the changelog since it was created before two releases that have gone out. Could you merge main into this to clear that up first?

boutell:
done

abea (approved):

PR #3366, opened 2021-08-30
PRO-1896 use req.clone consistently and fix bug that broke cloning twice with task reqs

boutell:


modules/@apostrophecms/task/index.js
```diff
@@ -222,11 +216,24 @@ module.exports = {
...
         return req;
+
+        function addCloneMethod(req) {
+          req.clone = (properties = {}) => {
+            const _req = {
+              ...req,
+              ...properties
+            };
+            self.apos.i18n.setPrefixUrls(_req);
+            addCloneMethod(_req);
```
abea, line 233: Why do we need to run this on the req again?

boutell: Calling addCloneMethod gives us a closure in which the request that will be cloned is the correct request and not the original one returned by `task.getReq`. This becomes important when a request is cloned and then cloned again. We don't want a copy of the original, we want a copy of the first clone, and the ability to make a third clone that copies the second with our own `clone` method, etc.

abea (approved):

PR #3370, opened 2021-08-30
Remove lodash wrapper around bodyParser.json options override

jogarijo:
`bodyParser.json` options passed to `@apostrophecms/express` are currently ignored because they are being wrapped by lodash before spread. This change should fix it. cf #3369


boutell (approved):
Thanks!

PR #3387, opened 2021-09-03
PRO-1997 cut and paste now works between locales that do not share a hostname

boutell:
A known limitation: this only works if you switch locales after cutting. I don't think there is a good way to do it for two tabs that are already open on separate locales with separate hostnames, short of nonstop communication between them, which would probably require websockets, etc. This is a good compromise at least for now.

If locales share a hostname using prefixes this is a non-issue.


modules/@apostrophecms/i18n/ui/apos/apps/AposI18nCrossDomainSession.js
```diff
@@ -1,23 +1,5 @@
 export default () => {
-  document.addEventListener('click', e => {
```
boutell, line 2: This turned out to be dead code because of ways the implementation was simplified to do more server-side work, however a small amount of new browser-side work is now required for the cross-domain clipboard because we must bridge LocalStorage between the domains.

abea (changes requested):
I didn't QA test, but it looks good other than the `const` bug. We'll want to add to the manual QA list for this.

abea (approved):

PR #3389, opened 2021-09-03
Pro-1810 lint module names

abea:


modules/@apostrophecms/util/index.js
```diff
@@ -555,7 +555,8 @@ module.exports = {
       // called for situations less dire than
       // `apos.util.error`.
       warn(msg) {
-        self.logger.warn.apply(self.logger, arguments);
+        const args = [ '\n⚠️', ...arguments ];
```
boutell, line 558: If you're going to do this all the way down at the level of any call to `util.warn` then you should move it to `logger.js` and make it part of how our standard logger handles presentation of `warn`. That allows custom loggers to have their own approach so it's not redundant to an already clear distinction.

boutell: You could also just do it at the level of `warnDev` in which case I wouldn't have this concern.

boutell (changes requested):

boutell (approved):

PR #3390, opened 2021-09-03
PRO-2026 resolve issues with unlocalized types not being pickable when browsing, yet unpublished pieces being pickable when autocompleting

boutell:


modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -32,16 +33,37 @@ export default {
...
+      type: String,
+      required: false
     }
   },
   emits: [ 'select' ],
   methods: {
-    select(item) {
+    select(item, $event) {
+      if (item.disabled) {
+        $event.stopPropagation();
```
boutell, line 46: It's right to do this, but it's not enough to prevent dismissal of the list because that's done via a focusOut handler another level up, and we can't completely refuse all interaction on these because that would prevent them from having a tooltip, and thus the self-documenting function that is the entire reason we show them even though they are unpublished. Fortunately it's unlikely folks will click on them much once they see the tooltip, and the disabled appearance, and the fact that they don't get selected. So calling it an improvement, at the least.

modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -32,16 +33,37 @@ export default {
...
       const selectedItems = this.selectedItems;
       if (!selectedItems.some(selectedItem => selectedItem._id === item._id)) {
         // Never modify a prop
         this.$emit('select', [ ...selectedItems, item ]);
       }
+    },
+    getClasses(item) {
+      const classes = {
+        'apos-search__item': true,
+        'item': true
```
stuartromanek, line 58: what is this, what is being styled with this selector? What can't `apos-search__item` do?

also, all classes have to be prefixed `apos-`.

modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -32,16 +33,37 @@ export default {
...
+      };
+      if (item.disabled) {
+        classes.disabled = true;
```
stuartromanek, line 61: needs apos namespace. either `apos-disabled` or `apos-search__item--disabled`

modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -81,7 +103,7 @@ export default {
     pointer-events: none;
   }
 
-  &:hover {
+  &:hover.enabled {
```
stuartromanek, line 106: enabled should be the default case, why do we need to qualify this?

modules/@apostrophecms/schema/ui/apos/components/AposSearchList.vue
```diff
@@ -32,16 +33,37 @@ export default {
...
+    getClasses(item) {
+      const classes = {
+        'apos-search__item': true,
+        'item': true
+      };
+      if (item.disabled) {
+        classes.disabled = true;
+      } else {
+        // Avoids writing brittle styles to shut off our other styles
+        classes.enabled = true;
```
stuartromanek, line 64: needs apos namespace. either `apos-enabled` or `apos-search__item--enabled`

stuartromanek (changes requested):

boutell:
Cleaned up that shiz. No more enabled class, used a mixin to avoid duplication.

stuartromanek (approved):

PR #3391, opened 2021-09-03
PRO-1670 parked defaults

abea:
Since pages were using the regular doc-type `allowedSchema` method they didn't get the parked page checks. So when a page was parked and we tried to edit it there were opportunities to fail, including if we were updating a page whose type was not in the main array of selectable types.

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -520,17 +520,13 @@ module.exports = {
 
       async convert(req, input, doc, options = {
         presentFieldsOnly: false,
-        copyingId: false,
-        isPage: false
+        copyingId: false
       }) {
-        let fullSchema;
+        const fullSchema = self.apos.doc.getManager(options.type || self.name)
+          .allowedSchema(req, doc);
         let schema;
         let copyOf;
-        if (options.isPage) {
-          fullSchema = self.apos.page.allowedSchema(req, doc);
-        } else {
-          fullSchema = self.apos.doc.getManager(options.type || self.name).allowedSchema(req);
-        }
+
         if (options.presentFieldsOnly) {
           schema = self.apos.schema.subset(fullSchema, self.fieldsPresent(input));
         } else {
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -520,11 +520,17 @@ module.exports = {
 
       async convert(req, input, doc, options = {
         presentFieldsOnly: false,
-        copyingId: false
+        copyingId: false,
+        isPage: false
       }) {
-        const fullSchema = self.apos.doc.getManager(options.type || self.name).allowedSchema(req);
+        let fullSchema;
         let schema;
         let copyOf;
+        if (options.isPage) {
+          fullSchema = self.apos.page.allowedSchema(req, doc);
+        } else {
+          fullSchema = self.apos.doc.getManager(options.type || self.name).allowedSchema(req);
+        }
         if (options.presentFieldsOnly) {
           schema = self.apos.schema.subset(fullSchema, self.fieldsPresent(input));
         } else {
@@ -540,13 +546,14 @@ module.exports = {
             ...input
           };
         }
+
         await self.apos.schema.convert(req, schema, input, doc);
+
         doc.copyOfId = copyOf && copyOf._id;
         if (copyOf) {
           self.apos.schema.regenerateIds(req, fullSchema, doc);
         }
       },
-
       // Return the names of all schema fields present in the `input` object,
       // taking into account issues like relationship fields keeping their data in
       // a separate ids property, etc.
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/search/index.js
```diff
@@ -50,8 +50,8 @@ module.exports = {
   extend: '@apostrophecms/page-type',
   options: {
     alias: 'search',
-    name: '@apostrophecms/search',
-    perPage: 10
+    perPage: 10,
+    label: 'apostrophe:searchLabel'
```
abea, line 54: No label made the editor sad. We should consider doing this for the home page as well. It's conceivable someone will park the home page and not allow editors to create new ones or change the type of the parked one.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -520,11 +520,17 @@ module.exports = {
 
       async convert(req, input, doc, options = {
         presentFieldsOnly: false,
-        copyingId: false
+        copyingId: false,
+        isPage: false
```
abea, line 524: Eh? I considered doing a `__meta.chain` thing to see if it was a page, but this is only called twice for pages, I think.

boutell: A page type will never stop being a page type, even if the particular page type changes.

So we can write `self.apos.instanceOf(module, '@apostrophecms/piece-type')` here, and avoid requiring the level above to inject something we should already know as a manager of a specific type.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -520,11 +520,17 @@ module.exports = {
...
       }) {
-        const fullSchema = self.apos.doc.getManager(options.type || self.name).allowedSchema(req);
+        let fullSchema;
         let schema;
         let copyOf;
+        if (options.isPage) {
```
boutell, line 529: It would be better classical OOP practice if this were an invocation of a new doc-type method that is overridden for `page-type`, rather than "if this type else this type" logic.

In fact, probably `allowedSchema` should just take an `options` object to allow override of `type`.

modules/@apostrophecms/doc-type/index.js
boutell: Never mind, better idea below.

modules/@apostrophecms/search/index.js
boutell: Makes sense.

boutell (changes requested):

modules/@apostrophecms/doc-type/index.js
abea: @boutell It seems like `allowedSchema` should be that method. The problem I'm running into is that the page module's `allowedSchema` calls the doc-type's `allowedSchema`, setting up an infinite loop.

What I think we should do is move the page module's `allowedSchema` functionality into `page-type`, using  `extendMethods` to use the original `doc-type` method (since the page version already does), but wrapped in the extra page functionality. There are three or four page module methods in there that we could simply use with `apos.page.whatever`. There's one place in the page module where we'd change `self.allowedSchema` to `manager.allowedSchema`.

Do you see any issues with that?

boutell: I agree that's the best route, with a compatibility wrapper for anyone who does call `apos.page.allowedSchema`.

modules/@apostrophecms/page-type/index.js
```diff
@@ -215,6 +215,49 @@ module.exports = {
   },
   methods(self) {
     return {
+      // Given a page and its parent (if any), returns a schema that  is
+      // filtered appropriately to that page's type, taking into account whether
+      // the page is new, whether it is parked, and the parent's allowed subpage
+      // types.
+      allowedSchema(req, page = {}, parentPage = {}) {
+        let schema = self.schema;
```
boutell, line 223: By not calling the super version of this method you're losing the permissions-related filtering that happens there. You still need to do that to get your initial schema for further processing.

boutell: So you should be using extendMethods here.

modules/@apostrophecms/page/index.js
```diff
@@ -2062,47 +2062,11 @@ database.`);
           await change(req, page, data);
         }
       },
-      // Given a page and its parent (if any), returns a schema that
-      // is filtered appropriately to that page's type, taking into
-      // account whether the page is new and the parent's allowed
-      // subpage types
+      // Backward compatible method following moving this to page-type module.
+      // This page module method may be deprecated in the next major version.
```
boutell, line 2066: OK

boutell: (I was going to say keep it as a wrapper that takes care of finding the right manager, but you already have the manager for another reason in the one place you'd use it, so.)

boutell (changes requested):
Losing too much of the base class functionality, but shouldn't be hard to fix.

modules/@apostrophecms/page-type/index.js
abea: my bad. I forgot my own plan.

boutell (approved):

PR #3394, opened 2021-09-07
Hotfix for possible SVG spearsphishing attack vector

boutell:
This is low probability, but possible. This initial PR is against a base branch that matches 3.3.1. I like this procedure better than starting out with a PR against main, because this way no error-prone cherry-picking is required, just a merge to main after release.


modules/@apostrophecms/attachment/index.js
```diff
@@ -707,7 +734,10 @@ module.exports = {
         const batchSize = 100;
         let lastId = '';
         while (true) {
-          const docs = await self.db.find({ _id: { $gt: lastId } }).limit(batchSize).sort({ _id: 1 }).toArray();
+          const docs = await self.db.find({
+            ...(criteria || {}),
```
boutell, line 738: A bug that got in the way: the `each` method for attachment migrations was ignoring its `criteria` argument.

modules/@apostrophecms/attachment/index.js
```diff
@@ -402,6 +405,16 @@ module.exports = {
         }
         info.length = await self.apos.util.fileLength(file.path);
         info.md5 = await self.apos.util.md5File(file.path);
+        if (info.extension === 'svg') {
+          try {
+            await self.sanitizeSvg(file, info);
```
abea, line 410: The `info` parameter isn't used in the method.

abea (changes requested):
Also an eslint error

abea (dismissed):

abea (approved):

PR #3396, opened 2021-09-08
Localized media info, fixed file size display

boutell:


modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -171,6 +171,12 @@
   "manageDocType": "Manage {{ type }}",
   "manageDraftSubmissions": "Manage Draft Submissions",
   "managePages": "Manage Pages",
+  "mediaCreatedDateDayjsFormat": "MMM Do, YYYY",
+  "mediaCreatedDate": "Uploaded: {{ createdDate }}",
+  "mediaDimensions": "Dimensions: {{ width }} x {{ height }}",
```
abea, line 176: This is a different `𝗑` than the original one. I think the one originally used is more associated with multiplication rather than being a letter.

modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -171,6 +171,12 @@
   "manageDocType": "Manage {{ type }}",
   "manageDraftSubmissions": "Manage Draft Submissions",
   "managePages": "Manage Pages",
+  "mediaCreatedDateDayjsFormat": "MMM Do, YYYY",
```
abea, line 174: There is an established pattern for localized dayjs formats. Should be something like `dayjsMediaCreatedFormat`

abea (changes requested):

abea:
This should maybe hold for the next release at this point.

boutell:
OK we'll hold.

boutell:
Ready for review, a draft until today's release goes out so it doesn't accidentally sneak in without QA.

abea (dismissed):

abea (approved):

PR #3397, opened 2021-09-08
PRO-2012 new global event names, much more explicit startup sequence

boutell:


modules/@apostrophecms/module/lib/events.js
```diff
@@ -4,6 +4,15 @@ module.exports = function(self) {
 
   return {
 
+    // Register an old event name to be accepted for a new event name.
```
boutell, line 7: For now this only works for apostrophe core level events, but the API is designed so that if we need to we can extend it later to support aliases for module level events.

modules/@apostrophecms/module/lib/events.js
```diff
@@ -82,6 +91,7 @@ module.exports = function(self) {
       if (_.camelCase(eventName) === _.camelCase(handlerName)) {
         throw new Error('The handler name ' + handlerName + ' is essentially identical to the event name, ' + eventName + '. To prevent conflicts with other handlers, you must give it a distinct name expressing what it actually does with the event.');
       }
+      eventName = self.apos.eventAliases[`${moduleName}:${eventName}`] || eventName;
```
boutell, line 94: Evaluated here in `on` so that we don't impose a performance hit in actual handlers.

modules/@apostrophecms/page/index.js
```diff
@@ -2179,6 +2143,36 @@ database.`);
           input[field] = page[field];
         }
       },
+      async implementParkAllInDefaultLocale() {
```
boutell, line 2146: These are methods now so the core can call them.

scripts/lint-i18n.js
```diff
@@ -2,11 +2,11 @@ const glob = require('glob');
 const fs = require('fs');
 let keys = Object.keys(require('../modules/@apostrophecms/i18n/i18n/en.json'));
 // Core apostrophe events look like keys
-keys = [ ...keys, 'destroy', 'afterInit', 'modulesReady', 'run' ];
+keys = [ ...keys, 'destroy', 'ready', 'modulesRegistered', 'afterInit', 'modulesReady', 'run' ];
```
boutell, line 5: More apostrophe: strings for our i18n linter to not get confused by.

scripts/lint-i18n.js
```diff
@@ -2,11 +2,11 @@ const glob = require('glob');
...
 const files = glob.sync('**/*.@(js|vue|html)', { ignore: [ '**/node_modules/**/*', 'coverage/**/*' ] });
 
 const undeclared = new Set();
-const used = new Set();
+const used = new Set([ 'afterInit', 'modulesReady' ]);
```
boutell, line 9: New edge case: don't get confused if these are never used.

test/bootstrapping.js
```diff
@@ -13,10 +12,7 @@ describe('bootstrap of Apostrophe core', function() {
     try {
       apos = await t.create({
         root: module,
-        overrideTest: 'test', // overriden by data/local.js
-        __testDefaults: {
```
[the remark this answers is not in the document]
boutell: I kept the data/local.js tests as those don't get a lot of coverage elsewhere.

index.js
```diff
@@ -103,13 +103,23 @@ module.exports = async function(options) {
...
+    self.alias('afterInit', 'ready');
 
     defineModules();
 
     await instantiateModules();
     lintModules();
-    await self.emit('modulesReady');
-    await self.emit('afterInit');
+    await self.emit('modulesRegistered'); // formerly modulesReady
+    await self.apos.migration.migrate(); // emits before and after events, inside the lock
```
abea, line 115: I don't see an after event in the migrate method.

modules/@apostrophecms/module/lib/events.js
```diff
@@ -4,6 +4,15 @@ module.exports = function(self) {
 
   return {
 
+    // Register an old event name to be accepted for a new event name.
+    // If newName is emitted, oldName will be accepted as a synonym in handlers.
+    // Currently supported for Apostrophe core module events only. Not to be
+    // used for module-specific events. Call with just the event name,
+    // not the `apostrophe:` part.
+    alias(oldName, newName) {
```
abea, line 12: Let's call it `aliasEvent` since when it's called there's no connection to this file.

abea (changes requested):

index.js
boutell: Probably because that one is not new, so it's not in the diff. It's there.

modules/@apostrophecms/module/lib/events.js
boutell: Done.

index.js
abea: true.

abea (approved):

PR #3399, opened 2021-09-08
Nested widget spacing

stuartromanek:
Give the people what they want.

This tees up allowing developers to configure a margin value that spaces nested widgets within areas.

I'm open to baking in a default.


modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -196,6 +196,7 @@ export default {
     apos.bus.$on('widget-hover', this.updateWidgetHovered);
     apos.bus.$on('widget-focus', this.updateWidgetFocused);
     this.bindEventListeners();
+    document.documentElement.style.setProperty('--a-widget-margin', apos.ui.widgetMargin);
```
stuartromanek, line 199: Not where this should go, but where is correct?

For context, this is where we set/override a developer preference for the margin value as a CSS variable. This only needs to be set once per logged-in Edit mode-enabled view.

modules/@apostrophecms/ui/index.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   options: {
-    alias: 'ui'
+    alias: 'ui',
+    widgetMargin: '0'
```
stuartromanek, line 4: this will is a CSS value, so string is fine, allows for a CSS'y format like `20px 0`

modules/@apostrophecms/ui/index.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   options: {
-    alias: 'ui'
+    alias: 'ui',
```
stuartromanek, line 3: This feels like the right module to put this in. The other contender was `@apostrophecms/area`

modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
abea: It should be in something that only runs once per page load, probably.

stuartromanek: @abea yes, only logged in

boutell: So AposAreas.js would probably make more sense.

boutell (changes requested):
AposAreas.js would make more sense as a home for a one-time setting of a CSS variable based on a module property.

stuartromanek:
Changed where the margin gets applied and also added a default value (20px 0)

modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
stuartromanek: moved this to `AposAreas.js`

boutell (approved):

PR #3400, opened 2021-09-09
PRO-2030 svg attack mitigation

boutell:


test-lib/build-test-package-json.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
```
boutell, line 1: This new utility avoids the need to create a placeholder test/package.json manually, which is a particular hassle for 2.0 core now that we have strict checks for modules being dependencies in moog-require. Specifically, 2.0 core has lots of dependencies that get required directly by tests, which count as "project level" in that context.

This utility has the potential to be required and used from npm scripts in other modules like apostrophe-workflow if their test/package.json files become irritating to maintain.

abea (approved):

PR #3406, opened 2021-09-10
PRO-2057 fixed the regression tests

boutell:
The scoped pieces test does have to be removed, see separate PR.

### DIFF
--- a/lib/modules/apostrophe-modal/public/js/modal.js
+++ b/lib/modules/apostrophe-modal/public/js/modal.js
@@ -461,16 +461,18 @@ apos.define('apostrophe-modal', {
       setImmediate(function() {
         $wrapper.find('[data-modal-content]').removeClass('apos-modal-slide-current');
         var fired = false;
-        $content.one('webkitTransitionEnd otransitionend oTransitionEnd msTransitionEnd transitionend', function() {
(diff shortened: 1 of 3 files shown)


lib/modules/apostrophe-modal/public/js/modal.js
```diff
@@ -461,16 +461,18 @@ apos.define('apostrophe-modal', {
...
         $wrapper.find('[data-modal-content]').removeClass('apos-modal-slide-current');
         var fired = false;
-        $content.one('webkitTransitionEnd otransitionend oTransitionEnd msTransitionEnd transitionend', function() {
+        $content.one('webkitTransitionEnd otransitionend oTransitionEnd msTransitionEnd transitionend', handler);
+        // Fallback because the transition has been observed to never happen, at least in nightwatch testing
+        setTimeout(handler, 1000);
+        $content.addClass('apos-modal-slide-current');
+        function handler() {
           if (fired) {
-            // Has been seen firing twice in nightwatch tests, in spite of `one`
```
boutell, line 466: This now has a more obvious purpose, in that the timeout and the transition event are clearly capable of producing two invocations, so the comment feels superfluous.

abea (approved):

PR #3409, opened 2021-09-10
PRO-2063: fix bug that broke Undo and created duplicate widget _ids, prevent future duplicate widget _ids and migrate

boutell:


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -512,7 +512,7 @@ export default {
 
       if (refreshable) {
         refreshable.innerHTML = content;
-        if (!this.original) {
+        if (this.editMode && (!this.original)) {
```
boutell, line 515: At some point we started using this refresh method even in cases where we're not in edit mode (or not yet), which means we were grabbing an "original" that might lack content, resulting in an improper basis for the patches from undo, and... scene.

test/pieces.js
```diff
@@ -568,8 +572,15 @@ describe('Pieces', function() {
               {
                 metaType: 'widget',
                 type: '@apostrophecms/rich-text',
-                id: cuid(),
+                _id: (i === 1) ? widgetId : null,
```
boutell, line 575: Test consequences of an explicit widget id that duplicates another, but only in the first document in the set.

abea (approved):

PR #3418, opened 2021-09-14
 PRO-1114 Updates cascade grouping to combine like-named groups' fields

abea:
I couldn't let this go anymore. Seems like a simple fix for a big quality of life improvement. Devs no longer need to know what the inherited fields in a group are.

### DIFF
--- a/lib/moog.js
+++ b/lib/moog.js
@@ -192,17 +192,24 @@ module.exports = function(options) {
           }
           if (properties.group) {
             const groups = klona(that[`${cascade}Groups`]);
-            for (const value of Object.values(properties.group)) {
+            for (const [ key, value ] of Object.entries(properties.group)) {
               for (const field of value.fields || []) {
-                for (const value of Object.values(groups)) {
-                  if (value.fields) {
-                    if (value.fields.includes(field)) {
-                      value.fields = value.fields.filter(_field => _field !== field);
+                // Remove fields from existing groups if they're added to a new
+                // group.
+                for (const val of Object.values(groups)) {
+                  if (val.fields) {
+                    if (val.fields.includes(field)) {
+                      val.fields = val.fields.filter(_field => _field !== field);
                     }
                   }
                 }
               }
+              // Combine groups of the same name
+              if (groups[key] && Array.isArray(groups[key].fields)) {
+                value.fields = groups[key].fields.concat(value.fields);
+              }
             }
+
             that[`${cascade}Groups`] = {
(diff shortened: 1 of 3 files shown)


lib/moog.js
```diff
@@ -192,17 +192,24 @@ module.exports = function(options) {
...
+                  if (val.fields) {
+                    if (val.fields.includes(field)) {
+                      val.fields = val.fields.filter(_field => _field !== field);
                     }
                   }
                 }
               }
+              // Combine groups of the same name
+              if (groups[key] && Array.isArray(groups[key].fields)) {
+                value.fields = groups[key].fields.concat(value.fields);
```
abea, line 209: Since we're giving the new groups priority when spreading them below, let's combine on the new group's fields rather than try to be more fancy about it.

lib/moog.js
```diff
@@ -192,17 +192,24 @@ module.exports = function(options) {
           }
           if (properties.group) {
             const groups = klona(that[`${cascade}Groups`]);
-            for (const value of Object.values(properties.group)) {
+            for (const [ key, value ] of Object.entries(properties.group)) {
               for (const field of value.fields || []) {
-                for (const value of Object.values(groups)) {
```
abea, line 197: This kind of thing makes me cross eyed (the same constant name in a nested loop).

boutell (changes requested):
Is this a BC break?

That depends. What happens if another group in the same module *explicitly* poaches one of those fields? What if that group appears earlier in the file? It should still win because it *definitely cares* about that field.

In that case I think we might get away with this, because it's not as if anybody *intends* to leave fields in "Ungrouped."

boutell:
(We'd want test coverage on that scenario.)

abea:
@boutell If someone poaches a field earlier then it will have been removed from the original field array before getting to this step.

boutell (changes requested):
Please add unit tests. Specifically make sure you cover this case:

```
// assume we inherited basics: { fields: [ 'one', 'two', 'three' ] }
group: {
  other: {
    fields: [ 'one' ]
  },
  basics: {
    fields: [ 'four', 'five' ]
  }
}
```

`other` should successfully steal `one`, so `basics` should wind up with 2 through 5.

boutell (approved):

boutell:
Nice, thanks for adding the test coverage. If people bump into any oddities we'll add more coverage and tweak from there.

boutell (approved):

PR #3419, opened 2021-09-14
Fixes error messages to remove the options argument recommendation

abea:
The documented recommendation is to only use `self` as an argument to customization functions. This should be reflected in error messages as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-* The file size of uploaded media is visible again when selected in the editor, and media information such as upload date, dimensions and file size is now properly localized. 
+* The file size of uploaded media is visible again when selected in the editor, and media information such as upload date, dimensions and file size is now properly localized.
+* Fixes error moog error messages to reflect the recommended pattern of customization functions only taking `self` as an argument.
 
 ## 3.4.1 - 2021-09-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 * The file size of uploaded media is visible again when selected in the editor, and media information such as upload date, dimensions and file size is now properly localized.
-* Fixes error moog error messages to reflect the recommended pattern of customization functions only taking `self` as an argument.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-* The file size of uploaded media is visible again when selected in the editor, and media information such as upload date, dimensions and file size is now properly localized. 
+* The file size of uploaded media is visible again when selected in the editor, and media information such as upload date, dimensions and file size is now properly localized.
+* Fixes error moog error messages to reflect the recommended pattern of customization functions only taking `self` as an argument.
```
boutell, line 8: "error" appears twice, funny grammar

lib/moog.js
```diff
@@ -142,8 +142,8 @@ module.exports = function(options) {
 
     const upgradeHints = {
       construct: 'in Apostrophe 3.x, "construct" has been replaced with "methods", "routes", "apiRoutes", etc.',
-      beforeConstruct: 'in Apostrophe 3.x, "beforeConstruct" has been replaced with "beforeSuperClass". It takes (self, options) and should be solely concerned with modifying the options before the base class sees them. It must be synchronous. Check out the new fields section, you might not need beforeSuperClass.',
-      afterConstruct: 'in Apostrophe 3.x, "afterConstruct" has been replaced with "init". It takes (self, options) and may be an async function.'
+      beforeConstruct: 'in Apostrophe 3.x, "beforeConstruct" has been replaced with "beforeSuperClass". It takes (self) and should be solely concerned with modifying the options before the base class sees them. It must be synchronous. Check out the new fields section, you might not need beforeSuperClass.',
```
boutell, line 145: It actually does receive `self, options`, so we're fibbing a little bit, but I agree it's better to treat `self` as the documented parameter.

boutell: A pity we didn't remove `options` before release. Oh well, 4.0.

boutell (changes requested):

lib/moog.js
abea: Exactly. My goal was to get this in line with our docs.

boutell (approved):

PR #3421, opened 2021-09-15
Supports adding a test module namespace

abea:
The existing tests didn't support official Apostrophe modules well since they created a symlink test node module named after the package directory. A3 modules do not include the namespace in the package directory typically. This supports adding the namespace for a test module.

In our official modules in packages like `form` or `seo` we would use `testModule: '@apostrophecms'` on test apps to namespace the symlink.

### DIFF
--- a/index.js
+++ b/index.js
@@ -254,6 +254,8 @@ module.exports = async function(options) {
   // when options.testModule is true. There must be a
   // test/ or tests/ subdir of the module containing
   // a test.js file that runs under mocha via devDependencies.
+  // If `options.testModule` is a string it will be used as a
+  // namespace for the test module.
 
   function testModule() {
     if (!options.testModule) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -279,9 +279,17 @@ module.exports = async function(options) {
...
+    let pkgNamespace = '';
+    if (pkgName.includes('/')) {
+      const parts = pkgName.split('/');
+      pkgNamespace = '/' + parts.slice(0, parts.length - 1).join('/');
+    }
+
     if (!fs.existsSync(testDir + '/node_modules')) {
-      fs.mkdirSync(testDir + '/node_modules' + addDirNamespace(options.testModule), { recursive: true });
-      fs.symlinkSync(moduleDir, testDir + '/node_modules' + addDirNamespace(options.testModule) + '/' + require('path').basename(moduleDir), 'dir');
+      fs.mkdirSync(testDir + '/node_modules' + pkgNamespace, { recursive: true });
```
boutell, line 291: What version of Node did this recursive flag arrive in ?

boutell: Otherwise looks good.

boutell (changes requested):

index.js
abea: Added in Node 10.

boutell: Unfortunately A2 still runs on some Node 8 servers. 😢

boutell (changes requested):
Unfortunately A2 still runs on some Node 8 servers. 😢

abea:
This isn't for A2.

index.js
abea: This isn't for A2.

abea: I can implement this differently in A2.

boutell (approved):
Sorry, too many balls in the air.

PR #3422, opened 2021-09-16
Quiets the no users warning if a test Module

abea:
Quality of life improvement. No need to get the warning spam when running tests.


boutell (changes requested):
I'm with you on the goal here, but this doesn't help if it's a unit test in apostrophe core.

What I would suggest is that in `test-util/util.js`, which all the tests use, you pass a `test: true` option.

modules/@apostrophecms/login/index.js
```diff
@@ -390,7 +390,7 @@ module.exports = {
         const adminReq = self.apos.task.getReq();
         const user = await self.apos.user.find(adminReq, {}).relationships(false).limit(1).toObject();
 
-        if (!user) {
+        if (!user && !self.options.apos.options.testModule) {
```
boutell, line 393: Also it's way past late enough to just check `self.apos.options.testModule`.

boutell (approved):

PR #3423, opened 2021-09-16
def rt els

stuartromanek:


modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -28,7 +28,6 @@
     <div class="apos-rich-text-editor__editor" :class="editorModifiers">
       <editor-content :editor="editor" :class="moduleOptions.className" />
     </div>
-    <!-- Using actual DOM element rather than :after to ease localization -->
```
stuartromanek, line 31: unrelated, future generations won't wonder about this

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -219,7 +218,7 @@ export default {
       const self = this;
       const enhanced = [];
       (styles || []).forEach(style => {
-        style.options = {};
+        style.options = { attrs: {} };
```
stuartromanek, line 221: so that we can apply bulk html attributes to elements. right now we only care about classes though

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Document.js
```diff
@@ -0,0 +1,24 @@
+// Acts a custom Document extension
```
stuartromanek, line 1: specifying our own schema content gets us close to being able to 'specify the default rich text element' by prioritizing one node type over another.

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Heading.js
```diff
@@ -0,0 +1,39 @@
...
+    ? headings.filter(heading => heading.def)[0].options.level
+    : levels[0];
+  // Map of headings to their attributes
+  const attrs = {};
+  levels.forEach(level => {
+    attrs[level] = headings.find(heading => heading.options.level === level).options.attrs || {};
+  });
+  return Heading.extend({
+    defaultOptions: {
+      levels,
```
stuartromanek, line 17: let us spell out the valid heading levels for a RT widget

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Heading.js
```diff
@@ -0,0 +1,39 @@
...
+      attrs
```
stuartromanek, line 18: maps heading levels to their specified html attributes (classes)

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Heading.js
```diff
@@ -0,0 +1,39 @@
...
+    },
+    addAttributes() {
+      return {
+        level: {
+          default: defaultLevel,
```
stuartromanek, line 23: puts a finer point on _which_ heading should be default, the part 2 of the Document subclass

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * The file size of uploaded media is visible again when selected in the editor, and media information such as upload date, dimensions and file size is now properly localized.
 * Fixes moog error messages to reflect the recommended pattern of customization functions only taking `self` as an argument.
+* Rich Text widget's now instantiate with a valid element from the `styles` option rather than always starting with an unclassed `<p>` tag.
```
boutell, line 9: widgets, not widget's

CHANGELOG.md
```diff
@@ -14,6 +15,7 @@
 ### Adds
 
 * If `options.testModule` on the app is a string it will be used as an npm namespace when creating a symlink test module.
+* Rich Text widget's styles supports a `def` property for specifying the default style the editor should instantiate with.
```
boutell, line 18: support, not supports

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Document.js
boutell: Acts as a (sorry)

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Document.js
```diff
@@ -0,0 +1,24 @@
...
+import { Node } from '@tiptap/core';
+export default (options) => {
+  const defNode = options.styles.filter(style => style.def)[0];
```
boutell, line 4: Is the styles option mandatory / always set to a fallback value somewhere above this point?

boutell (changes requested):
This looks great. Only one real "hey is this right" question, and tiny text nitpicks. Gonna need quality QA

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Document.js
stuartromanek: Yes, we inherit a styles array from rich-text's `minimumDefaultOptions` if no styles are provided.

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -217,7 +216,6 @@ export default {
     // commands and parameters used internally.
     enhanceStyles(styles) {
       const self = this;
-      const enhanced = [];
```
stuartromanek, line 220: unrelated, weren't using

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -242,9 +240,7 @@ export default {
           style.options.class = style.class;
         }
 
-        if (style.type) {
```
stuartromanek, line 245: unrelated, weren't using

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Classes.js
```diff
@@ -36,13 +36,11 @@ export default (options) => {
...
-                    ? classes.join(' ')
-                    : (
-                      allow[tag].length ? allow[tag][0] : null
-                    )
-                };
+                return classes.length
+                  ? classes.join(' ')
+                  : (
+                    allow[tag].length ? allow[tag][0] : null
+                  );
```
stuartromanek, line 43: unrelated to the goal of this ticket but is the fix for [this one](https://linear.app/apostrophecms/issue/PRO-2091/tiptap-breaking-change).

This break was preventing attributes from being applied to elements parse, which resulted in attributes not being applied on reload

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
```diff
@@ -0,0 +1,64 @@
+// If a style has been marked `def`, we need to make sure it
+// and it's attributes are prioritized in our editor.
+// It's easier to create a new Node with our specifications and put it
+// at the front of the line than try to infer between editor instantiation
+// and the editor-user setting styles via the toolbar to know when its needed.
```
stuartromanek, line 5: This marks a different approach to the whole problem, a few things led up to this:
- We allow multiple styles that use the same tag (and different attributes/options). When one of these multiples was marked `def` we couldn't infer whether it was time to use the settings of our default or if the element settings were being applied explicitly from the toolbar (switching among styles)
- In Prosemirror all heading levels are part of the same Node extension, making it difficult to single out a specific level and apply attributes to, esp when compounded by the multiple styles/same tag problem in point one^

So instead of trying to neatly configure all our nodes we create a new node where we can reliably stamp all our requirements in, knowing that when we prioritize the this new node type in the document's schema, it will be the default element on first creation.

This node is a one-off, in that there are no controls to set any future selections to this type. The editor essentially hands it off to be transformed by the normal style controls, where it is converted among the normal base nodes/marks.

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Heading.js
stuartromanek: ^ This point is less relevant now that we make a special node for our default, but it still makes sense to hem headings in based on their configuration.

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Classes.js
boutell: Eyyy nice

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
```diff
@@ -0,0 +1,64 @@
...
+const nodeMap = {
+  heading: Heading,
+  paragraph: Paragraph
+};
+
+export default (options) => {
+  const def = options.styles.filter(style => style.def)[0];
+
+  // Configuration has a default style
+  if (def) {
```
boutell, line 20: It's OK to return nothing here when there's no default?

boutell (changes requested):
Enjoying your transmogrification into That One Guy Who Actually Gets Tiptap Completely

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
stuartromanek: We do ensure we have a `def` style even if we set it, but should the checks for a `def` style be removed to make that clearer?

where we make sure have a default 👉  https://github.com/apostrophecms/apostrophe/pull/3423/files#diff-19bda1b8bcbb44e43a07e5665b4030fc2604c5fe3fc632974c3b25289d9a00c1R256-R266

boutell (approved):

PR #3425, opened 2021-09-20
Exits lintModules if there are no modules to lint

abea:
This was throwing an error in the new form modules tests since there were no other modules in the test suite to lint. The `fs.readdirSync` ran into trying to do something with the non-existent directory.


abea:
The alternative might be to create an empty directory in the testModule function or do something else there.

boutell (approved):

boutell (approved):

PR #3428, opened 2021-09-21
Webpack css loader config

Dreamer93:
We are solving an issue where `css-loader` is trying to parse `url()` inside `modules/@apostrophecms/assets`. This is to no avail as webpack is the one that should compile paths. 

Solution established with @boutell


boutell:
In fact I think webpack leaves them alone too, which gives us the intended result: you can write a project-relative path (a URL starting with /images/whatever.jpg will actually hit the file served as /public/images/whatever.jpg). Grabbing this, thank you.

PR #3429, opened 2021-09-21
PRO-1739 fixed foreign area editing

boutell:
Also added a decent error message when you forget to nest `widgets` in `options`, because that was making it painful to test the other thing.

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -38,11 +38,19 @@ module.exports = {
   restApiRoutes(self) {
     return {
       // GET /api/v1/@apostrophecms/doc/_id supports only the universal query
-      // features, but works for any document type, Simplifies browser-side
-      // logic for redirects to foreign documents; the frontend only has to
+      // features, but works for any document type. Simplifies browser-side
+      // logic for redirects to foreign documents. The frontend only has to
       // know the doc _id.
+      //
+      // Since this API is solely for editing purposes you will receive
+      // a 404 if you request a document you cannot edit.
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/doc/index.js
```diff
@@ -38,11 +38,19 @@ module.exports = {
   restApiRoutes(self) {
     return {
       // GET /api/v1/@apostrophecms/doc/_id supports only the universal query
-      // features, but works for any document type, Simplifies browser-side
-      // logic for redirects to foreign documents; the frontend only has to
+      // features, but works for any document type. Simplifies browser-side
+      // logic for redirects to foreign documents. The frontend only has to
       // know the doc _id.
+      //
+      // Since this API is solely for editing purposes you will receive
```
boutell, line 45: Took the liberty of adding this rule because it makes sense for the use case and previously it was so busted there's no bc to worry about.

modules/@apostrophecms/doc/index.js
```diff
@@ -38,11 +38,19 @@ module.exports = {
...
+      // a 404 if you request a document you cannot edit.
       async getOne(req, _id) {
-        return self.find(req, { _id }).toObject();
+        _id = self.apos.i18n.inferIdLocaleAndMode(req, _id);
+        const doc = await self.find(req, { _id }).permission('edit').toObject();
+        if (!doc) {
+          throw self.apos.error('notfound');
```
boutell, line 51: Was returning a null document rather than a 404 for an allegedly RESTful API before, not cool.

abea (approved):

PR #3432, opened 2021-09-21
fix autocomplete

stuartromanek:
Regression: autocomplete in the locale searcher didn't work
Small UX bit around resetting the search when the remaining results are probably not actionable.


modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -581,7 +588,7 @@ export default {
     },
     updateFilter(event) {
       if (event && event.data) {
-        wizard.sections.selectLocales.filter = event.data
+        this.wizard.sections.selectLocales.filter = event.data
```
stuartromanek, line 591: this was the bug

boutell (approved):

PR #3434, opened 2021-09-22
Create pt-br.json

pietro-rutzen:
Brazilian Portuguese translation


boutell (changes requested):
Thank you so much for the PR!

Please rename this file `pt-BR.json` (note the uppercase) to follow the standard capitalization pattern for i18next:

"Make sure you use the 'en-US' format, instead of underscores or similar."

Also, I see you put down "pt-br" for some of the dayjs formats. I am curious: does that work? I wasn't aware dayjs could take a locale name as a format, but maybe it can.

And in general: have you tested this in a project yet with the locale enabled? Just so we know.

Thanks!

pietro-rutzen:
> Thank you so much for the PR!
> 
> Please rename this file `pt-BR.json` (note the uppercase) to follow the standard capitalization pattern for i18next:
> 
> "Make sure you use the 'en-US' format, instead of underscores or similar."
> 
> Also, I see you put down "pt-br" for some of the dayjs formats. I am curious: does that work? I wasn't aware dayjs could take a locale name as a format, but maybe it can.
> 
> And in general: have you tested this in a project yet with the locale enabled? Just so we know.
> 
> Thanks!

I've got the wrong impression that I could pass the 'pt-br' as format pattern but testing it with dayJs, it actually does not work like that, I've reverted it to the previous string. 

I've tested the formatting of the dates with a dayJs sandbox, and if apos is using the right i18n locale, then everything should be looking good.

I have not tested it in a project. Only modified the json.

Cheers!

boutell:
It's fine to provide an appropriate dayjs string for the ones you reverted to "apostrophe" FYI, or maybe what the apostrophe custom formatter is doing is OK for your audience or you're just not up for tangling with it, which is OK.

pietro-rutzen:
> It's fine to provide an appropriate dayjs string for the ones you reverted to "apostrophe" FYI, or maybe what the apostrophe custom formatter is doing is OK for your audience or you're just not up for tangling with it, which is OK.

I could not find the time to test it deeply, and I'm not sure I can do that so soon, so I figured it would be better to revert it as it looks like it's actually OK for my audience as you've mentioned.

I'll have a native speaker peer review the translations and me or this person will comment here as soon as it's done.

abea:
Very exciting!
Apostrophe media library translated into Portuguese

abea (commented):
Let us know if you have any problems getting a native speaker review. I did do a review of the strings through Google Translate as well as compared to the `en` set to check for interpolation. That all looks good so far.

pietro-rutzen:
> Let us know if you have any problems getting a native speaker review. I did do a review of the strings through Google Translate as well as compared to the `en` set to check for interpolation. That all looks good so far.

No one I've talked to seemed very inspired towards helping me with that.

But I believe it looks good!

Could you get an example where the formatted dates are used?

That is the only thing I'm not so sure it's good or not.

abea:
The piece and page "last updated" columns are the main place they're used and they show a "relative" date (default: _2 days ago_). Those use the `dayjsRelativeTimeFormat` localization key. `dayjsCellDateFormat` is used in the `AposCellDate` table cell component, which would be a normal date/time (default: _12/13/2021 at 10:32am_). That component isn't actually used in the UI yet.

The "apostrophe" value simply tells the UI to use the default formats (shown above). For a localization file you would enter a [Day.js format](https://day.js.org/docs/en/display/format) to replace those defaults. For example, from using Google translate, the AposCellDate in Portuguese might be "13/12/2021 às 10h32".

For the relative time format, Day.js does have [a plugin to do relative time](https://day.js.org/docs/en/display/from-now), but we're not using it. Implementing and localizing the way we're doing it might be pretty difficult, so it might be better for us to look into switching to using this plugin instead.

Screen Shot 2021-09-24 at 11 17 46

abea (approved):

abea:
@boutell Do you want to re-review this?

boutell:
Removing myself as reviewer, @abea's got this and I'm assigned on something else this week.

PR #3436, opened 2021-09-22
PRO-1924: project level should override "improvements" of the same module name

boutell:


.gitignore
```diff
@@ -13,9 +13,9 @@ npm-debug.log
 .vscode
 .out
 coverage
-node_modules
+/node_modules
```
boutell, line 16: Be more specific

index.js
```diff
@@ -334,6 +334,10 @@ module.exports = async function(options) {
       synth.define(name, options);
     });
 
+    // Apostrophe prefers that any improvements to @apostrophecms/global
+    // be applied before any project level version of @apostrophecms/global
+    synth.applyImprovementsBeforeProjectLevel();
```
boutell, line 339: This requires a new pass be made at the end after defining the modules. The cleanest way to do it was as an optional method of moog-require that apostrophe always calls. moog-require is internal now, so this isn't something third parties ever have to think about.

lib/moog-require.js
```diff
@@ -202,5 +202,28 @@ module.exports = function(options) {
     return _.has(self.improvements, name);
   };
 
+  self.applyImprovementsBeforeProjectLevel = () => {
+    for (const [ name, definition ] of Object.entries(self.definitions)) {
+      // At this stage the complete definition of a type is a linked list of `extend`
```
boutell, line 207: Patch that linked list!

test-lib/test.js
```diff
@@ -1,9 +1,16 @@
 const fs = require('fs-extra');
 const path = require('path');
 
-fs.removeSync(path.join(__dirname, '/../test/node_modules'));
-fs.mkdirSync(path.join(__dirname, '/../test/node_modules'));
-fs.symlinkSync(path.join(__dirname, '/..'), path.join(__dirname, '/../test/node_modules/apostrophe'), 'dir');
+const testNodeModules = path.join(__dirname, '/../test/node_modules');
```
boutell, line 4: We are now able to ship tests of what would happen if certain code existed in `node_modules` during unit tests, necessary to adequately test this scenario. We do it by putting it in `test/extra_node_modules`. That gets brought into the `test/node_modules` folder along with the symlink to apostrophe itself, which was already being built at test startup time.

test-lib/test.js
```diff
@@ -12,15 +19,19 @@ const packageJson = path.join(__dirname, '/../test/package.json');
...
-  "devDependencies": {
-    "test-bundle": {}
-    }
-}`);
+  devDependencies: {
+    'test-bundle': '1.0.0'
+  }
+};
+for (const dir of dirs) {
+  packageJsonInfo.dependencies[dir] = '1.0.0';
```
boutell, line 32: Satisfy the "must be a real dependency" rule for our test dependencies.

test/extra_node_modules/improve-global/index.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  improve: '@apostrophecms/global',
+  options: {
+    testGlobalLevelLoaded: true,
```
boutell, line 4: One flag just to say this even loaded, and another to try to change a value which we may or may not override as a test.

boutell:
Marked as a draft just to prevent release today since it hasn't been through QA, next iteration is fine.

abea (approved):

PR #3437, opened 2021-09-22
Bypasses field validation if the field is conditionally hidden

abea:
Fixes a bug where conditional and required fields that had not had their condition met where throwing `'required'` errors in the UI.

So in this scenario, if the `radioFields` was left null (or not `med`) the `colorField`  was throwing an error. No more.

```
      radioFields: {
        label: 'Radio',
        type: 'radio',
        choices: [
          {
            label: 'Small',
            value: 'small'
          },
          {
            label: 'Medium',
            value: 'med'
          },
          {
            label: 'Big',
            value: 'big'
          }
        ]
      },
      colorField: {
        type: 'color',
        label: 'Color',
        required: true,
        if: {
          radioFields: 'med'
        }
      },
```

### DIFF
--- a/modules/@apostrophecms/permission/ui/apos/components/AposInputRole.vue
+++ b/modules/@apostrophecms/permission/ui/apos/components/AposInputRole.vue
@@ -126,7 +126,6 @@ export default {
       const list = document.createElement('ul');
       const intro = document.createElement('p');
       const followUp = document.createElement('p');
-      const link = document.createElement('a');
       intro.appendChild(document.createTextNode(this.$t('apostrophe:piecePermissionsIntro')));
       followUp.appendChild(document.createTextNode(this.$t('apostrophe:piecePermissionsPieceTypeList')));
       html.appendChild(intro);
@@ -137,7 +136,10 @@ export default {
         list.appendChild(li);
       });
       html.appendChild(list);
-      return { content: html, localize: false };
+      return {
+        content: html,
+        localize: false
+      };
     },
     validate(value) {
       if (this.field.required && !value.length) {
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/permission/ui/apos/components/AposInputRole.vue
```diff
@@ -137,7 +136,10 @@ export default {
         list.appendChild(li);
       });
       html.appendChild(list);
-      return { content: html, localize: false };
+      return {
```
abea, line 139: eslint

modules/@apostrophecms/permission/ui/apos/components/AposInputRole.vue
```diff
@@ -126,7 +126,6 @@ export default {
       const list = document.createElement('ul');
       const intro = document.createElement('p');
       const followUp = document.createElement('p');
-      const link = document.createElement('a');
```
abea, line 129: not used, throwing an eslint error in the IDE, though our eslint checker clearly doesn't check Vue files.

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
```diff
@@ -22,6 +22,10 @@ export default {
       type: Object,
       required: false
     },
+    conditional: {
```
abea, line 25: This could be `conditionMet` or something to be more descriptive.

abea: I'll change that to be clearer.

PR #3442, opened 2021-09-22
Fix/array input l10n

abea:
Fixes stray l10n string

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
@@ -40,7 +40,7 @@ export default {
   },
   computed: {
     editLabel () {
-      return `Edit ${this.field.label}`;
+      return `Edit ${this.$t(this.field.label)}`;
     }
   },
   methods: {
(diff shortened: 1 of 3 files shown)


abea:
Not meant to merge until next release.

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -40,7 +40,7 @@ export default {
   },
   computed: {
     editLabel () {
-      return `Edit ${this.field.label}`;
+      return `Edit ${this.$t(this.field.label)}`;
```
boutell, line 43: The whole thing needs l10n! See how other strings with interpolation are set up.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
boutell: i.e. no use if we don't l10n "Edit"

abea: oh roit

boutell (approved):

PR #3447, opened 2021-09-24
PRO-2055 (the core bit)

boutell:
A small addition so that the `context-editing` event can explicitly manage the "hey the user is currently typing / manipulating, chill on the UI changes" state when desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 * Piece types with relationships to multiple other piece types may now be configured in any order, relative to the other piece types. This sometimes appeared to be a bug in reverse relationships.
 * Code at the project level now overrides code found in modules that use `improve` for the same module name. For example, options set by the `@apostrophecms/seo-global` improvement that ships with `@apostrophecms/seo` can now be overridden at project level by `/modules/@apostrophecms/global/index.js` in the way one would expect.
 
+## Adds
+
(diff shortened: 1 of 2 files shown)


stuartromanek (commented):
No problem w the code, but hammering on editing isn't what the bug necessarily _felt_ like. The case changing a select field once also tripped this up.

Is there ever a case where something might fail to shut `editing` off? Should there be some larger timeout to reset the field?

boutell:
If you look at the PR you'll see what the actual fix was. This change though fixes another UX concern with the palette code, which is that it's relying on content-editing to make really sure that update button isn't available for changes it hasn't pushed yet but making no actual guarantee they'll be ready in 1100ms. With this change it can make that promise.

boutell:
Maybe a more helpful answer: the palette module emits that "hey I'm typing" event on all changes because that's simpler than checking whether the user is dragging a slider (super necesary to emit it) or just pulling down a dropdown and picking one choice. Which is fine, except it's also getting emitted when nothing has really changed at all, i.e. on blur events, which was breaking the update button by clobbering it out of the DOM between blur and click! Nasty one, but the fix in the end was as a simple check for whether a real change has been made before emitting.

However in the process I discovered several other things that could negatively impact UX and fixed them, partly in the hope they would solve the original problem, and making this event's timeline more well-defined was one of those wins along with making the renderer simple and synchronous.

stuartromanek (approved):

PR #3459, opened 2021-09-29
Spanish locale

egonzalezg9:
Generic Spanish Locale


abea (changes requested):
@egonzalezg9 I found one label that may need capitalization.

I also noticed that we had not created keys for a couple of permission strings. Please add translations for these two strings that describe user permissions:

```
  "modify": "Modify",
  "modifyOrDelete": "Modify / Delete",
```

abea:
Since those additional keys were added after this PR was added I'll merge without those. We'd welcome another PR for only those two l10n keys. I can also take care of the capitalization to get this in the upcoming release.

Thanks for your work on this @egonzalezg9

PR #3502, opened 2021-10-25
PRO-2212 job cleaning

abea:
The legacy job module code brought form A2 included options that A2 used in the UI. Before building out A3 job module UI this PR removes that legacy documentation and configuration.


abea:
The tests are failing because uploadfs has a `azure-storage` dependency that has its own `validator` dependency, which is throwing an audit warning.

boutell (approved):

PR #3515, opened 2021-10-28
Pro 2232 batch operations buttons

ValJed:
[PRO-2232](https://linear.app/apostrophecms/issue/PRO-2232/manage-batch-operations-buttons)

## Purpose
Make batchOperations buttons configurable. We can now add buttons with icons from the module definition, we can group operations in one button with an icon too. We also changed the `unlessFilter` for `if` that accepts single value or array of values. We also don't show group buttons if all operations inside are hidden.

I also added a minor fix to remove the `all` keyword when all items are not selected from the `Select All Box`.

## How to test
Go to any pieces type module, add batch operations, group them, add and modify `if` values to test filters conditions.


modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
```diff
@@ -20,12 +20,30 @@
         :icon-only="true" icon="delete-icon"
         type="outline"
```
abea, line 21: You can remove this commented code now.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -388,20 +395,61 @@ module.exports = {
...
-            ...self.batchOperations[key]
-          })).filter(batchOperation => {
-            // If a `requiredField` is registered, only include the operation
-            // if that field is present on the schema.
-            if (batchOperation.requiredField && !_.find(self.schema, {
-              name: batchOperation.requiredField
-            })) {
-              return false;
+          const groupedOperations = Object.entries(self.batchOperations)
+            .reduce((acc, [ opeName, properties ]) => {
```
abea, line 399: A smaller thing, but using `ope` to abbreviate `operation` is confusing to me. A more common abbreviation for the word is `op`. There are a few variables this applies to below.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -388,20 +395,61 @@ module.exports = {
...
+              ...groupProperties,
+              operations: [
+                ...acc[groupName]?.operations || [],
+                currentOpe
+              ]
+            };
+          }
+
+          function getAssociatedGroup (operation) {
+            return Object.entries(self.batchOperationsGroups)
```
abea, line 448: I don't see `batchOperationsGroups` anywhere else. I'm not really clear on what is happening in this other than eventually returning the empty array. The use of `_` is throwing me as well since it is assigned to lodash above.

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
```diff
@@ -20,12 +20,30 @@
...
+        v-for="{name, label, icon, operations} in showedOperations"
+        :key="name"
+      >
+        <AposButton
+          v-if="!operations"
+          label="label"
+          :icon-only="true"
+          :icon="icon"
+          type="outline"
+          @click="() => {}"
```
abea, line 33: this should have the same click handling as the context menu. So run batchAction with the action.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -388,20 +395,61 @@ module.exports = {
...
+              const requiredFieldNotFound = properties.requiredField && !self.schema
+                .some((field) => field.name === properties.requiredField);
+
+              if (requiredFieldNotFound) {
+                return acc;
+              }
+
+              const associatedGroup = getAssociatedGroup(opeName);
+              const currentOperation = {
+                name: opeName,
```
abea, line 409: The context menu component emits the `action` property. At this point I think we need to use that instead of `name`.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -388,20 +395,61 @@ module.exports = {
...
+              };
+            }, {});
+
+          self.batchOperations = Object.entries(groupedOperations)
+            .map(([ name, properties ]) => ({
+              name,
+              ...properties
+            }));
+
+          function getCurrentOrGroup (currentOpe, [ groupName, groupProperties ], acc) {
```
abea, line 432: Could we call this `getOperationOrGroup`? "Current" makes sense in the context of the reducer, but doesn't communicate what this is doing as well.

abea (changes requested):

abea:
@ValJed There is also a Node 12 test failure. Probably a newer feature that you used server side. I couldn't tell exactly where it was coming from. This happens to me every time I try to use `?.` server side.

modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -308,8 +308,10 @@
   "selectedMenuItem": "✓ {{ label }}",
   "selectAll": "Select All",
   "selectBoxMessage": "{{ num }} {{ label }} on this page selected.",
```
abea, line 310: I just noticed that a user can select all on one page, switch to the next page and select all there (one at a time) then this reads something like "20 articles on this page are selected" even if there are only 10 on the page. That fix can be a separate PR.

abea (changes requested):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
...
 * Checkboxes for pieces are back, a main checkbox allows to select all page items. When all pieces on a page are checked, a banner where the user can select all pieces appears. A launder for mongo projections has been added.
 * Registered `batchOperations` on a piece-type will now become buttons in the manager batch operations "more menu" (styled as a kebab icon). Batch operations should include a label as well as either a `route`, `schema` or `modal` property to lead into action execution.
+* `batchOperations` can now be grouped into a single button with a menu, `unlessFilter` has been changed with `if`, and allows to pass a single value or an array of values.
```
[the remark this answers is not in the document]
ValJed: Modified

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
```diff
@@ -141,24 +164,46 @@ export default {
...
-      };
-
-      return config;
     }
   },
+  mounted () {
+    this.computeShowedOperations();
+  },
   methods: {
+    computeShowedOperations () {
```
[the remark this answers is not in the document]
ValJed: I just avoid using computed properties when I don't need it. In this case we only need to compute these values when component mounts and when a filter is applied. It would add a certain amount of data to watch for Vue which we can do without in this case. What do you mean by the `the property function`?

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
```diff
@@ -129,7 +151,8 @@ export default {
         },
         status: {},
         value: { data: '' }
-      }
+      },
+      showedOperations: []
```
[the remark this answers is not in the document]
ValJed: Yes modified.

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
ValJed: Removed

modules/@apostrophecms/piece-type/index.js
ValJed: I didn't dive really deep into but it seems that it's created by the cascading system, it automatically computes operations that are in groups... I guess it works the same way for all the cascading properties. 
I will rename the `_`, I use this convention for param that are not used (eslint doesn't throw error). I renamed it to `_key`.

modules/@apostrophecms/piece-type/index.js
ValJed: Yes that's right, I actually modified it in the next PR, but I'll do this here.

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
ValJed: Modified, but has been modified in the next PR.

modules/@apostrophecms/piece-type/index.js
ValJed: Yes right, I modify it.

modules/@apostrophecms/i18n/i18n/en.json
ValJed: Well seen, I fixed it.

ValJed:
@abea Ok I saw somewhere the optional chaining `?.` in the code so I thought I was able to use it. I remove it.

modules/@apostrophecms/piece-type/index.js
ValJed: Ok, changed.

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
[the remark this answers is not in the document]
abea: That sounds fine. I've not read about the impact of using more computed properties for this kind of thing, but I see what you're saying.

Regarding moving `showOperation`, I meant that you might need to have that be directly part of `showedOperations`, not only used in it, to make the whole thing computed since `showOperation` contains the filter property references.

modules/@apostrophecms/piece-type/index.js
abea: Oh, that's right. It's built with a template literal so the specific term isn't anywhere else.

abea:
You can use `?.` in front end code since it goes through webpack. Server side we're still supporting Node 12 and it isn't compiled (maybe it should be).

modules/@apostrophecms/piece-type/index.js
```diff
@@ -388,20 +395,61 @@ module.exports = {
...
+              };
+            }, {});
+
+          self.batchOperations = Object.entries(groupedOperations)
+            .map(([ action, properties ]) => ({
+              action,
+              ...properties
+            }));
+
+          function getOperationOrGroup (currentOpe, [ groupName, groupProperties ], acc) {
```
abea, line 432: Could you change `currentOpe` to `currentOp`? That is my last little change.

abea (changes requested):

abea (approved):

PR #3523, opened 2021-11-03
Pro 2235 pieces context menu

ValJed:
[PRO-2235](https://linear.app/apostrophecms/issue/PRO-2235/as-a-developer-i-can-configure-my-module-to-add-a-new-button-to-the)

## Purpose
Adds a configurable contextual menu to piece type modules, it allows to add multiple actions to a module.

## How to test
Add a moreMenu object to any piece type module, it should add it to your manager modal contextual menu.
Ex:
image

PS: I named it `moreMenu`, I don't really like this name but found it somewhere, if you have better naming for this do not hesitate.


modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -145,6 +145,7 @@
   "imageTag": "Image Tag",
   "imageTags": "Image Tags",
   "images": "Images",
+  "importPieces": "Import {{ type }}",
```
abea, line 148: I would not bother localizing this. It will be added through the import module eventually.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -96,6 +96,13 @@ module.exports = {
       }
     }
   },
+  moreMenu: {
```
abea, line 99: Adding this as a cascade makes sense. Let's rename it to be less tied to the UI and more about the purpose. Something like `managerActions` (still a little about the UI, but less specific), simply `operations`, `utilityOperations` (maybe my favorite).

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -468,6 +459,24 @@ export default {
...
+          key: 'apostrophe:newDocType',
+          type: this.$t(this.moduleLabels.singular)
+        }
+      };
+
+      this.moreMenu.menu = [
+        ...this.relationshipField && this.moduleOptions.canEdit
+          ? [ newPiece ] : [],
+        ...this.moreMenu.menu,
+        ...Array.isArray(moreMenu) && moreMenu.length ? moreMenu : []
```
abea, line 478: I'm not sure if we need to check `moreMenu.length`. If it's 0 then you get the empty array either way.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -96,6 +96,13 @@ module.exports = {
...
+    add: {
+      import: {
```
abea, line 101: Label this `// TEMP` so we remember to remove it in the import work.

abea (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
ValJed: That's right 😅

modules/@apostrophecms/i18n/i18n/en.json
ValJed: Ok removed

modules/@apostrophecms/piece-type/index.js
ValJed: Like `utilityOperations` too because it's consistent with batchOperations 👍🏼

modules/@apostrophecms/piece-type/index.js
ValJed: Yes done

modules/@apostrophecms/piece-type/index.js
```diff
@@ -507,6 +515,18 @@ module.exports = {
                 return operations.includes(operation);
               }) || [];
           }
+        },
+        composeUtilityOperations() {
+          self.utilityOperations = Object.entries(self.utilityOperations || {})
+            .reduce((acc, [ action, properties ]) => {
```
abea, line 521: You use `reduce` well, but I don't know if I'll ever understand its pattern easily.

ValJed: Basically here I just change an object to an array, the last parameter is an empty array, that means that At my first iteration the accumulator is an empty array, and I simply return this array with the old value with a spread `...acc`, and I add a new object in this array. The `Object.entries` allows me to get the key (action) and its object (properties) to put it together in an object.
It's purpose is to iterate through an array and to reduce this array in something else. So this is a perfect way to change an object in array, or in string or in anything that is not an object.
Like map and filter it's functional oriented so it's always return a new element, that's why it fits perfectly with the spread pattern.

abea: I do understand what is happening, but I don't use it often so it usually takes me a few minutes to figure it out again.

ValJed: In this case you're right it's not necessary, a simple map is enough and more readable, changed

ValJed: Yes I understand in this case I definitely over used it.

abea: lol I wasn't trying to suggest you change it, but this works as well.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -233,27 +233,20 @@ export default {
...
-      // Add computed singular label to context menu
-      this.moreMenu.menu.unshift({
-        action: 'new',
-        label: {
-          key: 'apostrophe:newDocType',
-          type: this.$t(this.moduleLabels.singular)
-        }
-      });
-    }
+    this.batchOperations = this.flattenOperations();
```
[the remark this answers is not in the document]
abea: I'll do that in a separate PR. We can do it a different way, but right now export is broken.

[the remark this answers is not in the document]
abea: I'm removing this line since it breaks the component. `flattenOperations` does not exist in this code.

abea (approved):

abea:
Manually ran tests: passed.

PR #3525, opened 2021-11-03
PRO-2254 No attachments builder

ValJed:
## Purpose
Modifies `attachments` query builder to add a launder. REST requests can ask no attachments through query builders.
Doesn't show batch operations on relationShip modals.

## How to test
Try to select all pieces (with a large amount), it should not be as long as it was (I noticed the busy loader is blocked but didn't take time to check why).
Check that on relationship modals we don't have batch operations anymore


modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
```diff
@@ -163,6 +167,11 @@ export default {
   },
   methods: {
     computeActiveOperations () {
+      if (this.isRelationship) {
```
abea, line 170: I think we should clear the batch operations on the manager level rather than put the work in the toolbar.

abea, line 170: I think we should clear the batch operations on the manager level rather than put the work in the toolbar.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -1463,6 +1463,13 @@ module.exports = {
           }
         },
 
+        noAttachments: {
```
abea, line 1466: Following our pattern I think this should be `attachments` with `def: true`. The conditional later can explicitly look for `!== false` to add the attachments like normal.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -235,8 +235,6 @@ export default {
     await this.getPieces();
     await this.getAllPiecesTotal();
 
-    this.batchOperations = this.flattenOperations();
```
abea, line 238: This is still used in `handleBatchAction`. The purpose was to support finding the operation details from the operation name easily. We could check multiple levels of operations in `handleBatchAction` instead (top level then group operations), but we can't simply remove this.

abea (changes requested):

boutell (changes requested):
A builder is a good idea, but:

* As Bea says, it should be `attachments`, not `noAttachments`.
* However the default should be `false`, because we don't want this in every query, just the REST API queries. The API queries can invoke the new builder by chaining it in the `getRestQuery` methods.
* The code that does the actual work should migrate into the builder, as a `finalize` handler. Then anyone who wants it can do `.attachments(true)` on their query.
* The query string is then still able to override, as long as we chain `.attachments(true)` before the call to `applyBuildersSafely`.

boutell:
(I made some edits to my review, in particular it's `after`, not `finalize`, that we need here. `finalize` is for when you can do the work just by modifying criteria, `after` is for when you need the results fetched first.)

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
ValJed: Don't you think it would be easier to directly pass needed informations to `handleBatchAction` since we have everything available when confirming the modal instead of recomputing? It would avoid to compute the same thing multiple times. Maybe it would be better to merge the modal one first to integrate with.

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
ValJed: To me, since this code has impact only in the toolbar module it should stay in the concerned module, it makes the AposDocsManager component lighter.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Checkboxes for pieces are back, a main checkbox allows to select all page items. When all pieces on a page are checked, a banner where the user can select all pieces appears. A launder for mongo projections has been added.
 * Registered `batchOperations` on a piece-type will now become buttons in the manager batch operations "more menu" (styled as a kebab icon). Batch operations should include a label as well as either a `route`, `schema` or `modal` property to lead into action execution.
 * `batchOperations` can now be grouped into a single button with a menu, `unlessFilter` has been changed to `if`, and allows to pass a single value or an array of values.
+* Adds launder to attachments query builder, allows user to asks no attachments in REST route. No batch operations when on a relationShip modal.
```
boutell, line 13: How about:

* Adds support for using the `attachments` query builder in REST API calls via the query string.

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
boutell: I agree with Val, the toolbar is where the batch operations are displayed, logic concerning when not to show them seems reasonable to put here.

boutell (changes requested):

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
abea: I forgot we had the `if` check for filters below as well. Sounds good.

abea (approved):
I'm good with this once Tom's concerns are resolved.

PR #3526, opened 2021-11-04
PRO-2227 notification event

abea:
- Adds the ability to pass an event to a notification for it to emit.
- Hides the dismiss button in a notification if it's a running job. (not directly related)

-------------

- [ ] All tests must pass, including the linters. Run `npm run lint` to test code linting independently.

Thanks for contributing!


abea:
This does pass our test suite. Just not the `npm audit` part.

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -127,7 +133,10 @@ export default {
         console.error('Unable to find notification job:', this.notification.jobId);
         this.job = null;
       }
+    }
 
+    if (this.notification.event?.name) {
+      apos.bus.$emit(this.notification.event.name, this.notification.event.data);
```
boutell, line 139: This has to be preceded by an API call to remove the `event` property from the notification, so that it doesn't fire more than once as the user navigates from page to page, etc.

boutell (changes requested):

modules/@apostrophecms/notification/index.js
```diff
@@ -160,6 +160,35 @@ module.exports = {
       return self.db.deleteMany({ _id });
     }
   }),
+  apiRoutes(self) {
+    return {
+      post: {
+        'event-clear': async function (req) {
+          if (!req.body.id) {
```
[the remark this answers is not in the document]
boutell: Also, multiword route names in the piece type module are written in natural English order rather than "noun then verb". For example:

```
':_id/revert-draft-to-published': async (req) => { ... }
```

So I would recommend:

```
`:_id/clear-event`: async req => { ... }
```

boutell (changes requested):

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -127,7 +133,16 @@ export default {
...
       }
-
+    }
+    // Notifications may include events to emit.
+    if (this.notification.event?.name) {
+      // Clear the event to make sure it's only emitted once across browsers.
+      const result = await this.clearEvent(this.notification._id);
+      // The result includes the number of documents found with the event.
+      // The notification doc will only still have the event in one instance.
+      if (result.n) {
```
boutell, line 143: This is correct, but it's low-level mongo stuff, and there's no guarantee mongo won't reveal things we'd rather it didn't. It would be more appropriate to use an HTTP status code other than 200 in the case where it's already been cleared, or if there's no good one, to use a result property specific to this API rather than passing raw results from the mongo driver to the browser.

boutell (changes requested):
One small thing 🙏

boutell (changes requested):

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -138,10 +138,13 @@ export default {
...
       try {
         // Clear the event to make sure it's only emitted once across browsers.
-        await this.clearEvent(this.notification._id);
-        // The result includes the number of documents found with the event.
-        // The notification doc will only still have the event in one instance.
-        apos.bus.$emit(this.notification.event.name, this.notification.event.data);
+        const safe = await this.clearEvent(this.notification._id);
+
+        if (safe) {
+          // The result includes the number of documents found with the event.
```
boutell, line 144: Comment is no longer relevant.

boutell, line 144: Comment seems no longer accurate.

boutell (changes requested):

boutell (approved):

PR #3531, opened 2021-11-08
Fix/players run once

stuartromanek:
prevent players from running more than once in edit mode


modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
```diff
@@ -69,12 +69,13 @@ export default {
...
     },
     runPlayer() {
-      if (!this.playerOpts) {
+      if (!this.playerOpts || this.playerOpts.played) {
         return;
       }
       const el = this.$el.querySelector(this.playerOpts.selector);
       if (el && this.playerOpts.player) {
         this.playerOpts.player(el);
+        this.playerOpts.played = true;
```
boutell, line 78: This would prevent it from ever playing again for *any* widget of this type because `playerOpts` is just a reference to a shared object containing the player method for all widgets of this type.

And if we fixed that by cloning `playerOpts` that would still not be right for our needs because there might be situations where we do replace the element during the lifetime of this component, and it would be appropriate to allow the player to run again.

So attaching a property to `el` like `runPlayer` does is the right thing to do here because we need to be sure it's the same DOM element.

boutell (changes requested):

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
stuartromanek: in this context `el` is being queried fresh each `runPlayer` so that wouldn't persist, right? Do you mean add a data attr to the queried `el` that marks the play?

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
```diff
@@ -61,20 +63,26 @@ export default {
...
+        if (el && this.playerOpts.player) {
+          this.playerEl = el;
+        }
       }
-      const el = this.$el.querySelector(this.playerOpts.selector);
-      if (el && this.playerOpts.player) {
-        this.playerOpts.player(el);
+    },
+    runPlayer() {
+      if (this.playerEl && !this.playerEl.played) {
```
boutell, line 83: Yes this is the pattern — custom properties on DOM elements are allowed, and it's the same object in JS for as long as it's the same DOM element — but please namespace like we do in runPlayers with the name `aposWidgetPlayed` to avoid any conflict with future DOM properties or similar bookkeeping logic at project level.

boutell (changes requested):

boutell (approved):

PR #3534, opened 2021-11-09
PRO-2211 adds job tests

abea:

Thanks for contributing!

### DIFF
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -2,7 +2,12 @@ const _ = require('lodash');
 
 module.exports = {
   extend: '@apostrophecms/doc-type',
-  cascades: [ 'filters', 'columns', 'batchOperations', 'utilityOperations' ],
+  cascades: [
+    'filters',
+    'columns',
+    'batchOperations',
+    'utilityOperations'
+  ],
   options: {
     perPage: 10,
     quickCreate: true,
@@ -97,12 +102,12 @@ module.exports = {
     }
   },
   utilityOperations: {
-    add: {
-      // TEMP
-      import: {
-        label: 'Import pieces'
-      }
-    }
+    // add: {
+    //   // TEMP
+    //   import: {
+    //     label: 'Import pieces'
+    //   }
+    // }
   },
   batchOperations: {
     add: {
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -295,8 +270,6 @@ module.exports = {
           }
           return self.publish(req, draft);
         },
-        // TEMP - This works fine, but should be reviewed during work actually
-        // focused on batch archive/restore.
         async archive (req) {
```
boutell, line 273: This was flagged temp and now it's not, so I'm calling it here, although I should apologize for not spotting it earlier when it was first PR'd in:

The iteration function here is going straight to the database, which is not right because the afterArchive event will not fire. Also, permissions are skipped.

The correct approach is to find the document via the module, set the archived property of the document ot true, and then call `update` through Apostrophe's model layer. The model layer will automatically figure out if it's newly archived and emit appropriate events.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -316,8 +289,6 @@ module.exports = {
             }
           );
         },
-        // TEMP - This works fine, but should be reviewed during work actually
-        // focused on batch archive/restore.
         async restore (req) {
```
boutell, line 292: Same issue as above, must use Apostrophe's model layer so all lifecycle events fire.

test-lib/util.js
```diff
@@ -57,8 +57,40 @@ async function create(options) {
   return require('../index.js')(config);
 }
 
+// Create and admin user. By default the username and password are both 'admin'
```
boutell, line 60: and -> an

test/job.js
```diff
@@ -0,0 +1,229 @@
...
+
+  let jobTwo;
+  it('can run a batch job', async function () {
+    const req = apos.task.getReq();
+
+    jobTwo = await jobModule.runBatch(
+      req,
+      articleIds,
+      async function(req, id) {
+        await apos.doc.db.updateOne({
```
boutell, line 107: This is fine because we're just testing the jobs logic, so no need to be careful about lifecycle events.

test/job.js
```diff
@@ -0,0 +1,229 @@
...
+async function pollJob(job, { jar }) {
+  const {
+    processed,
+    total,
+    good,
+    bad
+  } = await apos.http.get(job.route, { jar });
+
+  if (processed < total) {
+    await new Promise(resolve => {
```
boutell, line 217: This is correct but you already brought in Bluebird, so use `await Promise.delay(msec)`.

boutell: Or chuck bluebird and put this in a function. You're only using it for Promise.delay (Promise.all is standard).

test/job.js
```diff
@@ -0,0 +1,229 @@
...
+  it('can insert many test articles', async function () {
+    const req = apos.task.getReq();
+
+    const promises = [];
+
+    for (let i = 1; i <= 500; i++) {
+      promises.push(insert(req, apos.modules.article, 'article', {}, i));
+    }
+
+    const inserted = await Promise.all(promises);
```
boutell, line 92: Hmm, sure. A little edgy, but nice to have at least one unit test of hitting the database as hard as the driver will let us.

boutell: If it became an issue we could use `Promise.each` with a `limit`.

boutell (changes requested):
These tests are very 💯 . I did catch an issue with the actual batch operations for archive/restore which is late, I know, but you were removing the comments saying they were preliminary so I checked them out.

modules/@apostrophecms/piece-type/index.js
abea: 🤦 duh

test/job.js
abea: 👨‍🎤

boutell (approved):

PR #3536, opened 2021-11-10
PRO-2262 batch refresh

abea:
Gets the manager refreshing following a batch action.

Thanks for contributing!

### DIFF
--- a/modules/@apostrophecms/job/index.js
+++ b/modules/@apostrophecms/job/index.js
@@ -121,7 +121,8 @@ module.exports = {
             // It's only relevant to pass a job ID to the notification if
             // the notification will show progress. Without a total number we
             // can't show progress.
-            jobId: total && job._id
+            jobId: total && job._id,
+            ids
           });
 
           return {
@@ -272,6 +273,7 @@ module.exports = {
           },
           dismiss: options.dismiss,
           jobId: options.jobId,
+          batchIds: options.ids,
           icon: req.body.messages.icon || 'database-export-icon',
           type: options.type || 'success',
           return: true
--- a/modules/@apostrophecms/job/index.js
+++ b/modules/@apostrophecms/job/index.js
@@ -98,7 +98,7 @@ module.exports = {
       // on a single type of piece.
       //
       // Notification messages should be included on a `req.body.messages` object. See `triggerNotification for details`.
-      async runBatch(req, ids, change, options) {
+      async runBatch(req, ids, change, options = {}) {
         let job;
         let notification;
         const total = ids.length;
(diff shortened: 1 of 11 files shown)


modules/@apostrophecms/job/index.js
```diff
@@ -118,7 +118,9 @@ module.exports = {
             // It's only relevant to pass a job ID to the notification if
             // the notification will show progress. Without a total number we
             // can't show progress.
-            jobId: total && job._id
+            jobId: total && job._id,
+            ids,
+            action: options.action
```
abea, line 123: we're now passing the action name in here to be used by handlers.

modules/@apostrophecms/job/index.js
```diff
@@ -276,7 +277,11 @@ module.exports = {
             type: req.body.type || req.t('apostrophe:document')
           },
           dismiss: options.dismiss,
-          jobId: options.jobId,
+          job: {
```
abea, line 280: giving us more flexibility.

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -164,7 +166,7 @@ export default {
       return result;
     },
     async pollJob () {
-      if (!this.job?.total || (this.job.processed >= this.job.total)) {
```
abea, line 167: Allow the final pass since it won't be called again.

modules/@apostrophecms/job/index.js
```diff
@@ -118,7 +118,9 @@ module.exports = {
             // It's only relevant to pass a job ID to the notification if
             // the notification will show progress. Without a total number we
             // can't show progress.
-            jobId: total && job._id
+            jobId: total && job._id,
+            ids,
```
boutell, line 122: There could be thousands of ids yeah? Pushing them all in every "in progress" notification could be a serious performance issue. How about pushing just those not previously sent in a notification?

boutell: Or how about pushing the ids only in the final notification? That could be a reasonable compromise.

All of this is making me wish batch operations still used a modal. Then it would be very predictable UX: you can't do anything (at least in this tab) until the batch operation is done at which point you can expect some things to refresh. How great an idea is it really to try to work while a batch operation is making changes all over your database?

boutell: I understand I'm probably not going to get my way on that last suggestion at least for beta, but we need to figure out how not to bang bang bang on the refresh of the piece manager.

boutell (changes requested):

modules/@apostrophecms/job/index.js
abea: There's only one progress notification per job. Once there the notification takes over  polling the job progress. This is used for the `content-changed` event.

boutell (approved):

PR #3539, opened 2021-11-11
Pro 2282 pro 2236 import button and modal (#3530)

ValJed:
* Uses total in job instead of array of IDs.
* Creates AposFile.vue component to abstract file dropzone UI.
* Uses AposFile.vue component in AposInputAttachment.vue
* Creates handleUtilityOperation method in AposDocsManager.vue to trigger a new modal or the confirmation modal.

This code has been reviewed already, just resolved two minor conflicts.

### DIFF
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -96,14 +96,6 @@ module.exports = {
       }
     }
   },
-  utilityOperations: {
-    add: {
-      // TEMP
-      import: {
-        label: 'Import pieces'
-      }
-    }
-  },
   batchOperations: {
     add: {
       archive: {
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -101,10 +101,7 @@ module.exports = {
       }
     }
   },
-<<<<<<< HEAD
-=======
   utilityOperations: {},
->>>>>>> origin/main
   batchOperations: {
     add: {
       archive: {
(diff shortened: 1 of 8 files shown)


boutell (dismissed):

modules/@apostrophecms/piece-type/index.js
```diff
@@ -101,7 +101,10 @@ module.exports = {
       }
     }
   },
+<<<<<<< HEAD
```
abea, line 104: commit merge artifacts

abea (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -258,6 +261,53 @@ export default {
...
+      }
+
+      const {
+        modal, body, ...modalOptions
+      } = operation.modalOptions || {};
+
+      if (modal) {
+        await apos.modal.execute(modal, {
+          action: this.moduleOptions.action,
+          route: operation.route,
```
abea, line 279: We need to decide if we're going to support the route option here before this gets published. We removed it for batch operations.

abea: I would suggest we remove it and use the action. We can always add this later for more flexibility.

abea (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
boutell: I agree. Remove it for now. Later it can be the default value of the option.

abea (commented):
I've made changes to address my concerns.

boutell (approved):

PR #3548, opened 2021-11-17
Select all adjustments

stuartromanek:
Nice work on this @ValJed, some minor style and behavior adjustments, notably the Select All box never appears if **all** pieces are displayed on page 1 of the manager

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManagerSelectBox.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManagerSelectBox.vue
@@ -4,26 +4,26 @@
     :duration="300"
   >
     <div
-      v-if="selectedState === 'checked' || allPiecesSelection.isSelected"
+      v-show="selectedState === 'checked' && allPiecesSelection.total > displayedItems"
       class="apos-select-box"
     >
       <div class="apos-select-box__content">
         <p class="apos-select-box__text">
           {{ selectBoxMessage }}
-          <button
+          <AposButton
             v-if="!allPiecesSelection.isSelected"
-            class="apos-select-box__select-all"
+            type="subtle" :modifiers="['inline', 'small', 'no-motion']"
+            :label="selectBoxMessageButton" class="apos-select-box__select-all"
             @click="$emit('select-all')"
-          >
-            {{ selectBoxMessageButton }}
-          </button>
-          <button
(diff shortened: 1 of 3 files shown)


ValJed (changes requested):
I think we should keep the check on `allPiecesSelection.isSelected`.
Because if you `Select all` but then you unselect just some items, you should still be able to clear the selection and to see how many pieces are currently checked.

image

stuartromanek:
@ValJed you clear them by using the Select All checkbox again and resetting the state. I think this is the correct hand off, the Select All Beyond This Page is more meant for selecting the unsee'able, not acting as a shortcut for deselecting whats on the current page

https://user-images.githubusercontent.com/1889830/143034238-0e153122-b1b9-4beb-842d-b8ab3c14b76c.mov

.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManagerSelectBox.vue
```diff
@@ -56,6 +60,18 @@ export default {
   },
   emits: [ 'select-all', 'clear-select', 'set-all-pieces-selection' ],
   computed: {
+    selectionBeyondCurrentPage() {
+      return !!this.checkedIds.filter(x => !this.currentItems.map(y => y._id).includes(x)).length;
+    },
+    showSelectAll() {
+      if (
+        this.selectionBeyondCurrentPage ||
```
ValJed, line 68: Here you can simply relace `this.selectionBeyondCurrentPage` by `this.allPiecesSelection.isSelected` and remove `selectionBeyondCurrentPage` to avoid more computation.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManagerSelectBox.vue
```diff
@@ -56,6 +60,18 @@ export default {
   },
   emits: [ 'select-all', 'clear-select', 'set-all-pieces-selection' ],
   computed: {
+    selectionBeyondCurrentPage() {
+      return !!this.checkedIds.filter(x => !this.currentItems.map(y => y._id).includes(x)).length;
```
ValJed, line 64: Could have write for more readability
```javascript
return this.checkedIds.some(x => !this.currentItems.some(y => y._id === x))
```

ValJed (changes requested):

PR #3549, opened 2021-11-18
Fix convert method

ValJed:
[PRO-2339](https://linear.app/apostrophecms/issue/PRO-2339/fixes-the-presentfieldsonly-option-of-the-convert-method)

## Purpose
`fieldsPresent` method in doc-type was broken, looking for properties in strings and wasn't returning anything.

## How to test
Check that when calling the method `self.convert` with the option `presentFieldsOnly` set to true you're getting the right properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 
 * The `apos.util.attachmentUrl` method now works correctly. To facilitate that, `apos.uploadsUrl` is now populated browser-side at all times as the frontend logic originally expected. For backwards compatibility `apos.attachment.uploadsUrl` is still populated when logged in.
 * Widget players are now prevented from being played twice by the implementing vue component.
-* `fieldsPresent` method in doc-type was broken, looking for properties in strings and wasn't returning anything.
+* The`fieldsPresent` method that is sued with the `presentFieldsOnly` option in doc-type was broken, looking for properties in strings and wasn't returning anything.
 
 ### Changes
 * Removes Apostrophe 2 documentation and UI configuration from the `@apostrophecms/job` module. These options were not yet in use for A3.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 * The search field of the pieces manager modal works properly. Thanks to [Miro Yovchev](https://github.com/myovchev) for pointing out the issue and providing a solution.
+* The`fieldsPresent` method that is used with the `presentFieldsOnly` option in doc-type was broken, looking for properties in strings and wasn't returning anything.
 
 ## 3.8.0 - 2021-11-15
 
@@ -27,7 +28,6 @@
 
 * The `apos.util.attachmentUrl` method now works correctly. To facilitate that, `apos.uploadsUrl` is now populated browser-side at all times as the frontend logic originally expected. For backwards compatibility `apos.attachment.uploadsUrl` is still populated when logged in.
 * Widget players are now prevented from being played twice by the implementing vue component.
-* The`fieldsPresent` method that is sued with the `presentFieldsOnly` option in doc-type was broken, looking for properties in strings and wasn't returning anything.
 
 ### Changes
 * Removes Apostrophe 2 documentation and UI configuration from the `@apostrophecms/job` module. These options were not yet in use for A3.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 
 * The `apos.util.attachmentUrl` method now works correctly. To facilitate that, `apos.uploadsUrl` is now populated browser-side at all times as the frontend logic originally expected. For backwards compatibility `apos.attachment.uploadsUrl` is still populated when logged in.
 * Widget players are now prevented from being played twice by the implementing vue component.
+* `fieldsPresent` method in doc-type was broken, looking for properties in strings and wasn't returning anything.
```
boutell, line 30: This should mention that the `presentFieldsOnly` option to `convert` was also broken.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -554,17 +554,9 @@ module.exports = {
...
-          if (field.type.name.substring(0, 5) === '_relationship') {
-            if (_.has(input, field.idsStorage)) {
-              output.push(field.name);
-            }
-          } else {
-            output.push(field.name);
-          }
-        }
+        return self.schema
+          .filter((field) => input[field.name])
```
boutell, line 558: This is just checking for truthiness which is not enough. `0` is a legitimate value. `_.has` should be used.

boutell (changes requested):
Please add a unit test for this option in `pieces.js`. Should be straightforward to create one similar to the `should be able to update a piece in the database` test.

CHANGELOG.md
ValJed: Done

modules/@apostrophecms/doc-type/index.js
ValJed: Right! Done

test/pieces.js
```diff
@@ -1263,4 +1263,19 @@ describe('Pieces', function() {
...
+
+    const productPiece = {
+      title: 'produce name',
+      color: 'red'
+    };
+
+    const converted = {};
+    await apos.modules.product.convert(req, productPiece, converted, { presentFieldsOnly: true });
+
+    assert(Object.keys(converted).length === 3);
```
ValJed, line 1277: 3 because there is a `copyOfId`  property always added I'm not sure if it's still used or a relic from A2.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 
 * The `apos.util.attachmentUrl` method now works correctly. To facilitate that, `apos.uploadsUrl` is now populated browser-side at all times as the frontend logic originally expected. For backwards compatibility `apos.attachment.uploadsUrl` is still populated when logged in.
 * Widget players are now prevented from being played twice by the implementing vue component.
+* The`fieldsPresent` method that is sued with the `presentFieldsOnly` option in doc-type was broken, looking for properties in strings and wasn't returning anything.
```
boutell, line 30: sued -> used

boutell, line 30: sued -> used

test/pieces.js
boutell: Yes copyOfId is fine, although it would be nice if we did not set it at all when it would be false/null/undefined.

CHANGELOG.md
abea, line 30: "sued" should be "used"

test/pieces.js
```diff
@@ -1263,4 +1263,19 @@ describe('Pieces', function() {
     assert(fs.readFileSync(path.join(__dirname, 'public', resume.attachment._url), 'utf8') === fs.readFileSync(path.join(__dirname, '/public/static-test.txt'), 'utf8'));
   });
 
+  it('should convert a piece keeping only the present fields', async () => {
+    const req = apos.task.getReq();
+
+    const productPiece = {
```
boutell, line 1269: This is a little bit confusing. productPiece is the input, not the existing piece, while "converted" is the existing piece.

So please make that change and give "existingPiece" some existing values, especially one for a property that is not present in "input", and verify that property does not get lost after convert.

boutell (changes requested):

CHANGELOG.md
abea: This is currently in the 3.8.0 section. It should be in the Unreleased section.

ValJed: Oups yes thanks.

test/pieces.js
ValJed: Naming modified to make it clearer

test/pieces.js
ValJed: I removed the key if undefined.

boutell (approved):

PR #3551, opened 2021-11-23
PRO-2349 fix style check

boutell:


modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -101,8 +101,10 @@ export default {
         ? activeOptions.toolbar : this.defaultOptions.toolbar;
 
       activeOptions.styles = this.enhanceStyles(
-        (activeOptions.styles.length && activeOptions.styles) ||
-        this.defaultOptions.styles);
+        activeOptions?.styles?.length ?
```
abea, line 104: I don't think the first `?.` is necessary. we know that `activeOperations` is defined for sure.

abea (approved):

PR #3564, opened 2021-11-29
PRO-2357 fix "cannot set headers"

boutell:
No more "cannot set headers after they are sent to the client" and "req.res.redirect not defined" messages when handling URLs with extra trailing slashes.


modules/@apostrophecms/page/index.js
```diff
@@ -1365,6 +1365,7 @@ database.`);
       // Route that serves pages. See afterInit in
       // index.js for the wildcard argument and the app.get call
       async serve(req, res) {
+        const superEnd = req.res.end;
```
abea, line 1368: Unused variable causing an eslint error

boutell:
oops thanks

abea (approved):

PR #3570, opened 2021-12-01
Pro 2283 dismiss progress bar

ValJed:
[PRO-2283](https://linear.app/apostrophecms/issue/PRO-2283/as-an-editor-if-theres-an-error-with-my-job-it-should-cause-my)

## Purpose
When running a job, the progress notification is now available and can be dismissed from inside the job if needed.

## How to test
Can be tested with the `piece-type-importer` with [this PR](https://github.com/apostrophecms/piece-type-importer/pull/5).
It uses the passed notification to dimiss it if needed. (Some csv files to test are provided in this PR)


modules/@apostrophecms/job/index.js
```diff
@@ -240,7 +241,7 @@ module.exports = {
               setResults (_results) {
                 results = _results;
               }
-            });
+            }, notif);
```
abea, line 244: probably the same thing. Pass as part of an object. it could be read as `info` or something like that on the receiving side.

abea (changes requested):

modules/@apostrophecms/job/index.js
```diff
@@ -194,17 +194,18 @@ module.exports = {
...
-          run();
 
           // Trigger the "in progress" notification.
           notification = await self.triggerNotification(req, 'progress', {
             // It's only relevant to pass a job ID to the notification if
             // the notification will show progress. Without a total number we
             // can't show progress.
-            jobId: total && job._id,
+            jobId: job._id,
             count: total
```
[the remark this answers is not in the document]
ValJed: Yes definitely, removed.

modules/@apostrophecms/job/index.js
```diff
@@ -194,17 +194,18 @@ module.exports = {
...
           });
 
+          run(notification);
```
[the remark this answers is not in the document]
ValJed: Yes, passed as object, and passing only the notif ID.

modules/@apostrophecms/job/index.js
ValJed: Passing the info object here!

modules/@apostrophecms/job/index.js
```diff
@@ -189,38 +189,31 @@ module.exports = {
...
           };
         } catch (err) {
           self.apos.util.error(err);
-          if (job) {
-            // Not a lot we can do about this since we already
-            // stopped talking to the user
-            self.apos.util.error(err);
-          } else {
+
+          if (!job) {
```
ValJed, line 209: We was logging twice the error here, modified it.

abea (approved):

PR #3571, opened 2021-12-01
Fixes the server-side and template l10n debugging plugin

abea:
During documentation I noticed the server-side and template l10n strings were not debugged the same as UI strings. The plugin needed to be activated in the init method.

+++++++++++++++

Thanks for contributing!

### DIFF
--- a/modules/@apostrophecms/i18n/index.js
+++ b/modules/@apostrophecms/i18n/index.js
@@ -13,7 +13,7 @@ const ExpressSessionCookie = require('express-session/session/cookie');
 
 const apostropheI18nDebugPlugin = {
   type: 'postProcessor',
-  name: 'apostrophei18nDebugPlugin',
+  name: 'apostropheI18nDebugPlugin',
   process(value, key, options, translator) {
     // The key is passed as an array (theoretically to include multiple keys).
     // We confirm that and grab the primary one for comparison.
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/i18n/index.js
```diff
@@ -15,9 +15,25 @@ const apostropheI18nDebugPlugin = {
...
   name: 'apostrophei18nDebugPlugin',
   process(value, key, options, translator) {
-    // For ease of tracking down which phrases were
-    // actually passed through i18next
-    return `🌍 ${value}`;
+    // The key is passed as an array (theoretically to include multiple keys).
+    // We confirm that and grab the primary one for comparison.
+    const l10nKey = Array.isArray(key) ? key[0] : key;
+
+    if (value === l10nKey) {
```
abea, line 22: This is copied from the i18next.js file

modules/@apostrophecms/i18n/index.js
```diff
@@ -73,7 +89,12 @@ module.exports = {
     if (self.show) {
       self.i18next.use(apostropheI18nDebugPlugin);
     }
-    await self.i18next.init();
+
+    const i18nextOptions = self.show ? {
+      postProcess: 'apostrophei18nDebugPlugin'
```
boutell, line 94: Do you have a second to fix the capitalization to be consistent (the I should be capitalized in all versions of the name apostropheI18nDebugPlugin)? I know it's not a new problem but we're perpetuating it.

boutell (changes requested):

boutell (approved):

PR #3574, opened 2021-12-02
PRO-2368: fix for widget players running before sub-widgets are in the DOM when you click the edit button

boutell:
This led to problems for an enterprise client with the reasonable expectation that by the time the widget player for a parent widget runs, the markup for child widgets should at least be present in the DOM, as it is on a normal page load.


modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
```diff
@@ -22,17 +22,11 @@ export default {
...
-      playerOpts: null,
-      playerEl: null
+      rendered: '...'
     };
   },
   mounted() {
     this.renderContent();
-    this.playerOpts = apos.util.widgetPlayers[this.type] || null;
-  },
-  updated () {
```
boutell, line 34: I couldn't find any invocation of this.

modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -3,27 +3,38 @@ import { klona } from 'klona';
...
   });
 
+  function prepareAreasAndRunPlayersWhenDone() {
+    prepareAreas();
+    if (widgetsRendering === 0) {
+      apos.util.runPlayers();
+    }
+  }
+
   function prepareAreas() {
```
stuartromanek, line 34: this fn is now just a wrapper to run `createAreaApps` .. can we just run `createAreaApps` directly inside `prepareAreasAndRunPlayersWhenDone` to reduce this scaffolding?

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
stuartromanek: this is a vue lifecycle hook, not something we invoke directly. You're sure it can go?

stuartromanek (changes requested):
As a refresher for me, the AposAreas app is only run once per, what, page?

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
boutell: Ahhh I missed that, but we always call `renderContent` on a change of value, so I think we're OK without it.

stuartromanek (approved):
verified behavior against a client project, looks good +1 for simplifying this `:)`

boutell:
eslint caught some dead code.

stuartromanek (approved):

PR #3576, opened 2021-12-06
github action to replace circleci, borrowed from my earlier work on apostrophe-static-i18n

boutell:
We're using a third party action for mongodb support, not that it would be very hard to do our own using "m", but it's there.

Works great, currently picks up an incompatibility that's breaking our unit tests in node 16.


boutell:
I guess we don't merge this until that other bug is fixed, but it's a real bug this PR enabled us to detect.

.github/workflows/main.yml
```diff
@@ -0,0 +1,45 @@
...
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
abea, line 22: ❤️

abea (commented):
It looks great. It's a little hard to fully ✅ before the tests are passing.

boutell:
@abea this now passes everything except circleci which fails for reasons external to the PR (github is still configured to try to run it, and it doesn't find its folder, which, well, it shouldn't after this PR is merged). Pls review. Thanks!

abea (approved):

PR #3579, opened 2021-12-07
remove circleci from readme

boutell:


abea (changes requested):
we should replace this with one for the action test status. I think this is the relevant information https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

abea (approved):

boutell:
@abea I did this per the docs. It says "no status" when I peep it in the branch, but I'm hoping that will change in the next merge to main. If not we will have to fuss with it more.

PR #3585, opened 2021-12-10
Adds a field option to put a field last in its group

abea:
We didn't have a way to put fields at the end of a widget's schema UI since we can't put them in their own group. This new field option, `last: true` puts the field at the end of its group (for any schema). This is especially useful for improving modules, which don't know the rest of the schema to do reordering. This also parallels the similar field group option.

-------------------

Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Adds
 
-* Adds a `last` option to fields. Setting `last: true` on a field puts that field at the end of the field's group order (unless more than one field has that option active).
+* Adds a `last` option to fields. Setting `last: true` on a field puts that field at the end of the field's widget order. If more than one field has that option active the true last item will depend on general field registration order. If the field is ordered with the `fields.order` array, that specified order will take precedence.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Adds
+
+* Adds a `last` option to fields. Setting `last: true` on a field puts that field at the end of the field's group order (unless more than one field has that option active).
+
 ## 3.9.0 - 2021-12-08
 
 ### Adds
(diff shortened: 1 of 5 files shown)


abea:
Bumping this for review @boutell

boutell (changes requested):
What about the `order` subproperty of `fields`? It's intended for this use case (the type doesn't use groups but does benefit from ordering). Perhaps this has not been documented yet?

As for other use cases, you can impose an order when you use `group`. And nothing gets into a group without an order being specified in `group`.

So I'm not sure when it would be the right practice to use this, exactly.

abea:
@boutell The idea behind this is the same as the `last` option for groups. Unless I'm wrong, improving modules would typically (or at least sometimes) run before project code, so there would not be a way to ensure the field is at the bottom of the list (by checking the fields length, for example). Just like the SEO module wants to make sure its field group is last (or one of the last), but we don't want to make it do a bunch of `init` work to do that, the same is true here for improving module fields.

abea:
@boutell This is updated to let `order` take precedence.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 
 ### Adds
 
-* Adds a `last` option to fields. Setting `last: true` on a field puts that field at the end of the field's group order (unless more than one field has that option active).
+* Adds a `last` option to fields. Setting `last: true` on a field puts that field at the end of the field's widget order. If more than one field has that option active the true last item will depend on general field registration order. If the field is ordered with the `fields.order` array, that specified order will take precedence.
```
boutell, line 13: The docs don't speak to what happens if the project-level developer configures the field in an explicit group at project level (should be the same outcome, an explicit order mentioning the field beats "last").

boutell:
```
fields: {
  Groups: {
     seo: {
       // if 'two' was last: true it doesn't matter because the developer now has a preference
       fields: [ 'one', 'two', 'three' ]
    }
  }
}  
```

boutell (changes requested):

boutell (approved):

PR #3588, opened 2021-12-10
Applies slug field default values to their slugify options

### LINKED ISSUE #3586: Impossible to have slug fields that stay empty
PLEASE NOTE: make sure the bug exists in the latest patch level of the project. For instance, if you are running a 2.x version of Apostrophe, you should use the latest in that major version to confirm the bug.

## To Reproduce

1. Create a schema (on a piece type, for example) with a `slug` type field separate from any doc type slug. It should not follow any other fields.
2. Include `def: ''` on the field.
3. Edit that document in the UI and don't enter anything in the slug field.

## Expected behavior
The field value should be empty. Either `null` or an empty string.

## Describe the bug
The string "none" is saved instead. This is because the slug schema field settings run the empty entry through sluggo without passing the field's default value as the sluggo string default.

Arguably the field should stay empty regardless of the field's `def` value (including no value) unless it is following another field.

abea:
Closes #3586

------------------

Thanks for contributing!

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -142,11 +142,14 @@ module.exports = {
       // leading slash required). Otherwise, expect a object-style slug
       // (no slashes at all)
       convert: function (req, field, data, destination) {
-        const options = {};
+        const options = {
+          def: field.def !== undefined ? field.def : undefined
+        };
         if (field.page) {
           options.allow = '/';
         }
         destination[field.name] = self.apos.util.slugify(self.apos.launder.string(data[field.name], field.def), options);
+
         if (field.page) {
           if (!(destination[field.name].charAt(0) === '/')) {
             destination[field.name] = '/' + destination[field.name];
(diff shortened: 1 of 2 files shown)


abea:
I realize this might seem like a BC break. It really seems like a bug if "none" starts populating in fields that had no value.

modules/@apostrophecms/schema/index.js
```diff
@@ -142,11 +142,14 @@ module.exports = {
       // leading slash required). Otherwise, expect a object-style slug
       // (no slashes at all)
       convert: function (req, field, data, destination) {
-        const options = {};
+        const options = {
+          def: field.def !== undefined ? field.def : undefined
```
boutell, line 146: Seems like this must be equivalent to `def: field.def` since you handle undefined by passing undefined.

boutell (changes requested):

modules/@apostrophecms/schema/index.js
abea: oh, true.

boutell (approved):

PR #3591, opened 2021-12-15
Part of PRO-2387: supporting changes for content-security-policy

boutell:
See the PR for the security-headers module for testing details.

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/apps/AposAdminBar.js
+++ b/modules/@apostrophecms/admin-bar/ui/apos/apps/AposAdminBar.js
@@ -10,7 +10,13 @@ export default function() {
           return window.apos;
         }
       },
-      template: '<component :is="`TheAposAdminBar`" :items="apos.adminBar.items" />'
+      render: function (h) {
+        return h('TheAposAdminBar', {
+          props: {
+            items: apos.adminBar.items
+          }
+        });
+      }
     });
   }
 };
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -752,8 +752,7 @@ module.exports = {
         if (!self.shouldRefreshOnRestart()) {
           return '';
         }
-        const script = fs.readFileSync(path.join(__dirname, '/lib/refresh-on-restart.js'), 'utf8');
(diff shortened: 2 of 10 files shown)


modules/@apostrophecms/admin-bar/ui/apos/apps/AposAdminBar.js
```diff
@@ -10,7 +10,13 @@ export default function() {
           return window.apos;
         }
       },
-      template: '<component :is="`TheAposAdminBar`" :items="apos.adminBar.items" />'
+      render: function (h) {
```
boutell, line 13: `h` is a Vue convention frequently seen in the documentation of the render function, so I'm following that convention. The verbose name of this argument is `createElement`.

modules/@apostrophecms/asset/index.js
```diff
@@ -752,8 +752,7 @@ module.exports = {
         if (!self.shouldRefreshOnRestart()) {
           return '';
         }
-        const script = fs.readFileSync(path.join(__dirname, '/lib/refresh-on-restart.js'), 'utf8');
-        return self.apos.template.safe(`<script data-apos-refresh-on-restart="${self.action}/restart-id">\n${script}</script>`);
+        return self.apos.template.safe(`<script data-apos-refresh-on-restart="${self.action}/restart-id" src="${self.action}/refresh-on-restart"></script>`);
```
boutell, line 755: Refactored to use a route to deliver this script so that we don't run afoul of the `unsafe-inline` content security policy rule.

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -28,7 +28,7 @@ module.exports = ({
     optimization: {
       minimize: process.env.NODE_ENV === 'production'
     },
-    devtool: 'eval-source-map',
+    devtool: 'source-map',
```
boutell, line 31: Without this change Vue generates eval() calls. I did a git log search and was unable to find any evidence that we ever tried `source-map` before trying `eval-source-map`, which I believe is for really jurassic bc when debugging.

abea (approved):

PR #3596, opened 2021-12-21
PRO-2398: necessary fixes and a new event to enable passport-bridge module

boutell:
This can wait for after the break. It is a draft just to ensure it does not get inadvertently released tomorrow.


modules/@apostrophecms/page/index.js
```diff
@@ -1977,14 +1977,13 @@ database.`);
         );
       },
       // Returns the effective base URL for the given request.
-      // If Apostrophe's top-level `baseUrl` option is set, it is returned,
-      // otherwise the empty string. This makes it easier to build absolute
+      // If Apostrophe's top-level `baseUrl` option is set, or a hostname for the
```
abea, line 1980: "or a hostname for the active locale" seems missing a verb. "or a hostname is defined for the active locale"?

abea (changes requested):
Minor cleanup bits

abea (approved):

PR #3602, opened 2022-01-03
PRO-2399 port

abea:

Thanks for contributing!


test-lib/util.js
```diff
@@ -43,12 +43,14 @@ async function create(options) {
...
+  // TODO: Remove __testDefaults references in 4.x major version or formalize
+  // intended usage with documentation.
   if (!config.__testDefaults) {
     config.modules = config.modules || {};
     const express = config.modules['@apostrophecms/express'] || {};
     express.options = express.options || {};
-    // Allow OS to choose open port
-    express.options.port = null;
+    // Allow OS to choose open port if not explicitly set.
+    express.options.port = express.options.port || null;
```
abea, line 53: If `null` is not special here I can simply delete these lines.

boutell: `null` is special here. `undefined` defaults to 3000.

boutell: See the `@apostrophecms/express` module:

```
if (process.env.PORT) {
          port = process.env.PORT;
        } else if (fs.existsSync(self.apos.rootDir + '/data/port')) {
          // Stagecoach option
          port = fs.readFileSync(self.apos.rootDir + '/data/port', 'UTF-8').replace(/\s+$/, '');
        } else {
          port = self.options.port;
        }

        if (port === undefined) {
          port = 3000;
        }

        if (typeof port === 'string') {
          port = Number.isNaN(parseInt(port)) ? port : parseInt(port);
        }
```

boutell (approved):
Correct as written, we need to continue to pass explicit null when we're not specifying a port, otherwise the `@apostrophecms/express` module defaults to 3000.

boutell (dismissed):

boutell (approved):

PR #3604, opened 2022-01-05
PRO-2410 validate widget type names in area fields

boutell:


boutell:
This is a draft only to keep it from merging before today's release.

modules/@apostrophecms/schema/index.js
```diff
@@ -2247,7 +2263,7 @@ module.exports = {
           self.apos.util.error(format(s));
         }
         function format(s) {
-          return stripIndent`
+          return stripIndents`
```
boutell, line 2266: I changed to stripIndents because stripIndent only catches the first line, which is not the intent when looking at the pattern below.

abea (approved):

abea:
is this waiting on something, @boutell ?

PR #3609, opened 2022-01-07
PRO-2173 Properly displays UI strings with colons.

### LINKED ISSUE #3475: 3.5: help/htmlHelp broken by http(s):
## To Reproduce

Add html help to a field `htmlHelp: '<a href="https://google.com">Test html</a>'`
The result in admin UI is everything before `https:` is cut off. The visible result is `//google.com">Test html`
The same applies if just`http:` or `https:` are present in both `help` and `htmlHelp`.
Example: `help: 'A link to https://google.com'` renders `//google.com` 

## Expected behavior
I guess not to do the above :)

## Details
Tested against the `main` branch, post v3.5.0 tag

abea:
## Summary

Closes #3475 
Closes PRO-2173

Use `htmlHelp` help text on a doc type field, such as:
```
htmlHelp: 'This link to the <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea" target="_blank">MDN textarea reference</a> should display properly.',
```

In the main branch this will only display the part after the colon. With this branch it will display that properly as HTML with a link. 

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/lib/i18next.js
+++ b/modules/@apostrophecms/ui/ui/apos/lib/i18next.js
@@ -26,6 +26,15 @@ export default {
       },
       appendNamespaceToMissingKey: true,
       parseMissingKeyHandler (key) {
+        // We include namespaces with unrecognized l10n keys using
+        // `appendNamespaceToMissingKey: true`. This passes strings containing
+        // colons that were never meant to be localized through to the UI.
+        //
+        // Strings that do not include colons ("Content area") are given the
+        // default namespace by i18next ("translation," by default). Here we
+        // check if the key starts with that default namespace, meaning it
+        // belongs to no other registered namespace, then remove that default
+        // namespace before passing this through to be processed and displayed.
         if (key.startsWith(`${this.defaultNS[0]}:`)) {
           return key.slice(this.defaultNS[0].length + 1);
         } else {
--- a/modules/@apostrophecms/ui/ui/apos/lib/i18next.js
+++ b/modules/@apostrophecms/ui/ui/apos/lib/i18next.js
@@ -23,6 +23,14 @@ export default {
       debug: i18n.debug,
       interpolation: {
         escapeValue: false
+      },
+      appendNamespaceToMissingKey: true,
+      parseMissingKeyHandler (key) {
+        if (key.startsWith(`${this.defaultNS[0]}:`)) {
+          return key.slice(this.defaultNS[0].length + 1);
+        } else {
+          return key;
+        }
       }
     });
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## UNRELEASED
+
+### Fixes
+
+* UI strings that are not registered localization keys will now display properly when they contain a colon (`:`). These were previously interpreted as i18next namespace/key pairs and the "namespace" portion was left out.
+
 ## 3.11.0 - 2022-01-06
 
 ### Adds


modules/@apostrophecms/ui/ui/apos/lib/i18next.js
```diff
@@ -23,6 +23,14 @@ export default {
       debug: i18n.debug,
       interpolation: {
         escapeValue: false
+      },
+      appendNamespaceToMissingKey: true,
+      parseMissingKeyHandler (key) {
+        if (key.startsWith(`${this.defaultNS[0]}:`)) {
```
boutell, line 29: This only checks for the default namespace. We must make sure it is not any of the namespaces that were pushed to the browser (think of admin UI for non-core modules). Which can be done, see these loops from the same file, note the `addResourceBundle` calls all of which take an `ns` parameter, wrap that to make note of these and you'll have something to check against that covers all the cases.

```
    for (const [ ns, phrases ] of Object.entries(i18n.i18n[i18n.locale])) {
      i18next.addResourceBundle(i18n.locale, ns, phrases, true, true);
    }
    if (i18n.locale !== i18n.defaultLocale) {
      for (const [ ns, phrases ] of Object.entries(i18n.i18n[i18n.defaultLocale])) {
        i18next.addResourceBundle(i18n.defaultLocale, ns, phrases, true, true);
      }
    }
    if ((i18n.locale !== 'en') && (i18n.defaultLocale !== 'en')) {
      for (const [ ns, phrases ] of Object.entries(i18n.i18n.en)) {
        i18next.addResourceBundle('en', ns, phrases, true, true);
      }
    }
```

boutell (changes requested):

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
abea: The bug only occurs when the key has no namespace and is therefore using the default namespace. If it's an actual l10n key registered on a namespace then it continues through the normal i18n flow and is displayed correctly as the intended text.

If we did apply this to every namespace, then I think we would break localization since it would hit i18next for conversion without the namespace information.

boutell: OK so this code runs only if the key is not found. I get that this doesn't break other namespaces. I still don't quite get what's going on with this if statement though, can you explain?

boutell (changes requested):
Yes, we know the string should be treated as having the default namespace,

modules/@apostrophecms/ui/ui/apos/lib/i18next.js
boutell: Also how will this fix play out if the argument *is* a namespaced key, but the translation just hasn't been written yet?

abea: Normally i18next displays unrecognized l10n keys by themselves. This includes strings never meant to be keys (where l10n isn't actively in use), e.g., "Not meant to be localized."

With `appendNamespaceToMissingKey: true`, both the key and the namespace are displayed. Example: "translation:Not meant to be localized"). That means that strings containing colons will be passed through fully and not chopped at the colon (such as URLs).

When no namespace is provided, the default namespace is used ("translation," by default). If a namespace is used, i18next will attempt to find the key in the namespace and display the proper value.

We don't want to print something like "translation:Not meant to be localized." So if we see the "key" starts with the default namespace, we slice that off the string and pass the rest of the string along to be displayed.

abea: > Also how will this fix play out if the argument is a namespaced key, but the translation just hasn't been written yet?

It will use the fallback locales as normal. Removing the Spanish `managePages` key entry will use the English entry.

boutell (approved):

boutell:
Cool, sounds like you managed to same some effort over my original plan.

boutell:
Cool, sounds like you managed to same some effort over my original plan.

boutell (approved):

PR #3610, opened 2022-01-07
PRO-2393 server side of PRO-2392, support for imposing custom login requirements

boutell:
Did this as TDD (Test Driven Development for anyone peeping in), feel good about the correctness as a result.

### DIFF
--- a/modules/@apostrophecms/login/index.js
+++ b/modules/@apostrophecms/login/index.js
@@ -107,8 +107,8 @@ module.exports = {
     return {
       post: {
         async login(req) {
-          const earlyRequirements = Object.fromEntries(Object.entries(self.requirements).filter((name, requirement) => requirement.phase !== 'afterPasswordVerified'));
-          const lateRequirements = Object.fromEntries(Object.entries(self.requirements).filter((name, requirement) => requirement.phase === 'afterPasswordVerified'));
+          const earlyRequirements = Object.fromEntries(Object.entries(self.requirements).filter(([ name, requirement ]) => requirement.phase !== 'afterPasswordVerified'));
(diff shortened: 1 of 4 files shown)


test/login-requirements.js
```diff
@@ -0,0 +1,324 @@
...
+    page = await apos.http.get(
+      '/',
+      {
+        jar
+      }
+    );
+
+    assert(page.match(/logged out/));
+
+
```
abea, line 259: eslint error here

modules/@apostrophecms/login/index.js
```diff
@@ -106,34 +107,118 @@ module.exports = {
...
+          const passportLogin = (user) => {
+            return require('util').promisify(function(user, callback) {
+              return req.login(user, callback);
+            })(user);
+          };
+          // Don't make verify functions worry about whether this object
+          // is present, just the value of their own sub-property
+          req.body.requirements = req.body.requirements || {};
+
+          if (req.body.incompleteToken) {
```
abea, line 122: The conditional flow would be easier to follow with the logic broken into separate functions.

abea (changes requested):
Part of the tech design is:

> TheAposLogin should support configuration of custom Vue components to be brought into play in the login application.

That's followed by some additional detail about the browser experience. I don't see a separate ticket for those hooks. What is the plan for that? Am I misunderstanding the tech design? I don't see guidance for that layer.

modules/@apostrophecms/login/index.js
boutell: Agreed, I will refactor this route. See slack for a response about the tech design and tickets etc. This ticket is purely the server side.

boutell:
Refactored so the degree of complexity per method is similar to that of the original route.

abea (approved):

PR #3615, opened 2022-01-11
PRO-2394 Custom login requirements (browser side)

boutell:
## Summary

Implements the browser side of the tech design for custom login requirements (see PRO-2392). Contact with the real thing revealed the need for one design change, the addition of a server-side "props" function for each requirement (see new revised design at the end of PRO-2392).

I bumped the documented `block` event for `beforeSubmit` requirements to possible follow-on ticket PRO-2421 because we have no immediate need for the feature.

I broke out some frontend niceties, including the fade-up of individual afterSubmit and afterPasswordVerified components, to separate ticket PRO-2420.

Check out the `login-requirements` branch of `@apostrophecms/testbed`. Log out and try logging in. You will be prompted to solve a very lame CAPTCHA (just type the letters shown), a math problem (note you are not shown the answer, only the problem), and a simple 2FA check similar in spirit to TOTP.

For the 2FA check, on the first login, you'll be shown the code. *Write it down.* After successfully logging in and logging out, on your next login attempt you will *not* be shown the code, but it is still correctly checked. This is analogous to what happens with google authenticator.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

- [N/A (already happened in PRO-2393)] The changelog is updated
- [see new ticket PRO-2419] Related documentation has been updated
- [X] Related tests have been updated (added coverage of props functions)

**Other information:**

### DIFF
--- a/modules/@apostrophecms/attachment/index.js
+++ b/modules/@apostrophecms/attachment/index.js
@@ -1114,7 +1114,7 @@ module.exports = {
               await copyOut(uploadfsPath, tempFile);
               await self.sanitizeSvg(tempFile);
               await copyIn(tempFile, uploadfsPath);
-              await self.db.update({
+              await self.db.updateOne({
                 _id: attachment._id
               }, {
                 $set: {
--- a/modules/@apostrophecms/login/index.js
+++ b/modules/@apostrophecms/login/index.js
@@ -348,7 +348,16 @@ module.exports = {
               username: req.user.username,
               email: req.user.email
             }
-          } : {})
+          } : {}),
+          requirements: Object.fromEntries(
+            Object.entries(self.requirements).map(([name, requirement]) => {
+              // server-side function should not be pushed to browser
+              const {
+                verify, ...browserRequirement
+              } = requirement;
+              return [ name, browserRequirement ];
+            })
+          )
         };
       },
 
--- a/modules/@apostrophecms/login/index.js
+++ b/modules/@apostrophecms/login/index.js
@@ -121,7 +121,7 @@ module.exports = {
             throw self.apos.error('forbidden', req.t('apostrophe:logOutNotLoggedIn'));
           }
           if (req.token) {
-            await self.bearerTokens.remove({
+            await self.bearerTokens.removeOne({
               userId: req.user._id,
               _id: req.token
             });
@@ -135,6 +135,27 @@ module.exports = {
             };
             await destroySession();
           }
+        },
+        // invokes the precheck function for the requirement specified by
+        // `body.name`. Invoked before displaying each `afterPasswordVerified`
+        // requirement. The return value of the precheck function is delivered
+        // as the API response
+        async precheck(req) {
+          const { token, user } = await self.findIncompleteTokenAndUser(req, req.body.incompleteToken);
+
+          const name = self.apos.launder.string(req.body.name);
+
+          const requirement = self.requirements[name];
+          if (!requirement) {
+            throw self.apos.error('notfound');
+          }
+          if (!requirement.precheck) {
+            return {};
+          }
+          return requirement.precheck(req, user);
+        },
+        async context(req) {
+          return self.getContext(req);
         },
         ...(self.options.passwordReset ? {
           async resetRequest(req) {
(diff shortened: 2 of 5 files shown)


modules/@apostrophecms/attachment/index.js
```diff
@@ -1114,7 +1114,7 @@ module.exports = {
               await copyOut(uploadfsPath, tempFile);
               await self.sanitizeSvg(tempFile);
               await copyIn(tempFile, uploadfsPath);
-              await self.db.update({
+              await self.db.updateOne({
```
boutell, line 1117: Unrelated: cleaned up legacy mongodb method names that produce warnings. Extremely uninteresting, but it's nice to kill warnings.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -89,15 +97,53 @@ export default {
...
+    beforeSubmitRequirements() {
+      return this.requirements.filter(requirement => requirement.phase === 'beforeSubmit');
+    },
+    activeRequirement() {
+      return (this.phase !== 'beforeSubmit') && this.requirements.find(requirement => (requirement.phase !== 'beforeSubmit') && !requirement.done);
     }
   },
-  async beforeCreate () {
+  watch: {
+    async activeRequirement() {
```
abea, line 121: I think you should be using the `newVal, oldVal` parameters here instead of looking for `this.activeRequirement` in  the function.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -89,15 +97,53 @@ export default {
...
+      return this.mounted && this.beforeCreateFinished;
+    },
+    disabled() {
+      return this.doc.hasErrors || !!this.beforeSubmitRequirements.find(requirement => !requirement.done);
+    },
+    beforeSubmitRequirements() {
+      return this.requirements.filter(requirement => requirement.phase === 'beforeSubmit');
+    },
+    activeRequirement() {
+      return (this.phase !== 'beforeSubmit') && this.requirements.find(requirement => (requirement.phase !== 'beforeSubmit') && !requirement.done);
```
abea, line 117: please break this into separate lines to be easier to review.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -89,15 +97,53 @@ export default {
...
+    loaded() {
+      return this.mounted && this.beforeCreateFinished;
+    },
+    disabled() {
+      return this.doc.hasErrors || !!this.beforeSubmitRequirements.find(requirement => !requirement.done);
+    },
+    beforeSubmitRequirements() {
+      return this.requirements.filter(requirement => requirement.phase === 'beforeSubmit');
+    },
+    activeRequirement() {
```
abea, line 116: The name here is confusing. these are `afterSubmit` requirements one at a time, yeah?  I'm confused why the `find` method isn't looking for the specific phase name via `this.phase`.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -89,15 +97,53 @@ export default {
...
+      return (this.phase !== 'beforeSubmit') && this.requirements.find(requirement => (requirement.phase !== 'beforeSubmit') && !requirement.done);
     }
   },
-  async beforeCreate () {
+  watch: {
+    async activeRequirement() {
+      if ((this.phase === 'afterPasswordVerified') && (this?.activeRequirement?.phase === 'afterPasswordVerified') && this.activeRequirement.propsRequired) {
```
abea, line 122: I find this confusing myself, but I think this is checking if `this` exists, which definitely wouldn't be necessary:
`this?.activeRequirement?.phase`

abea: still better to use the watcher's parameter instead.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -89,15 +97,53 @@ export default {
...
+              name: this.activeRequirement.name,
+              incompleteToken: this.incompleteToken
+            }
+          });
+          this.requirementProps = {
+            ...this.requirementProps,
+            [this.activeRequirement.name]: data
+          };
+        } catch (e) {
+          this.error = e.message || 'An error occurred. Please try again.';
```
abea, line 137: This error message needs to be registered as a l10n string.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -127,27 +176,121 @@ export default {
       }
       this.busy = true;
       this.error = '';
+      const activeRequirement = this.activeRequirement;
```
abea, line 179: This looks to be unused.

test/login-requirements.js
```diff
@@ -87,6 +99,19 @@ describe('Login', function() {
...
 
+    const context = await apos.http.post(
+      '/api/v1/@apostrophecms/login/context',
+      {
+        method: 'POST',
+        body: {},
+        jar
+      }
+    );
+    console.log(context);
```
abea, line 110: stray log.

test/login-requirements.js
```diff
@@ -39,6 +44,13 @@ describe('Login', function() {
                 },
                 ExtraSecret: {
                   phase: 'afterPasswordVerified',
+                  async props(req, user) {
+                    console.log('we were invoked');
```
abea, line 48: stray log.

abea (changes requested):

modules/@apostrophecms/login/index.js
```diff
@@ -348,7 +394,21 @@ module.exports = {
...
+          } : {}),
+          requirements: Object.fromEntries(
+            Object.entries(self.requirements).map(([ name, requirement ]) => {
+              // server-side functions should not be pushed to browser
+              const {
+                // eslint-disable-next-line no-unused-vars
+                verify,
+                // eslint-disable-next-line no-unused-vars
+                props,
+                ...browserRequirement
```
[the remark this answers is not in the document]
boutell: Originally everything else was going to be treated as a prop for the component. That changed when I realized we needed support for async, dynamic props and introduced the props function. So I could switch this to a "what do we want" approach instead of a "what do we not want" approach, sure.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
boutell: Both `afterSubmit` and `afterPasswordVerified` components are displayed with the "one at a time, sequentially fading up" presentation. Only the `beforeSubmit` components are displayed (or, well, allowed to be present) while the form is up. The difference between `afterSubmit` and `afterPasswordVerified` is that the actual initial password verification API call is made after the `afterSubmit` requirements have been handled. We have both of these because some are intended to reduce denial of service pressure on the login API, while others are intended as a second factor of authentication and only make sense after the user's identity is provisionally confirmed.

boutell: So `activeRequirement` is appropriate because it refers to the current singular, presented-on-its-own requirement, which might be from any phase but `beforeSubmit`.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
boutell: I added one too many `?.`s.

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
boutell: It's not new, but yes.

abea (approved):

PR #3635, opened 2022-01-21
PRO-2449 Fixes i18n processing of login throttle message

abea:
## Summary

Resolves ﻿PRO-2449. The string was using a pluralization namespace method, but due to the "(s)" part was not taking advantage of pluralization, so it does not need to be using `__ns_n()`.

1. Pull and install the https://github.com/apostrophecms/apostrophe-boilerplate/tree/temp-pro-2449 branch of apostrophe-boilerplate for A2.
2. Create a user, then attempt to log in three times with an incorrect password. Without this change, you will get a 500 error, shutting down the site. With this change, you see an error message that you'll have to wait 120 minutes.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/lib/modules/apostrophe-login/index.js
+++ b/lib/modules/apostrophe-login/index.js
@@ -455,7 +455,7 @@ module.exports = {
     }
 
     self.getThrottleLoginErr = function (req, minutes) {
-      return req.__ns_n(
+      return req.__ns(
         'apostrophe',
         'Too many login attempts. You may try again in %s minute(s).',
         minutes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Changelog
 
-## 2.220.8 (2022-01-20)
+## Unreleased
 
 ## Fixes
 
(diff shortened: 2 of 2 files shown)


lib/modules/apostrophe-login/index.js
```diff
@@ -455,7 +455,7 @@ module.exports = {
     }
 
     self.getThrottleLoginErr = function (req, minutes) {
-      return req.__ns_n(
```
abea, line 458: We're not really using `__ns_n` anywhere else. looked at the i18n docs for the related method and I don't know if our string here would work. We could change the arguments passed to the original method, though I wasn't clear on that using our wrapper method and there weren't relevant other usages to use as a guide. Ultimately using the simpler method gets this message the same place it woudl have got to before.

boutell (approved):
This being 2.x, we should do the simplest non-buggy thing that has no significant impact, and this fix appears to be that.

PR #3668, opened 2022-02-25
fix #3667

### LINKED ISSUE #3667: Parked pages' children are not localized properly
## Describe the bug

I added a parked page with children in this way:
```
{
    parkedId: '',
    type: '',
    slug: '',
    _defaults: {
      title: ''
    },
    _children: [
         {
            parkedId: '',
            type: '',
            slug: ',
            _defaults: {
               title: ''
          }
     }]
 }
```
In my project I have a few locales. It properly created pages in every locale and these pages were properly localized.

In CMS:
obraz

But I found a problem with these pages children. Pages were properly added, but they're not localized correctly. In the CMS I see
obraz

## Expected behavior

Children pages have the same localization as parents

## Details
* Apostrophe: 3.14.0
* Node.js: 17.3.0

boutell:
Also the needed migration resolves a different one-off problem for an enterprise client. Try to prevent the one-off from recurring with better locking

## Summary

* Fixes a nasty bug where use of `_children` with parked pages broke replication in multi-locale websites.
* Introduces tons of tests to make sure we avoid regressions in the parked page logic.
* The necessary migration also resolves a once-seen problem where the draft and published aposDocIds were not in sync on a single-locale site for a parked page. But we have been unable to replicate that problem on a new site or via highly creative unit tests. So there could be more work later if it ever recurs and we nail down the cause.
* Wraps a lock around more of the startup logic to further reduce the risk of race conditions.

Sync down WL, npm run dev, the sandbox site works now. (I tested this, you don't have to, we will validate it on the demo after merge.)

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/migration/index.js
+++ b/modules/@apostrophecms/migration/index.js
@@ -231,41 +231,36 @@ module.exports = {
       // Perform the actual migrations. Implementation of
       // the @apostrophecms/migration:migrate task
       async migrate(options) {
-        await self.apos.lock.lock(self.__meta.name);
         await self.emit('before');
-        try {
-          if (self.apos.isNew) {
-            // Since the site is brand new (zero documents), we may assume
-            // it requires no migrations. Mark them all as "done" but note
-            // that they were skipped, just in case we decide that's an issue later
-            const at = new Date();
-            // Just in case the db has no documents but did
-            // start to run migrations on a previous attempt,
-            // which causes an occasional unique key error if not
-            // corrected for here
-            await self.db.removeMany({});
-            await self.db.insertMany(self.migrations.map(migration => ({
-              _id: migration.name,
-              at,
-              skipped: true
-            })));
-          } else {
-            for (const migration of self.migrations) {
-              await self.runOne(migration);
-            }
+        if (self.apos.isNew) {
+          // Since the site is brand new (zero documents), we may assume
+          // it requires no migrations. Mark them all as "done" but note
+          // that they were skipped, just in case we decide that's an issue later
+          const at = new Date();
+          // Just in case the db has no documents but did
+          // start to run migrations on a previous attempt,
+          // which causes an occasional unique key error if not
+          // corrected for here
+          await self.db.removeMany({});
+          await self.db.insertMany(self.migrations.map(migration => ({
+            _id: migration.name,
+            at,
+            skipped: true
+          })));
+        } else {
+          for (const migration of self.migrations) {
+            await self.runOne(migration);
           }
-          // In production, this event is emitted only at the end of the migrate command line task.
-          // In dev it is emitted at every startup after the automatic migration.
-          //
-          // Intentionally emitted regardless of whether the site is new or not.
-          //
-          // This is the right time to park pages, for instance, because the
-          // database is guaranteed to be in a stable state, whether because the
-          // site is new or because migrations ran successfully.
-          await self.emit('after');
-        } finally {
-          await self.apos.lock.unlock(self.__meta.name);
         }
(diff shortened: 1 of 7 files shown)


modules/@apostrophecms/migration/index.js
```diff
@@ -231,41 +231,36 @@ module.exports = {
       // Perform the actual migrations. Implementation of
       // the @apostrophecms/migration:migrate task
       async migrate(options) {
-        await self.apos.lock.lock(self.__meta.name);
```
boutell, line 234: All I did here was remove the lock because I moved it up a level. Removing the try block resulted in github thinking this is very exciting but it's not.

modules/@apostrophecms/page/index.js
```diff
@@ -2187,7 +2188,47 @@ database.`);
...
+      addMisreplicatedParkedPagesMigration() {
+        self.apos.migration.add('misreplicated-parked-pages', async () => {
+          const parkedPages = await self.apos.doc.db.find({
+            parkedId: {
+              $exists: 1
+            }
+          }).toArray();
+          let locales = Object.keys(self.apos.i18n.locales);
+          if (locales[0] !== self.apos.i18n.defaultLocale) {
+            locales = [ self.apos.i18n.defaultLocale, ...locales.filter(locale => locale !== self.apos.i18n.defaultLocale) ];
```
gregvanbrug, line 2201: What you have is fine but believe you could also do something like below if you wanted to avoid filtering since you're just rearranging the array.
```
locales = [ ...new Set([ self.apos.i18n.defaultLocale, ...locales ])
```
Might not even need to do the check if you're setting the order yourself.

modules/@apostrophecms/page/index.js
```diff
@@ -2187,7 +2188,47 @@ database.`);
...
+          }
+          const parkedIds = [ ...new Set(parkedPages.map(page => page.parkedId)) ];
```
gregvanbrug, line 2203: Aren't `parkedIds` supposed to be unique?

test/parked-pages.js
```diff
@@ -58,5 +88,246 @@ describe('Parked Pages', function() {
...
+
+async function validate(apos, expected) {
+  const locales = Object.keys(apos.i18n.locales);
+  const slugs = await apos.doc.db.distinct('slug', {
+    slug: /^\//
+  });
+  slugs.sort();
+  assert.deepStrictEqual(slugs, expected);
+  const pages = await apos.doc.db.find({
+    slug: /^\//
```
gregvanbrug, line 284: Maybe only relevant to the tests but does a parked slug always start with `/`?

gregvanbrug (changes requested):
Looks good. Just couple little things to confirm.

modules/@apostrophecms/page/index.js
boutell: Right about all of the above, I didn't internalize enough everything set does for me.

modules/@apostrophecms/page/index.js
boutell: They are unique within a locale + mode combo, but intentionally not so across locales, as there's always a draft for every published version, and (for parked pages, not everything) always an fr for every en, etc.

test/parked-pages.js
boutell: Yes, only pages can be parked and "slug starts with a slash" is a guarantee we stand by for pages.

gregvanbrug (dismissed):

gregvanbrug (approved):

PR #3675, opened 2022-03-07
PRO-2601 simple selectors for QA tests

boutell:
These should allow you to get very close to what you asked for. To avoid pushing props too deeply there are some cases where you'll need a simple descendant selector too. For the username and password fields, use the existing `data-apos-field` selectors as a starting point to reach those fields by name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## UNRELEASED
+
+### Adds
+
+* `data-apos-test=""` selectors for certain elements frequently selected in QA tests, such as `data-apos-test="adminBar"`.
+
 ## 3.15.0 (2022-03-02)
 
 ### Adds
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposAdminBar.vue
@@ -1,12 +1,12 @@
 <template>
-  <div class="apos-admin-bar-wrapper" :class="themeClass">
+  <div data-apos-test="adminBar" class="apos-admin-bar-wrapper" :class="themeClass">
     <div class="apos-admin-bar-spacer" ref="spacer" />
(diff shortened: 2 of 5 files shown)


boutell:
With regard to the logout action, you can access that one as `data-apos-test="logout"` after you fire `data-apos-test="authenticatedUserMenuTrigger"`. The name of the logout one is based on the name of the action in apostrophe, again avoiding the need to pass a lot of props for this.

myovchev (changes requested):
Just the context menu item attribute needs a gentle touch. The rest is tested and works as expected.

boutell:
@myovchev I renamed it as requested. Please note that you will still need to use a parent selector (the admin bar one) to guarantee it isn't a context menu item somewhere else that is unrelated to the admin bar and has its own namespace of possible values. Otherwise we would have to do a lot of deep prop pushing just for this to be globally unique. I would prefer the slight awkwardness of a parent selector in the tests to a big change in the structure of the Vue code.

myovchev (approved):

myovchev:
> @myovchev I renamed it as requested. Please note that you will still need to use a parent selector (the admin bar one) to guarantee it isn't a context menu item somewhere else that is unrelated to the admin bar and has its own namespace of possible values. Otherwise we would have to do a lot of deep prop pushing just for this to be globally unique. I would prefer the slight awkwardness of a parent selector in the tests to a big change in the structure of the Vue code.

I'll keep the `.apos-popover` for the menu items, which is a direct ascendant of `body` and it's outside of the Vue apps. I think Vue uses something called `portals` in ReactJS and the current selector is more than enough, no need of any prop drilling. I'm almost sure the selector you've added alone should be globally unique, as only one popover at a time is written to the DOM, but I'll keep the popover root (it shouldn't be a problem as it is not a presentation class). Additionally, we always match a desired menu item by containing text, so we have double safety.

boutell:
Oh good point, I didn't think of that.

PR #3676, opened 2022-03-09
trashes moved page if parent is trashed

ValJed:
[PRO-2366/](https://linear.app/apostrophecms/issue/PRO-2366/[michelin-reported]-pages-not-listingshowing-in-page-tree-on-target)

## Summary

Since a page cannot be `trash: false` if its parent is trashed, makes the moved page trash if its parent is in every cases.

> 1. Export a not trashed page to another locale where its parent is trashed
> 2. Make sure the moved page is trashed.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## unreleased
+
+## Fixes
+
+* When moving a page, makes it trashed if its parent is trashed too.
+
 ## 2.220.9 (2022-02-04)
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


boutell (approved):
This feels good. When trashInSchema is set (when workflow is present, basically) a page might be a trashed child of a live parent, but a page should never be a live child of a trashed parent, ever. 👍 

Do we know that it fixes the bug in question following their reproduction instructions?

ETLaurent (approved):

PR #3678, opened 2022-03-14
add cache invalidate field

ETLaurent:
Part of [PRO-2576](https://linear.app/apostrophecms/issue/PRO-2576/caching-emit-etags-for-the-context-document-and-invalidate-them-based)

## Summary

- Add a `cacheInvalidatedAt` field via a migration for docs that don't have it yet, and set it to equal their `updatedAt` field value
- Set `cacheInvalidatedAt` value to the current date when saving a doc

- Run the website

👉 `add-cache-invalidated-at-field-for-*` migration should be done only once:

```
Running database migration: add-cache-invalidated-at-field
```

👉 `cacheInvalidatedAt` field should have been added to all documents from the `aposDocs` collection, and set to equal the docs' `updatedAt` field.

- Add a piece, or a page

👉 The document created (piece or page for instance) should have a `cacheInvalidatedAt` that equals the `updatedAt` field.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -90,6 +90,7 @@ module.exports = {
     self.composeSchema();
     self.apos.doc.setManager(self.name, self);
     self.enableBrowserData();
+    self.addCacheFieldMigration();
   },
   handlers(self) {
     return {
@@ -504,19 +505,19 @@ module.exports = {
       },
       // Add the "cacheInvalidatedAt" field to the documents that do not have it yet,
       // and set it to equal doc.updatedAt.
-      // The migration will apply only for the documents of the same type as
-      // the modules' that use this method, via `self.apos.migration.add`.
-      addCacheMigration() {
-        self.apos.migration.eachDoc({ type: self.__meta.name }, 5, async doc => {
-          await self.apos.doc.db.updateOne(
-            {
-              _id: doc._id,
-              cacheInvalidatedAt: { $exists: 0 }
-            },
-            {
-              $set: { cacheInvalidatedAt: doc.updatedAt }
-            }
-          );
+      addCacheFieldMigration() {
+        self.apos.migration.add(`add-cache-invalidated-at-field-for-${self.__meta.name}`, () => {
+          self.apos.migration.eachDoc({ type: self.__meta.name }, 5, async doc => {
+            await self.apos.doc.db.updateOne(
+              {
+                _id: doc._id,
+                cacheInvalidatedAt: { $exists: 0 }
+              },
+              {
+                $set: { cacheInvalidatedAt: doc.updatedAt }
+              }
+            );
+          });
         });
       },
       // Convert the untrusted data supplied in `input` via the schema and
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -502,6 +502,23 @@ module.exports = {
         }
         return self.apos.migration.addSortify(self.__meta.name, { type: self.name }, field);
       },
+      // Add the "cacheInvalidatedAt" field to the documents that do not have it yet,
+      // and set it to equal doc.updatedAt.
+      // The migration will apply only for the documents of the same type as
+      // the modules' that use this method, via `self.apos.migration.add`.
+      addCacheMigration() {
+        self.apos.migration.eachDoc({ type: self.__meta.name }, 5, async doc => {
+          await self.apos.doc.db.updateOne(
+            {
+              _id: doc._id,
+              cacheInvalidatedAt: { $exists: 0 }
+            },
+            {
+              $set: { cacheInvalidatedAt: doc.updatedAt }
+            }
+          );
+        });
+      },
       // Convert the untrusted data supplied in `input` via the schema and
       // update the doc object accordingly.
       //
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -900,6 +900,7 @@ module.exports = {
         // Setting it this way rather than setting it to published.updatedAt
         // guarantees no small discrepancy breaking equality comparisons
         draft.updatedAt = draft.lastPublishedAt;
+        draft.cacheInvalidatedAt = draft.lastPublishedAt;
         draft.updatedBy = published.updatedBy;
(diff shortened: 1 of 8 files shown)


test/page-type.js
```diff
@@ -20,13 +21,27 @@ describe('page-type', function() {
...
+          }
         }
       }
     });
 
     assert(apos && apos.__meta.name === 'apostrophe');
   });
 
+  it('should add the migration that sets the cache field', () => {
+    assert(migrations.includes('add-cache-invalidated-at-field-for-nifty-page'));
```
ETLaurent, line 42: very sketchy...

modules/@apostrophecms/doc-type/index.js
```diff
@@ -502,6 +502,23 @@ module.exports = {
         }
         return self.apos.migration.addSortify(self.__meta.name, { type: self.name }, field);
       },
+      // Add the "cacheInvalidatedAt" field to the documents that do not have it yet,
+      // and set it to equal doc.updatedAt.
+      // The migration will apply only for the documents of the same type as
+      // the modules' that use this method, via `self.apos.migration.add`.
+      addCacheMigration() {
+        self.apos.migration.eachDoc({ type: self.__meta.name }, 5, async doc => {
```
ETLaurent, line 510: should we exclude the `@apostrophecms/*` pieces?

ETLaurent: should we exclude drafts?

modules/@apostrophecms/doc-type/index.js
```diff
@@ -502,6 +502,23 @@ module.exports = {
...
+          await self.apos.doc.db.updateOne(
+            {
+              _id: doc._id,
+              cacheInvalidatedAt: { $exists: 0 }
```
ETLaurent, line 514: should we exclude drafts?

modules/@apostrophecms/doc-type/index.js
boutell: No, there's no harm in being consistent and someone else might decide to make a piece-page for them.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -883,6 +900,7 @@ module.exports = {
         // Setting it this way rather than setting it to published.updatedAt
         // guarantees no small discrepancy breaking equality comparisons
         draft.updatedAt = draft.lastPublishedAt;
+        draft.cacheInvalidatedAt = draft.lastPublishedAt;
```
boutell, line 903: 👍

boutell (changes requested):

modules/@apostrophecms/doc-type/index.js
```diff
@@ -502,6 +503,23 @@ module.exports = {
         }
         return self.apos.migration.addSortify(self.__meta.name, { type: self.name }, field);
       },
+      // Add the "cacheInvalidatedAt" field to the documents that do not have it yet,
+      // and set it to equal doc.updatedAt.
+      addCacheFieldMigration() {
+        self.apos.migration.add(`add-cache-invalidated-at-field-for-${self.__meta.name}`, () => {
```
boutell, line 509: That's a lot of migrations and the behavior doesn't differ from one doc type to another. So write just one migration in `@apostrophecms/doc` which operates on all docs.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -502,6 +503,23 @@ module.exports = {
...
+          self.apos.migration.eachDoc({ type: self.__meta.name }, 5, async doc => {
```
boutell, line 510: This is the right place to do `cacheInvalidatedAt: { $exists: 0 }`. Doing that in `updateOne` means we pay for an update call for every doc, whether it already has this property or not.

boutell (changes requested):

test/docs.js
```diff
@@ -30,6 +35,37 @@ describe('Docs', function() {
...
+        '@apostrophecms/migration': {
+          extendMethods(self) {
+            return {
+              add(_super, name, migrationFn, options) {
+                migrations.addList.push(name);
+                migrationFn();
+                return _super(name, migrationFn, options);
+              },
+              eachDoc(_super, criteria, limit, iterator) {
+                if (criteria.cacheInvalidatedAt) {
```
boutell, line 48: This is going to alter the behavior of every eachDoc call in every migration, which could be quite a lot of unexpected behavior.

Also your `add` call adds a migration for every other migration.

You should call `self.apos.migration.add` in one of the modules of your test, i.e. call that from `init` of that module or some method invoked by `init`. 

Here is an example:

https://github.com/apostrophecms/apostrophe/blob/main/modules/%40apostrophecms/area/index.js#L581-L606

boutell: Your migration will call the regular `eachDoc` method. There is no need to alter the behavior of all `eachDoc` methods for every migration, unless I am missing something.

test/docs.js
```diff
@@ -30,6 +35,37 @@ describe('Docs', function() {
...
+                  const _updateOne = self.apos.doc.db.updateOne.bind(self.apos.doc.db);
+                  self.apos.doc.db.updateOne = (filter, update, options) => {
```
boutell, line 50: This is a really big override of every mongo update call to add migrations. There is definitely no need for that. Please reach out if you have questions after reviewing some of the regular migrations in apostrophe core.

boutell (changes requested):

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1096,17 @@ module.exports = {
           }
         });
       },
+      // Add the "cacheInvalidatedAt" field to the documents that do not have it yet,
+      // and set it to equal doc.updatedAt.
+      addCacheFieldMigration() {
+        self.apos.migration.add('add-cache-invalidated-at-field', () => {
```
boutell, line 1102: 👍

test/docs.js
boutell: I see that you wrote a new migration which is correct, but this code in the tests to override all migrations in a strange way is still here. I think you meant to remove it?

boutell (changes requested):

boutell (approved):

PR #3679, opened 2022-03-15
Pro 2506 Pro 2585 modules webpack section

ValJed:
[PRO-2506](https://linear.app/apostrophecms/issue/PRO-2506/implement-new-module-format-section-for-configuring-bundles-required)
[PRO-2585](https://linear.app/apostrophecms/issue/PRO-2585/as-a-developer-i-can-contribute-custom-webpack-configuration-via-any)

## Summary
* Accepts webpack configs from modules
* Check that this config is well formatted.
* If two modules with the same config name, with take only the latest to merge it with the main config
* Tests added for the two utils methods.

> 1. `npm link` Apostrophe this branch into a project like testbed.
> 2. Add webpack configs in your modules and check there are added to the main config
> 3. It should throw error if a webpack module section is not an object or contain other properties than `bundles` and `extensions`.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/asset/lib/webpack/utils.js
+++ b/modules/@apostrophecms/asset/lib/webpack/utils.js
@@ -1,0 +1,47 @@
+const { merge: webpackMerge } = require('webpack-merge');
+
+module.exports = {
+  checkModulesWebpackConfig(modules, t) {
+    const allowedProperties = [ 'extensions', 'bundles' ];
+
+    for (const mod of Object.values(modules)) {
+      const webpackConfig = mod.__meta.webpack[mod.__meta.name];
+
+      if (!webpackConfig) {
+        continue;
+      }
+
+      if (
+        typeof webpackConfig !== 'object' ||
+        webpackConfig === null ||
+        Array.isArray(webpackConfig) ||
+        Object.keys(webpackConfig).some((prop) => !allowedProperties.includes(prop))
+      ) {
+        const error = t('apostrophe:assetWebpackConfigWarning', {
+          module: mod.__meta.name
+        });
+
+        throw new Error(error);
+      }
+    }
+  },
+  mergeWebpackConfigs (modules, config) {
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/asset/lib/webpack/utils.js
```diff
@@ -0,0 +1,47 @@
...
+          module: mod.__meta.name
+        });
+
+        throw new Error(error);
+      }
+    }
+  },
+  mergeWebpackConfigs (modules, config) {
+    const extensions = Object.values(modules).reduce((acc, mod) => {
+      const { name, webpack } = mod.__meta;
```
boutell, line 30: At this point, `webpack` could have a subproperty for every subclass that makes up the inheritance tree for `name`. That's intentional and we want all of them. So your merge should walk through all of them in order, rather than taking only `webpack[name]`.

Otherwise a base class and a subclass cannot contribute separate webpack extensions, which is allowed.

test/assets.js
```diff
@@ -21,4 +89,60 @@ describe('Assets', function() {
...
+        checkModulesWebpackConfig(modules, t);
+
+        return [ true, false ];
+      } catch (err) {
+        return [ false, err ];
+      }
+    }
+  });
+
+  it('shoud merge webpack configs from modules with the Apostrophe one', async function () {
```
boutell, line 129: The unit tests must show that if a base class and a subclass both contribute webpack extensions with different names, they both make it through, even if the base class module is never instantiated on its own.

This is similar to the way the `i18n` section allows base classes and subclasses both to contribute without conflicting with each other.

boutell (changes requested):
See comments. Need to make sure base classes and subclasses can both contribute webpack extensions (as long as the names are different) and they both come through.

boutell (approved):

PR #3683, opened 2022-03-16
handle max-age cache-control

ETLaurent:
[PRO-2573/](https://linear.app/apostrophecms/issue/PRO-2573/caching-offer-a-simple-way-to-set-max-age-for-apostrophe-page-and-get)

https://github.com/apostrophecms/apostrophe/pull/3674 was reverted, because the case when a user is logged-in and gets some logged-out content from cache was not handled.

## Summary

Add a simple way to cache pages and pieces responses.

- Add the following config to a project using A3, and `npm link` `apostrophe` (on the `2573-max-age` branch)

```diff 
diff --git a/modules/@apostrophecms/page/index.js b/modules/@apostrophecms/page/index.js
index 6580174..fe2ebbc 100644
--- a/modules/@apostrophecms/page/index.js
+++ b/modules/@apostrophecms/page/index.js
@@ -31,6 +31,18 @@ module.exports = {
         title: 'Search',
         type: '@apostrophecms/search'
       }
-    ]
+    ],
+    publicApiProjection: {
+      title: 1,
+      _url: 1
+    },
+    cache: {
+      api: {
+        maxAge: 3000
+      },
+      page: {
+        maxAge: 6000
+      }
+    }
   }
 };
diff --git a/modules/<any-piece>/index.js b/modules/<any-piece>/index.js
index 5982eb5..becd771 100644
--- a/modules/<any-piece>/index.js
+++ b/modules/<any-piece>/index.js
@@ -2,7 +2,16 @@ module.exports = {
   extend: '@apostrophecms/piece-type',
   options: {
     label: 'testNs:topicLabel',
-    pluralLabel: 'testNs:topicPlural'
+    pluralLabel: 'testNs:topicPlural',
+    publicApiProjection: {
+      title: 1,
+      _url: 1
+    },
+    cache: {
+      api: {
+        maxAge: 2000
+      }
+    }
   },
   i18n: {
     testNs: {
```

- Run the project and check that the different scenarios match with the ones described in the [issue](https://linear.app/apostrophecms/issue/PRO-2573/caching-offer-a-simple-way-to-set-max-age-for-apostrophe-page-and-get), for pieces and pages using the REST API, and pages using the regular routing

image

- Removing the cache config (`api`, `page`, or the whole `cache` config) should not add the `cache-control` value in the HTTP response

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/admin-bar/ui/apos/apps/force-refresh.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+export default () => {
+  const isLoggedInPageContent = document.body.getAttribute('data-apos-user-logged-in');
+  console.log('isLoggedInPageContent', isLoggedInPageContent);
+
+  if (isLoggedInPageContent) {
+    return;
+  }
+
+  if (apos.util.getCookie('loggedIn')) {
```
boutell, line 13: This cookie name should be namespaced, `aposLoggedIn` would be good. Usually we namespace cookies to the shortname to avoid confusion in dev since "localhost" is always the host there, but in this case the worst thing that can happen is a one-time extra refresh in a dev environment if you switch between testing two sites on localhost.

modules/@apostrophecms/admin-bar/ui/apos/apps/force-refresh.js
```diff
@@ -0,0 +1,18 @@
...
+    console.log('loggedIn cookie, RELOADING...');
+    apos.util.deleteCookie('loggedIn');
```
boutell, line 15: You could encounter more than one cached page as you move through the site, so deleting the cookie is potentially problematic. But I recognize that the cookie could get out of sync with your actual login status, so we can't keep refreshing forever in a loop. So I think this is the best we can do — as long as we also *set* the cookie server-side when not already present on every newly generated response for a logged-in user. I see you haven't done that part yet.

boutell (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/apps/force-refresh.js
ETLaurent: noted, you're right!

modules/@apostrophecms/admin-bar/ui/src/index.js
```diff
@@ -0,0 +1,26 @@
...
+  if (!isLoggedOutPageContent || !isLoggedInCookie) {
+    sessionStorage.setItem('aposRefreshedPages', '{}');
+
+    return;
+  }
+
+  const refreshedPages = sessionStorage.aposRefreshedPages ? JSON.parse(sessionStorage.aposRefreshedPages) : {};
+
+  // Avoid potential refresh loops
+  if (!refreshedPages[location.pathname]) {
```
boutell, line 18: the query string could matter, so I would use all of the URL as the key.

modules/@apostrophecms/login/index.js
```diff
@@ -151,6 +151,7 @@ module.exports = {
             expireCookie.expires = new Date(0);
             const name = self.apos.modules['@apostrophecms/express'].sessionOptions.name;
             req.res.header('set-cookie', expireCookie.serialize(name, 'deleted'));
+            req.res.cookie(`${self.apos.shortName}.loggedIn`, 'false');
```
boutell, line 154: We might make this configurable later but it's fine to hardcode it this way for now since it's neatly namespaced.

modules/@apostrophecms/login/index.js
```diff
@@ -678,6 +679,7 @@ module.exports = {
 
       // Awaitable wrapper for req.login. An implementation detail of the login route
       async passportLogin(req, user) {
+        req.res.cookie(`${self.apos.shortName}.loggedIn`, 'true');
```
boutell, line 682: Unfortunately passport modules don't always call this. So to be on the safe side just set it on every request if it is not already set and the user is logged in.

boutell: That means putting it into one of our middleware functions.

boutell (changes requested):

modules/@apostrophecms/login/index.js
ETLaurent: Ok I'll change that 👍

modules/@apostrophecms/module/index.js
```diff
@@ -177,6 +177,11 @@ module.exports = {
           return async function(req, res) {
             try {
               const result = await fn(req);
+
+              if (req.method === 'GET' && req.user) {
+                res.header('Cache-Control', 'no-store');
```
ETLaurent, line 182: I first re-used the `setMaxAge` function, but that was too much of a refactoring to fit in this case, so I decided to keep it simple (and quick), by only setting the Cache-Control directly.

But if we want the session to be checked as well, as it is done in `setMaxAge`, then `setMaxAge` will have to be re-adapted.

@boutell WDYT?

modules/@apostrophecms/template/index.js
```diff
@@ -643,6 +643,7 @@ module.exports = {
           modules: {},
           prefix: req.prefix,
           sitePrefix: self.apos.prefix,
+          shortName: self.apos.shortName,
```
ETLaurent, line 646: _ℹ️  pass the `shortName` to the frontend so it's used to retrieve the "logged-in flag" cookie_

test/pages.js
```diff
@@ -4,6 +4,7 @@ const _ = require('lodash');
 
 let apos;
 let homeId;
+const apiKey = 'this is a test api key';
```
ETLaurent, line 7: TODO: update all tests

modules/@apostrophecms/login/index.js
```diff
@@ -817,6 +821,17 @@ module.exports = {
             return next();
           }
         }
+      },
+      addLoggedInCookie: {
+        before: '@apostrophecms/i18n',
```
ETLaurent, line 826: needed before `'@apostrophecms/i18n'`, like the other middleware from this module?

modules/@apostrophecms/module/index.js
boutell: It's OK to check just `req.user` for this.

boutell (approved):

modules/@apostrophecms/module/index.js
ETLaurent: tests for this added in f9566b2ff

modules/@apostrophecms/template/index.js
ETLaurent: tests added in f9566b2ff

modules/@apostrophecms/template/index.js
```diff
@@ -656,6 +657,7 @@ module.exports = {
             _id: req.user._id,
             username: req.user.username
           };
+          self.addBodyDataAttribute(req, 'apos-user-logged-in', true);
```
ETLaurent, line 660: tests added in f9566b2ff

modules/@apostrophecms/login/index.js
```diff
@@ -151,6 +152,9 @@ module.exports = {
             expireCookie.expires = new Date(0);
             const name = self.apos.modules['@apostrophecms/express'].sessionOptions.name;
             req.res.header('set-cookie', expireCookie.serialize(name, 'deleted'));
+
+            // TODO: get cookie name from config
+            req.res.cookie(`${self.apos.shortName}.${loggedInCookieName}`, 'false');
```
ETLaurent, line 157: tested in 961152ecf

modules/@apostrophecms/login/index.js
```diff
@@ -817,6 +821,17 @@ module.exports = {
...
           }
         }
+      },
+      addLoggedInCookie: {
+        before: '@apostrophecms/i18n',
+        middleware(req, res, next) {
+          // TODO: get cookie name from config
+          const cookieName = `${self.apos.shortName}.${loggedInCookieName}`;
+          if (req.user && req.cookies[cookieName] !== 'true') {
+            res.cookie(cookieName, 'true');
```
ETLaurent, line 831: tested in 961152ecf

boutell (changes requested):
Etienne, this PR looks great. I have one request and it is because I thought of a more efficient way to do something, and I wouldn't ask but we want to keep the logged-out namespace as clean as possible 🙏 

Would you please get rid of `data-apos-logged-in` and replace it with just checking that `apos.login && apos.login.user` exists in frontend JS?

We already have that, so we can avoid the overhead of setting another header.

Otherwise good to go.

Thank you!

ETLaurent:
> Etienne, this PR looks great. I have one request and it is because I thought of a more efficient way to do something, and I wouldn't ask but we want to keep the logged-out namespace as clean as possible 🙏
> 
> Would you please get rid of `data-apos-logged-in` and replace it with just checking that `apos.login && apos.login.user` exists in frontend JS?
> 
> We already have that, so we can avoid the overhead of setting another header.
> 
> Otherwise good to go.
> 
> Thank you!

Makes total sense, changing it...

boutell (approved):
Thanks!

While it's true that `apos.user` currently exists only when logged in, it is a module not the user info, so there's a chance we'll decide later it needs to expose something. `apos.login.user` is only ever going to be the user info itself.

PR #3684, opened 2022-03-17
Pro 2493 bundles compilation

ValJed:
[PRO-2493](https://linear.app/apostrophecms/issue/PRO-2493/implement-compilation-of-separate-bundles)

## Summary

* Implements the transpilation of extra bundles
* Updates methods in utils to avoid parsing multiple times all the modules chains, this way we get the webpack extensions and the bundles config at the same time for src builds.
* Verifies that bundles entry points exists before to pass them to webpack.
* Modifies slightly the src webpack config to take extra bundles as entries, modifies outputs (for css also)  to use the entry name as the output one.
* Adds `extraBundles` to the `deployFiles` array to be copied into the release folder.
* Updates and add tests.

* Configures some webpack bundles in a A3 project (testbed).
* If the file exists for a bundle (js or scss) they should be bundled into `public/apos-frontend/default`.
* If in prod mode, these bundles should be copied to the release folder.
* Tests should pass

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/asset/index.js
```diff
@@ -360,43 +384,49 @@ module.exports = {
...
-              ).map(([ name, options ]) => {
-                const file = `${bundleDir}/${name}-build.css`;
-                if (fs.existsSync(file)) {
-                  return `/* BUILD: ${name} */\n` + fs.readFileSync(file, 'utf8');
-                } else {
-                  return '';
-                }
-              }).join('\n')
-            );
+            writeAssetFile({
```
ValJed, line 387: I thought originally that I had to write some files from this function for my extra bundles, that's why I refactored this part. I realized later that it wasn't needed.
Kept the new version since it avoid code redundancy and make it more readable.

modules/@apostrophecms/asset/index.js
```diff
@@ -6,7 +6,12 @@ const globalIcons = require('./lib/globalIcons');
 const path = require('path');
 const express = require('express');
 const { stripIndent } = require('common-tags');
-const { checkModulesWebpackConfig, mergeWebpackConfigs } = require('./lib/webpack/utils');
+const { merge: webpackMerge } = require('webpack-merge');
```
boutell, line 9: Nice I need to start using this

boutell (approved):
This looks great.

PR #3690, opened 2022-03-22
implement backlinks of related docs

ETLaurent:
[PRO-2575](https://linear.app/apostrophecms/issue/PRO-2575/as-a-developer-i-can-find-the-documents-that-have-relationships-to-a)

## Summary

When saving a doc, update its related documents' `relatedReverseIds` field accordingly.
For instance, when adding related topic(s) to an article, we add the article ID to the topic(s)' `relatedReverseIds` field.

This is done by the `updateBacklinks` beforeSave handler.

- Add some related docs to a doc's `relationship` field (for instance, add some topics to an article, if you follow the example from the [doc](https://v3.docs.apostrophecms.org/guide/relationships.html#the-anatomy-of-a-relationship-field))

image

- Check that for all these related docs, their `relatedReverseIds` field now have the doc ID in the DB

image

- Removing related docs should also remove the doc ID from their `relatedReverseIds` field :

image

image

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Adds
 
 * Full support for the `object` field type, which works much like `array` but stores just one sub-object as a property, rather than an array of objects.
+* To help find documents that reference related ones via `relationship` fields, implement backlinks of related documents by adding a `relatedReverseIds` field to them and keeping it up to date.
 
 # 3.16.1 (2022-03-21)
 
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -289,7 +289,7 @@ module.exports = {
           }
         };
 
-        self.apos.doc.walkThrough(req, doc, callbacks);
+        self.apos.doc.walkThrough(doc, callbacks);
 
         return relatedDocsIds;
       },
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -100,13 +100,6 @@ module.exports = {
         async updateBacklinks(req, doc) {
           const relatedDocsIds = self.getRelatedDocsIds(req, doc);
 
-          // TODO: don't forget to remove these logs:
-          console.log('---');
-          console.log('doc._id', doc._id);
-          console.log('doc.aposDocId', doc.aposDocId);
-          console.log('relatedDocsIds', relatedDocsIds);
-          console.log('---');
-
           // Remove all references to the doc
           await self.apos.doc.db.updateMany({
             relatedReverseIds: { $in: [ doc.aposDocId ] },
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -109,7 +109,8 @@ module.exports = {
 
           // Remove all references to the doc
           await self.apos.doc.db.updateMany({
-            relatedReverseIds: { $in: [ doc.aposDocId ] }
+            relatedReverseIds: { $in: [ doc.aposDocId ] },
+            aposLocale: { $in: [ doc.aposLocale, null ] }
           }, {
             $pull: { relatedReverseIds: doc.aposDocId }
           });
@@ -120,7 +121,8 @@ module.exports = {
 
           // Add doc reference to all related docs
           await self.apos.doc.db.updateMany({
-            aposDocId: { $in: relatedDocsIds }
+            aposDocId: { $in: relatedDocsIds },
+            aposLocale: { $in: [ doc.aposLocale, null ] }
           }, {
             $push: { relatedReverseIds: doc.aposDocId }
           });
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -94,9 +94,6 @@ module.exports = {
   handlers(self) {
     return {
       beforeSave: {
-        prepareForStorage(req, doc) {
-          self.apos.schema.prepareForStorage(req, doc);
-        },
         async updateBacklinks(req, doc) {
           const relatedDocsIds = self.getRelatedDocsIds(req, doc);
 
@@ -119,6 +116,9 @@ module.exports = {
           }, {
             $push: { relatedReverseIds: doc.aposDocId }
           });
+        },
+        prepareForStorage(req, doc) {
+          self.apos.schema.prepareForStorage(req, doc);
         },
         slugPrefix(req, doc) {
           const prefix = self.options.slugPrefix;
@@ -280,7 +280,9 @@ module.exports = {
         const relatedDocsIds = [];
         const handlers = {
           relationshipHandler: (doc, field) => {
-            relatedDocsIds.push(...doc[field.name].map(relatedDoc => self.apos.doc.toAposDocId(relatedDoc)));
+            if (Array.isArray(doc[field.name])) {
+              relatedDocsIds.push(...doc[field.name].map(relatedDoc => self.apos.doc.toAposDocId(relatedDoc)));
+            }
           }
         };
 
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -97,6 +97,34 @@ module.exports = {
         prepareForStorage(req, doc) {
           self.apos.schema.prepareForStorage(req, doc);
         },
+        async updateBacklinks(req, doc) {
+          const relatedDocsIds = self.getRelatedDocsIds(req, doc);
+
+          // TODO: don't forget to remove these logs:
+          console.log('---');
+          console.log('doc._id', doc._id);
+          console.log('doc.aposDocId', doc.aposDocId);
+          console.log('relatedDocsIds', relatedDocsIds);
+          console.log('---');
+
+          // Remove all references to the doc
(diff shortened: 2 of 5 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -97,6 +97,34 @@ module.exports = {
...
         },
+        async updateBacklinks(req, doc) {
+          const relatedDocsIds = self.getRelatedDocsIds(req, doc);
+
+          // TODO: don't forget to remove these logs:
+          console.log('---');
+          console.log('doc._id', doc._id);
+          console.log('doc.aposDocId', doc.aposDocId);
+          console.log('relatedDocsIds', relatedDocsIds);
+          console.log('---');
```
ETLaurent, line 108: ⚠️

modules/@apostrophecms/doc-type/index.js
```diff
@@ -97,6 +97,34 @@ module.exports = {
...
+
+          if (!relatedDocsIds.length) {
+            return;
+          }
+
+          // Add doc reference to all related docs
+          await self.apos.doc.db.updateMany({
+            aposDocId: { $in: relatedDocsIds }
+          }, {
+            $push: { relatedReverseIds: doc.aposDocId }
```
ETLaurent, line 125: Are ok to push the `aposDocId` of the doc, and not its `_id`?

modules/@apostrophecms/schema/index.js
```diff
@@ -1930,81 +1930,8 @@ module.exports = {
...
-                      fieldsById[self.apos.doc.toAposDocId(relatedDoc)] = relatedDoc._fields;
-                    }
-                  }
-                  doc[field.fieldsStorage] = fieldsById;
-                }
-              }
-            }
-          }
-        }
+        self.apos.doc.walkThrough(req, doc);
```
ETLaurent, line 1934: wanted to keep `prepareForStorage`, so that it still makes sense that we're preparing the doc when calling it elsewhere, rather than calling `self.apos.doc.walkThrough` directly.

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1093,97 @@ module.exports = {
...
+              if (Array.isArray(doc[field.name])) {
+                const relatedDocsIds = doc[field.name].map(relatedDoc => self.toAposDocId(relatedDoc));
+
+                if (Array.isArray(allRelatedDocsIds)) {
+                  allRelatedDocsIds.push(...relatedDocsIds);
+
+                  // Stop there, we only want to add the related IDs
+                  // to `allRelatedDocsIds` array reference, if provided, nothing more.
+                  return;
+                }
```
ETLaurent, line 1170: That's really how I managed to reuse the code, without changing the whole function, since `doc` is mutated a lot in this recursion.

I'd be happy to have a better way to do that

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+# UNRELEASED
+
+# Adds
+
+* To help find documents that reference related ones via `relationship` fields, implement backlinks of related documents by adding a `relatedReverseIds` field to them and keeping it up to date.
```
ETLaurent, line 7: Is it clear?

ETLaurent:
@boutell should we add a migration to fill the `relatedReverseIds` fields on every already-existing related docs, or is it ok to just fill it when docs will be saved?

CHANGELOG.md
boutell: 👍

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1097,67 @@ module.exports = {
...
         });
       },
+
+      // Iterate through the document's field, and execute the provided handlers
+      // for each array, object and relationship field type.
+      // The handlers take the `doc` and the current `field` values as arguments,
+      // letting you edit them if needed.
+      //
+      // Note that the provided array and object handlers must call the `recursiveFunc`
+      // so that every nested fields are walked through.
```
boutell, line 1107: so that every nested field is walked through.

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1097,67 @@ module.exports = {
...
+      // This this done by default, when array or object handlers are not provided.
```
boutell, line 1108: This is done by default...

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1097,67 @@ module.exports = {
...
+
+      walkThrough(
+        doc,
+        {
+          arrayHandler = (doc, field, recursiveFunc) => {
```
boutell, line 1113: I think writing an entire function inline as a default is not good for readability, please set it if the argument is null inside the method.

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1097,67 @@ module.exports = {
           }
         });
       },
+
+      // Iterate through the document's field, and execute the provided handlers
+      // for each array, object and relationship field type.
+      // The handlers take the `doc` and the current `field` values as arguments,
+      // letting you edit them if needed.
+      //
+      // Note that the provided array and object handlers must call the `recursiveFunc`
```
boutell, line 1106: Why is `recursiveFunc` necessary — can't you do that part automatically after invoking each handler? The only reason I can think of why you wouldn't do that is if one of your applications of this method does *not* want to recurse. In which case, explicitly returning `false` (not undefined) could be a way to signal that, similar to how `false` is used in some lodash functions.

CHANGELOG.md
boutell: Yes.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -253,6 +276,20 @@ module.exports = {
 
   methods(self) {
     return {
+      getRelatedDocsIds(req, doc) {
+        const relatedDocsIds = [];
+        const handlers = {
+          relationshipHandler: (doc, field) => {
```
boutell, line 282: The more I look at it, the more I think the word "Handler" is redundant here. It would be better if it were just "relationship", "object" etc.

modules/@apostrophecms/doc/index.js
boutell: I think the default handlers will wind up doing nothing because the recursion should be provided all the time after the handler is called.

boutell (changes requested):

modules/@apostrophecms/doc/index.js
ETLaurent: Normally they will be taken into account, as they should in `getRelatedDocsIds()`:

`walkThrough` is called with the `relationship` handler only, so the `array` and `object` handlers will be the default ones.

And in the case of a recursion (i.e area field type), `walkThrough` takes the provided handlers, or the default ones that are not provided.

Unless I'm missing something 🤔

modules/@apostrophecms/doc/index.js
ETLaurent: We do want to recurse, but since we want to give the possibility to call `recursiveFunc` with mutated param (`doc`, 2nd argument), as it's done in `prepareForStorage`, I think that handlers do need to provided it.

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1097,72 @@ module.exports = {
...
       },
+
+      // Iterate through the document's field, and execute the provided handlers
+      // for each array, object and relationship field type.
+      // The handlers take the `doc` and the current `field` values as arguments,
+      // letting you edit them if needed.
+      //
+      // Note that the provided array and object handlers must call `recursiveFunc`
+      // so that every nested field is walked through.
+      // This this done by default... when array or object handlers are not provided.
```
ETLaurent, line 1108: Did I get that right?

modules/@apostrophecms/doc-type/index.js
ETLaurent: Found another way, by merging the provided handlers with the default ones

boutell (changes requested):
Please see the `apos.doc.walk` method for an example of a method that calls a handler for everything of interest that it finds, recursively, but does not require that handler to invoke recursion by itself:

https://github.com/apostrophecms/apostrophe/blob/main/modules/%40apostrophecms/doc/index.js#L546-L577

modules/@apostrophecms/doc/index.js
```diff
@@ -1093,6 +1097,66 @@ module.exports = {
           }
         });
       },
+
+      // Iterate through the document fields and call the provided handlers
+      // for each item of an array, object and relationship field type.
+      walkThrough(doc, handlers) {
```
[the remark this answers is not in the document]
boutell: Later we could add support for invoking handlers for the other meta types (not today though).

boutell (changes requested):
Pattern feels really good now. Small request re: naming.

modules/@apostrophecms/doc/index.js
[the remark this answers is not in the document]
ETLaurent: `walkByMetaType` sounds good, I was also concerned about having `walk` and `walkThrough`

boutell (approved):

PR #3692, opened 2022-03-24
emit and handle etags

ETLaurent:
[PRO-2576](https://linear.app/apostrophecms/issue/PRO-2576/caching-emit-etags-for-the-context-document-and-invalidate-them-based)

## Summary

Emit `ETags` that are used to cache pages and pieces.

When the `Etag` value of a doc is unchanged, the client will use the cached version version of the page or REST GET response (of a single piece or page, i.e: `getOne()`).

`ETags` value rely on the `cacheInvalidatedAt` docs field, which is updated every time the doc or one of its related docs is updated (thanks to backlinks: the `relatedReverseIds` field).
The related docs of a doc have also their `cacheInvalidatedAt` field updated, when the doc is updated

First, ensure that HTTP cache is not disabled:

image

**Pages**

- In private window tabs, go to:
  - the GET REST API route of the homepage (`/api/v1/@apostrophecms/page/<homepage-id>`)
  - the homepage (`/`)

- Refresh the pages above, you should get a `304` status code, with the `ETag` response header matching the `If-None-Match` request header:

image

- In a non-private window, login to the website and edit the homepage
- Go back to the private window tabs and refresh the pages, you should have a `200` status code, with a mismatch between `ETag` response header and `If-None-Match` request header (ETag has changed), and of course your change ("edit 4" --> "edit 5" in this example):

image

- Refresh the pages, you should have the `304` status code back, and the `ETag` header equaling the `If-None-Match` one.

**Pieces**

In a project using apostrophe, create pieces of 2 piece types that are related, like an `article` piece that has a relationship with a `topic` piece type.

Example:

`Article 1` has 2 related topics, `Article 2` only has one:

```
Article 1 -> Topic 1
          -> Topic 2

Article 2 -> Topic 2
```

- In private window tabs, go to:
  - the GET REST API route of the `Article 1` (eg: `/api/v1/article/<article-1-id>`)
  - the piece page of `Article 1` (eg: `/articles/<article-1-slug>`)
  - the GET REST API route of the `Topic 1` and `Topic 2` (eg: `/api/v1/topic/<topic-x-id>`)
  - the piece page of `Topic 1` and `Topic 2` (eg: `/topics/<topic-x-slug>`)

- Refresh the pages above, you should get a `304` status code, with the `ETag` response header matching the `If-None-Match` request header:

image

- In a non-private window, login to the website and edit `Article 1`
- Go back to the private window tabs and refresh the pages, you should have a `200` status code, with a mismatch between `ETag` response header and `If-None-Match` request header (ETag has changed), **for `Article 1`, `Topic 1` and `Topic 2` pages**:

image

- Refresh the pages, you should have the `304` status code back, and the `ETag` header equaling the `If-None-Match` one.

- ☝️ Run the same tests as above but this time, edit `Topic 1` or `Topic 2`, not `Article 1` directly
  - 👉  If you edit `Topic 1`, then only `Topic 1` and `Article 1` should have their cache invalidated
  - 👉  If you edit `Topic 2`, then `Topic 1`, `Article 1` and `Article 2` should have their cache invalidated

- Editing a piece should invalidate the cache of the piece index page
- After the max-age period set in the cache config, the cache should be invalidated automatically and the response status code should be 200.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,6 @@ indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * Adds possibility for modules to [extend the webpack configuration](https://v3.docs.apostrophecms.org/guide/webpack.html).
 * Adds possibility for modules to [add extra frontend bundles for scss and js](https://v3.docs.apostrophecms.org/guide/webpack.html). This is useful when the `ui/src` build would otherwise be very large due to code used on rarely accessed pages.
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
-* Implement custom ETags emission. It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above. For now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular pages).
+* Implement custom ETags emission. It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above. For now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular served pages).
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Adds possibility for modules to [add extra frontend bundles for scss and js](https://v3.docs.apostrophecms.org/guide/webpack.html). This is useful when the `ui/src` build would otherwise be very large due to code used on rarely accessed pages.
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
 * Fixes deprecation warnings displayed after running `npm install`, for dependencies that are directly included by this package.
+* Implement custom ETags emission allowing caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Adds possibility for modules to [extend the webpack configuration](https://v3.docs.apostrophecms.org/guide/webpack.html).
 * Adds possibility for modules to [add extra frontend bundles for scss and js](https://v3.docs.apostrophecms.org/guide/webpack.html). This is useful when the `ui/src` build would otherwise be very large due to code used on rarely accessed pages.
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
+* Fixes deprecation warnings displayed after running `npm install`, for dependencies that are directly included by this package.
 * Implement custom ETags emission. It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above. For now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular served pages).
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,10 @@
 * Adds possibility for modules to [add extra frontend bundles for scss and js](https://v3.docs.apostrophecms.org/guide/webpack.html). This is useful when the `ui/src` build would otherwise be very large due to code used on rarely accessed pages.
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
 * Fixes deprecation warnings displayed after running `npm install`, for dependencies that are directly included by this package.
-* Implement custom ETags emission. It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above. For now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular served pages).
+* Implement custom ETags emission when `etags` cache option is enabled. [See the documentation for more information](https://v3.docs.apostrophecms.org/guide/caching.html).  
+It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) document updates, thanks to backlinks mentioned above.  
+Note that for now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular served pages).  
+The cache of an index page corresponding to the type of a piece that was just saved will automatically be invalidated. However, please consider that it won't be effective when a related piece is saved, therefore the cache will automatically be invalidated _after_ the cache lifetime set in `maxAge` cache option.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,7 @@
 * Adds possibility for modules to [extend the webpack configuration](https://v3.docs.apostrophecms.org/guide/webpack.html).
 * Adds possibility for modules to [add extra frontend bundles for scss and js](https://v3.docs.apostrophecms.org/guide/webpack.html). This is useful when the `ui/src` build would otherwise be very large due to code used on rarely accessed pages.
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
-* Fixes deprecation warnings displayed after running `npm install`, for dependencies that are directly included by this package.
-* Implement custom ETags emission allowing caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above.
+* Implement custom ETags emission. It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above. For now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular pages).
 
 ### Fixes
 
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -96,8 +96,10 @@ module.exports = {
       beforeSave: {
         async updateCacheFields(req, doc) {
           const relatedDocsIds = self.getRelatedDocsIds(req, doc);
+          console.log('relatedDocsIds', relatedDocsIds);
 
-          // Remove all references to the doc
+          // - Remove current doc reference from docs that include it
+          // - Update these docs' cache field
           await self.apos.doc.db.updateMany({
             relatedReverseIds: { $in: [ doc.aposDocId ] },
             aposLocale: { $in: [ doc.aposLocale, null ] }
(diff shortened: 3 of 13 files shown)


modules/@apostrophecms/page/index.js
```diff
@@ -1393,17 +1394,43 @@ database.`);
...
+
+      shouldSendUnmodifiedResponse(req) {
+        if (!self.options.cache || !self.options.cache.page || !self.options.cache.page.maxAge || !req.headers || !req.headers['if-none-match']) {
+          return false;
+        }
+
+        const releaseId = self.apos.asset.getReleaseId();
+        const cacheInvalidatedAtTimestamp = (new Date(req.data.page.cacheInvalidatedAt)).getTime();
+
+        const expectedETag = `"${releaseId}:${cacheInvalidatedAtTimestamp}"`;
```
ETLaurent, line 1406: TODO: factorize

modules/@apostrophecms/page/index.js
```diff
@@ -1393,17 +1394,43 @@ database.`);
...
           await self.emit('serve', req);
           await self.serveNotFound(req);
         } catch (err) {
           return await self.serve500Error(req, err);
         }
+
+        // TODO: stop render at the right place
+        if (self.shouldSendUnmodifiedResponse(req)) {
+          console.log('304');
+          req.res.statusCode = 304;
```
ETLaurent, line 1431: Seems to be automatically done by express

modules/@apostrophecms/module/index.js
```diff
@@ -182,7 +182,10 @@ module.exports = {
                 res.header('Cache-Control', 'no-store');
               }
 
-              res.status(200);
+              // Keep "304 Not Modified" status if previously set
+              const statusCode = res.statusCode === 304 ? 304 : 200;
```
ETLaurent, line 186: express was still sending a `304` status code in the end (browser received 304), probably thanks to its work on ETags matching, but I think that would be better not to force a `200` here, when we're actually sending a `304` earlier (in [here](https://github.com/apostrophecms/apostrophe/pull/3692/files#diff-5e6c997e317e3d9d022272dd5b800a9ddd9965d767280831bd1148f202c1dc1dR186) and [there](https://github.com/apostrophecms/apostrophe/pull/3692/files#diff-b0584effe98676702427f4558a0ea60d8993af875d2879d2f44893e1b7f8aa58R223))

modules/@apostrophecms/page/index.js
```diff
@@ -178,8 +178,16 @@ module.exports = {
           const criteria = self.getIdCriteria(_id);
           const result = await self.getRestQuery(req).and(criteria).toObject();
 
-          if (self.options.cache && self.options.cache.api) {
+          if (self.options.cache && self.options.cache.api && self.options.cache.api.maxAge) {
             self.setMaxAge(req, self.options.cache.api.maxAge);
+            self.sendETag(req, result);
+
+            if (self.doesETagMatch(req, result)) {
+              req.res.status(304);
```
ETLaurent, line 186: As mentioned in https://github.com/apostrophecms/apostrophe/pull/3692/files#r838668454, we're not forced to set the `304` status as Express seems to automatically send it (using ETag and If-None-Match).
Should we keep sending this `304` manually here, to be explicit on why we're sending an empty response below?

modules/@apostrophecms/page/index.js
```diff
@@ -2108,7 +2123,10 @@ database.`);
               _id: null
             });
           } else {
-            query.project(self.options.publicApiProjection);
+            query.project({
+              ...self.options.publicApiProjection,
+              cacheInvalidatedAt: 1
```
ETLaurent, line 2128: We need to have this value in order to [construct the ETag](https://github.com/apostrophecms/apostrophe/pull/3692/files#diff-0f2d52f6b9cfe649f1012ea4dc4af8bbdd7a0b8cba9db6a9748729b51e8e76d6R478) value, but that's not ideal to always expose it....

I'm open to suggestions!

ETLaurent:
Question:

How should we handle the cache invalidation of index pieces pages when a piece is edited?

image

modules/@apostrophecms/page/index.js
```diff
@@ -1404,6 +1412,17 @@ database.`);
...
         }
+
+        if (self.options.cache && self.options.cache.page && self.options.cache.page.maxAge) {
+          self.setMaxAge(req, self.options.cache.page.maxAge);
+          self.sendETag(req);
+
+          if (self.doesETagMatch(req)) {
+            // Stop there and send a 304 status code; the cached response will be used
+            return res.sendStatus(304);
+          }
```
ETLaurent, line 1423: FIXME: max-age isn't taken into account, the response is still cached even after the period of time set in max-age

boutell: I am not sure I understand. Are you saying we're setting the header but the browser doesn't respect it? Are we setting the header correctly?

boutell (changes requested):
Etienne, let's discuss your remaining concerns after the release of other things tomorrow, since this is not a rush item and you made some points about exposing data and cache headers not seeing to really work that we should look at.

modules/@apostrophecms/page/index.js
ETLaurent: It does but after the max-age period, it will just stop loading the cached content from the disk, and make the actual request.

But since Etag still matches (no modification has been made), the response status code 
will be `304`, and the content won't be fresh.

I might be missing an obvious Etag/Cache-Control mechanism here

test/docs.js
```diff
@@ -687,4 +688,95 @@ describe('Docs', function() {
...
+
+    const response = await apos.doc.insert(apos.task.getReq(), object);
+    const draft = await apos.doc.db.findOne({
+      _id: `${response.aposDocId}:en:draft`
+    });
+
+    assert(response.cacheInvalidatedAt.toString() === response.updatedAt.toString());
+    assert(draft.cacheInvalidatedAt.toString() === draft.updatedAt.toString());
+  });
+
```
ETLaurent, line 781: code from https://github.com/apostrophecms/apostrophe/pull/3678

modules/@apostrophecms/doc/index.js
```diff
@@ -1157,6 +1160,18 @@ module.exports = {
...
+      setCacheField() {
+        return self.apos.migration.eachDoc({ cacheInvalidatedAt: { $exists: 0 } }, 5, async doc => {
+          await self.apos.doc.db.updateOne({ _id: doc._id }, {
+            $set: { cacheInvalidatedAt: doc.updatedAt }
+          });
+        });
+      },
+      addCacheFieldMigration() {
+        self.apos.migration.add('add-cache-invalidated-at-field', self.setCacheField);
+      },
```
ETLaurent, line 1174: code from https://github.com/apostrophecms/apostrophe/pull/3678

modules/@apostrophecms/doc-type/index.js
```diff
@@ -918,6 +940,7 @@ module.exports = {
         // Setting it this way rather than setting it to published.updatedAt
         // guarantees no small discrepancy breaking equality comparisons
         draft.updatedAt = draft.lastPublishedAt;
+        draft.cacheInvalidatedAt = draft.lastPublishedAt;
```
ETLaurent, line 943: code from https://github.com/apostrophecms/apostrophe/pull/3678

test/pages.js
```diff
@@ -657,4 +689,341 @@ describe('Pages', function() {
...
+
+    // New ETag has been generated, with the new value of the edited homepage's `cacheInvalidatedAt` field...
+    assert(eTag2Parts[1] === pageUpdateResponse.cacheInvalidatedAt.getTime().toString());
+    // ...and a new timestamp
+    assert(eTag2Parts[2] !== eTag1Parts[2]);
+
+    delete apos.page.options.cache;
+  });
+
+  it('should not return a 304 status code when requesting a page with an outdated release id', async () => {
```
ETLaurent, line 907: The test that checks release id in the ETag has only be set for regular page serve.
We could set it also for the pieces and pages REST API, but I think it covers it enough.

ValJed (dismissed):
gg!

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 * Adds possibility for modules to [add extra frontend bundles for scss and js](https://v3.docs.apostrophecms.org/guide/webpack.html). This is useful when the `ui/src` build would otherwise be very large due to code used on rarely accessed pages.
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
 * Fixes deprecation warnings displayed after running `npm install`, for dependencies that are directly included by this package.
+* Implement custom ETags emission. It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) documents updates, thanks to backlinks mentioned above. For now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular served pages).
```
boutell, line 25: documents updates -> document updates

I think we need to make sure you can't enable this without also enabling the cache lifetime, and we should clarify even here in the changelog that there are still many possible situations where a change won't invalidate the cache and you will have to wait for the change lifetime.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -94,28 +94,50 @@ module.exports = {
...
+              aposLocale: { $in: [ doc.aposLocale, null ] }
+            }, {
+              $set: { cacheInvalidatedAt: doc.updatedAt }
+            });
+          }
+
+          if (doc._parentUrl) {
+            // Update piece index page's cache field
+            await self.apos.doc.db.updateOne({
+              slug: doc._parentUrl,
```
boutell, line 135: This will not work in the general case, the URL is not guaranteed to be identical to the slug, consider sitewide prefixes for instance.

It's OK to use this mechanism but you should do it by adding a `doc._parentPage` property to go along with `_parentUrl` so this can be done correctly (make this change in `addUrlsToPieces`).

modules/@apostrophecms/page/index.js
boutell: Just expose it, we add some other things preemptively because we know the user needs them for success. We could delete it afterwards but it would be slow and bug-prone.

boutell (changes requested):
Generally 💯 just some minor stuff

modules/@apostrophecms/doc-type/index.js
ETLaurent: Sitewide prefixes like `en/` in `somesite.com/en/some-slug`?

I'll add `piece._parentPage = parentPage.slug;` in `addUrlsToPieces ` then.

.editorconfig
```diff
@@ -9,3 +9,6 @@ indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
ETLaurent, line 14: allows line break in markdown documents (see `  ` double spaces in changelog entry from this PR 👇 )

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
...
 * Loads the right bundles on the right pages depending on the page template and the loaded widgets. Logged-in users have all the bundles on every page, because they might introduce widgets at any time.
 * Fixes deprecation warnings displayed after running `npm install`, for dependencies that are directly included by this package.
+* Implement custom ETags emission when `etags` cache option is enabled. [See the documentation for more information](https://v3.docs.apostrophecms.org/guide/caching.html).  
+It allows caching of pages and pieces, using a cache invalidation mechanism that takes into account related (and reverse related) document updates, thanks to backlinks mentioned above.  
+Note that for now, only single pages and pieces benefit from the ETags caching system (pages' and pieces' `getOne` REST API route, and regular served pages).  
+The cache of an index page corresponding to the type of a piece that was just saved will automatically be invalidated. However, please consider that it won't be effective when a related piece is saved, therefore the cache will automatically be invalidated _after_ the cache lifetime set in `maxAge` cache option.
```
ETLaurent, line 28: @boutell feel free to rephrase ;)

modules/@apostrophecms/doc-type/index.js
```diff
@@ -129,10 +129,10 @@ module.exports = {
             });
           }
 
-          if (doc._parentUrl) {
+          if (doc._parentPage) {
```
boutell, line 132: If you are only going to capture the slug in this property then please call it `_parentSlug`.

boutell (approved):

PR #3699, opened 2022-03-31
Fixes bundles contribution

ValJed:
## Summary
Fixes missing piece of the feature. Now multiple modules can contribute to the same bundle, they now export default functions. We reuse logic to write the import files.
Tweak webpack config to create bundles with the right name (`-module-bundle`) without touching the src-build bundle. and without having to copy manually our bundles...

Contribute to the same bundle from different modules.
You must export default function in your bundle files.
A module extending another that has a bundle will load this bundle too.
Check the order of apparition of your log adn see that the order of modules is respected.

*(Check at least one)*

- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -380,27 +380,27 @@ module.exports = {
             indexSass
           }) {
             fs.writeFileSync(importFile, (prologue || '') + stripIndent`
-            ${(icon && icon.importCode) || ''}
-            ${(icon && icon.registerCode) || ''}
-            ${(components && components.importCode) || ''}
-            ${(tiptap && tiptap.importCode) || ''}
-            ${(app && app.importCode) || ''}
-            ${(indexJs && indexJs.importCode) || ''}
-            ${(indexSass && indexSass.importCode) || ''}
-            ${(icon && icon.registerCode) || ''}
-            ${(components && components.registerCode) || ''}
-            ${(tiptap && tiptap.registerCode) || ''}
-          ` +
-            (app ? stripIndent`
-              setTimeout(() => {
-                ${app.invokeCode}
-              }, 0);
-            ` : '') +
-            // No delay on these, they expect to run early like ui/public code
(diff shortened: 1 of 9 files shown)


modules/@apostrophecms/asset/index.js
```diff
@@ -599,8 +643,11 @@ module.exports = {
             return pkgTimestamp > parseInt(timestamp);
           }
 
-          function getComponentName(component, options, i) {
-            return require('path').basename(component).replace(/\.\w+/, '') + (options.enumerateImports ? `_${i}` : '');
+          function getComponentName(component, { enumerateImports } = {}, i) {
+            return path
+              .basename(component)
+              .replace('-', '_')
```
ValJed, line 649: This is because JS does not allow to call functions with hyphens.

boutell (changes requested):
This code writes exactly the same thing to "module" and "nomodule":

```
              writeSceneBundle({
                scene,
                filePath: jsModules
              });
              writeSceneBundle({
                scene,
                filePath: jsNoModules
              });
```

Looking at the original code, I think this is what got lost (see the old merge function):

```
(!options.condition || options.condition === 'module')
...
(!options.condition || options.condition === 'nomodule')
```

modules/@apostrophecms/asset/index.js
ValJed: Right 🤦🏼  just fixed it

modules/@apostrophecms/asset/index.js
```diff
@@ -374,8 +369,42 @@ module.exports = {
...
+            prologue,
+            icon,
+            components,
+            tiptap,
+            app,
+            indexJs,
+            indexSass
+          }) {
+            fs.writeFileSync(importFile, (prologue || '') + stripIndent`
+            ${(icon && icon.importCode) || ''}
```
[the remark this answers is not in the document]
ValJed: Done

boutell (approved):

PR #3708, opened 2022-04-06
Pro 1974 crop image

ValJed:
[PRO-1974](https://linear.app/apostrophecms/issue/PRO-1974/as-a-user-i-can-crop-an-image-in-an-image-field)

## Summary

* Starts implementing grouping for relationship (not the purpose of the ticket right, but I thought it was needed for focal points because of some designs having a focal point tab, I guess we will need it in the future so kept it)
* Uses nice and lightweight `vue-advanced-cropper` to perform cropping on attachment images. cropper has been styled and the data is synced between the cropper and the schema fields.
* Fixes issue of relationships fields not synced when widget modifications are canceled.

* Link branch in testbed
* Add a widget image on any page. Click on the adjust image option:
image
* Play with the cropper, modifying the size through the fields should update the cropper stencil with a debounce of 500ms (could be changed).
* Cropping directly with the cropper stencil should update size fields.
* Also test with huge images, they should have a normal size but the crop should reflect the real size of the image. 
* You can zoom on image while cropping

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/ui/ui/apos/utils/index.js
```diff
@@ -0,0 +1,9 @@
+export const debounce = (func, timeout) => {
```
ValJed, line 1: Simply modernizes the function, same behavior (immediate param was never used)

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -4,10 +4,43 @@
...
+          {{ field.label }}
+        </label>
+        <div class="apos-schema__aligned-fields">
+          <div
+            class="apos-schema__aligned-field"
+            v-for="subField in field.fields"
+            :key="subField.name"
+            :data-apos-field="subField.name"
+          >
+            <component
```
ValJed, line 25: I'm not a huge fan of repeating these components twice but I don't see a perfect way to do it. Extracting it in another component would force us to still pass a lot of props each time.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -4,10 +4,43 @@
 <template>
   <div class="apos-schema">
     <div
-      v-for="field in schema" :key="field.name"
+      v-for="(field, index) in schemaToParse"
+      :key="`${field.name}${index}`"
       :data-apos-field="field.name"
     >
+      <div
+        v-if="field.type === 'alignedFields'"
```
ValJed, line 12: This type could be renamed, the system for making fields inline could be rethink, but it wasn't the purpose of this PR and it's where I lost a lot of time. As well as for grouped fields in relationships, that does not exist for now.

modules/@apostrophecms/schema/index.js
```diff
@@ -2507,11 +2512,16 @@ module.exports = {
                   throw new Error(`${context}: the subfield ${name} must have a 'fields' property with an 'add' subproperty containing its own fields.`);
                 }
               }
-              field.schema = self.fieldsToArray(context, field.fields.add || {});
+
+              field.schema = self.compose({
```
ValJed, line 2516: Should be kept for later in order to group fields in relationships.

modules/@apostrophecms/schema/index.js
```diff
@@ -1349,6 +1304,56 @@ module.exports = {
         return schema;
       },
 
+      composeGroups (schema, arrangeFields) {
```
ValJed, line 1307: Just moved this code to a function because I thought I would need it. I kept it because the initial function is big and complex enough.. Didn't touch any line of the code inside so no risk here.

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
```diff
@@ -90,24 +101,155 @@ export default {
...
-      this.docFields = value;
+    updateDocFields(value, repopulateFields = false) {
+      if (value.hasErrors) {
+        return;
+      }
+
+      this.docFields.data = {
+        ...this.docFields.data,
+        ...value.data,
+        ...repopulateFields && { _id: cuid() }
```
ValJed, line 137: This ID is needed for `AposSchema` to re populate the fields with the new values. We pass it only when dcoFields is updated when cropping the image, we do not repopulate fields when updating size from inputs.

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
```diff
@@ -90,24 +101,155 @@ export default {
...
     };
   },
   async mounted() {
     this.modal.active = true;
     this.docReady = true;
+
+    this.setVisibleSchema();
+    this.setNestedSchema();
+
+    // this.setGroups();
```
ValJed, line 120: Started implementing groups but realized we don't need it for focal points. but seeing the designs we will in the future so I'm gonna create a ticket.

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -204,11 +206,20 @@ export default {
         ? this.imageRelationshipComponent
         : 'AposRelationshipEditor';
 
+      const imgInfos = item.attachment ? {
+        url: item.attachment._urls.original,
+        width: item.attachment.width,
```
boutell, line 211: These would be the width and height of the entire attachment, not the existing crop. I think you are trying to get at the existing crop here.

modules/@apostrophecms/piece-type/ui/apos/components/AposImageCropper.vue
```diff
@@ -0,0 +1,107 @@
...
+      (coordinates) => {
+        this.$refs.cropper.setCoordinates(coordinates);
+      }, 500
+    );
+  },
+  methods: {
+    defaultSize({ imageSize, visibleArea }) {
+      return {
+        width: (visibleArea || imageSize).width,
+        height: (visibleArea || imageSize).height
```
ValJed, line 66: I think this part should be modified, but wanna be sure with you @boutell. We want to apply a cropping that would exist on the attachment document itself right? If no cropping we take the initial size of the image as we do here.

boutell (changes requested):

ETLaurent:
TODO:

- [ ] unit tests for `_urls.uncropped` object
- [ ] missing translation keys
- [ ] get the current image cropped values to set the cropper to the right position without having to reload the page

modules/@apostrophecms/attachment/index.js
```diff
@@ -712,11 +709,26 @@ module.exports = {
             if (options.annotate) {
               // Add URLs
               value._urls = {};
+              if (value._crop) {
+                value._urls.uncropped = {};
```
ETLaurent, line 713: Feels weird to have the `uncropped` property inside `_urls`, which results in adding deep complex structure.

I did that because I thought that was a "dirty" workaround at first, but since we're going to keep that, maybe we should have a `_uncroppedUrls` ou `_urlsUncropped` property next to `_urls` to keep it cleaner:

```
{
  _urls: {
    // ...
  },
  _urlsUncropped: {
    // ...
  },
```

wdyt?

modules/@apostrophecms/piece-type/ui/apos/components/AposImageCropper.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+  <div class="apos-image-cropper">
+    <cropper
+      ref="cropper"
+      :src="attachment._urls.uncropped ? attachment._urls.uncropped.original : attachment._urls.original"
```
ETLaurent, line 5: If we do https://github.com/apostrophecms/apostrophe/pull/3708/files#r853039376:

```suggestion
      :src="attachment._urlsUncropped ? attachment._urlsUncropped.original : attachment._urls.original"
```

modules/@apostrophecms/attachment/index.js
ValJed: I'm even not sure we need all the uncropped version here, but feeled logical.
I would say both are ok for me, if you feel more logical to have a specific property for uncropped ones it's good for me.

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
boutell: We are not using AposSchema here so I am not sure why we would need it unless perhaps you mean a separate implementation of groups for the specific need here.

boutell (approved):
`._urls.cropped` makes sense as-is. Thanks for working that out.

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
```diff
@@ -1,46 +1,103 @@
...
+          </div>
+        </div>
+        <!-- TODO: Implements Grouping for Relationships -->
+        <!-- <AposModalTabs
+          v-if="tabs.length"
+          :current="currentTab"
+          :tabs="tabs"
+          :errors="fieldErrors"
+          @select-tab="switchPane"
+        /> -->
```
ETLaurent, line 83: should we remove this TODO?

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
ETLaurent: should we remove the comments then?

ValJed: Yes I thought of a better implementation since it's used in different components. That was as a reference but I agree we could remove these comments.

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
```diff
@@ -1,46 +1,103 @@
...
+          </label>
+          <div
+            v-if="errors.width || errors.height"
+            class="apos-field__size-error"
+            :class="{'apos-field__size-error--focused': inputFocused}"
+          >
+            {{
+              $t('apostrophe:minSize', {
+                width: item.attachment.width,
+                height: item.attachment.height })
```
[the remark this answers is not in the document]
ValJed: Done

modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
ValJed: Removed it.

PR #3709, opened 2022-04-06
call crop route when saving editor modal

ETLaurent:
[PRO-2680](https://linear.app/apostrophecms/issue/PRO-2680/actual-cropped-images-are-generated-when-the-cropping-modal-is-saved)

## Summary

Call `/attachment/crop` route when `AposImageRelationshipEditor` modale is saved.

- add an image 

image

- click on "Image Adjustments"

image

- enter cropping data manually

image

- save the modal

- verify appropriate image filenames and sizes appear in public/uploads/attachments after cropping

image

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -206,6 +206,7 @@ export default {
 
       const result = await apos.modal.execute(editor, {
         schema: this.field.schema,
+        id: item._id,
         title: item.title,
         value: item._fields
       });
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -206,7 +206,7 @@ export default {
 
       const result = await apos.modal.execute(editor, {
         schema: this.field.schema,
-        id: item._id,
+        id: item.attachment._id,
         title: item.title,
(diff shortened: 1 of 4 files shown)


ValJed (approved):

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -206,7 +206,7 @@ export default {
 
       const result = await apos.modal.execute(editor, {
         schema: this.field.schema,
-        id: item._id,
+        id: item.attachment._id,
```
boutell, line 209: We cannot hardcode the assumption that every document has an `attachment` property. I think you are using this to figure out which attachment is being cropped. In which case passing all of `item` here would provide the needed flexibility and remove assumptions that break relationship editors for other types that are not images.

boutell (commented):
Hardcoded assumptions about the relationship type must be removed.

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
ETLaurent: @boutell yes indeed, that's totally true, now that I see it I realise that's a dumb mistake.
I'll pass the whole item to keep this relationship component agnostic of the its properties

PR #3710, opened 2022-04-06
OpenTelemetry

myovchev:
## Summary

**Integrate OpenTelemetry instrumentation in Apostrophe** 

The main goal is to make Apostrophe "telemetry ready" without or with the little possible overhead while not used.

**Features:**
- Boot tracing
- Per request tracing
- Tasks tracing
- Key points instrumentation with a high effort to add standard and Apostrophe owned attributes, allowing trace backend filtering
- Relationships (graph, based on active context) to allow operations (including filtered) to be viewed in the appropriate context
- Allow application level instrumentation

**Integration details:**

Single dependency point with OpenTelemetry API - `apos.telemetry` pseudo module. Direct dependency with OpenTelemetry Semantic Attributes per module. `apos.telemetry` exposes a simplified API (OTel API wrappers).

**Backward compatibility**

The integration is fully backward compatible with the previous platform functionality and exposed public APIs. 
The only change that might be considered non BC is `@apostrophecms/task:usage() signature. However this shouldn't be considered a public API and affects only the internal development so it can't be declared as breaking change.
Another notable (BC) change is the `run` emitter passing down the application Root span the event listeners.

**Operation prefixes:**

In order to keep the visual inspection experience high, the following prefixes for operation names were introduced (additionally to the naming convention standards):

- `aspostrophe:` boot sequence operations - `apostrophe:boot`, `apostrophe:boot:primary`, `apostrophe:boot:worker`
- `event:` event emitter operations
- `handler:` event listeners operations
- `db:` query related operations
- `render:` template render related operations
- `task:` task related operations

Closes PRO-2660, PRO-2664, PRO-2665, PRO-2666, PRO-2661

- Introduce OpenTelemetry SDK dependencies and configure the application. Pass down a shutdown handler as `beforeExit` configuration.
- Install collector backend (e.g. Jaeger).
- Boot up the application, login, manage, navigate. Execute Apostrophe tasks.
- Check the collector backend. 

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Documentation
- [ ] Build-related changes
- [ ] Other


index.js
```diff
@@ -39,6 +41,22 @@ let defaults = require('./defaults.js');
 // can be set to a number, which will effectively set the cluster
 // option to `cluster: { processes: n }`.
 //
+// `openTelemetryProvider`
+//
+// If set, Apostrophe will register it as a global OpenTelemetry tracer provider.
+// The expected value is an object, istance of TracerProvider.
```
boutell, line 47: istance -> an instance

index.js
```diff
@@ -59,8 +77,15 @@ let defaults = require('./defaults.js');
 // `null` in the primary process. In the child process it resolves as
 // documented above.
 
+// The actual entry point, a wrapper that enables the telemetry and starts the
+// root span
 module.exports = async function(options) {
+  const telemetry = opentelemetry(options);
+  let spanName = 'apostrophe:boot';
```
boutell, line 84: should be `const` I think.

index.js
```diff
@@ -70,48 +95,112 @@ module.exports = async function(options) {
...
+      return res;
+    } catch (err) {
+      telemetry.aposHandleError(span, err);
+      error = err;
+    } finally {
+      span.end();
+    }
+  });
+
+  if (self && self.taskRan && options.openTelemetrySDK) {
```
boutell, line 172: This can go away if we factor the process.exit() calls to `apos._exit()` and allow passing in a `beforeExit` handler.

lib/opentelemetry.js
```diff
@@ -0,0 +1,186 @@
...
+   * // Provide options for the newly created span
+   * const span = self.apos.telemetry.aposStartSpan('event:someEvent', true, {
+   *   attributes: [
+   *     [SemanticAttributes.HTTP_METHOD]: "GET",
+   *     [SemanticAttributes.HTTP_FLAVOR]: "1.1",
+   *     [SemanticAttributes.HTTP_URL]: req.url
+   *   ]
+   * });
+   */
+  function aposStartSpan(name, parentSpan, options) {
```
boutell, line 58: I am still trying to understand the need for this. Is it because the root span for the startup process is always out there as a default "parent" and that messes up the tracing for regular requests if we rely on the normal autodetection of parents via AsyncLocalStorage that happens in OpenTelemetry?

Would that issue go away more simply if we ended the root span at the end of the initialization process?

boutell, line 58: The apos prefix is unnecessary since this is already an apos pseudo-module.

lib/opentelemetry.js
```diff
@@ -0,0 +1,186 @@
...
+    // This shouldn't be needed, but having it doesn't hurt
+    if (options.openTelemetryProvider) {
+      api.trace.setGlobalTracerProvider(options.openTelemetryProvider);
+    }
+  }
+
+  /**
+   * Start and return a new span. Optionally provide a parent span or allow the parent
+   * span to be auto-detected.
+   * Use this when the current code block does the tracing and it doesn't expect
```
boutell, line 35: I don't think I understand why we need both this and aposStartActiveSpan.

lib/opentelemetry.js
```diff
@@ -0,0 +1,186 @@
...
+    if (parentSpan) {
+      const ctx = api.trace.setSpan(api.context.active(), parentSpan);
+      return tracer.startActiveSpan(name, options || undefined, ctx, fn);
+    }
+
+    return tracer.startActiveSpan(name, options || undefined, fn);
+  }
+
+  /**
+   * Invoke a callback within a new context (merged with a parent span).
```
boutell, line 104: Just call it a function rather than a callback (because it doesn't follow the traditional async callback argument pattern and I just don't want to confuse anyone).

lib/opentelemetry.js
```diff
@@ -0,0 +1,186 @@
...
+   * const value = await self.apos.telemetry.aposParentContextWith(
+   *   spanName,
+   *   parentSpan,
+   *   { kind: 1 },
+   *   self.someMethod,
+   *   arg1,
+   *   arg2
+   * );
+   */
+  function aposParentContextWith(name, parentSpan, options, fn, ...fnArgs) {
```
boutell, line 124: Are we actually using this method, or `aposContextWith`? I'm not sure there are any examples of it being used so far.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -2050,41 +2051,61 @@ module.exports = {
...
+              return results.map(result => result._id);
+            }
+
+            try {
+              await query.finalize();
+              let result;
+
+              if (!query.get('distinctCounts')) {
+                const criteria = query.get('criteria');
+                span.setAttribute(SemanticAttributes.DB_STATEMENT, JSON.stringify({ criteria }, null, 2));
```
boutell, line 2094: util.stringify is more appropriate because it provides a readable representation of regular expressions.

modules/@apostrophecms/task/index.js
```diff
@@ -15,62 +15,90 @@
...
           const cmd = self.apos.argv._[0];
-          if (!cmd) {
-            throw new Error('There is no command line argument to serve as a task name, should never happen');
-          }
-
-          if (cmd === 'help') {
+          const telemetry = self.apos.telemetry;
+          const spanName = `task:${cmd}`;
+          await telemetry.aposStartActiveSpan(spanName, async (span) => {
+            span.setAttribute(SemanticAttributes.CODE_FUNCTION, 'runTask');
```
boutell, line 35: The attributes seem a little redundant in combination with the span name, but if they improve the presentation it's worth it.

boutell (changes requested):
I was learning about your code as I went along. At this point I see the value of the telemetry pseudomodule, but we can definitely get rid of the `apos` prefixes on the method names, as that is already implied from context. I still think we should replace the hardcoded option of passing in the SDK with an option to pass in a `beforeExit` handler, which is better separation of concerns.

index.js
myovchev: An advanced setup would not involve OTel SDK (manual setup, manual initialization of resource, provider, exporters, etc). Adding an option to pass a provider gives the app implementation the freedom to do whatever is needed on an application level. The global provider registration is the only bridge between the core instrumentation and the application implementation.

Here is how it works (see the comments) - https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/trace/ProxyTracerProvider.ts#L33

index.js
myovchev: It is a mutable variable, see the `options.cluster`. eslint is normally auto fixing such things for me :)

myovchev:
> I was learning about your code as I went along. At this point I see the value of the telemetry pseudomodule, but we can definitely get rid of the `apos` prefixes on the method names, as that is already implied from context. I still think we should replace the hardcoded option of passing in the SDK with an option to pass in a `beforeExit` handler, which is better separation of concerns.

Going for `options.beforeExit` is definitely the cleanest way to do it. I didn't like the SDK dependency at all (although it's not a direct dependency, but any object having async shutdown() function). `beforeExit` handler has a general purpose and can be used in any scenarios, definitely a winner.

`apos` prefix indicates it is "an apos wrapper and not the original OpenTelemetry method".

The rationale behind the `aposStartSpan` and co. naming is the fact we are exporting directly parts of the OTel API (see `context`, `trace`, etc) and not the entire `api.*` object. The `apos` prefix makes it clear to the developer who invokes it that this is a wrapper of the OTel API `api.startSpan`, `api.startActiveSpan`, etc. Additionally, if the need arises (I doubt it though) the telemetry namespace could directly export the core OTel methods (e.g. `startSpan`) without namespace collisions. 

Of course we could always do `const api  = require('@opentelemetry/api')` anywhere but I wanted to avoid direct external dependency throughout our modules.

My first thought was to directly export the entire opentelemetry API as `apos.telemetry.api` but it haven't felt right. So the possible way to approach this is:

1. Refactor the `lib/opentelemetry.js` export like this:
```js
return {
    // contains the entire `@opentelemetry/api` export
    api,
    tracer,
    Attributes,
    handleError,
    startSpan,
    startActiveSpan,
    // those are obsolete and will be removed
    contextWith,
    parentContextWith
  };
```
2. This would require a usage refactoring as well, but it shouldn't be that high of an effort. 

Now when I see it, I start really like it, a lot. So I'll go for it as a last commit so that it would be easy revertable (in case you wanna keep the old way) before the merge.

lib/opentelemetry.js
myovchev: This is just fixing the not so good public API (IMO) of OTel. It gets rid of tons of boilerplate code from the core and makes it more readable.

lib/opentelemetry.js
myovchev: They both are needed. This one only detects the active context and does tracing, without carrying about the next (deeper) active state - it doesn't explicitly wrap other spans. This would be used with e.g. attribute drilling to be passed down for implicitly set the relationships. `startActiveSpan` does both, tracing and activates the nested tracing operation context (explicit usage, it shouldn't be passed down, next span will retrieve the active context). This is related with how span relationships are built with OTel.

lib/opentelemetry.js
myovchev: Not really, those are leftover artifacts of low level experiments, that I have abandoned. I'm removing those now.

modules/@apostrophecms/task/index.js
myovchev: Attributes are more important than span names (called `operations` in the most back-ends). They are used to do the actual data slices, while names are the context and are representing the relationships.

You can think of a span name as "mongodb collection" and "attributes" as "the mongodb criteria", but there is more to that. 
Span name is the graph node name, and you could filter everything related to this node via attributes query and visualize it in all possible ways (table, grid, graph, sky is the limit).

myovchev: Example (Jaeger back-end):  
Operation: all  
Tags: `code.function="toArray" code.namespace="@apostrophecms/redirect"`  
Results: All GET/POST/etc root traces having `@apostrophecms/redirect:toArray` invocation.

modules/@apostrophecms/doc-type/index.js
myovchev: NodeJS `util.inspect` I assume as I'm unable to find any `apos.util.stringify` method.

myovchev:
@boutell I have added the `beforeExit` handler feature. I'll wait for us to finish the whole implementation and I'll do the refactoring of the telemetry exports as we spoke earlier.

I'd like to discuss another problem which I've addressed with the last commit. When the process exits early (tasks) we need a way to close already opened tasks in order to not lose telemetry data. This is a challenge mostly related to the events. 
I came up with a `_onExitQueue`, which is exhausted just before the `beforeExit` handler in our exit handler - https://github.com/apostrophecms/apostrophe/pull/3710/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R174
To illustrate the problem, here is the flow
- emitter event:run - start span
  - emitter handler runTask startSpan
    - runTask - startSpan, do work, endSpan, call _exit handler
      - _exit handler -> emitter event:beforeExit - startSpan, end span
      - _exit handler -> exhaust `_onExitQueue`
      - _exit handler -> invoke `beforeExit` handler
      - _exit handler -> exit
  - emitter handler span not ended, process gone
- emitter event:run span not ended, process gone
The solution is - event emitter adds handlers to `_onExitQueue` in the right order that do span.end() when it expects the process to exit early. And it's a hackish solution.

What also bothers me is the check in the emit method - https://github.com/apostrophecms/apostrophe/pull/3710/files#diff-df59eb711db93cf5a1320347ab898116151b296de47557ad84ac6de3e204cd2cR43
It requires knowledge of the emitter about what would exit early (bad). Do you think there is a better way to approach this whole problem?

With this implementation:
- normal boot works fine
- CLI tasks have a warning at the very end (beforeExit event emitter) in the telemetry backend, which means we are missing (somehow) to close a tag, still hunting it down with no success so far. The warning is somehow related to the last event emitted `beforeExit`, and it makes no sense as everything should be actually cleaned up. I've left some debug output if you decide to test it.
- the special CLI tasks (e.g. build) looks good, no warnings.

index.js
```diff
@@ -39,6 +41,20 @@ let defaults = require('./defaults.js');
 // can be set to a number, which will effectively set the cluster
 // option to `cluster: { processes: n }`.
 //
+// `openTelemetryProvider`
+//
+// If set, Apostrophe will register it as a global OpenTelemetry tracer provider.
+// The expected value is an object, an istance of TracerProvider.
```
boutell, line 47: istance -> instance

index.js
```diff
@@ -39,6 +41,20 @@ let defaults = require('./defaults.js');
...
+// If the Node SDK is used in the application instead of manual configuration,
+// the provider instance is only available as a
+// private property: `sdkInstance._tracerProvider`. An issue can be opened
+// to discuss the exposure of a public getter with the OpenTelemetry developers.
+//
+// `beforeExit`
+//
+// If set, Apostrophe will invoke it (await) before exit.
```
boutell, line 55: before invoking `process.exit`.

index.js
```diff
@@ -39,6 +41,20 @@ let defaults = require('./defaults.js');
...
+// Expected type signature is `{ beforeExit: () => Promise<void> }`.
```
boutell, line 56: To be consistent with the tone of our documentation, which emphasizes async/await syntax over promises, just say "`beforeExit` may be an async function, will be awaited, and takes no arguments" rather than making an explicit reference to promises here (of course I know it's the same thing).

index.js
```diff
@@ -120,6 +166,65 @@ module.exports = async function(options) {
...
+  // Invokes and awaits `options.beforeExit` async function if available,
+  // passing as arguments the exit code and error (if any).
+  self._exit = async function(code = 0, error = null) {
+    try {
+      if (self.emit) {
+        await self.emit('beforeExit');
+      }
+    } catch (e) {
+      // we are at the point, where errors are ignored,
+      // if emitter is already regitered, all handler errors
```
boutell, line 189: registered

index.js
```diff
@@ -120,6 +166,65 @@ module.exports = async function(options) {
...
+  // Exhausts _onExitQueue.
+  // Invokes and awaits `options.beforeExit` async function if available,
+  // passing as arguments the exit code and error (if any).
+  self._exit = async function(code = 0, error = null) {
+    try {
+      if (self.emit) {
+        await self.emit('beforeExit');
+      }
+    } catch (e) {
+      // we are at the point, where errors are ignored,
```
boutell, line 188: no comma

index.js
```diff
@@ -130,6 +235,9 @@ module.exports = async function(options) {
     // promise event emitter code
     self.apos = self;
 
+    // Register the telemetry API as a pseudo module
+    Object.assign(self.apos, { telemetry });
```
boutell, line 239: `self.apos.telemetry = telemetry` is clearer and simpler.

modules/@apostrophecms/module/index.js
```diff
@@ -444,11 +445,27 @@ module.exports = {
...
-        await self.apos.area.loadDeferredWidgets(req);
-        req.res.send(
-          await self.apos.template.renderPageForModule(req, template, data, self)
-        );
+        const telemetry = self.apos.telemetry;
+        const spanName = `${self.__meta.name}:sendPage`;
+        await telemetry.aposStartActiveSpan(spanName, async (span) => {
+          span.setAttribute(SemanticAttributes.CODE_FUNCTION, 'sendPage');
+          span.setAttribute(SemanticAttributes.CODE_NAMESPACE, self.__meta.name);
+          span.setAttribute(telemetry.AposAttributes.TEMPLATE, template + '.html');
```
boutell, line 453: Might be a .njk file, so don't try to add .html here, it might not be true.

boutell: Also .html is only added if needed... just log it as passed.

modules/@apostrophecms/module/lib/events.js
```diff
@@ -38,18 +39,69 @@ module.exports = function(self) {
...
+          });
+        }
+
+        for (const entry of chain) {
+          const handlers = self.apos.eventHandlers[entry.name] && self.apos.eventHandlers[entry.name][name];
+          if (handlers) {
+            for (const handler of handlers) {
+
+              // Create an active "inner" span for each handler using the parent as a context
+              const spanName = `handler:${handler.moduleName}:${handler.handlerName}`;
```
boutell, line 66: handler names are not guaranteed to be unique outside of a given event name so please include the event name here as well.

```
`event:${moduleName}:${name}:handler:${handler.moduleName}:${handler.handlerName}`
```

modules/@apostrophecms/template/index.js
```diff
@@ -631,91 +632,125 @@ module.exports = {
...
-            stylesheetsPlaceholder: req.stylesheetsPlaceholder,
-            widgetsBundles: req.widgetsBundles
-          });
+          try {
+            const spanRenderName = `render:${module.__meta.name}:render`;
+            const content = await telemetry.aposStartActiveSpan(spanRenderName, async (spanRender) => {
+              spanRender.setAttribute(SemanticAttributes.CODE_FUNCTION, 'render');
+              spanRender.setAttribute(SemanticAttributes.CODE_NAMESPACE, module.__meta.name);
+              spanRender.setAttribute(telemetry.AposAttributes.SCENE, scene);
+              spanRender.setAttribute(telemetry.AposAttributes.TEMPLATE, template + '.html');
```
boutell, line 719: Do not add `.html`

modules/@apostrophecms/template/lib/custom-tags/component.js
```diff
@@ -27,29 +29,48 @@ module.exports = function(self) {
...
+            throw new Error(`{% component %} was invoked with the name of a component that does not exist.\nModule name: ${moduleName} Component name: ${componentName}`);
+          }
+          const result = await self.apos.util.recursionGuard(req, `component:${moduleName}:${componentName}`, async () => {
+            const input = await module.components[componentName](req, data);
+            return module.render(req, componentName, input);
+          });
+          span.setStatus({ code: telemetry.SpanStatusCode.OK });
+
+          if (result === undefined) {
+            // XXX should the recursion guard outcome to be sent as an error?
```
boutell, line 61: A separate ticket for this could make sense but don't worry about it here

boutell (changes requested):

boutell (changes requested):
Hi Miro,

I think a better fix for this problem is to remove the exit calls from the `apostrophe:run` event handler for tasks.

Instead, index.js should emit the `run` event with an additional object argument, `{ exit }`, where `exit` is initially a null property.

If the event handler chooses to change `exit` to an object with a `status` property, then `index.js` should take care of exiting, after awaiting `beforeExit`.

This way the event is able to completely finish its processing, including the tracing of all spans.

index.js
```diff
@@ -190,40 +182,37 @@ async function apostrophe(options, telemetry, rootSpan) {
...
-        await options.beforeExit(code, error);
+        await options.beforeExit(code, message);
       } catch (e) {
         console.error('beforeExit handler error', e);
       }
     }
     process.exit(code);
   };
 
+  process.once('SIGINT', (code) => {
```
boutell, line 206: We should not set these handlers at least not at this time. There would be interactions with multisite and we are way past our scope already in covering the startup part of tracing so thoroughly for this week.

index.js
```diff
@@ -323,7 +312,11 @@ async function apostrophe(options, telemetry, rootSpan) {
     if (self.taskRan) {
       await self._exit();
     } else {
-      await self.emit('run', self.isTask());
+      const SIGNAL = { exit: null };
```
boutell, line 315: This isn't a simple constant, so the uppercase feels inappropriate. The name is also a bit confusing. Its purpose is to decide what happens after, so let's call it `after`.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -2103,10 +2103,10 @@ module.exports = {
                 result = await distinctCounts(span);
               }
 
-              span.setStatus({ code: telemetry.SpanStatusCode.OK });
+              span.setStatus({ code: telemetry.api.SpanStatusCode.OK });
```
boutell, line 2106: Accessing it this way feels good.

boutell (changes requested):
Two minor things.

boutell (approved):

PR #3712, opened 2022-04-11
2681 image widget

ETLaurent:
## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/piece-type/ui/apos/components/AposImageRelationshipEditor.vue
```diff
@@ -103,12 +105,14 @@ export default {
...
-        }
-      });
+      if (this.item.attachment) {
+        await apos.http.post(`${apos.attachment.action}/crop`, {
+          body: {
+            _id: this.item.attachment._id,
+            crop: this.docFields.data
+          }
+        });
+      }
```
ETLaurent, line 115: see https://github.com/apostrophecms/apostrophe/pull/3709#discussion_r846337725

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -206,7 +206,7 @@ export default {
 
       const result = await apos.modal.execute(editor, {
         schema: this.field.schema,
-        id: item.attachment._id,
+        item,
```
ETLaurent, line 209: see https://github.com/apostrophecms/apostrophe/pull/3709#discussion_r846337725

modules/@apostrophecms/attachment/index.js
```diff
@@ -694,11 +694,11 @@ module.exports = {
             // apos.attachment.url with the returned object
             for (let i = ancestors.length - 1; i >= 0; i--) {
               const ancestor = ancestors[i];
-              const fields = ancestor.imagesFields && ancestor.imagesFields[o._id];
```
ETLaurent, line 697: A bug was there, thanks to @boutell's help, we changed it to make it work.

- `imagesFields` did not exist anymore, it was changed to `imageFields`
- passing the image document in `widget.html` (via `{% set attachment = apos.image.first(data.widget._image) %}`), would not even work with the right `imageFields` because no ancestors would have this property.

So we could have passed the entire widget instead (that has the property), but we decided that it would be clearer to pass the image and use its fields `_fields` directly to retrieve the cropping and focal point values

modules/@apostrophecms/attachment/index.js
```diff
@@ -778,14 +778,22 @@ module.exports = {
...
-      focalPointToBackgroundPosition(attachment) {
+      focalPointToObjectPosition(attachment) {
         if (!self.hasFocalPoint(attachment)) {
           return 'center center';
         }
         const point = self.getFocalPoint(attachment);
-        return point.x + '% ' + point.y + '%';
+        return `${point.x}% ${point.y}%`;
+      },
+      // Returns the attachment width.
```
boutell, line 790: "the effective attachment width."

boutell (changes requested):
Nice tests! One minor note.

ETLaurent:
Putting back the PR in wip, as @ValJed spotted a minor bug.

The slat media preview disappears after reloading the page:

**before**

image

**after**

image

modules/@apostrophecms/attachment/index.js
```diff
@@ -694,11 +694,12 @@ module.exports = {
...
             for (let i = ancestors.length - 1; i >= 0; i--) {
               const ancestor = ancestors[i];
-              const fields = ancestor.imagesFields && ancestor.imagesFields[o._id];
-              if (fields) {
+              const ancestorFields = ancestor.attachment && ancestor.attachment._id === value._id && ancestor._fields;
+              if (ancestorFields) {
                 value = _.clone(value);
-                value._crop = _.pick(fields, 'top', 'left', 'width', 'height');
-                value._focalPoint = _.pick(fields, 'x', 'y');
+                o.attachment = value;
```
ETLaurent, line 700: fixes [PRO-2729](https://linear.app/apostrophecms/issue/PRO-2729/annotate-option-of-attachmentall-and-headless-are-incompatible-with) and https://github.com/apostrophecms/apostrophe/pull/3712#issuecomment-1097822579

ETLaurent: tested [here](https://github.com/apostrophecms/apostrophe/pull/3712/files#diff-b04d3a965201f5f181bf662cc61d52f33e149a0f86b52c470d0a68ded543f240R315)

modules/@apostrophecms/attachment/index.js
```diff
@@ -694,11 +694,17 @@ module.exports = {
...
+                value._crop = {
+                  top: ancestorFields.top || 0,
+                  left: ancestorFields.left || 0,
+                  width: ancestorFields.width || 0,
+                  height: ancestorFields.height || 0
```
ETLaurent, line 705: For some reason, the image's `_field`  has its `left` property set to `null` , when setting 0 or nothing in the model for `left` _and_ `top` (which is very weird).

I spent too much time trying to find the source of the problem, so I dirty-fixed that by setting a default 0 value, in order to avoid having `null` in path...

```js
_fields: { top: 0, left: null, width: 50, height: 50, x: 83, y: 56 }
```

boutell: Defaulting the width and height to 0 would not make sense, if we must default them then we should default them to the actual width and height of the image.

boutell (changes requested):

modules/@apostrophecms/attachment/index.js
ETLaurent: I you guys have an idea of what could set `top` to `0` (which is good) and leave `left` to `null` in the image `_fields`, I'd love to have a hint :)

ETLaurent: > Defaulting the width and height to 0 would not make sense, if we must default them then we should default them to the actual width and height of the image.

yes that's what I thought also, I'll change that 👍

ETLaurent: @boutell I couldn't use the default attachment width and height because the file does simply not exist (when no cropping has been done, the attachment path does not contain the width nor height values).

I just left them as they are

PR #3723, opened 2022-04-20
PRO-2749 support for node 17 and 18

boutell:
## Summary

Support for node 17 and, more importantly, 18 (see changelog for issues addressed)

Try out Apostrophe with node 18. Also verify still happy in node 16

*(Check at least one)*

- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [X] Build-related changes
- [ ] Other

**Other information:**


lib/mongodb-connect.js
```diff
@@ -0,0 +1,55 @@
...
+  const parsed = new URL(uri);
+  if ((parsed.protocol !== 'mongodb:') || (parsed.hostname !== 'localhost')) {
+    return mongo.MongoClient.connect(parsed.toString(), connectOptions);
+  }
+  const records = await dns.promises.lookup('localhost', { all: true });
+  if (!records.length) {
+    // The computer that reaches this point has bigger problems 😅
+    throw new Error('Unable to resolve localhost to an IP address.');
+  }
+  return new Promise((resolve, reject) => {
```
boutell, line 28: I can't use `Promise.all` because I need to succeed if any of them work and I need to clean up in a specific way if they don't. So there's custom logic here to resolve and reject.

lib/escape-host.js
```diff
@@ -0,0 +1,8 @@
+module.exports = host => {
+  if (host.includes(':')) {
+    // ipv6
+    return `[${host}]`;
+  } else {
+    return host;
+  }
```
ValJed, line 7:
```javascript
return host.includes(':') ? `[${host}]` : host
```

lib/mongodb-connect.js
ETLaurent: What about `Promise.any()`?

> This method is useful for returning the first promise that fulfills
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

ETLaurent: ...and it's only available on browsers

ValJed (approved):

PR #3728, opened 2022-04-25
pro-733 No cropping for svg

ValJed:
[PRO-733](https://linear.app/apostrophecms/issue/PRO-733/as-a-user-i-should-not-be-able-to-crop-an-svg-ie-it-should-not-try-and)

## Summary
Reuse logic from A2 to check if an image attachment is croppable or not.
Passes the info to frontend when requesting documents attachments.
Renames `labelKey` in `editRelationLabel` to make it understandable. Uses this one instead of the classic one when passed.

* Link the branch in an A3 project (testbed)
* Add image widget on a page
* Add normal images and see you can still open the cropping modal
* Add svg image in the widget and see the context menu disapear.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+# unreleased
+
+### Adds
+
+* Uses the isCroppable method to know if an image attachment is croppable or not. Passes this info to the frontend when requesting documents attachments. Uses this info to show the context menu or not.
+
 # 3.17.0 (2022-03-31)
 
 ### Adds
--- a/modules/@apostrophecms/image/index.js
+++ b/modules/@apostrophecms/image/index.js
@@ -30,6 +30,8 @@ module.exports = {
     // another document that references them
     relatedDocument: true,
     relationshipEditor: 'AposImageRelationshipEditor',
+    relationshipEditorLabel: 'apostrophe:editImageAdjustments',
+    relationshipEditorIcon: 'image-edit-outline',
     relationshipFields: {
       add: {
         top: {
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -213,6 +213,7 @@ export default {
         title: item.title,
         value: item._fields
       });
+
       if (result) {
         const index = this.next.findIndex(_item => _item._id === item._id);
         this.$set(this.next, index, {
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
@@ -43,7 +43,7 @@
           :value="next"
           :disabled="field.readOnly"
           :has-relationship-schema="!!field.schema"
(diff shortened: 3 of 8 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -221,7 +222,7 @@ export default {
         });
       }
     },
-    getSlatLabelKey () {
+    getEditRelationshipLabel () {
       if (this.field.editor === 'AposImageRelationshipEditor') {
```
boutell, line 226: There should be no hardcoding of a specific relationship type in `AposInputRelationship`. So this label override must become part of the relationship field type configuration on the serv er side. Same for the "edit relationship" icon which is different in this design from the generic one if I recall properly.

Then we can extend my previous work in schema/index.js to allow it to come from the image piece type so that every relationship field doesn't have to specifically configure it.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
ValJed: Made some modifs. Label and icon are set in the relationship config. There are passed as a fallback to field if none locally is set. if editorIcon is passed we use this as a button instead of the context menu (because for now we have no others links in the context menu.)
Let me know if it fits for you.

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
-# unreleased
+# UNRELEASED
 
 ### Adds
 
-* Uses the isCroppable method to know if an image attachment is croppable or not. Passes this info to the frontend when requesting documents attachments. Uses this info to show the context menu or not.
+* Images may now be cropped to suit a particular placement after selecting them.
+* Developers may now specify an alternate Vue component to be used for editing the subfields of relationships, either at the field level or as a default for all relationships with a particular piece type.
+* Adds ability to set `minSize` to areas for image widgets. When setup, browsing relationship images from `AposInputRelationship` or from `AposMediaManager` when it is a chooser, will emit `piece-relationship-query`. We pass the query string object, this one will be filled with appropriate data if needed (in this case minSize array).
+* Uses the `isCroppable` method to know if an image attachment is croppable or not. Passes this info to the frontend when requesting documents attachments. Uses this info to show the context menu or not.
```
boutell, line 10: As a changelog entry it would probably be better to just say "prevent cropping of SVG files as this is not possible int he general case."

modules/@apostrophecms/image/index.js
```diff
@@ -30,6 +30,8 @@ module.exports = {
     // another document that references them
     relatedDocument: true,
     relationshipEditor: 'AposImageRelationshipEditor',
+    relationshipEditorLabel: 'apostrophe:editImageAdjustments',
```
boutell, line 33: 💯

boutell (approved):

CHANGELOG.md
ValJed: I forgot to modify it, I do it right now in my next PR.

PR #3732, opened 2022-04-26
Pro 2683 cropping respects minsize

ValJed:
[PRO-2683](https://linear.app/apostrophecms/issue/PRO-2683/as-an-editor-when-cropping-i-must-not-be-able-to-go-below-the)

## Summary
Blocks users to set values under `minSize` if set in area (can set only width or only height or both). Auto fixes bad values in input when under `minSize` (with helper being purple when minSize fixed), or above image size during blur.
Allows users to set only min width or only min height. It would work the same way just for a size.

* `npm link` branch in A3 project (testbed)
* Set a minSize in your images areas this way (in testbed config is in area.js file):
```javascript
  '@apostrophecms/image': {
    minSize: [ 800, 600 ]
  },
```
* Go edit an image widget from this area, check that an helper is here with the minSize infos:
image
* Try try set values under, when leaving input it should fix the value, and accent the helper text in purple for 1.5 seconds.
* Try to set values above the image size, it should fix it at blur too, without any specific info.
* Try to set only a min width or a min height. It should work properly too:
```javascript
  '@apostrophecms/image': {
    minSize: [ 800 ] // or [ null, 600 ]
  },
```

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
-# Unreleased
+# UNRELEASED
+
 ### Adds
 
+* Images may now be cropped to suit a particular placement after selecting them.
+* Developers may now specify an alternate Vue component to be used for editing the subfields of relationships, either at the field level or as a default for all relationships with a particular piece type.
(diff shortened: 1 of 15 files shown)


ETLaurent (approved):
tested ✅

PR #3735, opened 2022-04-27
PRO-2768 workaround for not deep watching value of AposSchema

boutell:
## Summary

We don't really deep-watch the value prop of AposSchema for performance reasons, schemas can be deeply nested and the potential for this to cause severe performance problems if we check it every time `value` changes due to our own `input` event is too great. `value` is primarily consulted only at mount time so this is usually OK, however occasionally a truly external change to the data does occur and there must be a way to force a refresh of all fields. Implement and internally document that mechanism so our internal developers know we're not strictly respecting `v-model` in this component.

When npm linked to both this branch of apostrophe and the corresponding branch of Document Versions, open Document Versions for the home page, then try this in the browser console:

```
apos.bus.$emit('content-changed', {
  doc: { ...apos.adminBar.context, title: 'New Title' },
  action: 'someAction'
});
```

The title change will be respected in the displayed dialog.

*(Check at least one)*

- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [X] Documentation
- [ ] Build-related changes
- [ ] Other

- [X] Related documentation has been updated

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
@@ -1,5 +1,25 @@
 <!--
-  AposSchema takes an array of fields with their values, renders their inputs, and emits their `input` events
+  AposSchema takes an array of fields (`schema`), renders their inputs,
+  and emits a new object with a `value` subproperty and a `hasErrors`
+  subproperty via the input event whenever the value of a field
+  or subfield changes.
+
+  At mount time the fields are initialized from the subproperties of the
+  `value.data` prop.
+
+  For performance reasons, this component is not strictly v-model compliant.
+  While all changes will emit an outgoing `input` event, the
+  incoming `value` prop only updates the fields in three situations:
+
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -135,22 +162,23 @@ export default {
...
-        if (
-          // If the fieldState had been cleared and there's new populated data
-          (!this.fieldState._id && newVal.data._id) ||
-          // or if there *is* active fieldState, but the new data is a new doc
-          (this.fieldState._id && newVal.data._id !== this.fieldState._id.data)
-        ) {
-          // repopulate the schema.
-          this.populateDocData();
-        }
+    'value.data._id'(_id) {
```
boutell, line 165: This is all we were really doing with the deep watch before and it's more performant. I verified that switching docs in the media manager still correctly swaps what's in the schema view.

myovchev (approved):

PR #3736, opened 2022-04-29
PRO-2739 fix nestedModuleSubdirs bug

boutell:
## Summary

The `nestedModuleSubdirs` feature no longer throws an error and interrupts startup if a project contains both `@apostrophecms/asset` and `asset`, which should be considered separate module names.

* Check out `apostrophecms/utah-a3`, switch branches to `nestedModules`, npm install, npm link with this branch of apostrophe. Launching "node app" reaches "listening on port 3000" without issue.
* Switch back to the "main" branch of apostrophe and restart. Throws an error complaining that there are two folders for the asset module, which is wrong (the namespaced module `@apostrophecms/asset` should be considered completely unrelated to the module `asset`).

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/lib/moog-require.js
+++ b/lib/moog-require.js
@@ -5,6 +5,7 @@ const path = require('path');
 const glob = require('glob');
 const importFresh = require('import-fresh');
 const resolveFrom = require('resolve-from');
+const regExpQuote = require('regexp-quote');
 
 module.exports = function(options) {
   const self = require('./moog')(options);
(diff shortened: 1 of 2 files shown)


lib/moog-require.js
```diff
@@ -62,7 +63,12 @@ module.exports = function(options) {
         self._indexes = glob.sync(self.options.localModules + '/**/index.js');
```
ETLaurent, line 63: We could exclude all npm namespace folder from the list:

```suggestion
        self._indexes = glob.sync(`${self.options.localModules}/**/index.js`, {
          ignore: `${self.options.localModules}/@*/**`
        });
```

lib/moog-require.js
```diff
@@ -62,7 +63,12 @@ module.exports = function(options) {
...
       }
       const matches = self._indexes.filter(function(index) {
-        return index.endsWith('/' + type + '/index.js');
+        // Double-check that we're not confusing "@apostrophecms/asset" with "asset" by
+        // making sure "type" is not preceded by an npm namespace folder. If type is itself namespaced
+        // this never comes up (because npm namespaces don't nest). The risk is that a legitimate
+        // project level module that happens to end with the same name as a namespaced module
+        // will be rejected as a duplicate when nestedModuleSubdirs is present
+        return index.endsWith('/' + type + '/index.js') && !index.match(new RegExp(`/@[^/]+/${regExpQuote(type)}/index\\.js$`));
```
ETLaurent, line 71: If taking https://github.com/apostrophecms/apostrophe/pull/3736/files#r863610696 into account:

```suggestion
        return index.endsWith(`/${type}/index.js`);
```

ETLaurent (approved):

lib/moog-require.js
boutell: The legitimate issue could also occur with namespaced modules, they could appear under two parent dirs at project level using this feature.

PR #3745, opened 2022-05-04
provide original doc to editor component's children

ETLaurent:
[PRO-2746](https://linear.app/apostrophecms/issue/PRO-2746/aposdoceditor-should-inject-originaldoc-into-descendant-components)

## Summary

Provides the original document to `AposDocEditor` children.

Inside a child component (`AposModalRail.vue` for instance), inject `originalDoc` and display one of its properties:

```vue
<template>
  <div class="apos-modal__rail" :class="`apos-modal__rail--${type}`">
    <p>{{ originalDoc.ref && originalDoc.ref.lastPublishedAt }}</p> <!-- ACCESS -->
    <slot />
  </div>
</template>
<script>
export default {
  name: 'AposModalRail',
  inject: [ 'originalDoc' ], // INJECTION
  props: {
    type: {
      type: String,
      default: 'left'
    }
  },
  computed: {

  }
};
</script>
```

The property of the doc should be accessible:

image

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [x] Other

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -158,7 +158,9 @@ export default {
       },
       triggerValidation: false,
       original: null,
-      originalDoc: {},
+      originalDoc: {
+        ref: null
+      },
       published: null,
       errorCount: 0,
       restoreOnly: false,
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -153,6 +158,7 @@ export default {
       },
       triggerValidation: false,
       original: null,
+      originalDoc: {},
```
ETLaurent, line 161: ℹ️ Use an object in order to keep the ref "alive".
Providing `original` directly would result in having a `null` injected value.

boutell (approved):

PR #3749, opened 2022-05-04
PoC provide/inject of originalDoc

boutell:
Just a PoC for Etienne, not intended to be merged.

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -125,6 +125,11 @@ export default {
     AposAdvisoryLockMixin,
     AposArchiveMixin
   ],
+  provide() {
+    return {
+      originalDoc: this.originalDoc
+    };
+  },
   props: {
     moduleName: {
       type: String,
@@ -157,7 +162,10 @@ export default {
       errorCount: 0,
       restoreOnly: false,
       saveMenu: null,
-      generation: 0
+      generation: 0,
+      originalDoc: {
+        ref: null
+      }
     };
   },
   computed: {
@@ -363,8 +371,9 @@ export default {
     manuallyPublished() {
       this.saveMenu = this.computeSaveMenu();
     },
-    original() {
+    original(newValue) {
       this.saveMenu = this.computeSaveMenu();
+      this.originalDoc.ref = newValue;
     },
     tabs() {
       if ((!this.currentTab) || (!this.tabs.find(tab => tab.name === this.currentTab))) {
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -125,6 +125,11 @@ export default {
     AposAdvisoryLockMixin,
     AposArchiveMixin
   ],
+  provide() {
+    return {
+      originalDoc: this.originalDoc
```
boutell, line 130: To make it reactive, we must provide an object data property. As in your PR this requires an extra layer of nesting ("ref"), yes.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -157,7 +162,10 @@ export default {
       errorCount: 0,
       restoreOnly: false,
       saveMenu: null,
-      generation: 0
+      generation: 0,
+      originalDoc: {
```
boutell, line 166: The ref property should exist from the start, even if it's initially null. vue 2 reactivity can't detect a property being added for the first time (not without help). It can detect a property value change.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -363,8 +371,9 @@ export default {
     manuallyPublished() {
       this.saveMenu = this.computeSaveMenu();
     },
-    original() {
+    original(newValue) {
```
boutell, line 374: In the providing component we watch `original` and update `this.originalDoc.ref`, which is compatible with reactivity since `ref` already exists as a property (even if initially null).

modules/@apostrophecms/schema/ui/apos/components/AposInputWrapper.vue
```diff
@@ -141,6 +142,7 @@ export default {
     }
   },
   mounted: function () {
+    console.log('mount time:', JSON.stringify(this.originalDoc, null, '  '));
```
boutell, line 145: Works. It looks like by the time fields are mounted `originalDoc.ref` has already become an object, but if it hadn't we could have used `watch` to wait for that.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
ETLaurent: @boutell You're right, I'll initialise `ref` 👍

ETLaurent:
@boutell closing the PR :)

PR #3752, opened 2022-05-08
Add webpack cache

myovchev:
## Summary

Enable the cache option for both the src and the apos webpack configurations. The benefit of this becomes obvious on the second startup (50%+ time spent, 5-10% CPU improvements based on simple benchmarks).

In your application execute:
```sh
\time -f "Total time: %E\nUser Mode (s) %U\nKernel Mode (s): %S\nCPU: %P" node app @apostrophecms/asset:build
```
It should show a speed improvement due to caching on second and subsequent runs.

Execute:
```sh
node app @apostrophecms/asset:clear-cache
```
Verify `data/temp/webpack-chache` folder is empty and repeat the first command. You should see the first run taking around twice longer again (no cache).

**Acceptance Criteria deviations**
- Changes of webpack configuration's plugin values will not result in a Apostrophe cache invalidation - this should be handled by webpack itself. The reason is random values from plugins (as timestamps) appearing in the configuration and the lack of control over that. However our cache routine listens for changes of plugin root config values and the plugin list itself.
- The modification timestamp on the folder is updated AFTER the build run and not before that. This makes sense as the folder does not exist yet and is auto-created by the webpack.
- Additional meta is written per cache folder to ensure better testing and debugging. This also allows easier cache management (e.g. invalidation) and processes operating only on FS level.

Closes PRO-2778
*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -4,6 +4,7 @@ const Promise = require('bluebird');
 const webpackModule = require('webpack');
 const globalIcons = require('./lib/globalIcons');
 const path = require('path');
+const util = require('util');
 const express = require('express');
 const { stripIndent } = require('common-tags');
 const { merge: webpackMerge } = require('webpack-merge');
@@ -111,6 +112,8 @@ module.exports = {
         usage: 'Build Apostrophe frontend CSS and JS bundles',
         afterModuleInit: true,
         async task(argv) {
+          // The lock could become huge, cache it, see computeCacheMeta()
+          let _packageLockContentCached;
           const req = self.apos.task.getReq();
           const namespace = self.getNamespace();
           const buildDir = `${self.apos.rootDir}/apos-build/${namespace}`;
@@ -317,7 +320,13 @@ module.exports = {
                 ? webpackMerge(webpackInstanceConfig, ...Object.values(self.webpackExtensions))
                 : webpackInstanceConfig;
 
+              // Inject the cache location at the end - we need the merged
+              const cacheMeta = await computeCacheMeta(name, webpackInstanceConfigMerged);
+              webpackInstanceConfigMerged.cache.cacheLocation = cacheMeta.location;
+
               const result = await webpack(webpackInstanceConfigMerged);
+              await writeCacheMeta(name, cacheMeta);
+
               if (result.compilation.errors.length) {
                 // Throwing a string is appropriate in a command line task
                 throw cleanErrors(result.toString('errors'));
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Wepback cache for build performance in development and production mode.
```
boutell, line 7: Should mention the cache clearing task for edge bases where it is not automatically detected that the cache needs to be cleared.

modules/@apostrophecms/asset/index.js
```diff
@@ -111,6 +112,8 @@ module.exports = {
         usage: 'Build Apostrophe frontend CSS and JS bundles',
         afterModuleInit: true,
         async task(argv) {
+          // The lock could become huge, cache it, see computeCacheMeta()
+          let _packageLockContentCached;
```
boutell, line 116: In Apostrophe leading `_` means "temporary property not to be stored to the database." To avoid confusion we don't use it for other things (except `_id` since that comes from mongo).

modules/@apostrophecms/asset/index.js
```diff
@@ -667,6 +676,93 @@ module.exports = {
...
+          // In order to compute an accurate hash, this helper needs
+          // the final, merged webpack configuration.
+          async function computeCacheMeta(name, webpackConfig) {
+            const cacheBase = self.getCacheBasePath();
+
+            if (!_packageLockContentCached) {
+              const packageLock = await findPackageLock();
+              if (packageLock === false) {
+                // this should happen only when testing and
+                // we don't wanna break all non-core module tests
```
boutell, line 693: want to

modules/@apostrophecms/asset/index.js
```diff
@@ -667,6 +676,93 @@ module.exports = {
...
+              const lastModified = new Date();
+              await fs.writeFile(
+                cachePath,
+                `${lastModified.toISOString()} ${self.getNamespace()}:${name}`,
+                'utf8'
+              );
+              // should be the same as the meta
+              await fs.utimes(cachePath, lastModified, lastModified);
+            } catch (e) {
+              // Build probably faild, path is missing, ignore
```
boutell, line 750: failed

test/assets.js
```diff
@@ -284,10 +293,222 @@ describe('Assets', function() {
...
+    const { meta: meta2, folders: folders2 } = getCacheMeta();
+    assert.equal(folders2.length, 3);
+    assert.equal(Object.keys(meta2).length, 3);
+    assert(meta2['default:apos']);
+    assert(meta2['default:src']);
+    assert(meta2['default:src-es5']);
+
+    // Expect at least 40% gain, in reallity it should be 50+
+    const gain = (execTime - execTimeCached) / execTime * 100;
+    assert(gain >= 40, `Expected gain >=40%, got ${gain}%`);
```
boutell, line 349: This test is really cool but I don't think we can have such a strict standard for something that is highly dependent on the performance given to us by github actions from moment to moment. We will have too many false positives / negatives. This test could be enabled by an environment variable for local use.

boutell: Maybe set the expectation to 20% and we'll see if that is stable enough.

modules/@apostrophecms/asset/index.js
```diff
@@ -667,6 +676,93 @@ module.exports = {
...
+              }
+            }
+
+            // Plugins can output timestamps or other random information as
+            // configuration (e.g. StylelintWebpackPlugin). As we don't have
+            // control over plugins, we need to remove their configuration values.
+            // We keep only the plugin name and config keys sorted list.
+            // Additionally plugins are sorted by their constructor names.
+            // A shallow clone is enough to avoid modification of the original
+            // config.
```
boutell, line 706: This strategy is cool with me.

boutell (changes requested):

boutell (approved):

PR #3763, opened 2022-05-15
PRO-2788: Trigger only the relevant build in watch mode

myovchev:
## Summary

Improve the build task to accept a `changes` parameter and based on that to detect what build to execute. Improve the build watcher to send the paths of the detected changes and restart the application only when needed. Allow external chokidar watcher to be registered. The smart building of Apostrophe UI happens on relevant change, no matter if `APOS_DEV` is enabled.

Additionally fix a bug where some Apostrophe modules symlinked in node_modules are not being watched.

Closes  PRO-2788, PRO-2828

1. Start Apostrophe application in dev mode: `node app`
- the old build behavior is expected (execute `src` related and `public` builds, complie `apos` based on `package-lock.json`)
- a "watching files" message should be visible in the console
- a change to `ui/src` should re-concatenate all bundles but only recompile the `ui/src` portion
- a change to `ui/public` should re-concatenate everything but no webpack builds should be executed
- a change to `ui/apos` should only rebuild the apos bundle
2. Start Apostrophe application in dev mode: `APOS_DEV=1 node app`
- the old build behavior is expected (rebuild everything)
- same as 1.
3. Start Apostrophe application in dev mode: `APOS_ASSET_WATCH=0 node app`
- a "watching files" message should NOT be visible in the console
- the old build behavior is expected (execute `src` related and `public` builds, complie `apos` based on `package-lock.json`)
- after the initial build, changes should not trigger any build

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/task/index.js
+++ b/modules/@apostrophecms/task/index.js
@@ -117,6 +117,8 @@ module.exports = {
       // normally be hyphenated, i.e. at the command line you might write
       // `--total=20`.
       //
+      // This method will return the value returned by the task.
+      //
       // **Gotchas**
       //
       // If you can invoke a method directly rather than invoking a task, do
@@ -132,6 +134,7 @@ module.exports = {
       // task developer might assume they can exit the process directly.
 
       async invoke(name, args, options) {
+        let result;
         const telemetry = self.apos.telemetry;
         const spanName = `task:${self.__meta.name}:${name}`;
         await telemetry.startActiveSpan(spanName, async (span) => {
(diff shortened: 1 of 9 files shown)


modules/@apostrophecms/task/index.js
```diff
@@ -155,7 +158,7 @@ module.exports = {
             };
             span.setAttribute(telemetry.Attributes.ARGV, telemetry.stringify(argv));
             self.apos.argv = argv;
-            await task.task(argv);
+            result = await task.task(argv);
```
boutell, line 161: So tasks can have a return value on success now, although that is only accessible if they are invoked programmatically rather than at the command line. OK.

boutell (approved):

PR #3769, opened 2022-05-19
Advanced build optimization

myovchev:
## Summary

Add:
Improve the build task to accept a changes parameter and based on that to detect what build to execute. Improve the build watcher to send the paths of the detected changes and restart the application only when needed. Allow external chokidar watcher to be registered. The smart building of Apostrophe UI happens on relevant change, no matter if APOS_DEV is enabled.

Fix:
- Fix a bug where some Apostrophe modules symlinked in node_modules are not being watched.
- Recover after a build error when in a watch mode. 

1. Start Apostrophe application in dev mode: `node app`
- the old build behavior is expected (execute `src` related and `public` builds, complie `apos` based on `package-lock.json`)
- a "watching files" message should be visible in the console
- a change to `ui/src` should re-concatenate all bundles but only recompile the `ui/src` portion
- a change to `ui/public` should re-concatenate everything but no Webpack builds should be executed
- a change to `ui/apos` should only rebuild the apos bundle
- edit any watched asset and create a bad code to reproduce Webpack build error
- console log should show detailed information about the error
- the page should not be reloaded
- edit the same file and correct the error
- a normal rebuild should occur
- the page should be reloaded 
2. Start Apostrophe application in dev mode: `APOS_DEV=1 node app`
- the old build behavior is expected (rebuild everything)
- same as 1.
3. Start Apostrophe application in dev mode: `APOS_ASSET_WATCH=0 node app`
- a "watching files" message should NOT be visible in the console
- the old build behavior is expected (execute `src` related and `public` builds, complie `apos` based on `package-lock.json`)
- after the initial build, changes should not trigger any build

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/asset/index.js
```diff
@@ -116,14 +117,26 @@ module.exports = {
...
           // The lock could become huge, cache it, see computeCacheMeta()
           let packageLockContentCached;
           const req = self.apos.task.getReq();
           const namespace = self.getNamespace();
           const buildDir = `${self.apos.rootDir}/apos-build/${namespace}`;
           const bundleDir = `${self.apos.rootDir}/public/apos-frontend/${namespace}`;
           const modulesToInstantiate = self.apos.modulesToBeInstantiated();
+          // Make it clear if builds should detect changes
+          const detectChanges = Array.isArray(argv.changes);
+          // Remove invalid changes. `argv.changes` is an array of relative
```
boutell, line 130: It bothers me that a command line task would have options that aren't really usable on the command line (there's no way to pass an array). It's not a blocker for now though.

boutell (approved):
Approving. Did you fix something *in general* for tasks here? If so it would be a good idea to mention that fix in the changelog.

myovchev:
> Approving. Did you fix something _in general_ for tasks here? If so it would be a good idea to mention that fix in the changelog.

It's mouthful to explain. 
If `task.invoke` is preceded by `self.isTask()` check and the invoked task throws an error, every subsequent call in this order will make `self.isTask()` to return true (due to `apos.argv` not reverted back to empty).

boutell:
Yeah that's pretty rare for most developer situations, OK to leave it out of the changelog. 👍

PR #3779, opened 2022-06-02
Fix symlinked node modules not being rebuilt

myovchev:
## Summary

Fix a Webpack cache issue leading to modules symlinked in `node_modules` not being rebuilt. 

Copy the asset module index.js from this PR to `node_modules/apostrophe/modules/@apostrophecms/asset/index.js` in the dedicated test repo. Clear the Webpack cache, boot and do changes. The changes should be applied.

Fixes a random failing test due to wrong expectations.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+* Fix a Webpack cache issue leading to modules symlinked in `node_modules` not being rebuilt. 
+
 ## 3.21.0 (2022-05-25)
 
 ### Adds
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -125,6 +125,7 @@ module.exports = {
           const buildDir = `${self.apos.rootDir}/apos-build/${namespace}`;
           const bundleDir = `${self.apos.rootDir}/public/apos-frontend/${namespace}`;
           const modulesToInstantiate = self.apos.modulesToBeInstantiated();
+          const symLinkModules = await findNodeModulesSymlinks(self.apos.npmRootDir);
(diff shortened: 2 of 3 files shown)


boutell (changes requested):
Awesome solution Miro, I was looking for some kind of exclude mechanism, I didn't realize managedPaths supported regular expressions. One small request for changes to consistently escape regular expressions in the right way.

myovchev:
> @myovchev I don't think I understand the test directions. Why can't I just test with this branch npm linked, and make changes to another npm linked module with apos frontend code, and observe the effect?
> 
> I want to make sure I'm testing it properly and it's unclear to me right now. Thanks.


@boutell  The nature of this bug is that it's only available/reproducible when the the apostrophe dependency is NOT symlinked (hence automated testing not possible). 

1. In order to be validated it works as expected before it's merged with the main branch a hard copy/replace of the modified `./modules/@apostrophecms/asset/index.js` to a [local installation of the test repo](https://github.com/apostrophecms/assembly-webpack-cache-test) `node_modules/apostrophe/modules/@apostrophecms/asset/index.js` .  Alternatively in the test repo set the dependency to `github:apostrophecms/apostrophe#fix-webpack-cache-symlinks` (requires more work and manual validation that the git hash of the installed `apostrophe` dependency matches this branch latest commit hash).

2. After the setup is updated one way or another, the [steps describing the bug](https://github.com/apostrophecms/assembly-webpack-cache-test/pull/1) should be repeated and validated the fix works

boutell (approved):
Approved. I confirmed the problem and also confirmed the fix after copying your index.js into the normally installed apostrophe module.

PR #3791, opened 2022-06-08
Pro 2814 share draft routes

ValJed:
[PRO-2814](https://linear.app/apostrophecms/issue/PRO-2814/as-a-frontend-developer-i-can-invoke-a-draft-sharing-rest-api)
## Summary

Adds routes in `@apostrophecms/piece-type` and `@apostrophecms/page` to share and unshare draft pages.

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -438,6 +438,12 @@ module.exports = {
             throw self.apos.error('invalid');
           }
           return self.revertPublishedToPrevious(req, published);
+        },
+        ':_id/share': async (req) => {
+
+        },
+        ':_id/unshare': async (req) => {
+
         }
       }
     };
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -440,10 +440,19 @@ module.exports = {
           return self.revertPublishedToPrevious(req, published);
         },
         ':_id/share': async (req) => {
+          const draft = await self.getDraftToShare(req);
+
+          const shared = await self.share(req, draft);
+
+          return shared;
 
         },
         ':_id/unshare': async (req) => {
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -938,6 +953,28 @@ module.exports = {
...
+        const piece = await self.findOneForEditing(req, {
+          _id
+        });
+
+        if (!piece) {
+          throw self.apos.error('notfound');
+        }
+
+        if (piece.aposMode !== 'draft') {
+          throw self.apos.error('invalid');
```
haroun, line 974: will this allow someone without the appropriate permissions to discover that a piece exists?

modules/@apostrophecms/piece-type/index.js
```diff
@@ -938,6 +953,28 @@ module.exports = {
             throw self.apos.error('invalid', 'Document has neither slug nor title, giving up');
           }
         }
+      },
+
+      async getDraftToShare(req) {
```
haroun, line 958: On all other routes we repeat the checks with `nofound`.
What is the difference here with `:_id/publish`/`:_id/unpublish` routes?

Once implemented, this will be part of the public API.

ValJed: `notfound` check is done in the `getDraftToShare` method.

modules/@apostrophecms/piece-type/index.js
ValJed: That's right, was in the specs but maybe that a notfound would be enough here. User doesn't need more infos.

ETLaurent (approved):

PR #3793, opened 2022-06-08
Pro 2814 invoke share api

ValJed:
[PRO-2814](https://linear.app/apostrophecms/issue/PRO-2814/as-a-frontend-developer-i-can-invoke-a-draft-sharing-rest-api)

## Summary

Checks infos from the current document to see if it's sharable, otherwise close the modal and notify the user.
When toggling button, call the right API route to share or unshare a document, update url depending on the response.
Adds translation to manage error in notification

* Link branch in testbed (or another project)
* In a piece context menu click on `share draft`
* It should open the share draft modal disabled
* Try to enable it, it should create and url and allow to copy it through a link
* Close to modal, if you come back the share draft should be enabled when opening the modal
* When disabling enabling, the  `aposShareKey` should be new
* Do the same test for pages

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
@@ -96,7 +96,6 @@ export default {
   },
   methods: {
     async copy() {
-      console.log('copy');
       await navigator.clipboard.writeText(this.shareUrl);
     },
     async toggle() {
@@ -128,10 +127,10 @@ export default {
         this.shareUrl = this.generateShareUrl(aposShareKey);
 
       } catch (err) {
-        if (!this.disabled) {
-          this.errorNotif();
-        } else {
+        if (this.disabled) {
           this.shareUrl = '';
+        } else {
+          this.errorNotif();
         }
       }
     },
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
@@ -164,7 +164,7 @@ export default {
           this.shareUrl = this.generateShareUrl(aposShareKey);
         }
       } catch {
-       await this.errorNotif();
+        await this.errorNotif();
       }
     },
     async errorNotif() {
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
@@ -123,7 +123,7 @@ export default {
           return this.errorNotif();
         }
 
-        this.shareUrl = aposShareKey;
+        this.shareUrl = this.generateShareUrl();
 
       } catch (err) {
         if (routeToCall === 'share') {
@@ -154,6 +154,13 @@ export default {
       setTimeout(() => {
         this.close();
       }, 500);
+    },
+    generateShareUrl(aposShareKey) {
+      return apos.http.addQueryToUrl(this.doc._url, {
+        ...apos.http.parseQuery(this.doc._url),
+        aposShareKey,
+        aposShareId: this.doc._id
+      });
     }
   }
 };
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
@@ -36,8 +36,8 @@
               {{ $t('apostrophe:shareDraftDescription') }}
             </p>
             <input
+              v-model="shareUrl"
               type="text"
-              :value="shareUrl"
               disabled
               class="apos-share-draft__url"
             >
--- a/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
+++ b/modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
@@ -35,25 +35,34 @@
             <p class="apos-share-draft__description">
               {{ $t('apostrophe:shareDraftDescription') }}
             </p>
-            <input
-              v-model="shareUrl"
-              type="text"
-              disabled
-              class="apos-share-draft__url"
+            <transition
+              name="collapse"
+              :duration="200"
             >
-            <a
-              v-if="shareUrl"
-              href=""
-              class="apos-share-draft__link-copy"
-              @click.prevent="copy"
-            >
-              <LinkVariant
-                class="apos-share-draft__link-icon"
-                :title="$t('apostrophe:shareDraftCopyLink')"
-                :size="16"
-              />
-              &nbsp;{{ $t('apostrophe:shareDraftCopyLink') }}
-            </a>
+              <div
+                class="apos-share-draft__url-block"
+                v-show="!disabled"
+              >
+                <input
+                  v-model="shareUrl"
+                  type="text"
+                  disabled
+                  class="apos-share-draft__url"
+                >
+                <a
+                  href=""
+                  class="apos-share-draft__link-copy"
+                  @click.prevent="copy"
+                >
+                  <LinkVariant
+                    class="apos-share-draft__link-icon"
+                    :title="$t('apostrophe:shareDraftCopyLink')"
+                    :size="16"
+                  />
+                  &nbsp;{{ $t('apostrophe:shareDraftCopyLink') }}
+                </a>
+              </div>
+            </transition>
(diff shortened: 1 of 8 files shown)


modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -36,12 +36,13 @@
...
             </p>
             <input
+              v-model="shareUrl"
               type="text"
-              value="https://dummy-url.com/wefe3456wfewuidfki9wefe3456wf554ewfki9"
               disabled
               class="apos-share-draft__url"
             >
             <a
+              v-if="shareUrl"
```
ETLaurent, line 45: 👍

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +86,90 @@ export default {
...
     };
   },
   async mounted() {
     this.modal.active = true;
+    this.checkUrlprop();
   },
   methods: {
-    copy() {
+    async copy() {
       console.log('copy');
```
ETLaurent, line 99:
```suggestion
```

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +86,90 @@ export default {
...
+        }
+
+        this.shareUrl = this.generateShareUrl(aposShareKey);
+
+      } catch (err) {
+        if (!this.disabled) {
+          this.errorNotif();
+        } else {
+          this.shareUrl = '';
+        }
```
ETLaurent, line 135: I'd switch the conditions for a better readability:

```suggestion
        if (this.disabled) {
          this.shareUrl = '';
          return;
        }
        this.errorNotif();
```

🥖

ValJed: 👍🏼

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +86,89 @@ export default {
...
+    async errorNotif() {
+      await apos.notify('apostrophe:shareDraftError', {
+        type: 'danger',
+        icon: 'alert-circle-icon',
+        dismiss: true
+      });
+
+      setTimeout(() => {
+        this.close();
+      }, 500);
```
ETLaurent, line 157:
```suggestion
      setTimeout(this.close, 500);
```

this should work

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +86,89 @@ export default {
...
+          .filter(([ _, val ]) => val)
+      );
+
+      const slug = apos.http.addQueryToUrl(this.doc._url, {
+        ...qs,
+        aposShareKey,
+        aposShareId: this.doc._id
+      });
+
+      return location.origin + slug;
```
ETLaurent, line 171:
```suggestion
      return ${location.origin}${slug};
```
minor stuff

ETLaurent: arf..., never mind actually

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +86,89 @@ export default {
...
+        dismiss: true
+      });
+
+      setTimeout(() => {
+        this.close();
+      }, 500);
+    },
+    generateShareUrl(aposShareKey) {
+      const qs = Object.fromEntries(
+        Object.entries(apos.http.parseQuery(this.doc._url))
```
ETLaurent, line 161: `parseQuery` only takes the query, apparently (with or without the `?` prefix), not the entire piece url.

Anyhow, is there any case where the `_url` would contain a query string, I roughly went through the code and I didn't see any qs attached to `_url`, but I might miss something

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +86,89 @@ export default {
...
+          .filter(([ _, val ]) => val)
+      );
+
+      const slug = apos.http.addQueryToUrl(this.doc._url, {
+        ...qs,
```
ETLaurent, line 166:
```suggestion
```

if `this.doc._url` doesn't contain any query string, see https://github.com/apostrophecms/apostrophe/pull/3793/files#r893504423

ETLaurent (changes requested):
more questions than changes requested

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
ValJed: Not more readable for me but ok

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
ValJed: Maybe for the pages?

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
ValJed: image

ETLaurent: Still not sure about passing the whole URL to `apos.http.parseQuery`.
It might work though, but the function's documentation warns that we should not do that..

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +95,96 @@ export default {
...
       },
-      disabled: false
+      shareUrl: '',
+      disabled: true
     };
   },
   async mounted() {
     this.modal.active = true;
+    await this.checkUrlprop();
+    this.getAposShareKey();
```
ETLaurent, line 105:
```suggestion
    await this.getAposShareKey();
```

ETLaurent (changes requested):
last 2 comment

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
ETLaurent: pass only the query part

ETLaurent (changes requested):

ETLaurent (approved):

modules/@apostrophecms/modal/ui/apos/components/AposModalShareDraft.vue
```diff
@@ -79,22 +95,95 @@ export default {
...
+    },
+    async errorNotif() {
+      await apos.notify('apostrophe:shareDraftError', {
+        type: 'danger',
+        icon: 'alert-circle-icon',
+        dismiss: true
+      });
+    },
+    generateShareUrl(aposShareKey) {
+      const url = new URL(`${location.origin}${this.doc._url}`);
```
haroun, line 178: should we rely on `getBaseUrl` instead for instances where the backoffice and the website does not use the same domain name?

ETLaurent: `req.baseUrl` is passed to the frontend, and use instead (in #3798).

But I'm not sure that `baseUrl` will take the front-office URL.
Beside, I think that we should use `baseUrlWithPrefix`, but it's undefined.

I'll dig more on that, but first let's ask @boutell if there's a "right" way to get the front-office domain/base url!

boutell:
`req.baseUrl` is already set dynamically via `apos.page.getBaseUrl(req)` which can be extended if necessary.

However pushing `baseUrlWithPrefix` to the front end is a good idea.

ETLaurent:
> `req.baseUrl` is already set dynamically via `apos.page.getBaseUrl(req)` which can be extended if necessary.
> 
> However pushing `baseUrlWithPrefix` to the front end is a good idea.

@boutell we came up with this: https://github.com/apostrophecms/apostrophe/pull/3798/files

PR #3817, opened 2022-06-23
Make __t() globally available per request

myovchev:
## Summary

This patch adds the `__t()` as global helper so that it is available for macros too. This is possible because `getEnv()` handler receives a real request object when called by our parsers on request. This should be 100% backward compatible change - the fragments are using custom context which takes precedence over the global context. 

Use `{{ __t('Some phrase') }}` inside a macro. It should render properly (with no parse errors).

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 * Fix missing title translation in the "Array Editor" component.
+* Bring back support for i18n inside macros (`__t()` can be used inside macros again).
 
 ## 3.23.0 (2022-06-22)
 
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -346,6 +346,8 @@ module.exports = {
           // no guarantee newEnv will be called for every req
           self.envs[name] = self.newEnv(req, name, self.getViewFolders(module));
         }
+        // i18n is available per request
+        self.envs[name].addGlobal('__t', req.t);
         return self.envs[name];
       },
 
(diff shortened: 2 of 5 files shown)


myovchev:
@boutell The proposed solution won't work because the context does not have `__req` (possibly it would have if called with `with context` but this also fails when multiple cross imports are made). 

A second attempt - we don't mess up with the `getEnv`, instead we use `renderBody` as the point to introduce global context, which should be propagated down to any registered macros and custom tags not extending Apostrophe.

boutell:
This works out to the same thing as the original attempt. There is one global in an `env` object that may be in the middle of several overlapping asynchronous render calls.

Since per-request environments have a huge performance cost, and `context` doesn't make it through without `with context` and in the situations you mention, I think the proper answer is to deprecate macros more thoroughly. They are not async friendly. Fragments are the async-friendly, widet-friendly, Apostrophe-friendly replacement for macros.

boutell:
Actually, I thought of another way 😆 

Right now we have one `env` per module.

We can break it down more and have one `env` per module plus locale name.

Then there should be no problem with the `addGlobal` call because `req.__t` will always be a req with the same locale.

boutell (changes requested):
We can't win this one by adding a global that might be reused by two requests with different locales due to async.

But we can win this one by having one `env` per module-plus-locale tuple, rather than one `env` per module.

Please note that `getEnv` does not really promise to get a unique `req` for every invocation, that's a bc argument only. So the right implementation is to add a third argument, `locale`, and be sure to pass it in.

boutell:
Closing as stale. Note my above comments on how it could be reworked if desired.

PR #3829, opened 2022-07-05
Change area menu layout to a grid - to better space out widget items at higher quantities

stepanjakl:
## Summary

*Just a little UI update to the Area widget menu/list using CSS grid layout. I found out that after having more than 8-10 widgets, the list becomes vertically a bit too long. Additional scrolling is necessary to have it properly in the view.*

| From:  | To: |
| ------------- | ------------- |
| Screenshot 2022-04-12 at 12 40 56  | Screenshot 2022-04-12 at 12 28 09  |

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


boutell:
Thank you for this. However we have a ticket in the works to properly support grouping widgets into categories, which should resolve the issue in a more flexible way. Possibly we could return to this if we really need it within individual groups.

PR #3857, opened 2022-08-08
Tool to find heavy npm modules; moved stylelint to dev deps

boutell:
## Summary

Tool to find heavy npm modules; moved stylelint to dev deps

Apostrophe still builds and works normally

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "debounce-async": "0.0.2",
     "deep-get-set": "^1.1.1",
     "dompurify": "^2.3.1",
-    "eslint-plugin-promise": "^5.1.0",
     "express": "^4.16.4",
     "express-bearer-token": "^2.4.0",
     "express-cache-on-demand": "^1.0.3",
@@ -102,10 +101,6 @@
     "sass-loader": "^10.1.1",
     "server-destroy": "^1.0.1",
     "sluggo": "^0.3.0",
-    "stylelint": "^14.6.1",
-    "stylelint-declaration-strict-value": "^1.8.0",
-    "stylelint-order": "^5.0.0",
-    "stylelint-webpack-plugin": "^3.2.0",
     "tinycolor2": "^1.4.2",
     "tough-cookie": "^4.0.0",
     "underscore.string": "^3.3.4",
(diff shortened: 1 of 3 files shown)


scripts/find-heavy-npm-modules
```diff
@@ -0,0 +1,38 @@
...
+  return weight + 1;
+}
+
+const sorted = [ ...costs.entries() ].sort((a, b) => a[1] - b[1]);
+for (const [ name, cost ] of sorted) {
+  console.log(`${name} has ${cost} sub-dependencies`);
+}
+
+const nonDevDeps = Object.keys(trueDeps).filter(name => !trueDeps[name].dev).length;
+console.log(`Total dependencies: ${nonDevDeps}`);
```
myovchev, line 38: Neat little script!

package.json
```diff
@@ -134,7 +129,12 @@
...
-    "webpack-bundle-analyzer": "^3.9.0"
+    "webpack-bundle-analyzer": "^3.9.0",
+    "eslint-plugin-promise": "^5.1.0",
+    "stylelint": "^14.6.1",
+    "stylelint-declaration-strict-value": "^1.8.0",
+    "stylelint-order": "^5.0.0",
+    "stylelint-webpack-plugin": "^3.2.0"
   },
   "browserslist": [
     "ie >= 10"
```
myovchev, line 140: @Tom I think this slows things/adds up to the build size a lot. But BC...

myovchev (approved):

package.json
boutell: These dependencies were already here. I moved them to devDependencies so it reduces the build size unless you're actually debugging apostrophe core itself in which case they belong.

(Project level might depend on all of these anyway, but then again it might not, or it might want different major versions, so this can still be a reduction.)

myovchev: I'm sorry, I meant this part `"ie >= 10"`

boutell: Oh, I see.

I'm not sure if this is or was doing anything.

We don't ship es5 support with apostrophe core anymore, it's in a separate module. I don't know how exactly babel decides which package.json file to look for this setting in. But it's probably looking for a project level one, or not looking at all because we always set our targets this way in our ES5 stuff anyway:

```
                      targets: {
                        browsers: '> 1%, IE 11, not dead'
                      }
```

It's a separate issue unrelated to this PR, but you could do a PR to remove it, and that would pass through QA normally. I suspect it won't have an impact but dead code is bad to have.

myovchev: I need to research it.

PR #3871, opened 2022-08-31
release 3.27.1

boutell:
Release only

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apostrophe",
-  "version": "3.27.0",
+  "version": "3.27.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "apostrophe",
-  "version": "3.27.0",
+  "version": "3.27.1",
```
ETLaurent, line 3:
```suggestion
  "version": "3.28.0",
```

if I'm not wrong, regarding the added features (webp, closing modals and checkboxes min and max fields)

ETLaurent (changes requested):

PR #3882, opened 2022-09-16
display placeholder for modal-driven widgets

ETLaurent:
Closes:
- [PRO-3131](https://linear.app/apostrophecms/issue/PRO-3131/as-a-developer-i-can-tell-when-it-is-appropriate-to-display-the)
- [PRO-3149](https://linear.app/apostrophecms/issue/PRO-3149/implement-initialmodal-false-option-for-widget-type-modules)

## Summary

Add a `apos-placeholder` class to the widgets wrapper div when they have `initialModal: false` set in their options.
Use the class set in the `placeholderClass` option if set, or do not add any class if set to `false`.

When the widget is set actual data, the `widget.aposPlaceholder` is set to `false`, resulting in having the placeholder class removed.

For now, use a simple background as the placeholder:

image

## What are the specific steps to test this change?

In testbed:

- add `topic` widget to some area
- add the `initialModal: false` option to `topic-widget`
👉  you should have the `apos-placeholder` background when adding a topic widget **without having the initial modal displayed**
- add the `placeholderClass: 'some-class'` option to the widget
👉 you should have the wrapper div with `some-class` added instead of `apos-placeholder` (and thus no background)
- add the `placeholderClass: false` option to the widget
👉 you should have the wrapper div no class added

## What kind of change does this PR introduce?
*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

## Make sure the PR fulfills these requirements:

- [x] It includes a) the existing issue ID being resolved, b) a convincing reason for adding this feature, or c) a clear description of the bug it resolves
- [ ] The changelog is updated
- [ ] Related documentation has been updated
- [ ] Related tests have been updated

If adding a new feature without an already open issue, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

PR #3884, opened 2022-09-20
display placeholder for image-widget

ETLaurent:
Depends on #3882

## Summary

Display a default placeholder image when adding an image widget.

On testbed:

- add an image widget
👉 the initial modal should not appear
👉 the default placeholder image should have been added:

image

👉 when you actually choose an image for that widget, it should replace the placeholder image

- override the `placeholderUrl` option in image widget:

```js
// in testbed's "modules/@apostrophecms/image-widget/index.js"
// you should have a `another-placeholder.jpg` file in "modules/@apostrophecms/image-widget/public/_
options: {
    placeholderUrl: '/modules/@apostrophecms/my-image-widget/another-placeholder.jpg'
  }
```
👉 you should have your image as placeholder now, instead of the default one

👉 all of these tests should be effective inside another widget such as the `two-column-widget` (nesting)
👉 all of these tests should be effective in the page editor modal

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


ETLaurent:
🧪 Unit tests will be added after #3886 is merged.

modules/@apostrophecms/image-widget/views/widget.html
```diff
@@ -1,31 +1,39 @@
-{% if data.options.className %}
-  {% set className = data.options.className %}
-{% elif data.manager.options.className %}
-  {% set className = data.manager.options.className %}
-{% endif %}
+{% if apos.util.eqStrict(data.widget.aposPlaceholder, true) and data.manager.options.placeholderUrl %}
```
boutell, line 1: This util.eqStrict check feels like overkill to me, especially in a template.

boutell: I guess it's not wrong though so I won't flag it.

modules/@apostrophecms/image-widget/views/widget.html
```diff
@@ -1,31 +1,39 @@
...
+  <img
+    src="{{ apos.asset.url(data.manager.options.placeholderUrl) }}"
+    alt="image placeholder"
```
boutell, line 4: Needs to go through i18n.

modules/@apostrophecms/image-widget/views/widget.html
```diff
@@ -1,31 +1,39 @@
...
+    width="100%"
```
boutell, line 5: Should have a class that can be styled. We should supply default styles to the extent we supply them for the regular content of the image widget.

boutell: Using the width attribute is not appropriate because it more or less forces them to override all of widget.html if they want to do something else.

boutell (changes requested):

modules/@apostrophecms/image-widget/views/widget.html
ETLaurent: Hmm, I see what you mean, I was wondering also..
But since I've used strict comparisons for `aposPlaceholder` [elsewhere](https://github.com/apostrophecms/apostrophe/pull/3882/files#diff-8afe4cc61ac0fbcc6476c7d670671b940a728f08619ef63f47ba0621323cc33aR78), I'd wanted to keep it consistent

boutell (approved):

PR #3890, opened 2022-09-26
fix captcha bug when login failed

falkodev:
https://linear.app/apostrophecms/issue/PRO-3101/login-button-remains-disabled-after-invalid-login-attempt-when-using

Remove requirements check after a failed login attempt (because it has been validated before submission)

### DIFF
--- a/modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
+++ b/modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
@@ -232,7 +232,6 @@ export default {
       } catch (e) {
         this.error = e.message || 'An error occurred. Please try again.';
         this.phase = 'beforeSubmit';
-        this.requirements = getRequirements();
       } finally {
         this.busy = false;
       }
@@ -257,7 +256,6 @@ export default {
         this.redirectAfterLogin();
       } catch (e) {
         this.error = e.message || 'An error occurred. Please try again.';
-        this.requirements = getRequirements();
         this.phase = 'beforeSubmit';
       } finally {
         this.busy = false;
--- a/modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
+++ b/modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
@@ -237,10 +237,12 @@ export default {
       }
     },
     getInitialSubmitRequirementsData() {
-      return Object.fromEntries(this.requirements.filter(r => r.phase !== 'afterPasswordVerified').map(r => ([
-        r.name,
-        r.value
-      ])));
+      return Object.fromEntries(this.requirements
+        .filter(r => r.phase !== 'afterPasswordVerified' || !r.done)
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -257,7 +256,6 @@ export default {
         this.redirectAfterLogin();
       } catch (e) {
         this.error = e.message || 'An error occurred. Please try again.';
-        this.requirements = getRequirements();
```
boutell, line 260: I.... think this is right, but did you test it (do after-login requirements missed on the first try still at least work with this change)?

boutell (changes requested):

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
falkodev: Yes, tested with totp on afterPasswordVerified phase

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
```diff
@@ -237,10 +237,12 @@ export default {
       }
     },
     getInitialSubmitRequirementsData() {
-      return Object.fromEntries(this.requirements.filter(r => r.phase !== 'afterPasswordVerified').map(r => ([
-        r.name,
-        r.value
-      ])));
+      return Object.fromEntries(this.requirements
+        .filter(r => r.phase !== 'afterPasswordVerified' || !r.done)
```
boutell, line 241: This is going to return all requirements that are not "done," even if they *are* `afterPasswordVerified`. Does that make sense? It seems to me like you probably wanted `&&` here. Also please add parens around the `!==` clause for readability (assuming I have the intended grouping right, that is).

boutell (changes requested):

modules/@apostrophecms/login/ui/apos/components/TheAposLogin.vue
falkodev: I have this error message when trying to login with `&&` instead of `||`:

```
POST /api/v1/@apostrophecms/login/login: 

invalid: The hCaptcha token was missing while verifying login.
    at Object.error (/Users/anttarla/Dev/apostrophecms/apostrophe/modules/@apostrophecms/error/index.js:29:23)
    at Object.verify (/Users/anttarla/Dev/apostrophecms/test-project/node_modules/@apostrophecms/login-hcaptcha/index.js:19:31)
    at Object.initialLogin (/Users/anttarla/Dev/apostrophecms/apostrophe/modules/@apostrophecms/login/index.js:625:33)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /Users/anttarla/Dev/apostrophecms/apostrophe/modules/@apostrophecms/module/index.js:180:30
```

boutell (approved):

PR #3906, opened 2022-10-04
Pro 3112 extend attachment filegroups

BoDonkey:
## Summary

The `attachment` module currently has an option, `fileGroups` that will completely replace the default `fileGroups` of `office` and `images` with the user-provided groups. This is inconvenient if the user only wants to add a single new extension to the existing groups or add a new group while retaining the rest of the default groups. This PR modifies the `attachment` module to allow the addition of an extension(s) to the existing groups or a new group without discarding the default groups. It also contains two new tests for this behavior. Closes PRO-3112.

*For example:*

1. Create a `modules/@apostrophecms/attachment/index.js` file.
2. Add the following code:
```
module.exports = {
  options: {
    addFileGroups: [
      {
        name: 'office',
        extensions: [
          'odt'
        ]
      },
      {
        name: 'othermedia',
        label: 'media',
        extensions: [
          'mp4'
        ],
        extensionMaps: {}
      }
    ]
  }
};

```
3. Bring up a site and check that you can still upload an existing `office` group file like a 'pdf'.
4. Check you can upload an 'odt' file.
5. Check you can upload an 'mp4' file.
6. Run the tests and check that the new tests pass correctly.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**
The related documentation will be updated once this PR is approved. The documentation will not be merged until the release data of this improvement.


CHANGELOG.md
```diff
@@ -4,13 +4,31 @@
 
 ### Adds
 
-* Add "showQuery" in piece-page-type in order to override the query for the "show" page as "indexQuery" does it for the index page
+* New `APOS_LOG_ALL_ROUTES` environment variable. If set, Apostrophe logs information about all middleware functions and routes that are executed on behalf of a particular URL.
+* Adds the `addFileGroups` option to the `attachments` module. This allows easier addition of new file groups or extension of the existing groups.
```
boutell, line 8: `attachment` not `attachments`

modules/@apostrophecms/attachment/index.js
```diff
@@ -87,6 +87,28 @@ module.exports = {
...
+          };
+          if (newGroup.extensionMaps) {
+            existingGroup.extensionMaps = {
+              ...existingGroup.extensionMaps,
+              ...newGroup.extensionMaps
+            };
+          }
+          const groupIndex = self.fileGroups.findIndex(existingGroup => existingGroup.name === newGroup.name);
+          self.fileGroups.splice(groupIndex, 1);
+          self.fileGroups.push(existingGroup);
```
boutell, line 105: The order of the groups doesn't matter, they are never enumerated in the UI.

boutell: If you factor this inner bit out to an `addFileGroup` method, then we'll have something that modules can call to programmatically add a file group during initialization without breaking project level use of the `addFileGroups` option.

boutell (changes requested):

modules/@apostrophecms/attachment/index.js
BoDonkey: Re: the order. I did know that the order didn't matter. Embarrassingly, I did not know that find mutated in place. I thought it returned a new object that had to be used to replace the original, hence the splice and push.

Re: refactoring. I factored it out (I hope I understood the ask), but I wouldn't mind an explanation of "without breaking project level use of the addFileGroups option". Not sure why factoring out was needed.

boutell (approved):
Providing a method like this allows a reusable npm module, let's say `@apostrophecms/hosted-video-widget`, to add a group without messing about with an option the project level developer may also be trying to use. Options tend to be great for project level but something of a pain to use with `improve` and out of the question to set from another module.

... Yeah, in general you can assume you're not going to get a clone back from a method like `find`, JavaScript doesn't have a native way to deep clone after all and it's more performant not to clone when you don't need to.

PR #3918, opened 2022-10-25
QA errata for placeholders

boutell:
* do not try to "convert" while aposPlaceholder is still in effect (it breaks required fields)
* do not clone unnecessarily (performance)
* always respect placeholders regardless of field values until aposPlaceholder is cleared

See the `link-widget` branch of `testbed` which introduces a simple link widget that exercises the features that required the "no convert when aposPlaceholder is true" fix. The other changes were not critical for that particular QA test but are still improvements.

### DIFF
--- a/test/modules/bundle-page/ui/src/extra.js
+++ b/test/modules/bundle-page/ui/src/extra.js
@@ -1,1 +1,1 @@
-export default () => {};
+export default () => { 'bundle-page-watcher-test-30'; };
--- a/test/modules/bundle-page/ui/src/extra.js
+++ b/test/modules/bundle-page/ui/src/extra.js
@@ -1,1 +1,1 @@
-export default () => { 'bundle-page-watcher-test-30'; };
+export default () => {};
--- a/test/widgets.js
+++ b/test/widgets.js
@@ -212,11 +212,6 @@ describe('Widgets', function() {
           {
             _id: 'widget4',
             ...widgetData
-          },
-          {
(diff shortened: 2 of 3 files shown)


test/modules/bundle-page/ui/src/extra.js
```diff
@@ -1 +1 @@
-export default () => {};
+export default () => { 'bundle-page-watcher-test-30'; };
```
ETLaurent, line 1: related change?

test/widgets.js
```diff
@@ -263,14 +258,6 @@ describe('Widgets', function() {
...
       });
 
-      it('should not render the placeholders when widget\'s fields are defined', function() {
-        assert(result.includes('<li>widget5 - string: Some string</li>'));
-        assert(result.includes('<li>widget5 - integer: 2</li>'));
-        assert(result.includes('<li>widget5 - float: 2.2</li>'));
-        assert(result.includes('<li>widget5 - date: 2022-09-21</li>'));
-        assert(result.includes('<li>widget5 - time: 15:39:12</li>'));
-      });
-
```
ETLaurent, line 273: this test doesn't pass anymore? due to the change in sanitize method?

boutell: Correct. We don't care about user-entered values at all yet if `aposPlaceholder` is still set and the field has a `placeholder`.

ETLaurent (approved):

PR #3926, opened 2022-10-28
PRO-3170: Better i18n UI state management

myovchev:
## Summary

Solves the problem of navigating the modal wizard, while the state is not ready for that. No matter we introduce the global busy state (the blue screen of death) on every request to the back-end, it takes reasonable time before it shims the entire screen. With this patch the action buttons are always disabled when an action should not be performed.

See the ticket acceptance criteria.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
+++ b/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
@@ -500,12 +500,10 @@ export default {
           .map(doc => [ doc.aposLocale.split(':')[0], doc ])
       );
       await this.updateRelatedDocs();
-    } catch (e) {
-      //
+    } finally {
+      this.wizard.step = this.visibleStepNames[0];
+      this.wizard.busy = false;
     }
-
-    this.wizard.step = this.visibleStepNames[0];
-    this.wizard.busy = false;
   },
   methods: {
     close() {
@@ -742,8 +740,6 @@ export default {
         });
         const filtered = docs.filter(doc => result.editable.includes(doc._id));
         return filtered;
-      } catch (e) {
-        //
       } finally {
         this.wizard.busy = status;
       }
--- a/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
+++ b/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
@@ -224,7 +224,7 @@
             v-if="isLastStep()"
             type="primary"
             label="apostrophe:localizeContent"
-            :disabled="!complete()"
+            :disabled="!complete() || wizard.busy"
             @click="submit"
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -713,18 +727,26 @@ export default {
...
+      try {
+        const result = await apos.http.post(`${apos.doc.action}/editable?aposMode=draft`, {
+          body: {
+            ids: docs.map(doc => doc._id)
+          }
+        });
+        const filtered = docs.filter(doc => result.editable.includes(doc._id));
+        return filtered;
+      } catch (e) {
+        //
```
boutell, line 746: You don't have to have a catch block just to have a finally block. Since nothing useful is happening in this catch block it is better practice to let it throwwww to the calling function. You can still have a finally block.

modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -770,6 +792,8 @@ export default {
       if (this.wizard.values.toLocalize.data === 'thisDoc') {
         return;
       }
+      const status = this.wizard.busy;
```
boutell, line 795: Is this unused?

boutell (changes requested):
A great fix for this component. 💯 Minor requests.

Re: "busy: true" not blocking all interaction right away, if it's not doing that then we should open a ticket to fix it. Maybe due to an animation?

While it is designed not to be visually obvious for the first 50ms or so, to avoid the "strobe light" effect, it should block all interaction immediately.

modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
myovchev: Nope, it's smart enough to keep the previous busy state, see `this.wizard.busy = status;` at the end of the method

myovchev:
> A great fix for this component. 100 Minor requests.
> 
> Re: "busy: true" not blocking all interaction right away, if it's not doing that then we should open a ticket to fix it. Maybe due to an animation?
> 
> While it is designed not to be visually obvious for the first 50ms or so, to avoid the "strobe light" effect, it should block all interaction immediately.

I think busy is doing fine. It involves animation and IMO does the right thing UX-wise. The "lag" comes from what the browser does, and Cypress can "skip" it because of the "nextTick". The disabled state here makes Cypress very happy and is also good UX (some state lag is not necessary request related).

boutell (approved):

PR #3929, opened 2022-11-01
PRO-3214: Add module re-bundling feature

myovchev:
## Summary

Allows project-level developer to override bundling decisions by configuring the @apostrophecms/asset module, like this:
```js
// modules/@apostrophecms/asset/index.js
module.exports = {
  options: {
    rebundleModules: {
      // Everything from the fancy-form module should go in the regular "main" Apostrophe bundle
      'fancy-form': 'main',
      // Everything from the basic-product module should go in the "secondary" bundle,
      // including the module owned ui/src/index.js entrypoint
      'basic-product': 'secondary',
      // Only the ui/src/cart.js code from the advanced-product module should go 
      // in the "main" Apostrophe bundle
      'advanced-product:cart': 'main',
      // Only the ui/src/form.js code from the @dcad/form module should be retargeted to the
      // "secondary" bundle — but only that code, leave ui/src/index.js in the main bundle
      '@dcad/form:form': 'secondary'
    }
  }
};
```

**The following are considered invalid configurations:**
```js
// Sending everything to "main" and then sending 
// specific bundle "extra" to "new" in the same module doesn't make sense.
{
  'bundle-page': 'main',
  'bundle-page:extra': 'new'
}
// Sending everything to bundle "new" and then sending 
// specific bundle "extra" to "new" in the same module  doesn't make sense.
{
  'bundle-page': 'new',
  'bundle-page:extra': 'another'
}
```

**Edge cases**

- Targeting local override modules requires `my-` prefix as in this example: `'@apostrophecms/my-home-page:bundleName': 'main'`. Targeting the original (npm) module would be: `'@apostrophecms/home-page:bundleName': 'main'` 
- A bundle named "main" is just a regular bundle (no special treatment). However it conflicts with the special keyword `main` used in the rebundling configuration. `moduleName:bundleName = main` will send bundleName to the Apostrophe "main" public build. Keep in mind `moduleName:main = main` actually works - bundle named "main: from moduleName module will be moved to the main Apostrophe public build.

- Our asset build still works normally when this option is absent, for both the main bundle and an extra bundle.
- In the presence of the "moduleName": "bundleName" syntax above, all ui/src/someBundleName.js entry points in moduleName (including index.js) wind up mapped into bundleName. This is confirmed by a mocha test that checks whether strings are present in each bundle.
- In the presence of the "moduleName:bundleName": "newBundleName" syntax above, the code from ui/src/bundleName.js is remapped to newBundleName, but code from ui/src/index.js and any other bundles present in moduleName is not remapped.
- The use of the bundle name **main** as a target bundle value in any of the above configuration works as expected, i.e. contributing to the main bundle.
- Namespaced module names should work properly with the above syntax.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
 
 ## Unreleased
 
+Adds
+
+* Allow project-level developer to override bundling decisions by configuring the `@apostrophecms/asset` module.
```
boutell, line 7: using the new `rebundleModules` option. [Link to the reference page for the asset module, which will be updated around release time to cover this]

lib/bundle-helpers.js
```diff
@@ -0,0 +1,40 @@
+module.exports = {
+  reBundle(moduleName, bundles, rebundleConfigs) {
```
boutell, line 2: `rebundleModule` for consistency

lib/bundle-helpers.js
```diff
@@ -0,0 +1,40 @@
...
+          .reduce((all, opts) => ({
+            ...all,
+            ...opts
+          }), {});
+        result = { [entry.target]: options };
+        break;
+      }
+      // 3. Rename a single bundle.
+      if (entry.source) {
+      // 3.1. ... But it's sent to the main build
```
boutell, line 25: Indentation issue

lib/bundle-helpers.js
```diff
@@ -0,0 +1,40 @@
+module.exports = {
```
boutell, line 1: These are methods. So if you're going to break out some methods with a particular theme to a file in lib, go with `lib/bundle-methods.js`. Helpers implies something else (nunjucks helpers, which these are not).

lib/bundle-helpers.js
boutell, line 2: Explain the options in a comment, particularly the second and third — I'm not sure what I'm looking at.

modules/@apostrophecms/asset/lib/webpack/utils.js
```diff
@@ -276,7 +371,57 @@ async function verifyBundlesEntryPoints (bundles) {
   return packedFilesByBundle;
 };
 
-function formatConfigs (chain, webpackConfigs) {
+// Normalize config:
```
boutell, line 374: OK thanks for this comment.

boutell: Comment on reBundle could just refer to this function to explain its options.

modules/@apostrophecms/asset/lib/webpack/utils.js
```diff
@@ -276,7 +371,57 @@ async function verifyBundlesEntryPoints (bundles) {
...
+      catchAllModules.push(entry.name);
+    } else {
+      targeted.push(entry);
+    }
+  }
+  for (const entry of targeted) {
+    if (catchAllModules.includes(entry.name)) {
+      throw new Error(
+        'Invalid apos.asset.options.rebundleModules: ' +
+        `module ${entry.name} can't have more re-bundle options when` +
```
boutell, line 417: As a developer I wouldn't know what to change in my code after reading this. What's a catch-all directive? Can this error actually happen as far as you know or is this more of a "we have a bug, sorry" message?

modules/@apostrophecms/asset/lib/webpack/utils.js
```diff
@@ -276,7 +371,57 @@ async function verifyBundlesEntryPoints (bundles) {
...
+        'having a catch-all directive.'
+      );
+    }
+  }
+}
+
+function formatConfigs (chain, webpackConfigs, rebundleModules) {
```
boutell, line 424: What does this function do?

modules/@apostrophecms/template/lib/bundlesLoader.js
```diff
@@ -1,4 +1,5 @@
 const { stripIndent } = require('common-tags');
+const { reBundle } = require('../../../../lib/bundle-helpers');
```
boutell, line 2: Please follow the Apostrophe pattern of methods so that extension/override at project level or in an `improve` module is possible. If you want to break them out to separate files you can do that with `...` and `require` inside `methods`.

boutell (changes requested):

modules/@apostrophecms/asset/lib/webpack/utils.js
myovchev: This is an old function, merging and transforming webpack configs.

lib/bundle-helpers.js
myovchev: I think I just keep this function as a asset method as you suggested below. Makes much more sense.

modules/@apostrophecms/asset/lib/webpack/utils.js
myovchev: It is an old one. Merges and transforms webpack configs.

modules/@apostrophecms/asset/lib/webpack/utils.js
myovchev: Fixed to: `Error: Invalid apos.asset.options.rebundleModules: "bundle-page: main" conflicts with "bundle-page:extra: new"`

boutell (approved):

boutell (changes requested):
Looks good, but please address changelog conflicts.

CHANGELOG.md
```diff
@@ -5,11 +5,15 @@
 ### Adds
 
 * Allow project-level developer to override bundling decisions by configuring the `@apostrophecms/asset` module. Check the [module documentation](https://v3.docs.apostrophecms.org/reference/modules/asset.html#options) for more information.
+* Adds Reset Password feature to the login page. Note that the feature must be enabled and email delivery must be properly configured. See the [documentation](https://v3.docs.apostrophecms.org/) for more details.
```
boutell, line 8: Let's try to get this updated with a more specific link just before release.

boutell (approved):

PR #3930, opened 2022-11-01
Toggle errata

boutell:
## Summary

PR to your PR to my PR:

* If an item has errors always toggle it open
* If there is no title field open all the items and do not offer a toggle (probably most natural for single field scenarios)

* Make a subfield required and mess about with saving the document without completing it
* Check out the behavior with a single subfield and no titleField configured, see if you like my tweak there

*(Check at least one)*

- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
@@ -58,7 +58,7 @@
                   class="apos-input-array-inline-schema-wrapper"
                 >
                   <AposSchema
-                    :schema="effectiveSchema"
+                    :schema="field.schema"
                     v-model="item.schemaInput"
                     :trigger-validation="triggerValidation"
                     :utility-rail="false"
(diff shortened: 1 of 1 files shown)


stuartromanek:
Not sure I agree the absence of `titleField` connotes 'expand all items all the time' .. to me it would make sense to
- keep schemas expanded when the array item schema is under a certain length (2 or less items?)
- an explicit flag in the schema def, like
 ```
footerLinks: {
  label: 'Footer Links',
  type: 'array',
  inline: {
    expand: false
  },
  fields: { ... }
},
 ```
The flag feels a little extra since we're hustling on this but idk, thoughts?

stuartromanek:
To clarify ^ i think those two bullets would be used together.. 
If an inline array is present with under 3 child fields, auto expand the schemas unless the flag is included and says otherwise.
If an inline array is present with greater than 3 child fields, auto collapse the schemas unless the flag is included and says otherwise.

boutell:
* Implemented default behavior per Stu's suggestion
* Implemented the sub-option to override as suggested too, but called it `alwaysExpand` because I kept getting confused about whether `expand: true` meant "expandable" or "always expanded," this is unambiguous

stuartromanek (approved):

PR #3935, opened 2022-11-07
Pro 3244 command menu module

haroun:
## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/command-menu/index.js
```diff
@@ -0,0 +1,163 @@
...
+      'apostrophe:ready': {
+        composeCommands() {
+          self.rawCommands = Object.values(self.apos.modules).flatMap(self.composeCommandsForModule);
+          self.removes = [];
+          self.commands = {};
+          self.groups = {};
+
+          const composed = pipe(self.composeCommand, self.composeRemove, self.composeGroup)({ rawCommands: self.rawCommands });
+          try {
+            Object.entries(composed.command).every(([ name, command ]) => self.validateCommand({ name, command }));
```
boutell, line 19: You're using `every` but you don't seem to care about the return value. Use `forEach`, or fail in some appropriate way if you don't like the return value of `every`.

modules/@apostrophecms/command-menu/index.js
```diff
@@ -0,0 +1,163 @@
+const assert = require('assert').strict;
+const pipe = (...functions) => (initial) => functions.reduce((accumulator, current) => current(accumulator), initial);
```
boutell, line 2: 🧙

modules/@apostrophecms/command-menu/index.js
boutell: I think you don't care about it because your asserts will stop the whole show anyway. If that's the case use forEach.

boutell (changes requested):
Looks very elegant, minor concerns.

modules/@apostrophecms/command-menu/index.js
```diff
@@ -16,8 +16,8 @@ module.exports = {
 
           const composed = pipe(self.composeCommand, self.composeRemove, self.composeGroup)({ rawCommands: self.rawCommands });
           try {
-            Object.entries(composed.command).every(([ name, command ]) => self.validateCommand({ name, command }));
-            Object.entries(composed.group).every(([ name, group ]) => self.validateGroup({ name, group }));
+            Object.entries(composed.command).some(([ name, command ]) => self.validateCommand({ name, command }));
```
boutell, line 19: `some` stops on the first invocation that returns `true`, and `validateCommand` always does that, so I'm pretty sure it is wrong to use `some`.

As near as I can tell you don't care what the return value is, you're just stopping with an exception on any assert failures. That's fine, but don't return a value and use `some` if what you're really doing is iterating over everything unless an exception is thrown; it's confusing for later maintenance. Just use `forEach` and rely on the asserts as you have written them, unless I have misunderstood.

boutell (changes requested):

haroun:
I'm not sure about the get all errors at once thing I've just pushed. It's kind of in progress. I'll move the PR back to draft. Feedbacks are welcomed.

CHANGELOG.md
```diff
@@ -1,14 +1,41 @@
 # Changelog
 
-## Unreleased
+## UNRELEASED
 
-###  Fixes
+### Adds
+
+* Relationships load more quickly.
```
boutell, line 7: Please merge main into `keyboard-shortcut` also and let me know, and then I should be able to see just the relevant changes in this PR.

boutell (changes requested):

CHANGELOG.md
boutell: (Right now the github pull request view shows many recent changes in main because your new branch is more up to date with `main` than the merge target is.)

haroun: I'm sorry, I forgot to update the branch `feature/keyboard-shortcut` where this PR will be merged.

modules/@apostrophecms/command-menu/index.js
```diff
@@ -0,0 +1,199 @@
...
+  handlers(self) {
+    return {
+      'apostrophe:ready': {
+        composeCommands() {
+          self.rawCommands = Object.values(self.apos.modules).flatMap(self.composeCommandsForModule);
+          self.removes = [];
+          self.commands = {};
+          self.groups = {};
+
+          const composed = pipe(self.composeCommand, self.composeRemove, self.composeGroup)({ rawCommands: self.rawCommands });
```
boutell, line 17: I think this strategy might not permit a subclass to add back a "field" (a command) removed by a parent class, because you do all the adds and then all the removes. If you look at the implementation in moog.js it does permit that, because it loops through the steps and applies add, remove and group on each pass. You need equivalent logic.

If it's helpful feel free to factor that logic up to a helper that can also be called from your code.

Also consider whether you can just use a regular cascade rather than an unparsed section, I'm not quite sure if you actually need any differences in the behavior of the logic.  The piece module declares several cascades and has no custom logic to parse them. When you declare a cascade that way it just works like `fields` automatically.

boutell (changes requested):

haroun:
I've renamed `fields` into `commands` and `groups` into `modals`.
`groups` is still present because it will be used for Command Menu as is.

modules/@apostrophecms/command-menu/index.js
```diff
@@ -0,0 +1,310 @@
...
+          .filter(([ success ]) => !success)
+          .map(([ , error ]) => error);
+        if (errors.length) {
+          throw new Error('Invalid', { cause: errors });
+        }
+      },
+      buildCommands(initialState) {
+        const concatenate = self.apos.util.omit(
+          initialState.composed.commands,
+          initialState.composed.removes
```
boutell, line 191: This needs to respect inheritance order. If a base class adds a command, a subclass removes it, and a sub-subclass adds it back  (often project level for instance), then that needs to work. Make sure you include that unit test.

That means this has to happen iteratively through the layers of the initialization of the object — each layer must add, remove and group things, and then the next layer has a go at it.

Look at how it is done in `moog.js` where the regular cascades are implemented. If you can't just make this a regular cascade and benefit from that logic like fields, columns, etc. do (could you? Worth a look) then you need to make sure you have that same capability.

boutell (changes requested):

modules/@apostrophecms/command-menu/index.js
haroun: I've achieved that by building sequential removes in https://github.com/apostrophecms/apostrophe/pull/3935/files#diff-91dd88c4b9d7b9798c4f0b88ef58039a9adbd9383ec8b24a035694d85a344a91R79

and removing group commands duplicates in https://github.com/apostrophecms/apostrophe/pull/3935/files#diff-91dd88c4b9d7b9798c4f0b88ef58039a9adbd9383ec8b24a035694d85a344a91R113

As discussed yesterday, `cascades` won't achieve my goal because I'm building against a single pool of data shared across all the apostrophe instance instead of one pool of data per module like we do with moog.

boutell (approved):
Thanks for the unit test.

PR #3939, opened 2022-11-09
Add "publish" batch feature to manager modal

falkodev:
## Summary

Add "publish" batch feature in manager modal for pieces.

*For example:*
1. `npm link apostrophe` in testbed
2. Display a manager modal: a new icon is present
3. Select draft pieces and publish them in batch

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -104,6 +104,19 @@ module.exports = {
   },
   utilityOperations: {},
   batchOperations: {
+    publish: {
+      label: 'apostrophe:publish',
+      messages: {
+        progress: 'Publishing {{ type }}...',
+        completed: 'Published {{ count }} {{ type }}.'
+      },
+      icon: 'checkbox-marked-icon',
+      modalOptions: {
+        title: 'apostrophe:publishType',
+        description: 'apostrophe:publishBatchConfirmation',
+        confirmationButton: 'apostrophe:publishBatchConfirmationButton'
+      }
+    },
     add: {
       archive: {
         label: 'apostrophe:archive',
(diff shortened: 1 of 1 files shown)


boutell (approved):
Just for pieces for now? OK yeah, that bypasses the concerns about the page tree and I guess it meets your needs. Approved.

boutell (changes requested):
Tests are failing. Then I can approve.

boutell (changes requested):
A problem: "submitted drafts" shows all content types including pages although it is a virtual "piece type." And "batch publish" is clearly a desirable feature there. But, we don't have the logic to handle pages in the right order yet.

So I don't think we can delay writing that logic.

boutell:
There is no batch publish feature for pages right now.

I earlier shared the concern that if we implement one, then it has to publish things in the appropriate order, parent pages first, and skip altogether any pages whose parents are not published, and report any skipped pages to the user in a useful way, and publish all selected pieces before any selected pages.

You responded by implementing batch publish for pieces only, which seemed like a good workaround for now.

However since the "submitted drafts" piece type is actually polymorphic and shows all submitted pieces *and* pages, we will have to either (a) disable batch publish in that module, or (b) implement it correctly with the checks above.

falkodev:
Ok, thanks for the clarification. I disabled it for pages. I think this is the best and fastest solution currently.

boutell:
(Note that disabling it for the "manage submitted drafts" view is a pretty big bummer, as that's a major use case of the feature, but if we're not going to fix the handling of pages right now then we need to do that.)

boutell (approved):

PR #3941, opened 2022-11-18
"reusable content" widgets work properly

boutell:
## Summary

* It is now possible to add a widget beneath a widget that displays areas taken from a related document. See https://github.com/boutell/reusable-content-demo for a really simple demo. Without this branch, if your first act is to create a reusable content piece and then add a reusable content widget to the home page, you can't add any more widgets to the first area on the home page.

See above.

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -48,6 +48,7 @@
         :field-id="fieldId"
         :disabled="field && field.readOnly"
         :widget-hovered="hoveredWidget"
+        :non-foreign-widget-hovered="hoveredNonForeignWidget"
```
boutell, line 51: If the currently hovered widget is foreign, then the "hovered non foreign widget" will be its closest non-foreign ancestor.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -5,6 +5,8 @@
     :class="{'apos-area-widget-wrapper--foreign': foreign}"
     :data-area-widget="widget._id"
     :data-area-label="widgetLabel"
+    :data-apos-widget-foreign="foreign ? 1 : 0"
```
boutell, line 8: Selecting an element with "0" as a value is easier than selecting an element with no value at all.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -395,11 +401,16 @@ export default {
       if (this.focused) {
         return;
       }
-      this.state.add.top.show = true;
-      this.state.add.bottom.show = true;
+      if (!this.foreign) {
+        this.state.add.top.show = true;
```
boutell, line 405: The + buttons are never desirable in a foreign widget (but are desirable in its closest non-foreign ancestor).

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -395,11 +401,16 @@ export default {
...
+        this.state.add.bottom.show = true;
+      }
       this.state.container.highlight = true;
       this.state.labels.show = true;
-      apos.bus.$emit('widget-hover', this.widget._id);
+      apos.bus.$emit('widget-hover', {
+        _id: this.widget._id,
+        nonForeignId: this.foreign ? this.$el.closest('[data-apos-widget-foreign="0"]').getAttribute('data-apos-widget-id') : null
```
boutell, line 412: This is a necessary compromise to discover the closest non-foreign widget.

I tried having the parent find the child, but this doesn't work well because of chicken and egg problems. Looking back up the tree is a sound strategy.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -432,7 +443,7 @@ export default {
         this.focused = false;
         this.resetState();
         this.highlightable = false;
-        document.removeEventListener('click', this.blurUnfocus);
+        document.removeEventListener('click', this.unfocus);
```
boutell, line 446: Unrelated bug leading to runaway event handler adds.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -299,9 +307,9 @@ export default {
         container: this.state.container.focus ? this.classes.focus
           : (this.state.container.highlight ? this.classes.highlight : null),
```
stuartromanek, line 308: this line also needs to know whether a nonForeign is hovered so we can make the dashed lines appear
`: ((this.state.container.highlight || this.nonForeignHovered) ? this.classes.highlight : null),`

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -152,6 +154,10 @@ export default {
       type: String,
       default: null
     },
+    nonForeignWidgetHovered: {
```
stuartromanek, line 157: this prop doesn't get cleaned up (as far as i can tell) so it can't be used to help the UI know when we've stopped hovering foreign widgets

stuartromanek (changes requested):
A UX wrinkle: the plus buttons don't get hidden when we've blurred the foreign widget (and entire area).. it just leaves them.
https://user-images.githubusercontent.com/1889830/203106041-2a211e31-6638-47c0-8fd1-ad8f11fe1a67.mov

Also, not sure this is the right place for this idea, but being able to hover the individual foreign widgets in a foreign area feels like it adds to the confusion of what this widget's mode or purpose is. It seems like it would be better if you could only hover the entire foreign widget and be given its jump off point rather than each thing inside it. I think that would help emphasize this is different from normal, local widgets.

stuartromanek (approved):
needed to rebuild 👍

PR #3958, opened 2022-12-09
fix the search index

boutell:
Makes sure you can search i18n static pieces by value and not just by key. This was never properly ported from A2 to A3 before.

### DIFF
--- a/modules/@apostrophecms/schema/lib/addFieldTypes.js
+++ b/modules/@apostrophecms/schema/lib/addFieldTypes.js
@@ -78,6 +78,18 @@ module.exports = (self) => {
           }
         }
       }
+    },
+    index: function (value, field, texts) {
+      for (const item of ((value && value.items) || [])) {
+        const manager = self.apos.area.getWidgetManager(item.type);
+        if (!manager) {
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/search/index.js
```diff
@@ -316,29 +327,11 @@ module.exports = {
           text: doc.slug,
           silent: true
         });
-
-        // Areas can be schemaless so find them automatically
-        self.apos.area.walk(doc, function (area, dotPath) {
```
boutell, line 321: Areas are always in the schema now.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -78,6 +78,18 @@ module.exports = (self) => {
           }
         }
       }
+    },
+    index: function (value, field, texts) {
+      for (const item of ((value && value.items) || [])) {
+        const manager = self.apos.area.getWidgetManager(item.type);
```
falkodev, line 84: Should you check if `item.type` is defined?

falkodev (approved):

PR #3962, opened 2022-12-14
Pro 3412 remove duplicated content when localized false

falkodev:
## Summary

2 tasks added:
- unlocalize: remove duplicated content when localized false
- localize: create localized content from existing documents when localized true

1. Add option `localized: false` to a module 
2. Run the command `node app [moduleName]:unlocalize` and check the DB
3. Switch `localized` to `true` and run `node app  [moduleName]:localize` and check the DB
4. Run ``ode app help  [moduleName]unlocalize` and ``node app help  [moduleName]:unlocalize`

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * Relationships load more quickly.
 * Parked page checks at startup are faster.
-* Tasks to localize and unlocalize content (see `node app help [yourModuleName]:localize` and `node app help [yourModuleName]:unlocalize`).
+* Tasks to localize and unlocalize piece type content (see `node app help [yourModuleName]:localize` and `node app help [yourModuleName]:unlocalize`).
 ## 3.33.0 (2022-11-28)
 
 ### Adds
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * Relationships load more quickly.
 * Parked page checks at startup are faster.
-
+* Tasks to localize and unlocalize content (see `node app help [yourModuleName]:localize` and `node app help [yourModuleName]:unlocalize`).
 ## 3.33.0 (2022-11-28)
 
 ### Adds
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -92,6 +92,11 @@ module.exports = {
     self.apos.doc.setManager(self.name, self);
     self.enableBrowserData();
     self.addContextMenu();
+
+    // force autopublish to false when not localized to avoid bizarre configuration
+    if (!self.options.localized) {
+      self.options.autopublish = false;
+    }
   },
   handlers(self) {
     return {
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -1066,44 +1066,46 @@ module.exports = {
         usage: 'Add draft version documents for each locale when a module has the "localized" option.' +
         '\nExample: node app [moduleName]:localize',
         async task() {
-          if (self.options.localized) {
-            console.log('Adding draft version to documents');
-            const locales = Object.keys(self.apos.i18n.locales);
-
-            await self.apos.migration.eachDoc({ type: self.name }, async doc => {
-              await self.apos.doc.db.removeOne({ _id: doc._id });
-
-              if (doc.aposDocId && !doc._id.endsWith('published') && !doc._id.endsWith('draft')) {
-                for (const locale of locales) {
-                  const lastPublishedAt = new Date();
-                  const newDraft = {
-                    ...doc,
-                    aposLocale: `${locale}:draft`,
-                    aposMode: 'draft',
-                    aposDocId: doc._id,
-                    _id: `${doc.aposDocId}:${locale}:draft`,
-                    lastPublishedAt
-                  };
-                  const newPublished = {
-                    ...doc,
-                    aposLocale: `${locale}:published`,
-                    aposMode: 'published',
-                    aposDocId: doc._id,
-                    _id: `${doc.aposDocId}:${locale}:published`,
-                    lastPublishedAt
-                  };
-                  await self.apos.doc.db.insertOne(newDraft);
-                  await self.apos.doc.db.insertOne(newPublished);
-                }
-              }
-            });
-
-            await self.apos.attachment.recomputeAllDocReferences();
-
-            console.log(`Done localizing module ${self.name}`);
-          } else {
+          if (!self.options.localized) {
             throw new Error('Localized option not set to true, so the module cannot be localized.');
           }
+
+          console.log('Adding drafts and locales to documents');
+
+          const locales = Object.keys(self.apos.i18n.locales);
+          const lastPublishedAt = new Date();
+          const inserts = [];
+
+          await self.apos.migration.eachDoc({ type: self.name }, async doc => {
+            if (doc.aposDocId && !doc._id.endsWith('published') && !doc._id.endsWith('draft')) {
+              await self.apos.doc.db.removeOne({ _id: doc._id });
+
+              for (const locale of locales) {
+                const newDraft = {
+                  ...doc,
+                  aposLocale: `${locale}:draft`,
+                  aposMode: 'draft',
+                  aposDocId: doc._id,
+                  _id: `${doc.aposDocId}:${locale}:draft`
+                };
+                const newPublished = {
+                  ...doc,
+                  aposLocale: `${locale}:published`,
+                  aposMode: 'published',
+                  aposDocId: doc._id,
+                  _id: `${doc.aposDocId}:${locale}:published`,
+                  lastPublishedAt
+                };
+                inserts.push(newDraft);
+                inserts.push(newPublished);
+              }
+            }
+          });
+
+          await self.apos.doc.db.insertMany(inserts);
+          await self.apos.attachment.recomputeAllDocReferences();
+
+          console.log(`Done localizing module ${self.name}`);
         }
       },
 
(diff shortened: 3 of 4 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
+                  await self.apos.doc.db.insertOne(newPublished);
+                }
+              }
+            });
+
+            await self.apos.attachment.recomputeAllDocReferences();
+
+            console.log(`Done localizing module ${self.name}`);
+          } else {
+            throw new Error('Localized option not set to true, so the module cannot be localized.');
```
boutell, line 1105: Put this test and `throw` up front to save on nesting.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
           }
         }
+      },
+
+      localize: {
+        usage: 'Add draft version documents for each locale when a module has the "localized" option.' +
+        '\nExample: node app [moduleName]:localize',
+        async task() {
+          if (self.options.localized) {
+            console.log('Adding draft version to documents');
```
boutell, line 1070: Adding drafts and locales to documents

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
+            const locales = Object.keys(self.apos.i18n.locales);
+
+            await self.apos.migration.eachDoc({ type: self.name }, async doc => {
+              await self.apos.doc.db.removeOne({ _id: doc._id });
+
+              if (doc.aposDocId && !doc._id.endsWith('published') && !doc._id.endsWith('draft')) {
```
boutell, line 1076: Here you think about not tampering with docs that are somehow already localized if you come across them, which is a good idea, but on the previous line you still remove them. Make this check up front so you preserve them.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
+                for (const locale of locales) {
+                  const lastPublishedAt = new Date();
```
boutell, line 1078: Generate this value once outside the locales loop, for speed and consistency.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
+                  };
+                  const newPublished = {
+                    ...doc,
+                    aposLocale: `${locale}:published`,
+                    aposMode: 'published',
+                    aposDocId: doc._id,
+                    _id: `${doc.aposDocId}:${locale}:published`,
+                    lastPublishedAt
+                  };
+                  await self.apos.doc.db.insertOne(newDraft);
```
boutell, line 1095: You can use `insertMany` here for a lil' speed boost, array of two. Actually, you can do an array including all of them across all of the locales and make just one `insertMany` call.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,81 @@ module.exports = {
...
+      },
+
+      unlocalize: {
+        usage: 'Remove duplicate documents when a module has not "localized" and "autopublish" anymore.' +
+        '\nOptions are:' +
+        '\n- locale: if not set, it is the project\'s default locale' +
+        '\n- mode: by default, published' +
+        '\nExample: node app [moduleName]:unlocalize --mode=published --locale=en',
+        async task(argv) {
+          if (!self.options.autopublish && !self.options.localized) {
```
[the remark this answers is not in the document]
boutell: Yeah, I think it makes more sense to patch `doc-type` so that it shuts off `autopublish` in `init` if `localized` is seen to be false, otherwise this kind of check is going to proliferate.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
+                  _id: doc.aposDocId
+                };
+                await self.apos.doc.db.insertOne(newDoc);
+              }
+            });
+            await self.apos.attachment.recomputeAllDocReferences();
+
+            console.log(`Done unlocalizing module ${self.name}`);
+          } else {
+            throw new Error('Autopublish or localized option are not false, so the module cannot be unlocalized.');
```
boutell, line 1138: Same suggestion to factor this check and throw to the top and not nest the whole thing. Otherwise though this task seems 💯

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,84 @@ module.exports = {
...
+          if (!self.options.autopublish && !self.options.localized) {
+            const locale = argv.locale || self.apos.i18n.defaultLocale;
+            const mode = argv.mode || 'published';
+            console.log(`Removing duplicated documents and updating ${mode} ones`);
+
+            await self.apos.migration.eachDoc({ type: self.name }, async doc => {
+              await self.apos.doc.db.removeOne({ _id: doc._id });
+              if (doc.aposDocId && doc.aposLocale === `${locale}:${mode}` && doc.aposMode === mode) {
+                const newDoc = {
+                  ...doc,
```
boutell, line 1126: Remove `lastPublishedAt` just to be thorough

boutell (changes requested):

boutell:
This is very good, minor concerns overall.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 * Relationships load more quickly.
 * Parked page checks at startup are faster.
-
+* Tasks to localize and unlocalize content (see `node app help [yourModuleName]:localize` and `node app help [yourModuleName]:unlocalize`).
```
boutell, line 9: piece type content

modules/@apostrophecms/doc-type/index.js
```diff
@@ -92,6 +92,11 @@ module.exports = {
     self.apos.doc.setManager(self.name, self);
     self.enableBrowserData();
     self.addContextMenu();
+
+    // force autopublish to false when not localized to avoid bizarre configuration
+    if (!self.options.localized) {
+      self.options.autopublish = false;
```
boutell, line 98: 💯

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,94 @@ module.exports = {
...
+                  _id: `${doc.aposDocId}:${locale}:published`,
+                  lastPublishedAt
+                };
+                inserts.push(newDraft);
+                inserts.push(newPublished);
+              }
+            }
+          });
+
+          await self.apos.doc.db.deleteMany({ _id: { $in: deletes }});
```
boutell, line 1106: This is a little *too* efficient (: You could run out of RAM and fail on a large database. `insertMany` is good, but only once per document, after the loop over locales.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1060,94 @@ module.exports = {
...
+                ...doc,
+                aposLocale: undefined,
+                aposMode: undefined,
+                _id: doc.aposDocId
+              };
+              inserts.push(newDoc);
+            }
+          });
+
+          await self.apos.doc.db.deleteMany({ _id: { $in: deletes }});
```
boutell, line 1145: For this one, it's best not to do insertMany/deleteMany, for the reason I mentioned above.

boutell (changes requested):
also fix eslint issues reported

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1060,6 +1088,105 @@ module.exports = {
...
+            if (doc.aposDocId && doc.aposLocale === `${locale}:${mode}` && doc.aposMode === mode) {
+              const newDoc = {
+                ...doc,
+                aposLocale: undefined,
+                aposMode: undefined,
+                _id: doc.aposDocId
+              };
+              inserts.push(newDoc);
+
+              if (inserts.length > 100) {
```
boutell, line 1177: This is great, but how about moving this check into `self.flush` itself, along with the cleanup of the two arrays. This also makes it easier to check the length of both "inserts" and "deletes".

To clean them up even though they are passed in as arguments, take advantage of "splice".

The less code you have to write every time you call flush, the less can go wrong.

Also I suggest renaming it "flushInsertsAndDeletes" just to avoid camping on a potentially common method name.

boutell (changes requested):

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1174,15 +1176,11 @@ module.exports = {
...
-                self.flush(deletes, inserts);
-                deletes = [];
-                inserts = [];
-              }
+              await self.flushInsertsAndDeletes(inserts, deletes);
             }
           });
 
-          self.flush(deletes, inserts);
+          await self.flushInsertsAndDeletes(inserts, deletes);
```
boutell, line 1183: You need a third argument here to say "I don't care if there are 100 of them or not, we need to flush the rest now."

boutell (changes requested):

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1022,13 +1022,13 @@ module.exports = {
           }
         }
       },
-      async flushInsertsAndDeletes(inserts, deletes) {
-        if (inserts.length > 100) {
+      async flushInsertsAndDeletes(inserts, deletes, { force = false }) {
+        if (inserts.length > 100 || force) {
           await self.apos.doc.db.insertMany(inserts);
```
boutell, line 1027: Must not call if the array is empty (mongo is touchy about empty arguments).

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1022,13 +1022,13 @@ module.exports = {
...
           inserts.splice(0);
         }
 
-        if (deletes.length > 100) {
+        if (deletes.length > 100 || force) {
           await self.apos.doc.db.deleteMany({ _id: { $in: deletes } });
```
boutell, line 1032: Must not call if the array is empty (mongo is touchy about empty arguments).

boutell (changes requested):

boutell (approved):

PR #3968, opened 2022-12-16
PRO-3428 fix lists

boutell:
## Summary

* Fixed bulleted and ordered lists. You can select existing text and create a list from it
* You can then continue the list with the "enter" key and you get the default style of your particular rich text widget as expected inside each item

* You can select existing text and create a list from it (if you don't see any bullets make sure your project has styles, or check the inspector; but definitely make sure the markup is there)
* There will be paragraphs (or your other default style's tag) inside each list item, this is mandatory ProseMirror behavior, I did not come to catch that fish today

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
@@ -1,8 +1,16 @@
 // If a style has been marked `def`, we need to make sure it
-// and it's attributes are prioritized in our editor.
-// It's easier to create a new Node with our specifications and put it
-// at the front of the line than try to infer between editor instantiation
-// and the editor-user setting styles via the toolbar to know when its needed.
+// and its attributes are prioritized in our editor.
+//
+// Specifically, when the "enter" key is pressed, we don't want
+// tiptap/prosemirror to insert "paragraph" nodes in a rich
+// text widget in which they are not even configured as an option.
+//
+// The simplest way to ensure this is to create a new node type that
+// matches the tag and CSS class of the default style (the first
+// configured style, or the one marked with def: `true`).
+//
+// Matching extensions have been provided to the document and listItem types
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
```diff
@@ -1,8 +1,16 @@
 // If a style has been marked `def`, we need to make sure it
-// and it's attributes are prioritized in our editor.
-// It's easier to create a new Node with our specifications and put it
-// at the front of the line than try to infer between editor instantiation
-// and the editor-user setting styles via the toolbar to know when its needed.
+// and its attributes are prioritized in our editor.
+//
+// Specifically, when the "enter" key is pressed, we don't want
```
boutell, line 4: I did my best to beef up this already good summary, we'll see if it helps us next time 😆

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/ListItem.js
```diff
@@ -0,0 +1,6 @@
+import ListItem from '@tiptap/extension-list-item';
+export default (options) => {
+  return ListItem.extend({
+    content: 'defaultNode block*'
```
stuartromanek, line 4: looking at the original extension, so we *never* encounter `paragraph` the way this would expect because we always replace it, is that correct? is there any reason to add on to the default ie `defaultNode paragraph block*` ?

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
stuartromanek: much more comprehensible 😅

PR #3983, opened 2022-12-29
Fix and enhance the public build Webpack configuration merging

myovchev:
## Summary

- Do not merge module `webpack.extensions` configuration with the core `apos` build
- Allow applications and modules to control array and object Webpack configuration merging 

**A (Real World) Scenario**

A module `public-forms` is providing public forms via a Vue 3 UI. It extends webpack configuration in order to achieve that. Here is the relevant implementation, that is only possible with the changes made in this PR (explanations "How" can be found below).

```js
// possible npm module  'public-forms' - index.js
module.exports = {
  webpack: {
    extensions: {
      vue3: {
        resolve: {
          extensions: [ '*', '.js', '.ts', '.vue', '.json' ],
          alias: {
           // The application should introduce vue3 dependency, 
           // ensure to build without vue runtime duplicates.
            vue$: path.resolve(
              process.cwd(),
              './node_modules/vue/dist/vue.runtime.esm-bundler.js'
            )
          }
        },
        // Control the loaders resolution - this module, the app, finally apostrophe - in this order.
       // We will introduce `extensionOptions.vue3()` options in order to control this behavior outside.
        resolveLoader: {
          extensions: [ '*', '.js', '.ts', '.vue', '.json' ],
          modules: [
            path.join(__dirname, 'node_modules'),
            'node_modules',
            'node_modules/apostrophe/node_modules'
          ]
        },
        module: {
          rules: [
            // handle SFC
            {
              test: /\.vue$/,
              loader: 'vue-loader',
              options: {
                sourceMap: true
              }
            },
            // yup, Typescript too
            {
              test: /\.ts$/,
              loader: 'ts-loader',
              options: { appendTsSuffixTo: [ /\.vue$/ ] }
            },
            // for the <style  lang="css"> blocks, postcss can be added too
            {
              test: /\.css$/,
              use: [
                'vue-style-loader',
                {
                  loader: 'css-loader',
                  options: {
                    esModule: false,
                    sourceMap: true
                  }
                }
              ]
            }
          ]
        },
        plugins: [
          new VueLoaderPlugin(),
          // ... even more plugins
        ]
      }
    }
  },
  init(self) {
    self.enableBrowserData('public');
  },
  // ...components and other stuff
  methods(self) {
    return {
      getBrowserData(req) {
       // some useful Vue app configuration
        return {
          key: 'value'
        };
      }
    };
  }
};
```
The relevant portion of the core webpack configuration
```js
// apostrophe/modules/@apostrophecms/asset/lib/webpack/src/webpack.config.js
{
  // ...
  resolveLoader: {
    extensions: [ '*', '.js' ],
    modules: [ 'node_modules', 'node_modules/apostrophe/node_modules' ]
  },
  resolve: {
    extensions: [ '*', '.js' ],
    alias: {
      Modules: path.resolve(modulesDir)
    },
    modules: [
      'node_modules',
      `${apos.npmRootDir}/node_modules`,
      `${apos.npmRootDir}/node_modules/apostrophe/node_modules`
    ],
    symlinks: false
  },
  // ...
}
```

**The Benefits**
1. Unique configuration arrays for key parts of the webpack config
2. Before the change the resulting merged config will be:
```js
{
        resolve: {
          extensions: [ '*', '.js', '*',  '.js', '.ts', '.vue', '.json' ],
          // ...
        },
        resolveLoader: {
          extensions: [ '*', '.js', '*', '.js', '.ts', '.vue', '.json' ],
          modules: [
            'node_modules',
            'node_modules/apostrophe/node_modules'
            'node_modules/public/form/node_modules',
            'node_modules',
            'node_modules/apostrophe/node_modules'
          ]
        },
}
```
After the change and after the module has added a `srcCustomizeArray` handler it is:
```js
{
        resolve: {
          extensions: [ '*',  '.js', '.ts', '.vue', '.json' ],
          // ...
        },
        resolveLoader: {
          extensions: [ '*', '.js', '.ts', '.vue', '.json' ],
          modules: [
            'node_modules/public/form/node_modules',
            'node_modules',
            'node_modules/apostrophe/node_modules'
          ]
        },
}
```
The developer can fine tune very important parts - loaders resolving now defaults to the module that provides the relevant loaders (options could be provided to control that from outside), the application comes second, and the fallback is finally the Apostrophe owned modules.

This scenario is tested in a real life module and everything works as expected. 

- The `apos` build configuration should not contain anything from any `webpack.extensions` configuration
- `extensions` and `modules` Webpack configuration keys should not contain duplicates by default when  `webpack.extensions` provide such
- Extending or overriding the asset owned `srcCustomizeArray` and `srcCustomizeObject` methods should reflect in modified Webpack merge algorithm 

*(Check at least one)*

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
-* The module `webpack.extensions` configuration is not applied to the core Admin UI build anymore.
+* The module `webpack.extensions` configuration is not applied to the core Admin UI build anymore. This is the correct and intended behavior as explained in the [relevant documentation](https://v3.docs.apostrophecms.org/guide/webpack.html#extending-webpack-configuration).
 
 ## 3.36.0 (2022-12-22)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Adds
+
+* Use `mergeWithCustomize` when merging extended source Webpack configuration. Introduce overideable asset module methods `srcCustomizeArray` and `srcCustomizeObject`, with reasonable default behavior, for fine tuning Webpack config arrays and objects merging. More info - [the Webpack mergeWithCustomize docs](https://github.com/survivejs/webpack-merge#mergewithcustomize-customizearray-customizeobject-configuration--configuration)
+
+### Fixes
+
+* The module `webpack.extensions` configuration is not applied to the core Admin UI build anymore.
 
 ## 3.36.0 (2022-12-22)
 
--- a/modules/@apostrophecms/asset/index.js
+++ b/modules/@apostrophecms/asset/index.js
@@ -7,7 +7,7 @@ const path = require('path');
 const util = require('util');
 const express = require('express');
 const { stripIndent } = require('common-tags');
-const { merge: webpackMerge } = require('webpack-merge');
+const { mergeWithCustomize: webpackMerge } = require('webpack-merge');
 const cuid = require('cuid');
 const chokidar = require('chokidar');
 const _ = require('lodash');
@@ -73,7 +73,6 @@ module.exports = {
     self.buildWatcherEnable = process.env.APOS_ASSET_WATCH !== '0' && self.options.watch !== false;
     self.buildWatcherDebounceMs = parseInt(self.options.watchDebounceMs || 1000, 10);
     self.buildWatcher = null;
-
   },
   handlers (self) {
     return {
@@ -389,8 +388,11 @@ module.exports = {
                 es5TaskFn: self.es5TaskFn
               }, self.apos);
 
-              const webpackInstanceConfigMerged = self.webpackExtensions
-                ? webpackMerge(webpackInstanceConfig, ...Object.values(self.webpackExtensions))
+              const webpackInstanceConfigMerged = !options.apos && self.webpackExtensions
+                ? webpackMerge({
+                  customizeArray: self.srcCustomizeArray,
+                  customizeObject: self.srcCustomizeObject
+                })(webpackInstanceConfig, ...Object.values(self.webpackExtensions))
                 : webpackInstanceConfig;
 
               // Inject the cache location at the end - we need the merged config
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
 
+## Unreleased
+
+### Adds
+
+* Use `mergeWithCustomize` when merging extended source Webpack configuration. Introduce overideable asset module methods `srcCustomizeArray` and `srcCustomizeObject`, with reasonable default behavior, for fine tuning Webpack config arrays and objects merging. More info - [the Webpack mergeWithCustomize docs](https://github.com/survivejs/webpack-merge#mergewithcustomize-customizearray-customizeobject-configuration--configuration)
+
+### Fixes
+
+* The module `webpack.extensions` configuration is not applied to the core Admin UI build anymore.
```
boutell, line 11: 👍 Please clarify that this was never intended to be supported, as the "apos" build is intentionally separate to avoid ever breaking the expectations of official admin UI modules.

modules/@apostrophecms/asset/index.js
```diff
@@ -941,6 +943,28 @@ module.exports = {
       // Open the implementation for more dev comments.
       transformRebundledFor,
 
+      // Optional functions passed to webpack's mergeWithCustomize, allowing
```
boutell, line 946: We're changing the default behavior of webpack merge here, so document what our new "default" behavior is. I see your comment but it's not clear to me why this behavior is more "reasonable" than the standard default.

boutell (changes requested):

modules/@apostrophecms/asset/index.js
boutell: If it's a question of preventing duplicates then please document who "wins" — later settings become overrides? In what way?

myovchev: There is no change in the default behaviour. 
`['value', 'another', 'value']` becomes `['value', 'another']`. This is applied by default on extensions strings and resolve paths (arrays), no change in the order. The win is correct array values and in the case of the resolve paths, probably better performance. 
The developer may opt-in for change order of merging arrays as described above.

boutell (approved):
Thanks Miro.

PR #3987, opened 2023-01-05
Fix another expanded menu bug, also test issues and an option priority issue

boutell:
* Fix a bug that prevented placeholderUrl from being honored in the presence of placeholderImage
* Fix outdated test
* Fix nasty bug I found this morning that sometimes broke editing a widget added via the expanded menu

### DIFF
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -782,13 +782,15 @@ module.exports = {
       determineBestAssetUrl(name) {
         let urlOption = self.options[`${name}Url`];
         const imageOption = self.options[`${name}Image`];
-        if (imageOption) {
-          const chain = [ ...self.__meta.chain ].reverse();
-          for (const entry of chain) {
-            const path = `${entry.dirname}/public/${name}.${imageOption}`;
-            if (fs.existsSync(path)) {
-              urlOption = `/modules/${entry.name}/${name}.${imageOption}`;
-              break;
+        if (!urlOption) {
+          if (imageOption) {
+            const chain = [ ...self.__meta.chain ].reverse();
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/module/index.js
```diff
@@ -782,13 +782,15 @@ module.exports = {
...
         let urlOption = self.options[`${name}Url`];
         const imageOption = self.options[`${name}Image`];
-        if (imageOption) {
-          const chain = [ ...self.__meta.chain ].reverse();
-          for (const entry of chain) {
-            const path = `${entry.dirname}/public/${name}.${imageOption}`;
-            if (fs.existsSync(path)) {
-              urlOption = `/modules/${entry.name}/${name}.${imageOption}`;
-              break;
+        if (!urlOption) {
```
boutell, line 785: The existing placeholderUrl option for images should still take precedence if explicitly present, otherwise placeholderImage is considered. Caught by unit tests

modules/@apostrophecms/module/index.js
```diff
@@ -782,13 +782,15 @@ module.exports = {
...
+          if (imageOption) {
```
boutell, line 786: Just indented

test/widgets.js
```diff
@@ -409,7 +408,8 @@ describe('Widgets', function() {
                 'placeholder-page': {},
                 [`@apostrophecms/${type}-widget`]: {
                   options: {
-                    placeholderUrl: null
+                    placeholderUrl: null,
+                    placeholderImage: null
```
boutell, line 412: Both need to be off to test the scenario this is looking for

gregvanbrug (approved):

PR #3997, opened 2023-01-12
fix: add typescript definitions

### LINKED ISSUE #3994: Typescript definitions support
## The problem to solve

When developing modules based on the self pattern we often don't see if a function really exists, where it is declared etc.

## Proposed solution

Add typescript definition file at root project and 

## Alternatives

- rewrite apostrophe in typescript or other typed language that transpiles to javascript. This option would cost a lot with few benefits
- use jsdoc. This option is complicated because jsdoc does not handle all the type, however we can combine jsdoc and typescript definitions together.

## Additional context

I already started a POC with few definitions

SimonChaumet:
## Summary

This pull requests adds typescript definitions
Closes #3994

- check typescript type inference and compile

*(Check at least one)*

- [ ] Bug fix
- [ x] New feature
- [ ] Refactor
- [ x] Documentation
- [x ] Build-related changes
- [ ] Other

- [ x] It includes a) the existing issue ID being resolved, b) a convincing reason for adding this feature, or c) a clear description of the bug it resolves

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,38 @@
+declare module "apostrophe" {
+  interface ApostropheModulesOptions {
+    "apostrophe-module": {
+      playerData: Array<string>
+    }
+  }
+
+  interface ApostropheModulesApi {
+    "apostrophe-module": {
+      pushAsset(type: 'script' | 'stylesheet', name: string, options: { when: 'always' | 'lean' | 'user' }): void
+    }
+  }
+
+  type ApostropheModuleSelf<Name extends keyof ApostropheModulesApi> = ApostropheModulesApi[Name]
+  type ApostropheModuleOptions<Name extends keyof ApostropheModulesOptions> = ApostropheModulesOptions[Name]
+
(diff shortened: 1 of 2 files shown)


boutell:
I don't think we would accept a PR to add typings to Apostrophe 2.x because it is late in the Apostrophe 2.x life cycle. New development like this is more appropriate to Apostrophe 3.x.

I'm going to close this initial PR because it is on 2.x but this doesn't mean we're not interested in continuing this conversation about 3.x.

SimonChaumet:
Our project is still in 2.x and this feature will not create breaking changes or developments from your team

boutell:
I see. I appreciate the contribution. I am still inclined to think it is too much effort being invested in a version of Apostrophe that will be supported for some time yet but is not the primary focus. To really make this work well and make it possible to type-check, for instance, overridden methods in a piece type module, it would be necessary to make changes to A2 core which is unlikely at this stage. So I would suggest that migration to A3 would be a much more valuable project to pursue first.

(If you work with or for one of our enterprise clients, please do open support tickets within that project so we understand the urgency for the customer and can have this discussion with all the stakeholders concerned.)

PR #3998, opened 2023-01-12
emit beforeSaveSafe event providing the safe user and the user

ETLaurent:
## Summary

Emit event before inserting a safe user in order to be able to modify it.
Necessary for the enterprise [@apostrophe-pro/doc-global-sharding](https://github.com/apostrophecms/doc-global-sharding) module.

cf. https://github.com/apostrophecms/doc-global-sharding/pull/3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Adds
 
 * Emit a `beforeInsert` event from the `@apostrophecms:attachment` module, with `req` and the `doc` as arguments, in order to give the possibility to override the attachment.
-* Emit a `beforeSaveSafe` event from the `@apostrophecms:user` module, with `req`, `safeUser` and `user` as arguments, in order to give the possibility to override the safe user.
+* Emit a `beforeSaveSafe` event from the `@apostrophecms:user` module, with `req`, `safeUser` and `user` as arguments, in order to give the possibility to override properties of the 'safe' object which contains password hashes and other information too sensitive to be stored in the aposDocs collection.
 
 ## 3.37.0 (2023-01-06)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Adds
 
 * Emit a `beforeInsert` event from the `@apostrophecms:attachment` module, with `req` and the `doc` as arguments, in order to give the possibility to override the attachment.
+* Emit a `beforeSaveSafe` event from the `@apostrophecms:user` module, with `req`, `safeUser` and `user` as arguments, in order to give the possibility to override the safe user.
```
boutell, line 8: Let's explain a little bit more what this is: "give the possibility to override properties of the 'safe' object which contains password hashes and other information too sensitive to be stored in the aposDocs collection."

modules/@apostrophecms/user/index.js
```diff
@@ -292,6 +292,9 @@ module.exports = {
 
         await self.hashPassword(doc, safeUser);
         await self.hashSecrets(doc, safeUser);
+
+        await self.emit('beforeSaveSafe', req, safeUser, doc);
```
boutell, line 296: I argue with myself about this order of arguments but I think you are right because `safeUser` is the thing being saved, and the doc is just additional information. 👍

boutell (changes requested):

modules/@apostrophecms/user/index.js
ETLaurent: Yes at the beginning I had `user, safeUser` then I thought "wait, the document in being saved is safeUser, so user is only additional data".

boutell (approved):

PR #4001, opened 2023-01-16
Pro 3495 i18next locale format

falkodev:
## Summary

Enforce i18next locale format as xx-XX.

**Issue**: declaring a language in lowercase is ok for i18next (for example `en`). But having a locale (language-country) is only correct if the country code is in uppercase - e.g: `en-US` (i18next should still be able to read a locale `en-us` but actually no - it is not A3's fault, as it read the JSON correctly, but when adding resources to i18next, they are just ignored by i18next)

**Solution**: check if the locale contains a country code. If it is the case, format it to `xx-XX` while adding the resources to i18next, and do the same to `locales` configuration in A3 so everything works smoothly.

1. `npm link apostrophe` with this branch on testbed
2. In `modules/@apostrophecms/i18n/index.js`, add this to `options.locales`
```js
     'de-de': {
        label: 'German',
        prefix: '/de-de'
      }
```
3. Copy `node_modules/apostrophe/modules/@apostrophecms/i18n/i18n/de.json` to `de-de.json`
4. Run testbed, login and switch to `de-de`. The UI should be in German

image

**Is it a bug fix or a feature?** 
- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Emit a `beforeInsert` event from the `@apostrophecms:attachment` module, with `req` and the `doc` as arguments, in order to give the possibility to override the attachment.
 * Emit a `beforeSaveSafe` event from the `@apostrophecms:user` module, with `req`, `safeUser` and `user` as arguments, in order to give the possibility to override properties of the 'safe' object which contains password hashes and other information too sensitive to be stored in the aposDocs collection.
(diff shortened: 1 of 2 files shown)


boutell (changes requested):
A test is failing. It does appear locale related.

boutell (approved):

PR #4012, opened 2023-01-20
PRO-3558 download all attachments from any uploadfs backend

boutell:
## Summary

New command line task to download all attachments from any uploadfs backend.

```
mkdir /some/test/dir
node app @apostrophecms/attachment:download-all --to=/some/test/dir
```

If your existing site is using local uploadfs storage (the default) then it will finish really fast, but it's still a valid test because uploadfs abstracts away differences between storage backends.

Note that nonexistent files are skipped without a panic because it is super common for some expected files to be missing, inaccessible or of sizes the developer never bothered to rescale existing uploadfs to. We log the error code instead. We do try to throw an error and stop if we see `ENOSPC` (out of disk space).

You can use `--parallel=3` to download more than one attachment at a time, which makes sense only up to a point, 3 seems like a pretty good number so far today with Azure.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## UNRELEASED
+
+### Adds
+
+* The new `@apostrophecms/attachment:download-all --to=folder` command line task is useful to download all of your attachments from an uploadfs backend other than local storage, especially if you do not have a more powerful "sync" utility for that particular backend.
+
 ## 3.38.0 (2023-01-18)
 
 ### Adds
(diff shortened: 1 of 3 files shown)


boutell:
Someday I intend to expand `uploadfs` with an efficient method to fetch a complete catalogue of what exists, but that day is not today, so I'm still enumerating what Apostrophe *expects* to find. That's OK for now.

myovchev (approved):
One small thing, maybe it would be good to be pointed out (probably in the command help text) that the 'to' path is relative to the command execution process.

matthewhiggins415 (approved):

PR #4013, opened 2023-01-23
Bugfix | Extended menu error after widget deletion

waldemar-p:
## Summary

If I use the extended menu, after deleting a previously used widget, there appears an error upon opening the menu. The menu is not rendered because of it.

I am unsure how deep this issue goes, so I just created a quick fix that simply checks for the existence of the widget before pushing it into group.

*For example:*
1. Create a widget
  1.1. maybe use the extended menu
  1.2 maybe use the widget and delete the creation again
2. Delete the widget (removing it from the app.js should be enough)
3. Open the extended menu
4. Error should appear in the console and the menu should be empty.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
@@ -168,7 +168,8 @@ export default {
       }
 
       for (const item of Object.keys(config.widgets)) {
-        group.widgets.push(apos.modules[`${item}-widget`]);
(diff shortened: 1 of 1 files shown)


waldemar-p:
No wI checked again and it seems I forgot one instance to remove the widget from the configuration. Sorry and thank you.

boutell:
That's OK, I appreciate you digging into the code!

PR #4024, opened 2023-01-30
Pro 3517 view permission

haroun:
## Summary

Add support for `viewPermission`

1. `npm test`
2. Add `permission` or `viewPermission` to a widget schema and look for warnings in the server console
3. Add `permission` or `viewPermission` to a top level piece type schema field, you should see no warnings
4. Add `permission` or `viewPermission` to a sub level piece type schema field, you should see warnings in the server console
5. On the browser, open the `Network` panel and check the API response request when you open the manager modal for the piece having `permission` or `viewPermission` attributes. Try it with multiple credentials and verify that the fields is visible or not present for each case.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/test/pieces.js
+++ b/test/pieces.js
@@ -1918,7 +1918,7 @@ describe('Pieces', function() {
         assert.deepEqual(actual, expected);
       });
 
-      it('should be able to retrieve fields with permission and viewPermission when having appropriate credentials', async function() {
+      it('should be able to retrieve fields with permission and viewPermission when having appropriate credentials on rest API', async function() {
         await createUser('admin')();
         const jar = await loginAs('admin');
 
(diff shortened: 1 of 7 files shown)


test/pieces.js
```diff
@@ -1826,4 +1861,274 @@ describe('Pieces', function() {
...
+          'slug',
+          'visibility',
+          'archived',
+          'stock'
+        ];
+
+        assert.deepEqual(actual, expected);
+      });
+
+      it('should be able to retrieve fields with permission and viewPermission when having appropriate credentials', async function() {
```
boutell, line 1998: `permission` is not a test of view access, only `viewPermission` is.

boutell (changes requested):
Almost there, just need to account for the fact that `field.permission` does not imply no view access via the API. That's why we have a separate `viewPermission` for cases where we want both behaviors.

You were right, though, to check the reverse: if you do fail `viewPermission` for the field it doesn't make sense to let you edit the field.

boutell (approved):

boutell (changes requested):
This is actually approved, but I'm putting a hold on it to merge after this week's release just because it would be a little odd to add `viewPermission` before renaming `permission`.

haroun:
@boutell the target branch for this PR is `feature/advanced-permission-2`, I'm merging it.
When we are ready to do a partial or full release, I'll create a new PR to merge `feature/advanced-permission-2` against `main`.

PR #4038, opened 2023-02-08
Add "whenEmpty" prop to "array" field

falkodev:
## Summary

As an admin I can easily understand what is needed when the user or group permission list is empty

The visualization of the initially empty selections (see first image above) will be handled by introducing a new feature of inline array fields, activated by a new whenEmpty option. whenEmpty can have label and icon subproperties. When this option is configured for an inline array field and the array is currently empty, the configured icon and label are displayed per the design above. Icons themselves should be added to the system the same way widget icons already are (see the documentation and the source code for examples).

## Acceptance criteria

When an inline array is empty, a hint appears as described above.

When an inline array is no longer empty, the hint disappears.

Configure these per the design for the user and group permissions fields. 

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/asset/lib/globalIcons.js
+++ b/modules/@apostrophecms/asset/lib/globalIcons.js
@@ -1,8 +1,6 @@
 // Alphabetical collection of globally available vue-material-design-icons.
 module.exports = {
-  'account-icon': 'Account',
   'account-box-icon': 'AccountBox',
-  'account-group-icon': 'AccountGroup',
   'alert-circle-icon': 'AlertCircle',
   'alpha-x-icon': 'AlphaX',
   'anchor-icon': 'Anchor',
--- a/modules/@apostrophecms/asset/lib/globalIcons.js
+++ b/modules/@apostrophecms/asset/lib/globalIcons.js
@@ -1,6 +1,8 @@
 // Alphabetical collection of globally available vue-material-design-icons.
 module.exports = {
+  'account-icon': 'Account',
   'account-box-icon': 'AccountBox',
+  'account-group-icon': 'AccountGroup',
   'alert-circle-icon': 'AlertCircle',
   'alpha-x-icon': 'AlphaX',
   'anchor-icon': 'Anchor',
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/asset/lib/globalIcons.js
```diff
@@ -1,6 +1,8 @@
 // Alphabetical collection of globally available vue-material-design-icons.
 module.exports = {
+  'account-icon': 'Account',
```
boutell, line 3: ?

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -8,10 +8,27 @@
       <AposMinMaxCount
         :field="field"
         :value="next"
+        v-if="!field.inline"
```
boutell, line 11: Is this another thing to realize the design? Is the min and max information apparent in another way?

modules/@apostrophecms/asset/lib/globalIcons.js
boutell: I guess you are doing this for the convenience of the advanced permission module so it doesn't have to push separate icons. OK.

boutell: Wait, that's not necessary. You can register this using the `icons` cascade in the npm module and it will still work. 

See the submitted-draft module for an example.

You don't have to add the material design icon module to the module to use this as long as it's a name that the core can find in the material design icon module.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
haroun: yes but this is on the wrong PR, I'm moving it

boutell (approved):

PR #4042, opened 2023-02-10
Pro 3522 inline array

falkodev:
Currently inline array fields display labels again for each row. Change this to a more tabular view, per Stu's design:

Note that the table headers only appear once.

This is a good general improvement, so no need for an option flag.

Acceptance criteria

Headings appear on inline array fields only once at the top.

Inline array fields line up appropriately on each row as more are added.

## How to test?

You need https://github.com/apostrophecms/advanced-permission/pull/41

1. Check a piece document permissions and check the display
2. Play with the field options on advanced permissions side
   - `draggable`: `true` | `false`
   - `styles`: `'table'` | `''`
   - `inline`: `true` | `false`

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -108,11 +78,9 @@
 import AposInputMixin from 'Modules/@apostrophecms/schema/mixins/AposInputMixin.js';
 import cuid from 'cuid';
 import { klona } from 'klona';
-import draggable from 'vuedraggable';
```
boutell, line 111: Draggability is still a needed feature we can't remove from inline arrays, although it will be disabled with `draggable: false` for these particular fields because it does not make sense for advanced permissions lists (the order is not significant).

boutell:
@stuartromanek can you chime in on this specific point: does using table markup make sense to you? It makes sense to me, given that this is an extremely tabular approach of what is ultimately editable data, but I wanted to ask someone with better modern-CSS-fu than me.

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -1,92 +1,62 @@
...
-                :icon-only="true"
-                @click="closeInlineItem(item._id)"
-              />
-            </div>
-            <div class="apos-input-array-inline-content-wrapper">
-              <h3
-                class="apos-input-array-inline-label"
-                v-if="!item.open && !alwaysExpand"
-                @click="openInlineItem(item._id)"
+        <table class="apos-table" v-if="items.length">
```
boutell, line 14: An actual HTML table seems inevitable to me, but checking with Stu.

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -1,92 +1,62 @@
...
-              />
-            </div>
-          </div>
-        </draggable>
+                {{ $t(schema.label) }}
+              </th>
+            </tr>
+          </thead>
+          <tbody class="apos-input-array-inline">
+            <tr
```
boutell, line 28: Draggable on this list of tr's would make sense (when not disabled for the field).

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -252,108 +196,67 @@ export default {
...
       schemaInput: {
         data: item
-      },
-      open
+      }
     };
   });
 }
-
-function alwaysExpand(field) {
```
boutell, line 287: Hmm, switching to this tabular approach for all inline arrays does indeed imply that control over whether to "expand" each row is going away. I think that is probably OK, but @stuartromanek might see it differently, in which case we might need to revisit offering inline arrays in different formats.

boutell (changes requested):
I consulted with Stu about this:

* Yes, table markup is fine.
* However, we can't use this new layout and markup all the time. He shared a very convincing screenshot of why, see below. We just can't assume all inline arrays are a good fit for this.
* So this behavior should be triggered by `style: 'table'` being explicitly set on the field. Otherwise we should get the existing behavior.
* In either case, drag and drop should work unless `draggable: false` is set. However for the sake of time, we could leave implementing `draggable` for another ticket, and for now just say that `style: 'table'` does not support dragging and the regular style always does.

boutell (changes requested):
Sorry I missed this before: the option name should be `field.style`, not `field.styles`. Did I type this in the plural when we first discussed it? Again, sorry about that if I did so.

boutell:
OK, it does say `style: 'table'` in the ticket, but I should have flagged it earlier to save you time.

boutell (approved):

PR #4045, opened 2023-02-14
allow selected value as boolean

ETLaurent:
## Summary

A `Future Posts` filter with boolean as choices ("Future" choices has a value set to `true`):

before:
image (1)

after:
image

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposSelect.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposSelect.vue
@@ -55,7 +55,7 @@ export default {
       }
     },
     selected: {
-      type: [ String, Number ],
+      type: [ String, Number, Boolean ],
       default: ''
     },
     disabled: {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Adds
 
 * For devops purposes, the `APOS_BASE_URL` environment variable is now respected as an override of the `baseUrl` option.
+* Allow boolean values in `selected` prop of the `AposSelect.vue` component.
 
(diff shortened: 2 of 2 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposSelect.vue
```diff
@@ -55,7 +55,7 @@ export default {
       }
     },
     selected: {
-      type: [ String, Number ],
+      type: [ String, Number, Boolean ],
```
boutell, line 58: Can you remove this restriction by type entirely? The point here is to accept any JSON-compatible value, even `null` or an object or array.

boutell (changes requested):

modules/@apostrophecms/ui/ui/apos/components/AposSelect.vue
boutell: I would recommend a comment so no one jumps in to add it back.

ETLaurent: We have to let the type verification (eslint rule), which makes it hard to decide what to accept.

We could allow [all the types](https://vuejs.org/guide/components/props.html#prop-validation) but I don't see how select options would contain an array or an object...

If null or undefined is passed, then it will take the default `''` value.

I think that String, Number and Boolean make sense for a selected  value

boutell (approved):

PR #4057, opened 2023-02-20
add readOnlyFields option

ETLaurent:
## Summary

Add `readOnlyFields` option, following the `showFields` logic.

Add `readOnlyFields` to a boolean field `choices` option such as:

```js
    {
      type: 'boolean',
      name: 'boolean',
      label: 'Boolean',
      choices: [
        {
          value: true,
          readOnlyFields: [ 'foo', 'bar' ]
        },
        {
          value: false,
          readOnlyFields: [ 'baz' ]
        }
      ]
    },
```

When saving the document with the boolean field set to `true`, the `foo` and `bar` field should now be read-only (setting it to `false` will set `baz` read-only and set `foo` and `bar` editable again.

Same logic for `select` and `checkboxes` fields:

```js
    {
      type: 'select', // or checkboxes
      name: 'select',
      label: 'Select',
      choices: [
        {
          label: 'First choice'
          value: 'one',
          readOnlyFields: [ 'foo', 'bar' ]
        },
        {
          label: 'Second choice'
          value: 'two',
          readOnlyFields: [ 'baz' ]
        }
      ]
    },
```

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -607,7 +607,7 @@ module.exports = {
           }
         });
       });
-      return !readOnly[name];
+      return readOnly[name];
 
       function setReadOnly(name) {
         readOnly[name] = true;
--- a/lib/modules/apostrophe-schemas/index.js
+++ b/lib/modules/apostrophe-schemas/index.js
@@ -545,19 +545,21 @@ module.exports = {
     // based on showFields options of all fields
 
     self.isVisible = function(schema, object, name) {
+      const transform = value => !value;
       const buildWarningMessage = name => `⚠️ showFields misconfigured, attempts to show/hide ${name} which does not exist`;
-      return self.doesMatchConditionalChoices(schema, object, name, 'showFields', true, buildWarningMessage);
+      return self.doesMatchConditionalChoices(schema, object, name, 'showFields', transform, buildWarningMessage);
     };
 
     // Determine whether the given field is read-only
     // based on readOnlyFields options of all fields
 
     self.isReadOnly = function(schema, object, name) {
+      const transform = value => value;
       const buildWarningMessage = name => `⚠️ readOnlyFields misconfigured, attempts to set ${name} which does not exist as read only`;
-      return self.doesMatchConditionalChoices(schema, object, name, 'readOnlyFields', false, buildWarningMessage);
+      return self.doesMatchConditionalChoices(schema, object, name, 'readOnlyFields', transform, buildWarningMessage);
     };
 
-    self.doesMatchConditionalChoices = function(schema, object, name, actionName, reverse, buildWarningMessage) {
+    self.doesMatchConditionalChoices = function(schema, object, name, actionName, transform, buildWarningMessage) {
       var memo = {};
 
       _.each(schema, function(field) {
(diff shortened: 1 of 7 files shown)


lib/modules/apostrophe-schemas/index.js
```diff
@@ -584,6 +584,49 @@ module.exports = {
       }
     };
 
+    // Determine whether the given field is read-only
+    // based on readOnlyFields options of all fields
+
+    self.isReadOnly = function(schema, object, name) {
```
ETLaurent, line 590: ℹ️ logic replicated from `self.isVisible` above

lib/modules/apostrophe-schemas/public/js/user.js
```diff
@@ -761,6 +761,111 @@ apos.define('apostrophe-schemas', {
       }
     };
 
+    self.enableReadOnlyFields = function(data, name, $field, $el, field) {
```
ETLaurent, line 764: ℹ️ logic replicated and adapted from `self.enableShowFields` above.

lib/modules/apostrophe-schemas/views/macros.html
```diff
@@ -137,6 +137,7 @@
 {%- endmacro -%}
 
 {%- macro checkboxesBody(field) -%}
+  <div class="apos-field-readonly-overlay"></div>
```
ETLaurent, line 140: ℹ️ add a visual indicator that checkboxes are not editable

lib/modules/apostrophe-schemas/index.js
boutell: Can this be refactored so that both methods are short wrappers that pass a few property names to a shared method that implements the logic for both?

lib/modules/apostrophe-schemas/public/js/user.js
boutell: Same question re: making both short wrappers for a shared method that implements the logic when given the appropriate property names.

boutell (changes requested):

lib/modules/apostrophe-schemas/index.js
ETLaurent: This is a little more complicated than having just different options name `showFields` and `readOnlyFields`.
The conditions are different, making not just that easy to refactor.

Still doable, but I didn't want to spend more time on this and over-engineer it, because it's A2.
But if you think that should be refactored nonetheless, I can do it (but it'll impact showFields).

I'll give it a try

ETLaurent: @boutell I have factored out the code in `self.doesMatchConditionalChoices` for this file only first, just to see if it makes sense to you (if we're okay with that, I'll do the same with the frontend file).

To me, I find it a bit messy, to gain 11 lines of code in total, and I'm not happy with the way I handle the reverse condition to cope with the fact that `showFields` and `readOnlyFields` use a different logic (include vs. exclude).

We're also impacting `self.isVisible` implementation so we risk to break `showFields`, that'll need to be fully re-tested.

What do you think, is it worth it?

boutell (commented):
I think it's worth it. Thanks for taking it this far to find out. But I think it'll be more elegant if rather than passing in "reverse", you pass in a "transform" function that takes a value and either reverses it (in the case of isVisible) or returns it as-is (in the case of isReadOnly). Then you won't need the special-case logic you have now for "reverse", instead of an else statement and duplicated code you will just wrap the value in a call to the "transform" function. So you won't need a "condition" variable anymore and the code is almost as short as before.

After that the only increase in line count is caused by your decision not to nest "if" statements as much, which is fine but not a real increase in complexity caused by the refactoring.

boutell (approved):
LGTM!

PR #4058, opened 2023-02-20
Bugfix: locale soft-redirect

waldemar-p:
## Summary

get url with prefix to look after in `historicalUrls` and return the local version of the doc via `doc._url`

*For example:*
1. Create a new locale
2. Go into that locale and create a page
3. Rename that page and save
4. Go to the previous URL of the page and wonder why you are not redirected.

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

- [X] It includes a) the existing issue ID being resolved, b) a convincing reason for adding this feature, or **c) a clear description of the bug it resolves**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+* Fix the soft-redirect so that it works for localized content.
+
+
+
 ## 3.40.1 (2023-02-18)
 
 * No code change. Patch level bump for package update.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## UNRELEASED
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 * Replace `deep-get-set` dependency with `lodash`'s `get` and `set` functions to fix the [Prototype Pollution in deep-get-set](https://github.com/advisories/GHSA-mjjj-6p43-vhhv) vulnerability. There was no actual vulnerability in Apostrophe due to the way the module was actually used, and this was done to address vulnerability scan reports.
-* Fix the soft-redirect so that it works for localized content.
+* Add `parseurl` to packages and fix the soft-redirect so that it works for localized content.
 
 ## 3.40.1 (2023-02-18)
 
--- a/modules/@apostrophecms/soft-redirect/index.js
+++ b/modules/@apostrophecms/soft-redirect/index.js
@@ -21,6 +21,8 @@
 // }
 // ```
 
+const parseurl = require('parseurl');
+
 module.exports = {
   async init(self) {
     self.options.statusCode = self.options.statusCode || 302;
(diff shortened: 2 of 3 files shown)


modules/@apostrophecms/soft-redirect/index.js
```diff
@@ -37,7 +37,8 @@ module.exports = {
     return {
       '@apostrophecms/page:notFound': {
         async notFoundRedirect(req) {
-          const urlPathname = req.url.replace(/\?.*$/, '');
+          const url = req._parsedOriginalUrl?.pathname;
+          const urlPathname = url.replace(/\?.*$/, '');
```
boutell, line 41: This regex is redundant since pathname already excludes the ? part of the URL.

modules/@apostrophecms/soft-redirect/index.js
```diff
@@ -50,9 +51,9 @@ module.exports = {
           if (!(doc && doc._url)) {
             return;
           }
-          if (self.local(doc._url) !== req.url) {
+          if (doc._url !== url) {
```
boutell, line 54: `url` won't contain the protocol and hostname, just the pathname, so we still need `self.local` here to make a valid comparison by eliminating the protocol and hostname from `doc._url` (if you don't see them it's because you don't have `baseUrl` set, but it's just about always set in production).

boutell (changes requested):

modules/@apostrophecms/soft-redirect/index.js
waldemar-p: I wasn't sure. Thanks for the clarification.

modules/@apostrophecms/soft-redirect/index.js
waldemar-p: I missed that. Thanks!

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # Changelog
 
-## UNRELEASED
+## Unreleased
```
boutell, line 3: We usually leave this capitalized but we can handle it later no biggie.

boutell (changes requested):

modules/@apostrophecms/soft-redirect/index.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
     return {
       '@apostrophecms/page:notFound': {
         async notFoundRedirect(req) {
-          const urlPathname = req.url.replace(/\?.*$/, '');
+          const urlPathname = req._parsedOriginalUrl?.pathname;
```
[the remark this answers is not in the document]
waldemar-p: It comes from the `parseurl` that is coming as a dependency with `express`.

I am all ears if there is a more fitting variable or a way to achieve this closer to the apostrophe way.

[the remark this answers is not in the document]
waldemar-p: If it is cleaner I could use the `req.originalUrl` in that way:

`const urlPathname = req.originalUrl.replace(/\?.*$/, ''); `

boutell (changes requested):
Interesting! I did find it when I went digging through the module you mentioned. But that property is a cache and not something that anyone is guaranteeing will stick around as an API.

So the correct thing to do is add `parseurl` as a dependency of `apostrophe`, and use it explicitly, i.e. require it in this file and call `parseurl.original(req).pathname` per the `parseurl` docs. If Express used it already, then it will do its caching thing and use the property you found, but we are not explicitly depending on that existing already.

Would you mind adding that in?

waldemar-p:
Thanks for clarifying and proposing a cleaner method.

Of course, I am glad to edit my code according to your proposal.

boutell (approved):
Thanks!

PR #4059, opened 2023-02-20
[A2] upgrades passport version

ValJed:
[PRO-3654](https://linear.app/apostrophecms/issue/PRO-3654/[a2]-passport-package-vulnerability)

## Summary

Upgrades passport to 0.6 due to security vulnerabilities.

I tested, the login system that uses passport methods, and checked signatures of internally used methods:
* initialize
* session
* use
* authenticate
* serializeUser
* deserializeUser

Is still working on version 0.6. The Local strategy and TOTP one have been successfully tested.
I noticed that `apostrophe-passport` also interacts with passport instance through `self.apos.login.passport`.
Since it uses the same methods than the core and nothing more (authenticate, use), it should work as well (and we can't test every possible passport module).

A warning has been added in the changelog since I didn't test every passport method. Some projects may interact with the passport instance by accessing `self.apos.login.passport`. In this case they should check themselves if everything is still working. That why I bumped the minor version and not the patch one.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-* Upgrades passport to latest version in order to fix security vulnerability. For passport methods that are internally used by Apostrophe everything is still working. For projects that are accessing the passport instance directly through `self.apos.login.passport`, some verifications may be necessary to avoid any comptibility issue. The internally used methods are `authenticate`, `use`, `serializeUser`, `deserializeUser`, `initialize`, `session`.
+* Upgrades passport to the latest version in order to fix security vulnerability. For passport methods that are internally used by Apostrophe everything is still working. For projects that are accessing the passport instance directly through `self.apos.login.passport`, some verifications may be necessary to avoid any comptibility issue. The internally used methods are `authenticate`, `use`, `serializeUser`, `deserializeUser`, `initialize`, `session`.
 
 ## 2.225.0 (2023-02-17)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.226.0
+
+### Fixes
+
+* Upgrades passport to the latest version in order to fix security vulnerability. For passport methods that are internally used by Apostrophe everything is still working. For projects that are accessing the passport instance directly through `self.apos.login.passport`, some verifications may be necessary to avoid any comptibility issue. The internally used methods are `authenticate`, `use`, `serializeUser`, `deserializeUser`, `initialize`, `session`.
```
boutell, line 7: Let's put this under a "Security" heading, not just "Fixes".

However, let's also reword it:

"... latest version in order to ensure session regeneration when logging in or out. This adds additional security to logins by mitigating any risks due to XSS attacks. Apostrophe is already robust against XSS attacks."

I don't think there is a plausible scenario to exploit the issue on a typical Apostrophe site, but more security is good.

boutell (changes requested):

CHANGELOG.md
ValJed: 👍🏼

boutell (approved):

PR #4063, opened 2023-02-21
Pro-3521-unique-inline-array

falkodev:
https://linear.app/apostrophecms/issue/PRO-3521/as-an-admin-i-cannot-add-two-separate-permissions-for-the-same-user-or

In Apostrophe core, implement support for a new unique: true property in inline array subfields.

Only one subfield may be set to unique: true.

If an array item is edited such that its value for this field now duplicates that in another array item, a "Duplicate" error should be reported to the user via the standard error reporting space for that field.

This must also be rejected on the server side. Throwing an invalid error for the array field as a whole is sufficient since the UI will block well-intentioned actors.

relationship fields must be supported. For multi-valued field types like relationship, a unique error occurs if any value (e.g. any related document) is selected in more than one array item.

For purposes of this ticket and for time reasons, only support for string, integer, select and relationship fields is strictly required.

Unit tests must be written to verify the server-side support.


modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -752,6 +752,19 @@ module.exports = (self) => {
...
         throw self.apos.error('max');
       }
+      if (data.length && field.schema && field.schema.length) {
+        const { name: uniqueFieldName, label: uniqueFieldLabel } = field.schema.find(subfield => subfield.unique);
+        if (uniqueFieldName) {
+          const duplicates = data
+            .map(item => Array.isArray(item[uniqueFieldName])
+              ? item[uniqueFieldName][0]._id
+              : item[uniqueFieldName])
+            .filter((item, index, array) => array.indexOf(item) !== index);
```
haroun, line 762: `new Set(duplicates).size !== duplicates.length` will do the trick for you.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -752,6 +752,19 @@ module.exports = (self) => {
...
+          if (duplicates.length) {
```
haroun, line 763: we might need to find a way to flatten this levels of nested if statements

test/schemas.js
```diff
@@ -1330,6 +1330,56 @@ describe('Schemas', function() {
...
+    const result = {};
+    const req = apos.task.getReq();
+    await apos.schema.convert(req, schema, input, result);
+    assert(_.keys(result).length === 1);
+    assert(result.addresses);
+    assert(result.addresses.length === 2);
+    assert(result.addresses[0]._id);
+    assert(result.addresses[1]._id);
+    assert(result.addresses[0].address === '500 test lane');
+    assert(result.addresses[1].address === '602 test ave');
```
haroun, line 1371: Already tested in the previous scenario up to here. We should focus on duplicates only

test/schemas.js
```diff
@@ -1330,6 +1330,56 @@ describe('Schemas', function() {
...
+
+    input.addresses[1] = '500 test lane';
+    await apos.schema.convert(req, schema, input, result);
+    assert.throws(() => {
```
haroun, line 1375: I'm more in favor of the `actual` vs `expected` pattern we use now

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -244,6 +245,64 @@ export default {
...
+              : item[uniqueFieldName])
+            .filter((item, index, array) => array.indexOf(item) !== index);
+
+          if (duplicates.length) {
+            duplicates.forEach(duplicate => {
+              if (uniqueFieldSchema.type === 'relationship') {
+                nextTick(() => {
+                  duplicate.split('|').forEach(item => {
+                    this.$el.querySelectorAll(`[data-id="${item}"]`).forEach(relation => {
+                      relation.style.borderColor = errorColor;
```
haroun, line 266: Why don't we use classes instead of styles?

haroun (changes requested):

modules/@apostrophecms/schema/lib/addFieldTypes.js
falkodev: Nice, done.

modules/@apostrophecms/schema/lib/addFieldTypes.js
falkodev: Open to ideas

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
falkodev: Done

test/schemas.js
falkodev: It is to ensure nothing is broken with the `unique: true` without duplicates

test/schemas.js
falkodev: Is it 

```js
assert.throws(() => {
  throw apos.error(expected);
 }, actual);
```
?

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -244,6 +246,50 @@ export default {
...
+              ? item[uniqueFieldName].map(i => i._id).sort().join('|')
+              : item[uniqueFieldName])
+            .filter((item, index, array) => array.indexOf(item) !== index);
+
+          if (duplicates.length) {
+            duplicates.forEach(duplicate => {
+              if (uniqueFieldSchema.type === 'relationship') {
+                nextTick(() => {
+                  duplicate.split('|').forEach(item => {
+                    this.$el.querySelectorAll(`[data-id="${item}"]`).forEach(relation => {
```
boutell, line 263: This should be happening properly through Vue. We have existing capabilities to indicate the state and errors of an input field via what is provided with v-model etc. Direct DOM manipulation for this feels like a hint that something is wrong. It breaks encapsulation and could damage our ability to override the components with different visual presentation.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -244,6 +246,34 @@ export default {
...
+                  ? item.schemaInput.data[uniqueFieldName] && item.schemaInput.data[uniqueFieldName].forEach(datum => {
+                    item.schemaInput.fieldState[uniqueFieldName].duplicate = duplicate.split('|').find(i => i === datum._id);
+                  })
+                  : item.schemaInput.fieldState[uniqueFieldName].duplicate = item.schemaInput.data[uniqueFieldName] === duplicate;
+              });
+            });
+
+            return {
+              name: 'duplicate',
+              message: `duplicate ${this.$t(uniqueFieldSchema.label) || uniqueFieldName}`
```
boutell, line 272: If you are going to go the road of localizing here, then you need to localize "duplicate" too

boutell: We need to draw the user's attention to the specific array item that contains the duplicate. A new prop on the draggable would work for that in the inline case.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
falkodev: Ok, done localizing the error message. But the error name has to be "duplicate" in lowercase because A3 automatically creates a css class based on it and I use it to prevent every subfield to be marked as in error.

```css

::v-deep .apos-field--array.apos-field--error-duplicate {
    .apos-input {
      border-color: var(--a-base-8);
    }
}
```

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -753,7 +753,11 @@ module.exports = (self) => {
         throw self.apos.error('max');
       }
       if (data.length && field.schema && field.schema.length) {
-        const { name: uniqueFieldName, label: uniqueFieldLabel } = field.schema.find(subfield => subfield.unique);
+        console.log('field.schema', require('util').inspect(field.schema, {
```
boutell, line 756: Remove old logging

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -269,7 +268,7 @@ export default {
 
             return {
               name: 'duplicate',
-              message: `duplicate ${this.$t(uniqueFieldSchema.label) || uniqueFieldName}`
+              message: `${this.$t('apostrophe:duplicateError')} ${this.$t(uniqueFieldSchema.label) || uniqueFieldName}`
```
boutell, line 271: I think you can do `message: { key: 'apostrophe:duplicateError', name: $this.$t(uniqueFieldSchema.label) }` but if that's not a thing then this works OK for now

boutell (approved):

modules/@apostrophecms/schema/lib/addFieldTypes.js
falkodev: Done

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
falkodev: `key` works in `message` but `name` has to be outside to name correctly the error for the css class. I let it like that.

PR #4066, opened 2023-02-28
Pro 3833 UI perfs

ValJed:
[PRO-3833](https://linear.app/apostrophecms/issue/PRO-3833/fix-the-issue-with-very-slow-performance-in-editing-mode-when-editing)

## Summary

Research on UI performances issue. The page refresh that happens to update context of the page doesn't destroy appropriate Vue components and associates events. We basically have a Vue apps leak.

* Run a project with palette for example
* Open it and close it multiple times to refresh the page multiple times
* Check that page edition performances stay stable meaning that vue apps leak has been fixed 

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
@@ -53,6 +53,7 @@ export default {
       history.replaceState(null, '', apos.http.addQueryToUrl(location.href, newQuery));
     }
     return {
+      refreshed: false,
       patchesSinceLoaded: [],
       undone: [],
       patchesSinceSave: [],
@@ -136,6 +137,7 @@ export default {
     }
   },
   async mounted() {
+
     apos.bus.$on('revert-published-to-previous', this.onRevertPublishedToPrevious);
     apos.bus.$on('set-context', this.onSetContext);
     apos.bus.$on('push-context', this.onPushContext);
@@ -262,7 +264,7 @@ export default {
         doc: layer.doc
       });
       // So that areas revert to being editable
-      await this.refresh();
+      /* await this.refresh(); */
     },
     // Accept a hint that a user is actively typing and/or manipulating controls
     // and it would best not to enable a save button or a "...Saved" indication yet
@@ -291,7 +293,7 @@ export default {
         }, 1100);
       }
     },
-    async onPublish(e) {
+    async onPublish() {
       if (!this.canPublish) {
         const submitted = await this.submitDraft(this.context);
         if (submitted) {
@@ -504,8 +506,8 @@ export default {
         await this.refresh();
       }
     },
-    async refresh(options = {}) {
-      let url = window.location.href;
+    async refreshPage(refreshable) {
+
       const qs = {
         ...apos.http.parseQuery(window.location.search),
         aposRefresh: '1',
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -23,9 +24,23 @@ export default function() {
   });
 
   apos.bus.$on('refreshed', function() {
+    /* killObsoleteApps(); */
     createAreaAppsAndRunPlayersIfDone();
   });
 
+  /* function killObsoleteApps() { */
```
boutell, line 31: Good point about cleaning up the apps as well as the components. Given how messy this is, I think you would be better off using:

`this.$parent`

To access the app from the component's refreshed handler, and then you can destroy the app rather than the component, as it should recursively destroy the component for you. Worth a try!

boutell (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -261,8 +261,6 @@ export default {
       await this.setContext({
         doc: layer.doc
       });
```
ValJed, line 263: In this case, the refresh is performed from the `setContext` method when `contextStack` is empty. This is the case in the `onPopContext` method.

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -505,7 +503,18 @@ export default {
...
-      let url = window.location.href;
+      const refreshable = document.querySelector('[data-apos-refreshable]');
+      if (options.scrollcheck) {
+        window.apos.adminBar.scrollPosition = {
+          x: window.scrollX,
+          y: window.scrollY
+        };
+      }
+
+      if (!refreshable) {
```
ValJed, line 514: I just moved the logic here, to avoid to perform an http request when there is no refreshable element. tiny perf improvement.

ValJed: I kept the exact same logic, so if there is not refreshable element, it still call:
```
apos.bus.$emit('refreshed');
this.rememberLastBaseContext();
```
As it was before, not sure if it's usefu in this context though.

modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -93,11 +93,12 @@ export default function() {
     const component = window.apos.area.components.editor;
 
     if (apos.area.activeEditor && (apos.area.activeEditor.id === data._id)) {
-      // Editing a piece causes a refresh of the main content area,
-      // but this may contain the area we originally intended to add
-      // a widget to when we created a piece for that purpose. Preserve
-      // the editing experience by restoring that widget's editor to the DOM
-      // rather than creating a new one.
+    // Editing a piece causes a refresh of the main content area,
```
ValJed, line 96: Only linter change in this file.

boutell (changes requested):
We are very close but I think there is a race condition. Right now both the construction of new area apps and the removal of old ones happens on `refreshed`. This creates a risk that newly constructed area apps will be immediately removed, depending on the order in which things happen. Even if you're not observing this, I still think it's not safe to leave it that way.

So please emit a new `cleanup` event immediately prior to the `refreshed` event, and listen to that one for the cleanup stuff, and we can continue to use `refreshed` for "enhance new things" stuff. This ensures they happen in the right order.

Note that I do mean *immediately* before, because we really do want to refill the content div before these events are emitted. So this is a tiny change.

boutell (approved):

boutell:
(I was wrong about the race condition)

PR #4068, opened 2023-02-28
implement external field conditions

ETLaurent:
## Summary

Handle external conditions in frontend and backend.

In testbed (or other project), add external conditions, mixed with regular ones, as well as in `$or` statements.

In this example, we're adding external conditions to the fields of the article piece, but they can be also added in pages and widgets:

```js
  // in "article" piece
  fields: {
    add: {
      foo: {
        label: 'Foo',
        type: 'boolean'
      },
      foo2: {
        label: 'Foo 2',
        type: 'boolean',
        if: {
          foo: true,
          'a()': 'yes' // will be resolved once
        }
      },
      bar: {
        label: 'Bar',
        type: 'string',
        required: true,
        if: {
          $or: [
            { 'b()': 'no' },
            { 'b()': 'no' }, // will be squashed
            { foo2: true },
            { 'topic:a()': 'no' } // method from topic module
          ],
          'a()': 'yes' // will be resolved once
        }
      }
    }
  },
  methods(self) {
    return {
      async a(req, { docId }) {
        return new Promise(resolve => {
          setTimeout(() => {
            resolve('yes');
          }, 2000);
        });
      },
      async b(req, { docId }) {
        return new Promise(resolve => {
          setTimeout(() => {
            resolve('yes');
          }, 2000);
        });
      }
    };
  }
```

```js
  // in "topic" piece
  methods(self) {
    return {
      async a(req, { docId }) {
        return new Promise(resolve => {
          setTimeout(() => {
            resolve('yes');
          }, 2000);
        });
      }
    };
  }
```

The API calls made to resolve the external conditions should not run more than one for the same condition

In the example above, `a()` and `b()` should be resolved only once, as the former will used the previously stored result from `foo2` (or vice versa), and the latter will be squashed into one key.

When opening the modal of the document with a schema containing fields with external conditions, the calls should be made upfront, and a "busy" overlay should prevent users from interacting with the modal while API calls have not returned the result of the external conditions evaluation.

Once the external conditions have all been evaluated, the fields should be displayed according the results of their external condition(s) if any, _and_ according to the regular conditions.
The regular conditions should still work and have an impact on the display of the fields reactively.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -43,7 +44,94 @@ export default {
...
+              .filter(Boolean);
+
+            results = await Promise.all(promises);
+
+            this.externalConditionsResults = {
+              ...this.externalConditionsResults,
+              ...Object.fromEntries(results)
+            };
+          } catch (error) {
+            console.error(error);
```
boutell, line 97: Going to accept this for now. Should show an apostrophe error notification.

modules/@apostrophecms/schema/index.js
```diff
@@ -520,29 +523,35 @@ module.exports = {
...
+        const errorsList = [];
+
+        for (const error of errors) {
+          const isVisible = await self.isVisible(req, schema, destination, error.path);
+
+          if ((error.name === 'required' || error.name === 'mandatory') && !isVisible) {
             // It is not reasonable to enforce required for
             // fields hidden via conditional fields
-            return false;
+            continue;
```
boutell, line 534: Good catch.

boutell (approved):
Looks good, I noted a minor concern about error handling that should be fixed as a followup. I will test this afternoon.

PR #4069, opened 2023-03-02
Pro 3805 fix rich text

ValJed:
[https://linear.app/apostrophecms/issue/PRO-3805/rich-text-bug](https://linear.app/apostrophecms/issue/PRO-3805/rich-text-bug)

## Summary

Fix rich text issue where it's not possible to align new text nodes.
This is because text align is configured for `heading` and `paragraph` but when we create a new node, by default it's a defaultNode, which is a paragraph too but with another name. 

By the way I upgraded every extension to the last same beta version.

* Go to any site rich text
* create a new node by pressing enter
* write some text and verify you can align it
* pass some text in bullet list then back to default node and it should work also.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
```diff
@@ -22,57 +22,61 @@ const nodeMap = {
...
-            ...this.parent?.(),
-            HTMLAttributes: attrs,
-            levels: def.options.level ? [ def.options.level ] : null
-          };
-        }
-      });
-    }
+  if (def.type === 'heading' || def.type === 'paragraph') {
+    return nodeMap[def.type].extend({
+      name: nodeName,
```
ValJed, line 39: I'm not sure we need to set a special name to this node since it's still a paragraph.
I would personally manage it like a normal paragraph instead of a defaultNode. But I suppose there is a reason here?

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -138,17 +138,17 @@ export default {
...
-      if (!content.length) {
-        // If we don't supply a valid instance of the first style, then
-        // the text align control will not work until the user manually
-        // applies a style or refreshes the page
-        const defaultStyle = this.editorOptions.styles.find(style => style.def);
-
-        const _class = defaultStyle.class ? ` class="${defaultStyle.class}"` : '';
-        return `<${defaultStyle.tag}${_class}></${defaultStyle.tag}>`;
-      } else {
+      if (content.length) {
```
ValJed, line 141: early return

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -195,7 +195,7 @@ export default {
         listItem: false
       }),
       TextAlign.configure({
-        types: [ 'heading', 'paragraph' ]
+        types: [ 'heading', 'paragraph', 'defaultNode' ]
```
ValJed, line 198: The fix is here..

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
```diff
@@ -22,57 +22,53 @@ const nodeMap = {
...
-                  return false;
-                }
-                return {};
-              }
-            }
-          ];
-        }
-      });
-    }
+  if (def.type === 'textStyle') {
```
ValJed, line 50: Logic is unchanged here

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Default.js
```diff
@@ -22,57 +22,53 @@ const nodeMap = {
 };
 
 export default (options) => {
-  const def = options.styles.filter(style => style.def)[0];
+  const [ def ] = options.styles.filter(style => style.def);
+
+  if (!def) {
```
ValJed, line 27: early return

boutell (approved):
Thank you for keeping the changes so minimal here.

This relates to my recent work on list items, which were entirely broken in A3 until I introduced the idea that list items could contain defaultNode and defaultNode could be *whatever tag is used in the first styles option*.

See below:

```
+// Specifically, when the "enter" key is pressed, we don't want
+// tiptap/prosemirror to insert "paragraph" nodes in a rich
+// text widget in which they are not even configured as an option.
+//
+// The simplest way to ensure this is to create a new node type that
+// matches the tag and CSS class of the default style (the first
+// configured style, or the one marked with def: `true`).
+//
+// Matching extensions have been provided to the document and listItem types
+// to prefer defaultNode rather than paragraph.
```

And:

```
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/ListItem
.js
@@ -0,0 +1,6 @@
+import ListItem from '@tiptap/extension-list-item';
+  return ListItem.extend({
+    content: 'defaultNode block*'
+  });
+};
```

The first type mentioned in this configuration is what pressing enter in a list will insert, and it basically has to be hardcoded. Thus the need for a custom node type that in practice has flexibility in what element represents it. See even earlier commits from Stuart as well. This is not something we want to touch unless we have a lot of time to test and reconsider, which is not the case at the moment.

PR #4079, opened 2023-03-07
Pro 3800 checkboxes field combo

ValJed:
[PRO-3900](https://linear.app/apostrophecms/issue/PRO-3900/as-a-user-i-can-multiple-select-from-a-large-number-of-choices-without)

## Summary

Adds a new style for the  `checkboxes` field called `combo`.
When this one is selected it renders a sort of multi select. It looks like this
image

By default, when the checkboxes are in style `combo`, a button `Select All` or `field.all.label` key that is added as first element of select options.
If the field has explicit `all: false` this button is not added. It is also not added if a max is set to the field and is below the amount of choices.

You can link it in any A3 project,
create a new field of type `checkboxes` with a style `combo`
Try with `all: false`, you should have no `Select All`
Try with `all: { label: 'my-locale-key' }`, you should see the right translation in the first item in options list.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
@@ -6,7 +6,14 @@
     :display-options="displayOptions"
   >
     <template #body>
+      <AposCombo
+        v-if="field?.style === 'combo'"
+        :choices="choices"
+        :field="field"
+        :value="value"
+      />
       <AposCheckbox
+        v-else
         :for="getChoiceId(uid, choice.value)"
         v-for="choice in choices"
         :key="choice.value"
@@ -26,7 +33,7 @@ import AposInputChoicesMixin from 'Modules/@apostrophecms/schema/mixins/AposInpu
 export default {
   name: 'AposInputCheckboxes',
   mixins: [ AposInputMixin, AposInputChoicesMixin ],
-  beforeMount: function () {
+  beforeMount () {
     this.value.data = Array.isArray(this.value.data) ? this.value.data : [];
   },
   methods: {
@@ -69,6 +76,11 @@ export default {
       }
 
       return false;
+    },
+    getComponentName() {
+      return this.field?.style === 'combo'
+        ? 'AposCombo'
+        : 'AposCheckbox';
     }
   }
 };
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
@@ -80,7 +80,7 @@ export default {
       return false;
     },
     selectItems(choice) {
-      if (choice.value === 'all') {
+      if (choice.value === '__all') {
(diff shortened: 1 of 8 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
```diff
@@ -6,7 +6,16 @@
     :display-options="displayOptions"
   >
     <template #body>
+      <AposCombo
+        v-if="field?.style === 'combo' && choices.length"
```
ValJed, line 10: I kept it simple here since we have only 2 styles.
Also, AposInputCheckbox displays one AposCheckbox for each choice while we have a single component for AposCombo.

boutell: field will always exist, no need for `?.` here.

modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
```diff
@@ -49,12 +58,12 @@ export default {
 
       if (this.field.min) {
         if ((values != null) && (values.length < this.field.min)) {
-          return 'min';
+          return `min ${this.field.min}`;
```
boutell, line 61: does this break localization?

modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
```diff
@@ -69,6 +78,21 @@ export default {
       }
 
       return false;
+    },
+    selectItems(choice) {
+      if (choice.value === 'all') {
```
boutell, line 83: prepend `__` to that special value so it is unlikely to conflict with a reasonable value for a regular choice.

boutell (changes requested):
Generally very 💯 , one minor concern

modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
ValJed: No but the error message was giving no information about the min value which is weird and unclear for user. Works for all fields.

boutell: Doesn't this show the word "min" or "max" unlocalized though?

modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
ValJed: Nice observation, I fixed that.

boutell (approved):

modules/@apostrophecms/schema/ui/apos/components/AposInputCheckboxes.vue
ValJed: Oh yes actually it does

PR #4083, opened 2023-03-08
do not display errors for hidden fields

ETLaurent:
## Summary

"required" errors are not displayed anymore for fields that are hidden and for which the required rule does not apply.

This also fixes errors that are not correctly thrown back to the browser when a module or a method defined in an external condition key does not exist (a non-required field could have a badly formatted external condition -with an unknown module or method- and we could still save the doc).

Tests being covered in https://github.com/apostrophecms/testbed/pull/139

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
+++ b/modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
@@ -93,7 +93,7 @@ export default {
             };
           } catch (error) {
             await apos.notify(this.$t('apostrophe:errorEvaluatingExternalCondition', { name: field.name }), {
-              type: 'error',
+              type: 'danger',
               icon: 'alert-circle-icon',
               dismiss: true,
               localize: false
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -494,6 +494,11 @@ module.exports = {
             continue;
           }
 
+          const isVisible = await self.isVisible(req, schema, destination, field.name);
+          if (!isVisible) {
+            continue;
+          }
+
           // Fields that are contextual are left alone, not blanked out, if
           // they do not appear at all in the data object.
           if (field.contextual && !_.has(data, field.name)) {
@@ -523,21 +528,7 @@ module.exports = {
           }
         }
 
-        const errorsList = [];
-
-        for (const error of errors) {
-          const isVisible = await self.isVisible(req, schema, destination, error.path);
-
-          if ((error.name === 'required' || error.name === 'mandatory') && !isVisible) {
-            // It is not reasonable to enforce required for
-            // fields hidden via conditional fields
-            continue;
-          }
-
-          errorsList.push(error);
-        }
-
-        if (errorsList.length) {
+        if (errors.length) {
           throw errors;
         }
       },
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -506,38 +506,43 @@ module.exports = {
             try {
               await convert(req, field, data, destination);
             } catch (error) {
-              if ((error.name === 'required' || error.name === 'mandatory')) {
-                // `self.isVisible` will throw if an external condition key contains an unknown module
-                // or method, only for required fields
-                const isVisible = await self.isVisible(req, schema, data, field.name);
-                if (!isVisible) {
-                  // It is not reasonable to enforce required for
-                  // fields hidden via conditional fields
-                  continue;
-                }
-              }
-
               if (Array.isArray(error)) {
                 const invalid = self.apos.error('invalid', {
                   errors: error
                 });
                 invalid.path = field.name;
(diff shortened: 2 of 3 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -505,39 +505,38 @@ module.exports = {
           if (convert) {
             try {
               await convert(req, field, data, destination);
-            } catch (e) {
-              if (Array.isArray(e)) {
+            } catch (error) {
+              if ((error.name === 'required' || error.name === 'mandatory')) {
+                // `self.isVisible` will throw if an external condition key contains an unknown module
+                // or method, only for required fields
+                const isVisible = await self.isVisible(req, schema, data, field.name);
```
ETLaurent, line 512: Using `data` rather than `destination` because the latter won't contain all the fields, until the loop has ended.
We could do as before and check errors _after_ the loop but we'd have the errors displayed for hidden required fields... we're in a chicken and egg situation here.

We could still call the `self.apos.util.error(error + '\n\n' + error.stack);` line conditionally, after the loop, do avoid displaying an error for hidden required field, if using `data` is an issue (that from what I know isn't, but I can miss something).

modules/@apostrophecms/schema/index.js
```diff
@@ -505,19 +505,16 @@ module.exports = {
...
                 const invalid = self.apos.error('invalid', {
-                  errors: e
+                  errors: error
                 });
                 invalid.path = field.name;
                 errors.push(invalid);
               } else {
-                if ((typeof e) !== 'string') {
-                  self.apos.util.error(e + '\n\n' + e.stack);
-                }
```
ETLaurent, line 518: the guilty line

modules/@apostrophecms/schema/index.js
```diff
@@ -547,18 +551,30 @@ module.exports = {
...
+                errors[field.name] = error;
               }
-              conditionalFields[field.name] = result;
             }
           }
+
+          // send the error related to the given field via the `name` param
+          if (errors[name]) {
+            throw errors[name];
+          }
```
ETLaurent, line 576: otherwise, we get the first error found in the fields of the schema, not the error related to the field name given in parameters, and that's very confusing.

now we get the correct error for the field that's required:

image

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -93,7 +93,7 @@ export default {
             };
           } catch (error) {
             await apos.notify(this.$t('apostrophe:errorEvaluatingExternalCondition', { name: field.name }), {
-              type: 'error',
+              type: 'danger',
```
ETLaurent, line 96: we can now use "danger" in e2e tests, let's revert it back to it

cf https://github.com/apostrophecms/cypress-tools/pull/22

modules/@apostrophecms/schema/index.js
```diff
@@ -526,19 +523,26 @@ module.exports = {
...
 
+          if (!Array.isArray(error) && typeof error !== 'string') {
+            self.apos.util.error(error + '\n\n' + error.stack);
+          }
           errorsList.push(error);
         }
 
         if (errorsList.length) {
-          throw errors;
+          throw errorsList;
```
boutell, line 545: Good catch

modules/@apostrophecms/schema/index.js
```diff
@@ -547,18 +551,30 @@ module.exports = {
...
-                change = true;
+              try {
+                const result = await evaluate(field.if, field.name, field.moduleName);
+                const previous = conditionalFields[field.name];
+                if (previous !== result) {
+                  change = true;
+                }
+                conditionalFields[field.name] = result;
+              } catch (error) {
+                errors[field.name] = error;
```
boutell, line 568: Excellent

modules/@apostrophecms/schema/index.js
```diff
@@ -591,6 +607,8 @@ module.exports = {
               let externalConditionResult;
 
               try {
+                console.log('fieldName', fieldName);
```
boutell, line 610: Remove debugging logs

boutell (changes requested):
This is excellent. Some console.logging needs to be removed.

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
ETLaurent: Oh yes, 'danger' makes the icon red with an exclamation mark, whereas 'error' just shows a grey empty icon, as if this type does not exist...

image

image
image

image

modules/@apostrophecms/schema/index.js
ETLaurent: omg yes

boutell (approved):

PR #4085, opened 2023-03-11
Fix for issue 1365. Disallow "type" property name in fields

### LINKED ISSUE #1365: No warning for piece types adding a field called "type"

NuktukDev:
Fix for issue 1365.
Disallow "type" property name in _piece-type_ and _widget-type_ fields.adds.

Example:

```javascript
modules: {
  foo: {
    options: {
      name: 'foo',
      label: 'Foo'
    },
    fields: {
      add: {
        type: {
          label: 'Bar',
          type: 'string'
        }
      }
    }
  }
}
```

## Summary

* Added a check inside `modules/piece-type.js` and `modules/widget-type.js`. If there is a field property name as "type" it will error and exit.
* Closes #1365 

*For example:*

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**
* Unfortunately I could not write a test for this change. Reason being is when an error is thrown, I cannot check using `assert.throws` or `assert.rejects` as the error gets handled in `index.js:313-318`. `self.apos.util.error` only calls `console.error` and so I do not believe I am able to test for that either. A suggestion I have is to make a function that logs an error and rejects a promise, then in tests we could use `assert.rejects`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+### Fixes
+* Now errors and exits when a piece-type or widget-type module has a field object with the property "type"
+
 ### Adds
 
 * The `utilityOperations` feature of piece types now supports additional properties:
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -179,6 +179,10 @@ module.exports = {
     if (!self.options.name) {
       throw new Error('@apostrophecms/pieces require name option');
     }
+    const badFieldName = Object.keys(self.fields).indexOf('type') !== -1;
+    if (badFieldName) {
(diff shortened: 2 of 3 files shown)


boutell (approved):
Looks good. I believe @BoDonkey is working on tests.

BoDonkey (approved):
Merging into a feature branch to facilitate test writing before merging to main.

PR #4090, opened 2023-03-13
Pro 3986 qa visual notes

haroun:
## Summary

UI fixes for inline array displays

1. Check the visuals for inline array display in per document permissions

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
@@ -43,7 +43,7 @@
             >
               {{ $t(schema.label) }}
             </th>
-            <th class="apos-table-cell--hidden" />
+            <th />
           </thead>
           <draggable
             class="apos-input-array-inline"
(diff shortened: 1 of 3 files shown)


boutell (approved):

stuartromanek (changes requested):
- Allow the entire doc permissions field wrappers to take up the full width of the content column. Make a generic, full width class and replace area's specific override found in `AposSchema.vue:322`
- Make the `.apos-input-array-inline-table` visually full width, including it's -35px offset. The table and the 'Add Item' button should be flush
- Can we change the Add Item buttons? Add User, Add Group

Concerning the new relationship look + behavior (Might be out of scope for this PR, LMK)
- When an editor engages with a relationship field, as soon as they type **anything** the dark gray "Suggestions..." bar at the top of the list should disappear. It can reappear if the user removes all characters from the type ahead.
- The 'Can't find what you're looking for...' UI that appears at the bottom of the list should always offer a link to the manager (i think now it its dependent on the Browse button flag). It also should only appear at all in the following cases
-- When the populated list is the 'suggestions' (first opened, user has not begun typing)
-- When the user has begun searching and there are no matches to display
Otherwise the UI should disappear.

Otherwise this is great

stuartromanek (approved):

PR #4118, opened 2023-04-03
fix various issues on conditional fields

ETLaurent:
## Summary

Closes https://github.com/apostrophecms/apostrophe/issues/4049

Fixes:

**Empty choice** (covered by e2e tests https://github.com/apostrophecms/testbed/pull/150)

- Add empty choice for dynamic optional select field type
- Add empty choice for required select fields, if they have no `def` option

**Falsy value selection** (covered by e2e tests https://github.com/apostrophecms/testbed/pull/151)

- Choosing a falsy value should triggers a change, and update conditional display
  - when choosing `false` in a boolean field that had no preset value (ie. field not required)
  - when choosing a falsy value in a select after an empty choice was selected (ie. a `null` choice)

**Default values** (covered by e2e tests https://github.com/apostrophecms/testbed/pull/151)

- Populate new widget instances and existing widgets with default values
  - Set default `null` values in new widgets instances to be consistent with what's is done in `AposDocEditor`
- Populate existing docs with default values
- Add default arrays and objects values to `docFields.data` of the Doc and Widget
- Add default relationship subfields values to `docFields.data` of the Doc and Widget

For a widget, and a doc (piece for instance):

1. add a new widget or doc
- all the fields should be populated with default values.
- the editor `this.docFields.data` should contain the default values, even the ones of the array (inline or not), nested arrays, objects, nested objects, and relationship subfields.

2. add new conditional fields to the schema of the widget or doc (conditional field _and_ the field it's relying on)
- the defaults values for these new fields should be populated in `this.docFields.data`
- the conditional fields should work with those default values

3. add a select field with a conditional field relying on it (with 2 falsy choices such as `null` and `false`)
- the conditional field should be displayed if set to be displayed on `false` or `null`, after changing the select from `false` to `null` or vice versa.
4. instead of a select, use a boolean field with no default value
- selecting `No` from nothing (ie: boolean fields has no default value) should display the conditional fields configured to be shown with `false` on the boolean field

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

- [x] Related tests have been updated -> https://github.com/apostrophecms/testbed/pull/150, https://github.com/apostrophecms/testbed/pull/151

**Other information:**

### DIFF
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -356,13 +356,17 @@ export default {
       await this.loadDoc();
       try {
         if (this.manuallyPublished) {
-          this.published = await apos.http.get(this.getOnePath, {
+          const published = await apos.http.get(this.getOnePath, {
             busy: true,
             qs: {
               archived: 'any',
               aposMode: 'published'
             }
           });
+          this.published = {
+            ...this.getDefault(),
+            ...published
+          };
         }
       } catch (e) {
         if (e.name !== 'notfound') {
@@ -375,7 +379,10 @@ export default {
         }
       }
     } else if (this.copyOf) {
-      const newInstance = klona(this.copyOf);
+      const newInstance = klona({
+        ...this.getDefault(),
+        ...this.copyOf
+      });
       delete newInstance.parked;
       newInstance.title = `Copy of ${this.copyOf.title}`;
       if (this.copyOf.slug.startsWith('/')) {
@@ -440,6 +447,11 @@ export default {
           draft: true
         });
 
+        docData = {
+          ...this.getDefault(),
+          ...docData
+        };
+
         if (docData.archived) {
           this.restoreOnly = true;
         } else {
@@ -464,10 +476,7 @@ export default {
             this.docType = docData.type;
           }
           this.original = klona(docData);
-          this.docFields.data = {
-            ...this.getDefault(),
-            ...docData
-          };
+          this.docFields.data = docData;
           if (this.published) {
             this.changed = detectDocChange(this.schema, this.original, this.published, { differences: true });
           }
@@ -639,7 +648,13 @@ export default {
     },
     async loadNewInstance () {
       this.docReady = false;
-      const newInstance = await this.getNewInstance();
+
+      let newInstance = await this.getNewInstance();
+      newInstance = {
+        ...this.getDefault(),
+        ...newInstance
+      };
+
       this.original = newInstance;
       if (newInstance && newInstance.type !== this.docType) {
         this.docType = newInstance.type;
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -482,8 +482,8 @@ export default {
         if (field.name.startsWith('_')) {
           return;
         }
-        // Do not simply check if `field.def` is truthy, this in not enough
-        // since a def as an empty string must be considered valid:
+        // Using `hasOwn` here, not simply checking if `field.def` is truthy
+        // so that `false`, `null`, `''` or `0` are taken into account:
         const hasDefaultValue = Object.hasOwn(field, 'def');
         doc[field.name] = hasDefaultValue
           ? klona(field.def)
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -464,7 +464,10 @@ export default {
             this.docType = docData.type;
           }
           this.original = klona(docData);
-          this.docFields.data = docData;
+          this.docFields.data = {
+            ...this.getDefault(),
+            ...docData
+          };
           if (this.published) {
             this.changed = detectDocChange(this.schema, this.original, this.published, { differences: true });
           }
@@ -472,6 +475,18 @@ export default {
           this.prepErrors();
         }
       }
+    },
+    getDefault() {
+      const doc = {};
+      this.schema.forEach(field => {
+        // Do not simply check if `field.def` is truthy, this in not enough
+        // since a def as an empty string must be considered valid:
+        const hasDefaultValue = Object.hasOwn(field, 'def');
+        doc[field.name] = hasDefaultValue
+          ? klona(field.def)
+          : null;
+      });
+      return doc;
     },
     async preview() {
       if (!await this.confirmAndCancel()) {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -114,6 +114,7 @@ import AposArchiveMixin from 'Modules/@apostrophecms/ui/mixins/AposArchiveMixin'
 import AposAdvisoryLockMixin from 'Modules/@apostrophecms/ui/mixins/AposAdvisoryLockMixin';
 import AposDocErrorsMixin from 'Modules/@apostrophecms/modal/mixins/AposDocErrorsMixin';
 import { detectDocChange } from 'Modules/@apostrophecms/schema/lib/detectChange';
+import { has } from 'Modules/@apostrophecms/schema/lib/object';
 
 export default {
   name: 'AposDocEditor',
@@ -463,14 +464,24 @@ export default {
           if (docData.type !== this.docType) {
             this.docType = docData.type;
           }
+
           this.original = klona(docData);
-          this.docFields.data = docData;
+          this.docFields.data = this.schema.reduce(addDefaultValues, docData);
+
           if (this.published) {
             this.changed = detectDocChange(this.schema, this.original, this.published, { differences: true });
           }
+
           this.docReady = true;
           this.prepErrors();
         }
+      }
+
+      function addDefaultValues(memo, field) {
+        return {
+          ...memo,
+          ...(!has(memo, field.name) && has(field, 'def') && { [field.name]: field.def })
+        };
       }
     },
     async preview() {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -479,9 +479,9 @@ export default {
     getDefault() {
       const doc = {};
       this.schema.forEach(field => {
-        // Do not simply check if `field.def` is truthy, this in not enough
-        // since a def as an empty string must be considered valid:
-        const hasDefaultValue = Object.prototype.hasOwnProperty.call(field, 'def');
+        // Using `hasOwn` here, not simply checking if `field.def` is truthy
+        // so that `false`, `null`, `''` or `0` are taken into account:
+        const hasDefaultValue = Object.hasOwn(field, 'def');
         doc[field.name] = hasDefaultValue
           ? klona(field.def)
           : null;
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -464,7 +464,10 @@ export default {
             this.docType = docData.type;
           }
           this.original = klona(docData);
-          this.docFields.data = docData;
+          this.docFields.data = {
+            ...this.getDefault(),
+            ...docData
+          };
           if (this.published) {
             this.changed = detectDocChange(this.schema, this.original, this.published, { differences: true });
           }
@@ -472,6 +475,18 @@ export default {
           this.prepErrors();
         }
       }
+    },
+    getDefault() {
+      const doc = {};
+      this.schema.forEach(field => {
+        // Do not simply check if `field.def` is truthy, this in not enough
+        // since a def as an empty string must be considered valid:
+        const hasDefaultValue = Object.prototype.hasOwnProperty.call(field, 'def');
+        doc[field.name] = hasDefaultValue
+          ? klona(field.def)
+          : null;
+      });
+      return doc;
     },
     async preview() {
       if (!await this.confirmAndCancel()) {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -479,6 +479,9 @@ export default {
     getDefault() {
       const doc = {};
       this.schema.forEach(field => {
+        if (field.name.startsWith('_')) {
+          return;
+        }
         // Do not simply check if `field.def` is truthy, this in not enough
         // since a def as an empty string must be considered valid:
         const hasDefaultValue = Object.hasOwn(field, 'def');
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -464,10 +464,7 @@ export default {
             this.docType = docData.type;
           }
           this.original = klona(docData);
-          this.docFields.data = {
-            ...this.getDefault(),
-            ...docData
-          };
+          this.docFields.data = docData;
           if (this.published) {
             this.changed = detectDocChange(this.schema, this.original, this.published, { differences: true });
           }
@@ -475,18 +472,6 @@ export default {
           this.prepErrors();
         }
       }
-    },
-    getDefault() {
-      const doc = {};
-      this.schema.forEach(field => {
-        // Using `hasOwn` here, not simply checking if `field.def` is truthy
-        // so that `false`, `null`, `''` or `0` are taken into account:
-        const hasDefaultValue = Object.hasOwn(field, 'def');
-        doc[field.name] = hasDefaultValue
-          ? klona(field.def)
-          : null;
-      });
-      return doc;
     },
     async preview() {
       if (!await this.confirmAndCancel()) {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -356,17 +356,13 @@ export default {
       await this.loadDoc();
       try {
         if (this.manuallyPublished) {
-          const published = await apos.http.get(this.getOnePath, {
+          this.published = await apos.http.get(this.getOnePath, {
             busy: true,
             qs: {
               archived: 'any',
               aposMode: 'published'
             }
           });
-          this.published = {
-            ...this.getDefault(),
-            ...published
-          };
         }
       } catch (e) {
         if (e.name !== 'notfound') {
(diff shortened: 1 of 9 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputSelect.vue
```diff
@@ -43,24 +43,6 @@ export default {
...
-      });
-    }
-    this.$nextTick(() => {
-      // this has to happen on nextTick to avoid emitting before schemaReady is
-      // set in AposSchema
-      if (this.field.required && (this.next == null) && (this.choices[0] != null)) {
-        this.next = this.choices[0].value;
-      }
-    });
-  },
```
ETLaurent, line 63: moved to `AposInputChoicesMixin.js` in order to wait for async stuff to happen before doing that

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -450,15 +451,25 @@ export default {
           if (docData.type !== this.docType) {
             this.docType = docData.type;
           }
+
           this.original = klona(docData);
-          this.docFields.data = docData;
+          this.docFields.data = this.schema.reduce(addDefaultValues, docData);
```
ETLaurent, line 456: add the default values defined in the schema fields to fix conditional fields not showing up when editing existing docs, relying on new fields that were not saved in the document

ETLaurent:
TODO: fix in widgets

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -251,11 +251,18 @@ export default {
             result = false;
             break;
           }
-          if (Array.isArray(self.getFieldValue(key))) {
-            result = self.getFieldValue(key).includes(val);
+
+          const fieldValue = self.getFieldValue(key);
+          const fieldValueOrDef = fieldValue !== undefined
+            ? fieldValue
+            : self.schema.find(field => field.name === key).def;
```
ETLaurent, line 258: use the default values defined in the schema fields to fix conditional fields not showing up when editing existing docs, in the case where new fields have been added but not (yet) saved to the document

boutell (commented):
Is this happening in all of our schema editors or just some?

My first thought is that schema editors should be initially populated with default values in the first place, so this problem shouldn't exist. And I wonder if that's already true, for some of them.

ETLaurent:
> Is this happening in all of our schema editors or just some?
> 
> My first thought is that in the "new document" or "new array item" or "new widget" or "new object" case, schema editors should be initially populated with default values in the first place, so this problem shouldn't exist. And I wonder if that's already true, for some of them. Is it possible we are just not doing it right in some of those larger contexts, and there is no bug in the editor itself? Or are the values there and visible but not behaving like they are there with respect to this one feature?

@boutell I did it in aa07089d4f2ca0b3907cc818173bc66646cc193d and then reverted it to scope it only to the conditional method, because ArrayEditor, DocEditor, WidgetEditor (even some TipTap components) have different ways of populating `docFields`, and I did not want to risk breaking anything (ie: I wasn't sure that docFields should be populated with the default values), hence my revert to comparing the default values only in the conditional code...

If we want to populate docFields, we should do that in every component that uses `conditionalFields()`. I agree that it is cleaner and more logical to populate `docFields` with default values but the impact is huge

ETLaurent:
Adding those default values in docFields (or using them directly in the `conditionalFields` method) fixes a "rare" issue:

When we add some new fields in a doc schema, but the doc fields have not been saved in the DB. So if we edit a saved doc, the conditional rules for those fields (that do not exist yet in the DB) don't work.

ETLaurent:
@boutell I'll look into populating the doc field with def values beforehand, I'll just finish writing the e2e tests for that, in order to have a safe net

boutell:
@ETLaurent to be clear I think at least some editors already do it. In particular pages/pieces call a special API to get a new document with defaults already populated. And the widget editor has a `getDefault` method that ought to work:

```
    getDefault() {
      const widget = {};
        widget[field.name] = field.def ? klona(field.def) : field.def;
      });
      return widget;
    }
```

So I think it makes sense to question where this behavior is actually being seen and if that particular component (array items maybe? What was the context in the original request?) lacks this feature perhaps. But there could just as easily be something about our conditional fields logic that just fires "too late" to respect the defaults.

ETLaurent:
> @ETLaurent to be clear I think at least some editors already do it. In particular pages/pieces call a special API to get a new document with defaults already populated. And the widget editor has a `getDefault` method that ought to work:
> 
> ```
>     getDefault() {
>       const widget = {};
>       this.schema.forEach(field => {
>         widget[field.name] = field.def ? klona(field.def) : field.def;
>       });
>       return widget;
>     }
> ```
> 
> So I think it makes sense to question where this behavior is actually being seen and if that particular component (array items maybe? What was the context in the original request?) lacks this feature perhaps. But there could just as easily be something about our conditional fields logic that just fires "too late" to respect the defaults.

@boutell conditional system relies on `this.docFields`, not `this.original` (which relies on `getDefault()`).
And docFields does not have default values populated

boutell (changes requested):
I took a look at the source on this.

* `AposDocEditor` already populates the document with defaults before it becomes `docFields`. There's a special route used for it. Are you not seeing that — does it not work?
* `AposWidgetEditor` tries to do something similar, with a simple `getDefaults` method, but it only applies its defaults to its `original` property, and not to `docFields`. That is a bug that can be fixed in `AposWidgetEditor`.
* Please look at the object field, array field (inline and not) and relationship editors to see if they have similar discrepancies.

I believe this problem should be solved before `docFields` is populated, not after.

modules/@apostrophecms/schema/ui/apos/mixins/AposInputChoicesMixin.js
```diff
@@ -29,5 +29,24 @@ export default {
     } else {
       this.choices = this.field.choices;
     }
+
+    if (this.field.type === 'select') {
+      this.prependEmptyChoice();
+    }
```
ETLaurent, line 35: moved here to prepend the empty choice after async results have returned, in case of dynamic choices

modules/@apostrophecms/schema/ui/apos/mixins/AposInputChoicesMixin.js
```diff
@@ -29,5 +29,24 @@ export default {
...
+  },
+
+  methods: {
+    prependEmptyChoice() {
```
ETLaurent, line 39: fixed: add an empty choice, even for required fields, when field does not have a `def` option.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -142,7 +142,15 @@ export default {
...
-    this.original = this.value ? klona(this.value) : this.getDefault();
+    if (this.value) {
+      this.original = klona(this.value);
+      return;
+    }
+
+    const defaults = this.getDefault();
+
+    this.original = defaults;
+    this.docFields.data = defaults;
```
ETLaurent, line 153: add defaults to `this.docFields.data` for a new widget instance

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -177,7 +185,12 @@ export default {
     getDefault() {
       const widget = {};
       this.schema.forEach(field => {
-        widget[field.name] = field.def ? klona(field.def) : field.def;
+        // Do not simply check if `field.def` is truthy, this in not enough
+        // since a def as an empty string must be considered valid:
+        const hasDefaultValue = Object.prototype.hasOwnProperty.call(field, 'def');
+        widget[field.name] = hasDefaultValue
+          ? klona(field.def)
+          : null;
```
ETLaurent, line 193: this is a common annoying bug: we were simply checking if `field.def` was truthy, therefore `false`, `null`, `''`, `0` and every falsy values were not taken into account

ETLaurent:
> I took a look at the source on this.
> 
> * `AposDocEditor` already populates the document with defaults before it becomes `docFields`. There's a special route used for it. Are you not seeing that — does it not work?
> * `AposWidgetEditor` tries to do something similar, with a simple `getDefaults` method, but it only applies its defaults to its `original` property, and not to `docFields`. That is a bug that can be fixed in `AposWidgetEditor`.
> * Please look at the object field, array field (inline and not) and relationship editors to see if they have similar discrepancies.
> 
> I believe this problem should be solved before `docFields` is populated, not after.

@boutell 

- `AposDocEditor` already populates defaults for a new instance, I now the route returns the default values populated by the server.
- `AposWidgetEditor`: I already added the defaults for a new instance (see [comment 1](https://github.com/apostrophecms/apostrophe/pull/4118#discussion_r1163850767) and [comment 2](https://github.com/apostrophecms/apostrophe/pull/4118#discussion_r1163852668))

This only reason why I am looking at the `def` fields option in the `conditionalFields` method from `AposEditorMixin` is the following:

> Use the `def` fields option defined in the schema, directly in the `conditionalFields` method from `AposEditorMixin` to fix conditional fields not showing up when editing existing docs, in the case where new fields have been added to the schema but not (yet) saved to the document.

This is a "rare" case where we add new conditional fields to a piece or widget schema, and then we edit an existing piece or widget that was saved with the "old" schema (before those new fields were added).
Since a new instance won't be created but the values from the existing doc will be used, we have the new conditional fields not working with their default values since their saved data will be `undefined` (ie. not stored yet).
Therefore, populating `docFields` with the default values does not fix the issue in that case (because we're not creating a new instance). Hence the need to do a "last minute" check in the `conditionalFields` method.
That's not a pretty solution but this is the only I'm seeing for that "rare" case where fields values are not yet saved in the documents...

As for object field, array field (inline and not) and relationship editors, I'm checking

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -251,11 +251,14 @@ export default {
...
-            result = self.getFieldValue(key).includes(val);
+
+          const fieldValue = self.getFieldValue(key);
+
+          if (Array.isArray(fieldValue)) {
+            result = fieldValue.includes(val);
             break;
           }
-          if (val !== self.getFieldValue(key)) {
+          if (val !== fieldValue) {
```
ETLaurent, line 261: just some minor optimization

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -186,9 +186,12 @@ export default {
...
       }
       for (const doc of after) {
-        if (this.subfields[doc._id] && !Object.keys(doc._fields || {}).length) {
-          doc._fields = this.subfields[doc._id];
+        if (Object.keys(doc._fields || {}).length) {
+          continue;
         }
+        doc._fields = this.subfields[doc._id]
+          ? this.subfields[doc._id]
+          : this.getDefault();
```
ETLaurent, line 194: add default relationship subfields value

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -451,7 +451,10 @@ export default {
             this.docType = docData.type;
           }
           this.original = klona(docData);
-          this.docFields.data = docData;
+          this.docFields.data = {
+            ...this.getDefault(),
+            ...docData
+          };
```
ETLaurent, line 457: add defaults to `this.docFields.data` for an existing doc

ETLaurent:
There is an issue with changes detected when closing a modal (readonly) viewing of an archive doc, where no actual change has been made.

```js
// modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue

// reverting the following code to `this.docFields.data = docData;` fixes the issue.
// we have to find a way to add the default values without triggering a "content change"
this.docFields.data = {
            ...this.getDefault(),
            ...docData
          };
```

modules/@apostrophecms/schema/ui/apos/components/AposInputRelationship.vue
```diff
@@ -317,6 +320,18 @@ export default {
       if (this.field.editor === 'AposImageRelationshipEditor') {
         return 'apostrophe:editImageAdjustments';
       }
+    },
+    getDefault() {
+      const object = {};
+      this.field.schema.forEach(field => {
+        // Do not simply check if `field.def` is truthy, this in not enough
+        // since a def as an empty string must be considered valid:
+        const hasDefaultValue = Object.prototype.hasOwnProperty.call(field, 'def');
```
boutell, line 329: Use `Object.hasOwn` (not `hasOwnProperty`). It's supported across the board in modern browsers and the syntax is much simpler. Server side we'll get access to it in two weeks when Node 14 dies 😄

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
```diff
@@ -44,7 +44,7 @@ export function detectFieldChange(field, v1, v2) {
   if (isEqual(v1, v2)) {
     return false;
   } else if (!v1 && !v2) {
-    return false;
+    return v1 !== v2;
```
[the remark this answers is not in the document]
boutell: Please explicitly check for that one difference then and "else if" use the old test. This could have many knock-on effects and we want to minimize it.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -177,7 +189,12 @@ export default {
     getDefault() {
       const widget = {};
       this.schema.forEach(field => {
-        widget[field.name] = field.def ? klona(field.def) : field.def;
+        // Do not simply check if `field.def` is truthy, this in not enough
+        // since a def as an empty string must be considered valid:
+        const hasDefaultValue = Object.prototype.hasOwnProperty.call(field, 'def');
```
boutell, line 194: hasOwn, no need for prototype etc.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
boutell: 💯

boutell (changes requested):
This is great, some minor things.

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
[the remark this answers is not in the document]
ETLaurent: Do you mean that?

```js
  } else if (!v1 && !v2) {
    return v1 !== v2;
  } else if (!v1 && !v2) {
    return false;
  }
```

because it will never reach the 2nd `else if` statement if we enter the first one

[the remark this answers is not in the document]
ETLaurent: It's fixing the case when we have `null` and `false` (cf 1st [comment](https://github.com/apostrophecms/apostrophe/pull/4118/files#r1157339181) in this thread), so both of these will be falsy, but they are different.

ETLaurent:
> There is an issue with changes detected when closing a modal (readonly) viewing of an archive doc, where no actual change has been made.
> 
> ```js
> // modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
> 
> // reverting the following code to `this.docFields.data = docData;` fixes the issue.
> // we have to find a way to add the default values without triggering a "content change"
> this.docFields.data = {
>             ...this.getDefault(),
>             ...docData
>           };
> ```

@boutell regarding this ☝️, changes were detected when `docFields.data` contained default values and `original` did not.
So to be consistent, I figured that `published`, `original` and new instances should also be populated with default values.
New instance of a doc that's coming from the server should already contain the default values (which are set to `null` by default), but I still chose to add them in frontend in order to do the same as we do for new widget instances, and to existing doc.

cf. https://github.com/apostrophecms/apostrophe/pull/4118/commits/0b68537426aec17d236fe788678c66c71f71cd0e

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
ETLaurent: I was wondering, having not found its usage in the code, I chose `hasOwnProperty`.
Good thing that we can use `Object.hasOwn` now

ETLaurent:
@boutell this has been a cat and mouse game with the e2e tests... and they proved to be efficient one more time.

Seems like adding default values to docFields for an existing doc is bringing a lot of side effects (such as in detecting whether the doc has changed or not).

So I decided to revert adding default values for existing docs and widgets in the 2 last commits (e9bc084, 04bc8ff).

This **does not revert the main fixes made on new widget instances** and selecting falsy values, but only the ability to have the default values set to fields that do not exist yet in existing docs that were saved before adding these new fields.
For instance if we add new fields (with a conditional display) to the schema of a doc and then we edit an existing doc, `docFields` won't contain the `def` values for those fields which makes the conditional display not working for them.

We could still [go back](https://github.com/apostrophecms/apostrophe/pull/4118/commits/c39a2eae2b3f9a01055d1658a94706973d110186#diff-522dc6f501b45c72fd2e94bb9ec3d53adbc19ccc20b1adf4b878267bdd92f9aeL256) to looking at the default values directly in the `conditionalFields` method to fix that, but this case is rare (adding new conditional fields to a schema and expecting conditional display to work for them in a doc that was saved before adding them) and may not need as much time and energy fixing it.

What do you think?

boutell (changes requested):
If the only reason you're backing out those new defaults for undefined fields is because they cause the document to be considered to be "changed..." I say good! It *is* changed! Saving it *will* introduce these default values which were not there before.

Yes the experience could be better, but that's really on the developer who didn't write a proper migration when they introduced the new field (and they can if they don't like it).

If there are more reasons, let me know.

ETLaurent:
Okay, adding defaults to `docFields.data` only then, not `original` or `published`.

Now I have to figure out why we have issues with those 2 tests, _only when advanced permission is enabled_:

```
should try to submit something with no changes (modal)
should Disable edit of an archived page with diverse content (editor)
```

ETLaurent:
`should Disable edit of an archived page with diverse content (editor)` is fixed ✅ , the `large-page-archive` in tested needed to be updated with 2 missing default fields (scheduled published fields)

Now the `should try to submit something with no changes (modal)` that's failing only with advanced permissions 🚧

ETLaurent:
> `should Disable edit of an archived page with diverse content (editor)` is fixed ✅ , the `large-page-archive` in tested needed to be updated with 2 missing default fields (scheduled published fields)
> 
> Now the `should try to submit something with no changes (modal)` that's failing only with advanced permissions 🚧

@boutell should now be fixed, the [tests](https://github.com/apostrophecms/testbed/pull/151#issuecomment-1513407642) are running again.

I had to exclude the computed properties from being added to `docFields.data` and update the database archive we use on testbed with cypress, in order to add 2 missing fields (scheduled publishing) and create a specific archive when advanced permission is enabled, to add the `userPermissions` and `groupPermissions` default fields.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -473,6 +476,21 @@ export default {
         }
       }
     },
+    getDefault() {
+      const doc = {};
+      this.schema.forEach(field => {
+        if (field.name.startsWith('_')) {
+          return;
+        }
+        // Do not simply check if `field.def` is truthy, this in not enough
```
boutell, line 485: this in -> this is

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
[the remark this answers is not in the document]
boutell: I am concerned that there may be other pairs of falsy values that should not be considered different, but let's see how it goes this way, it's more correct ultimately.

boutell (approved):

boutell (approved):

PR #4122, opened 2023-04-07
Check user permission before displaying batch operations

falkodev:
https://linear.app/apostrophecms/issue/PRO-4128/batch-operations-are-missing-permission

Batch operations are always present in the manager modal.

This is an issue when the current user does not have the required permissions to perform the action (e.g. publish).

Within apostrophe core, we can set a permission to tell the frontend what are the required operation for the item to be visible.

## Acceptance criteria

• I should see Publish batch operation only if I have the permission to publish the related content

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This is useful when the goal is to allow REST API access to "guest" users who ha
 project-specific reasons to fetch access content via REST APIs.
 * `test-lib/utils.js` has new `createUser` and `loginAs` methods for the convenience of
 those writing mocha tests of Apostrophe modules.
-* `batchOperations` permissions: if a `permission` property is added to any entry in the `batchOperations` cascade of a piece-type module, this permission will be checked for every user. See `batchOperations` configuration in `modules/@apostrophecms/piece-type/index.js`. The check function `checkBatchOperationsPermissions` can be extended.
+* `batchOperations` permissions: if a `permission` property is added to any entry in the `batchOperations` cascade of a piece-type module, this permission will be checked for every user. See `batchOperations` configuration in `modules/@apostrophecms/piece-type/index.js`. The check function `checkBatchOperationsPermissions` can be extended. Please note that this permission is checked only to determine whether to offer the operation.
 
 ## 3.43.0 (2023-03-29)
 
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -1059,11 +1059,12 @@ module.exports = {
   extendMethods(self) {
     return {
       getBrowserData(_super, req) {
+        const userBatchOperations = self.batchOperations.filter(batchOperation => self.apos.permission.can(req, 'edit', self.name, batchOperation.action));
```
boutell, line 1062: It would be better to tag the batch operations themselves with an optional `permission` property, which is what would be checked for here, if it is present. Generally it will be `edit`, or nothing (if you can view it you can do it), but that is an extensible pattern.

boutell: Please factor this check to a method as it is likely to be extended often.

boutell (changes requested):

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1053,6 +1053,9 @@ module.exports = {
           await self.apos.doc.db.deleteMany({ _id: { $in: deletes } });
           deletes.splice(0);
         }
+      },
+      checkBatchOperationsPermissions(req, permission = 'edit') {
+        return self.batchOperations.filter(batchOperation => self.apos.permission.can(req, permission, self.name, batchOperation.action));
```
boutell, line 1058: The `permission` to be checked should be coming from `batchOperation.permission`, if present, otherwise that paticular operation should always be allowed. The fourth argument (`mode`) should not be needed here, because the manage view is only accessible in draft mode.

boutell (changes requested):

modules/@apostrophecms/piece-type/index.js
boutell:
```
      can(req, action, docOrType, mode)
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This is useful when the goal is to allow REST API access to "guest" users who ha
 project-specific reasons to fetch access content via REST APIs.
 * `test-lib/utils.js` has new `createUser` and `loginAs` methods for the convenience of
 those writing mocha tests of Apostrophe modules.
+* `batchOperations` permissions: if a `permission` property is added to any entry in the `batchOperations` cascade of a piece-type module, this permission will be checked for every user. See `batchOperations` configuration in `modules/@apostrophecms/piece-type/index.js`. The check function `checkBatchOperationsPermissions` can be extended.
```
boutell, line 14: "Please note that this permission is checked only to determine whether to offer the operation."

boutell (changes requested):

boutell (approved):

PR #4135, opened 2023-04-26
Fix document click issues

boutell:
Address the document click issues by replacing `v-click-outside-element` with a custom implementation that is aware of modal stack relationships.

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/apps/AposModals.js
+++ b/modules/@apostrophecms/modal/ui/apos/apps/AposModals.js
@@ -31,6 +31,38 @@ export default function() {
       },
       getProperties(id) {
         return this.$refs.modals.getProperties(id);
+      },
+      // Returns true if el1 is "on top of" el2 in the
+      // modal stack, as viewed by the user. If el1 is a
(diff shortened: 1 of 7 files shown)


modules/@apostrophecms/modal/ui/apos/apps/AposModals.js
```diff
@@ -31,6 +31,38 @@ export default function() {
...
+      // modal stack, as viewed by the user. If el1 is a
+      // modal that appears later in the stack than el2
+      // (visually stacked on top), this method returns true.
+      // If el2 is `document` and el1 is a modal, this
+      // method returns true. For convenenience, if el1
+      // or el2 is not a modal, it is treated as its DOM
+      // parent modal, or as `document`. If el1 has no
+      // parent modal this method always returns false.
+      onTopOf(el1, el2) {
+        if (!el1.matches('[data-apos-modal]')) {
```
boutell, line 45: This way we keep the ugliness of direct DOM inspection localized to one method, which we call from a directive. Everyone else gets to use the pretty directive 😆

PR #4136, opened 2023-04-27
Fix rich text placeholder behavior when there is no insert menu

boutell:
Behavior with no insert menu configured is now much like the current 3.44.0 behavior. Unfortunately the tests don't pass yet, they will have to be rewritten and it may be difficult because I am now relying on pseudo elements which Cypress isn't great at inspecting.


modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -97,9 +97,6 @@ import TableHeader from '@tiptap/extension-table-header';
 import TableRow from '@tiptap/extension-table-row';
 import Placeholder from '@tiptap/extension-placeholder';
 
-// Cleanly namespace it so we don't conflict with other uses somewhere
```
boutell, line 100: Moved this to namespace completely to the instance

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -586,7 +584,8 @@ function traverseNextNode(node) {
     outline: none;
   }
 
-  .apos-rich-text-editor__editor ::v-deep .ProseMirror:focus  p.is-empty::before {
+  .apos-rich-text-editor__editor ::v-deep .ProseMirror:focus p.apos-is-empty::before,
```
boutell, line 587: Different behavior when there is no insert menu

ETLaurent (approved):

PR #4140, opened 2023-04-28
Allow `@` when a document is a template (doc-template-library module).

falkodev:
Allow `@` when a document is a template (doc-template-library module).

https://linear.app/apostrophecms/issue/PRO-4125/create-the-template-library-module-with-template-query-builder

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="apos-media-manager-display">
-    <div class="apos-media-manager-display__grid" :class="classes">
+    <div class="apos-media-manager-display__grid">
       <AposMediaUploader
         v-if="canEdit"
         :disabled="maxReached"
@@ -116,10 +116,6 @@ export default {
       required: false,
       default: null
     },
-    classes: {
-      type: String,
-      default: null
-    },
     template: {
       type: Boolean,
       default: false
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
```diff
@@ -190,7 +190,9 @@ export default {
       const options = {
         def: ''
       };
-      if (this.field.page && !componentOnly) {
+      if (this.field.aposIsTemplate) {
+        options.allow = '@';
```
boutell, line 194: I am not sure this is sufficient. How will users edit the template in context if it is a page template and the slug does not contain a leading `/`? I would imagine a page template slug will probably start with `/@`. We might need to modify sluggo to allow more than one exception.

boutell (changes requested):

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -162,6 +171,14 @@ export default {
       }
 
     },
+    getCellStyles(item) {
+      if (this.template && item.dimensions) {
```
boutell, line 175: Let's take `this.template` out of it and arrange things so that `item.dimensions` just doesn't exist except when we want this kind of treatment. Then it becomes a legitimate general purpose feature of the core, and not an intrusion of a pro module into the core.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputSlug.vue
boutell: It seems we are not yet using the array support in sluggo to support both `/` and `@`. In the template module we'll want to ensure it is `/@` specifically at the beginning for pages.

boutell (changes requested):

boutell (approved):

PR #4147, opened 2023-04-30
Fields nested following

myovchev:
## Summary

Allow nested fields (array and object schema fields) to follow field values from the parent schema.

In order to be done, the nested fields should use syntax `<fieldToFollow`, which will result following of values of a field named `fieldToFollow` from the parent schema. `<<fieldToFollow` would follow a field from the grand parent, and so on.

Here is an example, that involves an imaginary field type `stringWithSuggestions` following a parent field:
```js
module.exports = {
  options: {
    name: 'household',
    fields: {
      add: {
        petPreference: {
          type: 'select',
          label: 'Pet Preference',
          choices: [
            {
              value: 'cats',
              label: 'Cats'
            },
            {
              value: 'dogs',
              label: 'Dogs'
            }
          ],
          required: true
        },
        pets: {
          type: 'array',
          label: 'Pets',
          fields: {
            add: {
              name: {
                // Custom schema field like "string" but
                // offers the user suggestions based on "preference"
                type: 'stringWithSuggestions',
                label: 'Name',
                following: [ '<petPreference' ]
              }
            },
            inline: true
          }
        }
      }
    }
  } 
};
``` 

Open the Testbed app -> Testing -> Households Piece. Change "Pet Preferences" field and see the chosen value change throughout the nested piece schema.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Allow sub-schema fields (array and object) to follow parent schema fields using the newly introduced `following: '<parentField'` syntax, where the starting `<` indicates the parent level. For example `<parentField` follows a field in the parent level, `<<grandParentField` follows a field in the grand parent level, etc. The change is fully backward compatible with the current syntax for following fields from the same schema level.
+
 ### Changes
 * Various size and spacing adjustments in the expanded Add Content modal UI
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Adds
 
-* Allow sub-schema fields (array and object) to follow parent schema fields using the newly introduced `following: '<parentField'` syntax, where the starting `<` indicates the parent level. For example `<parentField` follows a field in the parent level, `<<grandParentField` follows a field in the grand parent level, etc. The change is fully backward compatible with the current syntax for following fields from the same schema level.
+* Allow sub-schema fields (array and object) to follow parent schema fields using the newly introduced `following: '<parentField'` syntax, where the starting `<` indicates the parent level. For example `<parentField` follows a field in the parent level, `<<grandParentField` follows a field in the grandparent level, etc. The change is fully backward compatible with the current syntax for following fields from the same schema level.
 
 ### Changes
 * Various size and spacing adjustments in the expanded Add Content modal UI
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Allow sub-schema fields (array and object) to follow parent schema fields using the newly introduced `following: '<parentField'` syntax, where the starting `<` indicates the parent level. For example `<parentField` follows a field in the parent level, `<<grandParentField` follows a field in the grand parent level, etc. The change is fully backward compatible with the current syntax for following fields from the same schema level.
```
boutell, line 7: grandparent (no space)

boutell, line 7: grandparent level (no space)

modules/@apostrophecms/schema/index.js
```diff
@@ -1261,6 +1264,28 @@ module.exports = {
...
       },
 
+      // If a field has a following property and a parent,
+      // hoist that property values to the parent,
+      // if they start with `<`.
+      hoistFollowingFieldsToParent(field, parent) {
+        if (!parent || !field.following) {
+          return;
+        }
+        const following = typeof field.following === 'string'
```
boutell, line 1274: Is this a thing? I'm pretty sure we don't support `following` as anything other than an array.

boutell: But if there are counterexamples already in our code then that is how it is.

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
```diff
@@ -342,12 +343,21 @@ export default {
     },
     getFollowingValues(item) {
```
boutell, line 344: * Shouldn't there be code sharing on this between object and array via the mixin?
* Is it tested for both inline and non-inline arrays?

boutell (changes requested):
This is great, very close to what we discussed, a few concerns

modules/@apostrophecms/schema/index.js
myovchev: `const following = Array.isArray(field.following) ? field.following : [ field.following ];` is how it's done prior the change (code from the editor mixin).

modules/@apostrophecms/schema/ui/apos/components/AposInputArray.vue
myovchev: Testbed does test both inline and non-inline arrays in different depts.
I've went for completely new mixin, as it makes sense only for fields having their own schema.

PR #4148, opened 2023-05-01
Pro 4116 fix per doc UI

ValJed:
[PRO-4116](https://linear.app/apostrophecms/issue/PRO-4116/manager-ui-is-broken-for-per-doc-permissions-requires-too-much)

## Summary

Fixes `extendQueries` that wasn't working as expected.
Adds few tests.

* Try to override queries builders and methods in a module that extend another.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


test/queryBuilders.js
```diff
@@ -0,0 +1,219 @@
...
+          extendQueries(self, query) {
+            return {
+              builders: {
+                age: {
+                  def: 'adult',
+                  launder(_super, str) {
+                    return _super() || str === 'senior';
+                  },
+                  finalize(_super) {
+                    _super();
```
boutell, line 78: finalize is async and should be awaited. I know your examples aren't so you get away with it, but I don't want to give anyone the wrong idea about finalize.

boutell: e.g. if someone uses this as a guide to extending an existing finalize function they may get a surprise if they don't await. Always safe to make your subclass version async just in case the super becomes async.

boutell: This is just a concern about the tests. The code is fine.

test/queryBuilders.js
```diff
@@ -0,0 +1,219 @@
...
+          options: {
+            alias: 'person'
+          },
+          extendQueries(self, query) {
+            return {
+              builders: {
+                age: {
+                  def: 'adult',
+                  launder(_super, str) {
+                    return _super() || str === 'senior';
```
boutell, line 75: This seems wrong. launder returns the value, not a boolean. But your or clause here returns a boolean in the second half, and matches on 'senior' rather than 'seniors'. Something is not right.

boutell (changes requested):
Great stuff, some notes on the tests.

test/queryBuilders.js
ValJed: Makes sense, I modify that.

test/queryBuilders.js
ValJed: That's true, my bad.

boutell (dismissed):

boutell (dismissed):

boutell (approved):

PR #4150, opened 2023-05-02
Pro 4071 manager projection

haroun:
## Summary

Deprecate `autocomplete` method in favor of `autocomplete` query builder.

Add manager API projection

1. Open a manager modal
2. Open the network panel and look for calls to `GET /api/v1/%my-module%`
3. In the manager search field, type a word and take note of the response size
4. Edit the module options with the following
    ```
    module.exports = {
      options: {
        managerApiProjection: {
          _id: 1,
          aposLocale: 1,
          aposDocId: 1,
          title: 1,
          type: 1,
          _url: 1
        }
      }
    };
    ```
5. Search for the same word as in step 3 and compare the response size

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


CHANGELOG.md
```diff
@@ -2,10 +2,21 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Add `apos.modules['piece-type']`.`getManagerApiProjection` method to reduce the amount of data returned in the manager
```
boutell, line 7: will contain

modules/@apostrophecms/doc-type/index.js
```diff
@@ -506,7 +506,27 @@ module.exports = {
       //
       // `query.field` will contain the schema field definition for
       // the relationship the user is attempting to match titles from.
+      getRelationshipQueryBuilderChoicesProjection(query) {
+        return {
+          title: 1,
```
boutell, line 511: * Leaving out type is never a good idea.
* Include `_url` just for good measure. It might not seem obviously relevant, but its implementation substitutes all the fields necessary to really know what document you're dealing with, without bringing in anything heavy.
* I just recently added support for displaying thumbnails when image pieces are the choices. Please make sure that still works. Possibly by extending this method for the image piece type so that it includes the attachment. If I'm right that it is using this new code now.

modules/@apostrophecms/page-type/index.js
```diff
@@ -404,6 +409,11 @@ module.exports = {
         to.parked = from.parked;
       },
       getAutocompleteProjection(_super, query) {
+        // TODO Remove in next major version.
+        self.apos.util.warnDevOnce(
+          'deprecate-get-autocomplete-projection',
+          'self.getAutocompleteProjection() is deprecated. Use the autocomplete({ ... }) query builder instead. More info at https://v3.docs.apostrophecms.org/reference/query-builders.html#autocomplete. Use self.getRelationshipQueryBuilderChoicesProjection() instead if it was used for apos.schema.relationshipQueryBuilderChoices().'
```
boutell, line 415: Thanks for patiently deprecating everything we forgot to remove. 😆

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -314,6 +314,10 @@ export default {
       const {
         currentPage, pages, results, choices
       } = await this.request({
+        ...(
+          this.moduleOptions.managerApiProjection &&
+          { project: this.moduleOptions.managerApiProjection }
```
boutell, line 319: Nice.

modules/@apostrophecms/schema/index.js
```diff
@@ -1090,7 +1090,7 @@ module.exports = {
           const idsStorage = field.idsStorage;
           const ids = await query.toDistinct(idsStorage);
           const manager = self.apos.doc.getManager(field.withType);
-          const relationshipQuery = manager.find(query.req, { aposDocId: { $in: ids } }).project(manager.getAutocompleteProjection({ field: field }));
+          const relationshipQuery = manager.find(query.req, { aposDocId: { $in: ids } }).project(manager.getRelationshipQueryBuilderChoicesProjection({ field: field }));
```
boutell, line 1093: I think this is a step too far. Existing projects may be overriding getAutocompleteProjection, because it is used here. So I think we can't stop calling it altogether. We can call the new method, which can be documented as the preferred name, but the new method has to call `getAutocompleteProjection` so that existing overrides don't break. Unfortunately.

boutell (changes requested):

modules/@apostrophecms/any-page-type/index.js
```diff
@@ -26,6 +26,11 @@ module.exports = {
       // titles from. The default behavior is to return the `title` property,
       // but since this is a page we are including the slug as well.
       getAutocompleteTitle(doc, query) {
+        // TODO Remove in next major version.
+        self.apos.util.warnDevOnce(
+          'deprecate-get-autocomplete-title',
+          'self.getAutocompleteTitle() is deprecated. Use the autocomplete({ ... }) query builder instead. More info at https://v3.docs.apostrophecms.org/reference/query-builders.html#autocomplete'
```
boutell, line 32: The query builder takes a string, not an object.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -508,6 +508,26 @@ module.exports = {
       //
       // `query.field` will contain the schema field definition for
       // the relationship the user is attempting to match titles from.
+      getRelationshipQueryBuilderChoicesProjection(query) {
+        const projection = self.getAutocompleteProjection(query);
```
boutell, line 512: Good choice for bc.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -615,6 +640,12 @@ module.exports = {
       //
       // We don't launder the input here, see the 'autocomplete' route.
       async autocomplete(req, query) {
+        // TODO Remove in next major version.
+        self.apos.util.warnDevOnce(
+          'deprecate-autocomplete',
+          'self.autocomplete() is deprecated. Use the autocomplete({ ... }) query builder instead. More info at https://v3.docs.apostrophecms.org/reference/query-builders.html#autocomplete'
```
boutell, line 646: String argument

modules/@apostrophecms/image/index.js
```diff
@@ -462,6 +462,14 @@ module.exports = {
   },
   extendMethods(self) {
     return {
+      getRelationshipQueryBuilderChoicesProjection(_super, query) {
+        const projection = _super(query);
+
+        return {
+          ...projection,
+          thumbnail: 1
```
boutell, line 470: Should be `attachment`

modules/@apostrophecms/piece-type/index.js
```diff
@@ -1064,6 +1078,18 @@ module.exports = {
...
           return true;
         });
+      },
+      getManagerApiProjection(req) {
+        if (!self.options.managerApiProjection) {
+          return null;
+        }
+
+        const projection = { ...self.options.managerApiProjection };
+        self.columns.forEach(({ name }) => {
```
boutell, line 1088: Hmm. Not enough yet. Consider:

```
      'draft:submitted.at': {
        label: 'apostrophe:submitted',
        component: 'AposCellLastEdited'
      },
      'draft:submitted.by': {
        label: 'apostrophe:user',
        component: 'AposCellBasic'
      }
```

`draft:submitted.by` pulls the `by` subproperty of the `submitted` property.

If the name contains a `draft:` or `published:` part, remove that part. It's not related to the projection, so not your problem.

Keep the rest as the projection path and it should work.

You can confirm success by looking at the "submitted draft" manager which uses this feature to show who submitted a draft and when.

Technically a column can have a custom component and display any property of any item. It isn't possible for us to find all of the properties they might pull and include them in the projection. However this is probably rare so far, and I think users who are sophisticated enough to be writing those will be OK with adding to the projection. If we do get complaints we may have to add an option to enable the manager projection feature.

boutell (changes requested):

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Figure.js
gregvanbrug, line 1: Should the commented code here be removed?

gregvanbrug (changes requested):
One small question. Otherwise looks good to me.

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Figure.js
haroun: Good catch, this is some code from another PR. I've removed it. Thank you

gregvanbrug (approved):

PR #4159, opened 2023-05-05
Insert menu ux

stuartromanek:
## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -238,6 +238,11 @@ export default {
       }
     },
     focusParentEvent(event) {
+      if (event.key === 'Escape') {
+        if (document.body === document.activeElement) {
+          this.focusedWidget = null;
+        }
+      }
```
stuartromanek, line 245: esc will unfocus a widget and return Add Content UX

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -486,23 +506,35 @@ export default {
...
+        pos: state.selection.$anchor.pos,
+        parentOffset: state.selection.$anchor.parentOffset
+      };
+
+      if (
+        !this.insertMenu ||
+        !this.insert.length ||
+        this.dismissedPositions.filter(o => {
+          return o.pos === pos.pos && o.parentOffset === pos.parentOffset;
+        }).length ||
```
stuartromanek, line 520: ugly

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -536,6 +568,40 @@ export default {
...
+        return;
+      }
+      const pos = this.editor.view.state.selection.$anchor;
+      this.dismissedPositions.push({
+        pos: pos.pos,
+        parentOffset: pos.parentOffset
+      });
+      this.editor.commands.focus();
+      this.activeInsertMenuComponent = null;
+      // Only insert character keys
```
stuartromanek, line 586: This captures a character key press and appends it to the editor when returning focus, allowing uninterrupted typing when using a `/` character as real text. using `length === 1` to derive 'a normal character' might not be bulletproof ..

boutell (changes requested):
@haroun and I found a reproducible issue:

* Create a rich text widget
* Press /. The menu disappears
* Backspace until the menu disappears (I think you have to do it twice)
* Press /. The menu does not reappear

We saw some other weird stuff, it seems like it's happier giving you a / menu in a new element rather than an old one, but that was less reproducible than the above.

stuartromanek:
@boutell @haroun I fixed the bug mentioned, where if you typed a something with a slash, removed the text, and typed the exact same text (with the slash) the menu would not appear.

But simplifying that came at the expense of some UX where we could hit Esc while the insert menu was open and have the menu dismissed. (i abandoned the original approach when I couldn't figure when a Resolved Position was removed from the document, freeing us to take it out of the Dismissed list)

There doesn't seem to be a way to programmatically dismiss the floating the menu, and i'm running into a bunch of chicken-egg situations trying to sniff context inside the `showFloatingMenu` fn.. i'm sure there is a better way to approach this problem but i'm pretty stuck.

Do you all have any thoughts about how we could re do the `showFloatingMenu` fn so that we can directly tell it to close the menu? Or do you have any other 'clever' tiptap/prosemirror tricks for sussing this out?

boutell:
Yeah, that's a tough one.

Even if we could do that, how could the user reopen it? By backspacing over the / and typing it again, I guess...

Strings are not handled like objects so we can't do anything fun like checking if it's the "same" slash character.

I'm also not sure we could force an "escape got pressed" boolean into the state that `showFloatingMenu` considers.

In my experiments with what we are currently shipping, I can "just keep typing," including regular text or the enter key, with reasonable results (my input is kept and the menu goes away). I can also backspace over the slash and the menu goes away.

But, this is what might work:

* Establish a `this.editor.state.aposFloatingMenuDismissed = false` property in `AposRichTextWidgetEditor` as soon as `this.editor` is available.
* Add a Vue component inside the floating menu that has a `mounted` function that generates an id and sets `this.editor.state.aposFloatingMenuGeneration` to that id.
* When escape is pressed in the floating menu, set `this.editor.state.aposFloatingMenuDismissed` to `this.editor.state.aposFloatingMenuGeneration`.
* Inside `showFloatingMenu`, check whether the two properties are equal. If so, return `false`.
* If they are not equal consider other factors normally.

This way you know if the escape key pertained to the same "generation" of the floating menu. Each "generation" ought to only instantiate the subcomponent with the `mounted` function once, so you should... get away with it. If adding properties to `this.editor.state` is even a thing!

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -26,45 +26,56 @@
...
       v-if="editor"
+      class="apos-rich-text-insert-menu"
+      :tippy-options="{ duration: 100, zIndex: 2000, placement: 'bottom-start' }"
+      :should-show="showFloatingMenu"
+      :editor="editor"
+      role="listbox"
+      tabindex="0"
+      ref="insertMenu"
+      :id="`insert-menu-${value._id}`"
+      :key="insertMenuKey"
```
stuartromanek, line 38: use `key` as a reset mechanism. When a new key is generated this component will reset and reconsider the `should-show` fn. This got tricky to trip when programmatically changing element focus from the editor to the menu and back again

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -472,24 +504,34 @@ export default {
           types: this.tiptapTypes
         }));
     },
-    showFloatingMenu({ state }) {
-      if (!this.insertMenu || !this.insert.length) {
+    showFloatingMenu({
+      state, oldState
+    }) {
+      const hasChanges = JSON.stringify(state?.doc.toJSON()) !== JSON.stringify(oldState?.doc.toJSON());
```
stuartromanek, line 510: allows us to detect whether the content has been changed or the editor is just moving the cursor (or 'doing' other things in the editor that are not typing)

stuartromanek:
Got the desired behavior with a simple keyboard listener and 'dumb' flag that temporarily suppresses the insert menu from appearing. 'temporarily suppressing' the menu seems good enough because we just want to remove it from the screen the moment the editor hits esc, we don't need to track the fact we did it forever because we no longer consider showing the menu at all if the editor content hasn't changed (no navigating to a `/` character and having the menu appear) 😅

boutell (changes requested):
Hmm:

I can't click on any of the choices in the floating menu. Nothing happens and the floating menu stays up. I can activate them through the keyboard though.

I am also seeing an issue where I don't always get the hint on all blank lines, but I can't reproduce it after editing them all and saving and refreshing and stuff, so I suspect it's some kind of legacy thing in the markup that will sort itself but would be nice to clean up at load time if we can find the cause.

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -670,44 +751,75 @@ function traverseNextNode(node) {
...
   }
 
   ::v-deep .ProseMirror {
     > * + * {
       margin-top: 0.75em;
     }
-
-    > :last-child {
-      margin-bottom: 1.75em;
-    }
```
stuartromanek, line 710: @haroun you added this recently.. the uneven spacing in RT box was bothering me, but did this have some other utility I'm unaware of? some certain case it was helpful?

stuartromanek:
image
two changes
- Moved insert item components outside of the insert menu dialog to avoid click intent conflicts between the insert menu and the active component
- insert components are now expected to emit 'done' and 'cancel' events so that we can cleanup the editor with more nuance

boutell (changes requested):
Feels good just one tiny adjustment for bc 🙏

stuartromanek:
re-add support for `close` events, add changelog line

boutell (approved):

PR #4161, opened 2023-05-08
Pro 4076 pattern for string field

ValJed:
[PRO-4076](https://linear.app/apostrophecms/issue/PRO-4076/[assembly-a3]-putting-a-bad-prefix-in-a-locale-makes-everything-crash)

## Summary

Adds regex pattern feature for string fields.
Performs frontend and backend validation regex pattern validation if existing.

You can add a new property `pattern` to string fields.
This field must be an instance of regex so can be written as a regex literal or by calling the constructor function:
```javascript
    prefix: {
    type: 'string',
    label: 'Prefix',
    pattern: /^\//,
    // or
    pattern: new RegExp('^/')
}
```

The frontend validation should block the user from saving changes if the pattern is not respected.
A backend check also has been added.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/schema/index.js
```diff
@@ -1614,8 +1614,23 @@ module.exports = {
         } catch (error) {
           throw self.apos.error('invalid', error.message);
         }
-      }
+      },
+
+      stringifyFieldsPatterns(schema) {
```
ValJed, line 1619: Since we can pass to the frontend a regex type, we simply pass the source (a string) and the frontend will re generate the regex based on this.

modules/@apostrophecms/schema/ui/apos/components/AposInputString.vue
```diff
@@ -137,6 +135,13 @@ export default {
         'password'
       ];
 
+      if (typeof value === 'string' && this.field.pattern) {
+        const regex = new RegExp(this.field.pattern);
+        if (!value.match(regex)) {
+          return 'invalid';
```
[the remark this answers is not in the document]
boutell: I think this is probably good enough, I agree that the `help` property should do the work of communicating expectations.

modules/@apostrophecms/schema/index.js
boutell: Schemas are created and passed around in many places. So I suggest moving this to a `validate` function of `string` fields in `addFieldTypes.js`. They don't have a validate function right now, but other field types do, so you can find examples. This will take care of the recursion for you, simplifying your code and making it compatible with more places where schemas are used.

This does mean constructing the regular expressions when needed server side too at convert time, but this isn't a significant performance issue in the context of async code.

boutell (changes requested):

modules/@apostrophecms/schema/index.js
ValJed: I'll add a validate method to the string field type to verify the pattern is a regex.
But this method had another purpose, I can't pass a regex to the frontend, I need to mutate it in a string to pass it to the client, that is the purpose of this method.
And for the user experience I want this pattern field property to be a regex and not a string because it feels more natural and allow linter to yell if it's not valid.

boutell (changes requested):
To clarify, what I am recommending is that the `validate` function should verify it is a regex, and then convert it to a string, updating the `pattern` property. This addresses your concern about not being able to pass it to the browser without creating a redundant implementation of recursive schema exploration. It's equivalent to what you are doing, with less work.

ValJed:
@boutell That is a way more elegant solution, thank you!

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -105,26 +105,37 @@ module.exports = (self) => {
...
       return !value.length;
     },
-    addQueryBuilder: function (field, query) {
+    validate(field, options, warn, fail) {
+      if (!field.pattern) {
+        return;
+      }
+
+      if (!(field.pattern instanceof RegExp)) {
+        fail('The pattern property must be a RegExp');
```
boutell, line 133: If it is a string, I think we have to accept that because everything else in schemas currently is 100% json-friendly and we will want to be able to build schemas with JSON-based tools in the future. However, you can bounce that string through the RegExp constructor and complain if it throws an exception.

boutell (changes requested):

modules/@apostrophecms/schema/lib/addFieldTypes.js
boutell: It's about supporting both.

boutell (approved):

PR #4183, opened 2023-06-02
Style selected items in doc-template-library manager

falkodev:
https://linear.app/apostrophecms/issue/PRO-4019/as-a-user-i-would-like-to-view-a-library-of-templates-ive-created

Matching https://github.com/apostrophecms/doc-template-library/pull/2


modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -42,18 +45,20 @@
...
-            :src="item.attachment._urls['one-sixth']"
-            :alt="item.description || item.title"
-          >
+          <slot name="placeholder">
+            <div
+              v-if="item.dimensions"
+              class="apos-media-manager-display__placeholder"
+              :style="getPlaceholderStyles(item)"
+            />
+            <!-- TODO: make sure using TITLE is the correct alt tag application here. -->
```
ValJed, line 54: Still todo? Should we remove that if it's old?

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -42,18 +45,20 @@
...
-            :style="getPlaceholderStyles(item)"
-          />
-          <!-- TODO: make sure using TITLE is the correct alt tag application here. -->
-          <img
-            v-else
-            class="apos-media-manager-display__media"
-            :src="item.attachment._urls['one-sixth']"
-            :alt="item.description || item.title"
-          >
+          <slot name="placeholder">
```
ValJed, line 48: Why this new slot?

ValJed (changes requested):

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -31,6 +31,9 @@
             v-model="checkedProxy"
           />
         </div>
+
+        <slot name="actions" />
```
ValJed, line 35: Ok this strategy won't be enough because we will need the actual data of the current template we clicked on in order to open the right editor manager or to redirect to the piece _url if this one exists.

ValJed (changes requested):

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
ValJed: So a good solution would be to pass an array of buttons, each one would have an action name associated with it, and the card component will be able to emit the right event based on that action name, like that the parent component can listen to those events and know on which button the user clicked on.

modules/@apostrophecms/image/ui/apos/components/AposMediaManagerDisplay.vue
```diff
@@ -31,6 +31,9 @@
...
+
         <button
```
ValJed, line 37: Why not putting the placeholder slot around this button? It would allow to have the border on the actual parent of the card, and would be easier to style.

ValJed (changes requested):

PR #4193, opened 2023-06-07
PRO-4318: resolve several bugs when the default locale name changes to a new locale, make the locale rename task robust

boutell:
## Summary

* See changelog.

* Check out the `pro-4318` branch of `a3-assembly-client-demo` and `npm install` (important, you need the right dependencies for this test) and get the dashboard started as you normally would (add admin user etc).
* Create a site with two locales, en (the default) and fr (will need a prefix).
* * Edit the English and French home pages of the site in some visible way, and then click Update.
* Now go back to the dashboard and change the name (not just the label) of the `en` locale to `en-GB` and add a third locale, `de` (will need a prefix).
* Save and you should get a notification that the locale was renamed.
* Go back to the site and you should find that (1) the English home page still has its changes, and (2) you can switch to the German and French home pages via the switcher WITHOUT being prompted to localize them for the first time.

*(Check at least one)*

- [X] Bug fix
- [X] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/global/index.js
```diff
@@ -42,7 +42,10 @@ module.exports = {
     // Intentionally the same
     pluralLabel: 'apostrophe:globalDocLabel',
     searchable: false,
-    singleton: true,
+    singleton: {
```
haroun, line 45: we should migrate advanced permission module too or ideally support the previous format too. we can have customer using `singleton` as a boolean

gregvanbrug (approved):

PR #4199, opened 2023-06-09
task to reattach a page to the tree if the path property becomes invalid somehow

boutell:
Test procedure:

* Create a grandparent page in the page tree.
* Use `node app @apostrophecms/page:reattach slug-of-that-page --new-slug=/something-new` to move it up as the last child of the home page and give it a new slug as well.
* Note that this is what actually happens in the page tree manager.

The new-slug option is currently necessary because manual slug editing is rather busted at the moment (separate issue unrelated to this PR as such).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 3.50.0 (2023-06-09)
+
+### Adds
+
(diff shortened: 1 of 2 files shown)


BoDonkey (approved):
Tested and approved.

PR #4213, opened 2023-06-26
Fix page creation without target

ValJed:
## Summary

Like said the the comment before the page post route. It should be possible the create a page without defining the target ID, it doesn't work currently, this PR fixes it.
It will now take the home page by default.

You should be able to create a page from the page post route without specifying the `_targetId` property.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+* Allows to create page without defining the page target ID, by default it takes the Home page.
+
 ## 3.51.1 (2023-06-23)
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


boutell (changes requested):
This is not necessary because `targetId` can already be set to the special string `_home`. See:

https://v3.docs.apostrophecms.org/reference/api/pages.html#post-api-v1-apostrophecms-page

ValJed:
@boutell Unfortunately there is a bug and it currently doesn't work when you pass the option `_newInstance=true` that's why I created this PR. I'm not sure why the `newInstance` request has been integrated to the POST route since it doesn't write anything in DB but it should work the same. You should be able to get a new instance with the `_targetId='_home'` and to post your page with the same option.

boutell (approved):
I see that your patch makes `getIdCriteria` work properly for `_newInstance`, and you're right that was broken before. As for the new default behavior, I see no harm in it especially given the comment. Please share with Bob so he can update documentation.

PR #4215, opened 2023-06-29
Pro 4020 create from template

haroun:
## Summary

* Hide empty schema groups
* Support for `false` when checking `if`

You can use https://github.com/apostrophecms/doc-template-library/pull/17 or do it yourself by

1. Update the piece or page schema and add an `if` to all fields of a schema groups
2. Check that the schema group is now hidden
3. Try `if` with a matching `false` value

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Adds
 
 * Add `update-doc-fields` event to call `AposDocEditor.updateDocFields` method
-* Allow support for `false` value in conditional `if` checks
+* Allow support for `false` value in schema conditional `if` checks
 * Hide empty schema tabs in `AposDocEditor` when all fields are hidden due to `if` conditions
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 Pages/Docs Manager and Doc Editor modal now have better keyboard accessibility.
 They keep the focus on elements inside modals and give it back to their parent modal when closed.
 * Adds support for a new `if` property in `addContextOperation` in order to show or not a context operation based on the current document properties. 
+* Add `update-doc-fields` event to call `AposDocEditor.updateDocFields` method
+* Add schema field `hidden` property to always hide a field
+* Hide empty schema tabs in `AposDocEditor` when all fields are hidden due to `if` conditions
 
 ### Fixes
 - Fixes a problem in the rich text editor where the slash would not be deleted after item selectin from the insert menu.
@@ -20,9 +23,6 @@ They keep the focus on elements inside modals and give it back to their parent m
 
 ### Adds
 
-* Add `update-doc-fields` event to call `AposDocEditor.updateDocFields` method
-* Add schema field `hidden` property to always hide a field
-* Hide empty schema tabs in `AposDocEditor` when all fields are hidden due to `if` conditions
 * Allows users to double-click a nested widget's breadcrumb entry and open its editor.
 * Adds support for a new `conditions` property in `addContextOperation` and validation of `addContextOperation` configuration.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Adds
+
+* Add `update-doc-fields` event to call `AposDocEditor.updateDocFields` method
+* Allow support for `false` value in conditional `if` checks
+* Hide empty schema tabs in `AposDocEditor` when all fields are hidden due to `if` conditions
+
 ### Fixes
 
 * Allows to create page without defining the page target ID, by default it takes the Home page.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Adds
 
 * Add `update-doc-fields` event to call `AposDocEditor.updateDocFields` method
-* Allow support for `false` value in schema conditional `if` checks
+* Add schema field `hidden` property to always hide a field
 * Hide empty schema tabs in `AposDocEditor` when all fields are hidden due to `if` conditions
 
 ### Fixes
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -404,11 +404,11 @@ export default {
         this.loadNewInstance();
       });
     }
-    apos.bus.$on('update-doc-fields', this.updateDocFields);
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Add `update-doc-fields` event to call `AposDocEditor.updateDocFields` method
+* Allow support for `false` value in schema conditional `if` checks
```
ValJed, line 8: Isn't this one a fix?

modules/@apostrophecms/modal/ui/apos/mixins/AposModalTabsMixin.js
```diff
@@ -46,10 +46,13 @@ export default {
...
       for (const key in this.groups) {
         if (key !== 'utility') {
+          const conditionalFields = this.conditionalFields('other');
+          const fields = this.groups[key].fields;
           tabs.push({
             name: key,
             label: this.groups[key].label,
-            fields: this.groups[key].fields
+            fields,
+            isVisible: fields.filter(field => conditionalFields[field] !== false).length > 0
```
ValJed, line 55: You could use the `some` method here to directly get a boolean, no need to check the length.
More readable I think:
```javascript
isVisible: fields.some(field => conditionalFields[field] !== false)
```

modules/@apostrophecms/modal/ui/apos/mixins/AposModalTabsMixin.js
```diff
@@ -60,14 +63,16 @@ export default {
 
   watch: {
     tabs() {
-      if ((!this.currentTab) || (!this.tabs.find(tab => tab.name === this.currentTab))) {
-        this.currentTab = this.tabs[0] && this.tabs[0].name;
+      if ((!this.currentTab) || (!this.tabs.find(tab => tab.isVisible === true && tab.name === this.currentTab))) {
+        const firstVisibleTab = this.tabs.find(tab => tab.isVisible === true);
+        this.currentTab = firstVisibleTab && firstVisibleTab.name;
```
ValJed, line 68: Slightly modified for readability (indentation + uses some instead of find):
```javascript
    tabs() {
      if (
        (!this.currentTab) ||
        (!this.tabs.some(tab => tab.isVisible === true && tab.name === this.currentTab))
      ) {
        const { name: firstVisibleTabName } = this.tabs.find(tab => tab.isVisible === true) || {};
        this.currentTab = firstVisibleTabName;
      }
    }
```
Since the code inside the if is redundant, it could be moved in a specific method.

modules/@apostrophecms/modal/ui/apos/mixins/AposModalTabsMixin.js
```diff
@@ -60,14 +63,16 @@ export default {
...
       }
     }
   },
 
   mounted() {
-    this.currentTab = this.tabs[0] ? this.tabs[0].name : null;
+    const firstVisibleTab = this.tabs.find(tab => tab.isVisible === true);
+    this.currentTab = (firstVisibleTab && firstVisibleTab.name) || null;
```
ValJed, line 75: This code is redudant, you could move it in a method I guess, like `this.setCurrentTab()`?

ValJed (changes requested):
some code style feedback, functional looks ok.

CHANGELOG.md
haroun: Has been replaced with `hidden`

boutell (changes requested):
The stuff relating to `hidden` and hiding tabs correctly is 💯 

But, the new event is problematic because `AposDocEditor` can be nested:

* Open manager
* Start editing a doc
* Say "I want to add a related article to this doc"
* Say "hmm the article I am connecting this to needs a lil' edit first"
* Make that edit — at this point two AposDocEditors exist
* The new `update-doc-fields` event arrives and both editors think it is for them

So, this won't quite work as written. I'm not quite sure how to help because I can't see the code that is trying to send the event (from another module maybe?). But the solution can't just be to check the type, because you could encounter this scenario while trying to relate an article to another article.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -665,6 +667,10 @@ export default {
         itemName: `${this.moduleName}:editor`
       });
     },
+    onUpdateDocFields(value) {
```
haroun, line 670: check if `value.data._id` is equal to `this.docFields.data._id` to prevent multiple editor modal

boutell (approved):

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
ValJed: I removed the need for the event bus here, see my comment in code.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -67,11 +67,12 @@
...
               :doc-id="docId"
               :value="docFields"
-              @input="updateDocFields"
-              @validate="triggerValidate"
               :server-errors="serverErrors"
               :ref="tab.name"
               :generation="generation"
+              @input="updateDocFields"
+              @validate="triggerValidate"
+              @update-doc-data="onUpdateDocFields"
```
ValJed, line 75: This event is sent from `AposTemplateList` from Doc Template Library, and not anymore from the event bus.
Check this [comment](https://github.com/apostrophecms/doc-template-library/pull/22/files#r1268191545) in Doc Template Library

boutell (changes requested):
This is really clever, but it will fail when the piece type has configured an alternative doc editor component name because the recursive search for `AposDocEditor` won't work.

Instead this event should be listened to by `AposSchema` (the parent of the field) and emitted again to its parent (which will be the document editor).

(It's true that we're making other changes that may require customization in AposDocEditor substitutes, but we've always documented that this is something they will have to deal with from time to time. What we can't do is cut them out completely so they have no hope of implementing a feature.)

ValJed:
@boutell good point, I made the modification here to listen the event in `AposSchema` and made the update on doc template library to simply emit the event to the parent. Still working well. 👍🏼

boutell (dismissed):
Approving and merging. Please follow up with a changelog entry.

boutell:
Val you have to approve too it seems 😆

boutell (approved):

PR #4216, opened 2023-06-30
Permission support context operations

ValJed:
[PRO-4026](https://linear.app/apostrophecms/issue/PRO-4026/as-an-admin-i-want-to-decide-who-has-permission-to-create-delete)

## Summary

To finish the work of the ticket above we need to allow `permission` property to the method `addContextOperation`.
This method can be a string or an array.
The available values are the ones that have a can method associated in `AposDocContextMenu`.
A validate function has been added to the method `addContextOperation` to save precious time to the user in case of bad configuration.

What it does is that it checks the permission value, if I set it to `edit` it will run the method `canEdit` in the vue component and shows the operation button only if this one returns `true`. 
Since we supports arrays too, we can check multiple permissions here. eg. if I set permission too `['archive', 'restore']` it will check both computed: `canArchive` and `canRestore`.

I might have over-engineered this feature depending on the need but it feels like it offers to developers a complete way to handle permissions in `addContextOperation`. 

Configure `addContextOperation`  with the property `permission`.
This one should work with a string or an array.
An error should be thrown if the permission value is not valid.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Adds
 
-* Adds support for `permission` property in `addContextOperation`, this one can be a string or an array. Validates `addContextOperation` configuration.
+* Adds support for `conditions` property in `addContextOperation`, this one must be an array. Validates `addContextOperation` configuration.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Adds support for `permission` property in `addContextOperation`, this one can be a string or an array. Validates `addContextOperation` configuration.
+
 ### Fixes
 
 * Allows to create page without defining the page target ID, by default it takes the Home page.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Adds
+
+* Adds support for `permission` property in `addContextOperation`, this one can be a string or an array. Validates `addContextOperation` configuration.
```
boutell, line 7: This is great, but a lot of these conditions are not really Apostrophe permission names. So right now it looks too much like something it is not.

I think we should just call this property `conditions`, require it to be an array, and make the values in the array the actual property names: `canPublish`, `canDismissSubmission`, etc.

This will avoid misleading the developer into thinking they can test any Apostrophe permission, and eliminate the need for string manipulation.

boutell (changes requested):

CHANGELOG.md
ValJed: Ok yes makes sense to me 👍🏼

ValJed:
@boutell Code has been simplified here:
* Property is called `conditions` and must be an array.
* Allowed values are the proper action names (`canEdit`, `canLocalize` ...) to avoid string manimulation.
* Validation still prevent user from badly configuring `addContextOperation`.

falkodev (approved):

boutell (approved):

boutell (approved):

PR #4217, opened 2023-06-30
double click breadcrumb to edit, ui cleanup

stuartromanek:
## Summary

 - Adds double-click handler to widget breadcrumb to open their editor
 - Removes inverted theme style from foreign widgets
 - Hides area controls completely for foreign widgets (we were already skipping every widget control for foreign, so this just removes the ui fragment)
 - Fixes bug where a widget's breadcrumb was un-hoverable if the widget wasn't explicitly focused

1. link to a3-demo and `npm run dev`
2. find a local widget that isn't `contextual`
3. double click the breadcrumb to edit 

*(Check at least one)*

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
@@ -41,13 +41,13 @@
           <li class="apos-area-widget__breadcrumb" data-apos-widget-breadcrumb="0">
             <AposButton
               type="quiet"
-              @click="foreign ? $emit('edit', i) : false"
+              @click="foreign ? $emit('edit', i) : null"
+              @dblclick.native="(!foreign && !isContextual) ? $emit('edit', i) : null"
               :label="foreign ? {
                 key: 'apostrophe:editWidgetType',
                 label: $t(widgetLabel)
               } : widgetLabel"
               tooltip="apostrophe:editWidgetForeignTooltip"
-              :icon="foreign ? 'earth-icon' : null"
               :icon-size="11"
               :modifiers="['no-motion']"
             />
@@ -76,6 +76,7 @@
         :class="ui.controls"
       >
         <AposWidgetControls
+          v-if="!foreign"
           :first="i === 0"
           :last="i === next.length - 1"
           :options="{ contextual: isContextual }"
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -41,13 +41,13 @@
           <li class="apos-area-widget__breadcrumb" data-apos-widget-breadcrumb="0">
             <AposButton
               type="quiet"
-              @click="foreign ? $emit('edit', i) : false"
+              @click="foreign ? $emit('edit', i) : null"
```
ValJed, line 44: Can be written:
```javascript
foreign && $emit('edit', i)
```

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -41,13 +41,13 @@
...
+              @dblclick.native="(!foreign && !isContextual) ? $emit('edit', i) : null"
```
ValJed, line 45: There is a lint error here based on this [breaking change](https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed.html#v-on-native-modifier-removed).
Not sure how to replace it though but it would be better (since we will need to migrate to Vue 3 soon). If you can't I guess we need to modify the eslint config.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -41,13 +41,13 @@
...
               :label="foreign ? {
                 key: 'apostrophe:editWidgetType',
                 label: $t(widgetLabel)
               } : widgetLabel"
               tooltip="apostrophe:editWidgetForeignTooltip"
```
ValJed, line 50: We could remove this tooltip when contextual since we can't edit from de modal:
```javascript
:tooltip="!isContextual && 'apostrophe:editWidgetForeignTooltip'"
```

ValJed (changes requested):
Mainly a linter error to fix, few feedback. Working well otherwise.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
boutell: We should override this Vue 3 - specific rule in our linter config, rather than making Stu change correct Vue 2 code. This is the only way to do this in Vue 2.

We may need to fork the Vue linter rules module and maintain a Vue 2 - compatible set ourselves (although in practice, there's little reason to change them, so the "maintenance" would not be difficult).

ValJed (approved):
Seems working. 
I find the fact that we can only click on the tiny text inside the breadcrumb instead of the whole breadcrumb less comfortable.

Also you resolved this but I still find
```javascript
@click="foreign && $emit('edit', i)"
```
Cleaner than
```javascript
@click="foreign ? $emit('edit', i) : null"
```
🙃

boutell (approved):

PR #4221, opened 2023-07-13
Add if feature to context operations

ValJed:
## Summary

* Adds support for a new `if` property in `addContextOperation` in order to show or not a context operation based on the current document properties.
* Needed for doc template library in order to hide some context operations when `aposIsTemplate` is true for  document (like `Save As Template` or `Share Draft`).

This new feature allows configuration looking like this:
```javascript        
self.apos.doc.addContextOperation(self.__meta.name, {
          action: 'saveAsTemplate',
          context: 'update',
          label: 'aposTemplate:saveAsTemplate',
          modal: 'AposTemplateLibraryEditor',
          conditions: [ 'canPublish' ],
          if: {
            aposIsTemplate: { $ne: true },
            $not: [
              { archived: true }
            ]
          }
        });
```
It performs a `AND` operator for props that are in the same object, and `OR` between objects that are in a `$or` array.
`$ne` means `not equal`. It all works for primitives values only.

Check that both context operations described above are hidden for templates (whatever their type).
Must be tested with [this PR](https://github.com/apostrophecms/doc-template-library/pull/20).

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
+++ b/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
@@ -69,7 +69,8 @@ module.exports = ({
       alias: {
         vue$: 'vue/dist/vue.runtime.esm.js',
         // resolve apostrophe modules
-        Modules: path.resolve(modulesDir)
+        Modules: path.resolve(modulesDir),
+        Root: path.resolve(__dirname, '../../../../../..')
       },
       modules: [
         'node_modules',
--- a/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
+++ b/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
@@ -69,8 +69,7 @@ module.exports = ({
       alias: {
         vue$: 'vue/dist/vue.runtime.esm.js',
         // resolve apostrophe modules
-        Modules: path.resolve(modulesDir),
-        Root: path.resolve(__dirname, '../../../../../..')
+        Modules: path.resolve(modulesDir)
       },
       modules: [
         'node_modules',
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -421,7 +421,10 @@ module.exports = {
           label: 'apostrophe:shareDraft',
           modal: 'AposModalShareDraft',
           manuallyPublished: true,
-          hasUrl: true
+          hasUrl: true,
+          if: {
+            aposIsTemplate: { $ne: true }
+          }
         });
       },
       getRelatedDocsIds(req, doc) {
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -421,10 +421,7 @@ module.exports = {
           label: 'apostrophe:shareDraft',
           modal: 'AposModalShareDraft',
           manuallyPublished: true,
-          hasUrl: true,
-          if: {
-            aposIsTemplate: { $ne: true }
-          }
+          hasUrl: true
         });
       },
       getRelatedDocsIds(req, doc) {
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
@@ -23,7 +23,7 @@ import { detectDocChange } from 'Modules/@apostrophecms/schema/lib/detectChange'
 import AposPublishMixin from 'Modules/@apostrophecms/ui/mixins/AposPublishMixin';
 import AposArchiveMixin from 'Modules/@apostrophecms/ui/mixins/AposArchiveMixin';
 import AposModifiedMixin from 'Modules/@apostrophecms/ui/mixins/AposModifiedMixin';
-import checkIfConditions from '../../../../../../lib/check-if-conditions';
+import checkIfConditions from 'Root/lib/check-if-conditions';
 
 export default {
   name: 'AposDocContextMenu',
--- a/modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
+++ b/modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
@@ -196,7 +196,7 @@ export default {
     },
     customOperationsByContext() {
       return this.customOperations.filter(({
-        manuallyPublished, hasUrl, conditions, context, if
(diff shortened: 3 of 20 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -421,7 +421,10 @@ module.exports = {
           label: 'apostrophe:shareDraft',
           modal: 'AposModalShareDraft',
           manuallyPublished: true,
-          hasUrl: true
+          hasUrl: true,
+          if: {
+            aposIsTemplate: { $ne: true }
```
ETLaurent, line 426: could this `if` be added via an override of `addContextMenu` that resides in the template library module?

I see that there are other references to templates in the core, maybe we should move them to the module, if not too complicated?

image

modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
```diff
@@ -443,10 +451,40 @@ export default {
     },
     close() {
       this.$emit('close', this.doc);
+    },
+    checkIfConditions(doc, conditions) {
```
boutell, line 455: Can we factor this to an `import`-ed file so it is shared with the other implementation?

boutell (changes requested):
@ETLaurent makes a good point. Are we doing what we can to avoid `aposIsTemplate` in core?

In this particular case, I think Etienne is definitely right. I like the idea of the doc-template module manipulating apos.doc.contextOperations to avoid hardcoding this condition in core. It will work, and doesn't require hard overrides of any components, so we should do it. That's one less appearance of `aposIsTemplate` in core.

```
      const options = {
        def: field.def
      };
      if (field.page) {
        options.allow = '/';
      }
      if (data.aposIsTemplate) {
        options.allow = field.page ? [ '/', '@' ] : '@';
      }
```

We could factor this code out to a `getSlugifyOptions(req, field)` method.

But, actually it is always the same for the field. It does not depend on `req`.

So instead we should move it into the `validate()` method of the slug field, and set a `field.allow` property there, at startup, based on a call to a new `slugFieldAllows(field)` method that the doc template module can easily `improve`. Then we can access it on the front end too:

**AposInputSlug**

```
      if (this.field.aposIsTemplate) {
        options.allow = this.field.page ? [ '/', '@' ] : '@';
      } else if (this.field.page && !componentOnly) {
        options.allow = '/';
      }
```

 There's a bug here I think. We talked about the bug yesterday, perhaps it is already fixed. When the page is a template we are ignoring the `!componentOnly` check.

If `componentOnly` is true we should override `options.allow` to `undefined` and ignore `field.allow`.

The rest of the time we should use `field.allow` (see above).

EVERYTHING BELOW HERE IS LESS IMPORTANT 😄 

Here are my comments on other occurrences of `aposIsTemplate` in the source. Perhaps these are harder to fix. We should **not** delay the initial release over this.

```
      } else if (this.field.aposIsTemplate) {
        let prefix = '';
        if (this.field.page) {
          if (!updated.startsWith('/@')) {
            prefix = '/@';
          }
        } else {
          if (!updated.startsWith('@')) {
            prefix = '@';
          }
        }
        updated = prefix + updated;
      }
```

Can't we just use the existing `field.prefix` feature? A server side override in the doc template module would set this property appropriately when `aposIsTemplate` is true. Possibly prepending it to any existing `field.prefix` because that would just feel more consistent, although the slugs of templates don't really matter beyond being unique and having the `@` or `/@`.

** AposSchema**

```
        fields[item.name].field.aposIsTemplate = this.value?.data?.aposIsTemplate;
```

I am wondering if this is really necessary because I see server-side references to `field.aposIsTemplate` too. If it is already there then I don't know if we need to be setting it here. Maybe we do. Which would make all of my other suggestions less likely to work.

ValJed:
@ETLaurent @boutell 

I agree about this even if it's not the purpose of this PR.
Note that we can't use the `super` pattern for context operations.
Imagine that one day we add  context operation in this method:
```javascript
      addContextMenu() {
        self.apos.doc.addContextOperation(self.__meta.name, {
          action: 'shareDraft',
          context: 'update',
          label: 'apostrophe:shareDraft',
          modal: 'AposModalShareDraft',
          manuallyPublished: true,
          hasUrl: true,
          if: {
            aposIsTemplate: { $ne: true }
          }
        });
      },
```

If I override this one in my `doc template library` module this new method will be lost. So we might find a better solution to extend/override this kind of method or be sure that we always create a new method to create a new context operation for examples.

I will override the context operation that I modified in this PR.
For the rest I'll manage it in another PR.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
ValJed: We could be the other implementations are different, for example they don't support the `$ne` for now neither nested properties (because it doesn't exist for fields).
I will extract it but I won't refactoring now other implementations since they are not the same.

ValJed:
@boutell @ETLaurent 
I moves the reference to `aposIsTemplate`, we found a better way to modify already declared context operations.
I also moved the check if conditions logic into a specific file, since it's a specific implementation that has not a lot in common with fields I won't refacto nothing related right now.
@boutell I kept your comment about slugs and others references to `aposIsTemplate` in the core that we need to move but I'll do it in the main fixes branch.

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -69,7 +69,8 @@ module.exports = ({
       alias: {
         vue$: 'vue/dist/vue.runtime.esm.js',
         // resolve apostrophe modules
-        Modules: path.resolve(modulesDir)
+        Modules: path.resolve(modulesDir),
+        Root: path.resolve(__dirname, '../../../../../..')
```
boutell, line 73: An interesting idea, but I think `import { something} from 'apostrophe/lib/whatever.js'` is just as effective, labels what is happening more clearly, and doesn't require adding any new webpack configuration.

boutell: (That is, in the frontend build, importing stuff from sub-files of `apostrophe` is just as valid as importing stuff from sub-files of any other npm module.)

ValJed: That is way better indeed, thanks!

boutell (approved):

PR #4233, opened 2023-07-21
Fix save as template

boutell:
A way to indicate a doc should be treated as autopublished in the frontend UI


modules/@apostrophecms/ui/ui/apos/mixins/AposCellMixin.js
```diff
@@ -43,6 +43,15 @@ export default {
   computed: {
     item() {
       return this.published || this.draft;
+    },
+    moduleOptions() {
```
boutell, line 47: These make more sense here

stuartromanek (approved):

PR #4245, opened 2023-07-31
addContextOperation: Adds `$and` operator

falkodev:
Adds `$and` operator on `addContextOperation` `if` property in order to check multiple fields before showing or hiding a context operation

Related to https://github.com/apostrophecms/import-export/pull/1/files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ for piece and page types, allowing "virtual piece types" on the back end that
 deal with many content types to give better hints to the UI.
 * Respect the `_aposAutopublish` property of a document if present, otherwise
 fall back to module configuration.
-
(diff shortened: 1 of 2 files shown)


boutell (changes requested):
Why do you need this? Setting the condition to `{ prop1: true, prop2: true }` already requires both to be true (see the `every` call at the beginning).

falkodev:
> Why do you need this? Setting the condition to `{ prop1: true, prop2: true }` already requires both to be true (see the `every` call at the beginning).

Because `getNestedPropValue` checks for a value, not an array of values. And in my case, if `export: false`  is added to multiple modules, only the last one is taken in consideration. I could have modified `getNestedPropValue` to handle arrays, but adding `$and` seemed simpler and easy to reuse in other cases (and it's mirrors mongodb operators).

If you prefer having `getNestedPropValue` handling arrays and values, let me know and I'll change this.

PR #4246, opened 2023-07-31
Core API, formatting by env, req support, tear down

myovchev:
## Summary

- Add structured logging API - `logDebug`, `logInfo`, logWarn` and `logError` (can be used in any module):
```js
self.logError('event-type');
self.logError('event-type', { some: 'data' });
self.logError('event-type', 'some message', { some: 'data' });
// prepend `req` to also log valuable request data, followed by any of the above arguments
self.logError(req, ...);
```
- Custom loggers can now be destroyed - provide `async destroy()` method.
- Default logger now stringifies objects.
- Production and non-production default policy is applied for structured logging.
- Filtering by severity and event type per module can be configured for structured logging.
- Format logs for readability in non-production environments.

Closes PRO-4479, PRO-4480, PRO-4481, PRO-4482, PRO-4483, PRO-4486, PRO-4487.

See the above mentioned tickets acceptance criteria.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/log/index.js
+++ b/modules/@apostrophecms/log/index.js
@@ -55,10 +55,22 @@
 //       // Log all errors and warnings from any module
 //       '*': {
 //         severity: [ 'warn', 'error' ]
+//         // match all severity levels
+//         // severity: '*'
+//         // match event types
+//         // events: [ 'event-type-1', 'event-type-2' ]
+//         // match all event types
+//         // events: '*'
 //       },
 //       // Log specific event types from the login module
 //       '@apostrophecms/login': {
 //         events: [ 'incorrect-username', 'incorrect-password' ]
+//         // match all event types
+//         // events: '*'
+//         // match specific severity levels
+//         // severity: [ 'info' ]
+//         // match all severity levels
+//         // severity: '*'
 //       }
 //     }
 //   }
@@ -66,7 +78,8 @@
 // ```
 // In this example, all errors and warnings from any module, but
 // only the specific event types (no matter the severity) from the login
-// module, are logged.
+// module, are logged. The logs will be kept if *either* criterion is met.
+// `filter['*'] = true` enables logging of all events from all modules.
 //
 // ## Environment Variables
 //
@@ -77,7 +90,7 @@
 // ```sh
 // # same as the `filter` example above
 // export APOS_FILTER_LOGS='*:severity:warn,error;@apostrophecms/login:events:incorrect-username,incorrect-password'
-// # log everything
+// # log everything, analogous to `{ filter: { '*': true }}`
 // export APOS_FILTER_LOGS='*'
 // ```
 //
@@ -94,50 +107,34 @@ module.exports = {
   },
   methods(self) {
     return {
-      // Logger for a module factory.
-      // The created logger object contains `debug`, `info`, `warn` and `error` methods,
-      // each suuporting the following signature:
-      // - (eventType)
-      // - (eventType, data)
-      // - (eventType, message)
-      // - (eventType, message, data)
-      // - (req, eventType[, message, data]) message and data are optional
-      getLoggerForModule(moduleSelf) {
-        return {
-          debug: (...args) => {
-            self.logEntry(moduleSelf, 'debug', ...args);
-          },
-          info: (...args) => {
-            self.logEntry(moduleSelf, 'info', ...args);
-          },
-          warn: (...args) => {
-            self.logEntry(moduleSelf, 'warn', ...args);
-          },
-          error: (...args) => {
-            self.logEntry(moduleSelf, 'error', ...args);
-          }
-        };
-      },
-
       // Normalize the filters. Detect configuration and set defaults
       // per the current NODE_ENV if needed.
       // Convert `{ *: true }` to an object with all severity levels.
-      // Convert severity/event type wildcards to arrays.
-      // Override the configuration with the `APOS_FILTER_LOGS` environment variable.
+      // Convert severity wildcards to arrays of (all) severity levels. This
+      // speeds up the severity detection (no wildcards match).
+      // Convert eventType wildcards to `[ '*' ]` array.
+      // Override the configuration with the `APOS_FILTER_LOGS` environment variable
+      // if set.
       initFilters() {
         self.filters = self.options.filter || {};
         if (process.env.APOS_FILTER_LOGS) {
-          self.filters = self.processEnvFilter(process.env.APOS_FILTER_LOGS);
+          try {
+            self.filters = self.parseEnvFilter(process.env.APOS_FILTER_LOGS);
+          } catch (e) {
+            throw new Error(`Invalid APOS_FILTER_LOGS environment variable: ${e.message}`);
+          }
         }
         self.filters['*'] = self.filters['*'] || {};
 
-        // Transform *: true.
+        // Transform *: true - log absolutely everything.
         if (self.filters['*'] === true) {
           self.filters['*'] = {
             severity: self.getDefaultSeverity(false)
           };
-        } else if (!self.filters['*'].severity || self.filters['*'].severity.length === 0) {
-          // Add environment specific severity levels if no severity is specified.
+          return;
+        }
+        // Add environment specific severity levels if no severity is specified.
+        if (!self.filters['*'].severity || self.filters['*'].severity.length === 0) {
           self.filters['*'] = {
             ...self.filters['*'],
             severity: self.getDefaultSeverity(process.env.NODE_ENV === 'production')
@@ -172,24 +169,29 @@ module.exports = {
       //     events: [ 'success', 'failure' ]
       //   }
       // }
-      // Log all is just `*`.
-      processEnvFilter(envFilter) {
+      // Log all is just `*` and results in `{ '*': true }`.
+      parseEnvFilter(envFilter) {
         const filter = {};
         if (envFilter === '*') {
           filter['*'] = true;
           return filter;
         }
         envFilter.split(';').forEach((entry) => {
-          const [ module, ...criteria ] = entry.split(':');
-          const [ type, ...values ] = criteria;
-          const value = values.join(':');
-          if ([ 'severity', 'events' ].includes(type)) {
-            filter[module] = {
-              [type]: value.split(',')
-            };
-          } else {
-            throw new Error(`Unknown filter type: ${type}`);
-          }
+          const [ module, ...criteria ] = entry
+            .trim()
+            .split(':')
+            .map((str) => str.trim());
+          // Trnasform e.g. `events:ev1,ev2` to`{ events: [ 'ev1', 'ev2' ] }`.
+          filter[module] = criteria.reduce((acc, current, index, arr) => {
+            if (index % 2) {
+              return acc;
+            }
+            if (!arr[index + 1] || typeof arr[index + 1] !== 'string') {
+              throw new Error(`Malformed configuration for module "${module}".`);
+            }
+            acc[current] = arr[index + 1].split(',').map((str) => str.trim());
+            return acc;
+          }, {});
         });
         return filter;
       },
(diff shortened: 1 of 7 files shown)


modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+// This provides compatibility out of the box with many popular
+// logging modules, including`pino`, `winston`, etc.
+//
+// ## Options
+//
+// ### `messageAs`
+//
+// When the messageAs option is set, the message argument to apos.util.info, etc.
+// is bundled into the second, object - based argument as a property of the name
+// given, and only the object argument is passed to Pino. If there is
```
boutell, line 25: is passed to the `logger`, which is useful if using Pino

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+          req,
+          eventType,
+          message,
+          data
+        );
+        if (self.shouldKeepEntry(data)) {
+          self.apos.util[severity](...args);
+        }
+      },
+      // Do a minimal computation and validation of the arguemnts - logs should
```
boutell, line 132: arguments

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+            self.logEntry(moduleSelf, 'warn', ...args);
+          },
+          error: (...args) => {
+            self.logEntry(moduleSelf, 'error', ...args);
+          }
+        };
+      },
+      // Internal module, do not use it directly. See `@apostrophecms/module` for
+      // module level logging methods - logInfo, logError, etc.
+      logEntry(moduleSelf, severity, req, eventType, message, data) {
```
boutell, line 119: when is moduleSelf ever missing?

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+          },
+          warn: (...args) => {
+            self.logEntry(moduleSelf, 'warn', ...args);
+          },
+          error: (...args) => {
+            self.logEntry(moduleSelf, 'error', ...args);
+          }
+        };
+      },
+      // Internal module, do not use it directly. See `@apostrophecms/module` for
```
boutell, line 117: Internal  method

modules/@apostrophecms/log/index.js
boutell: Also should be marked as an internal method you shouldn't be calling normally

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+      // message is optional, if not provided it is generated from the eventType and data.module.
+      // Optional message values:
+      // - `module: eventType: message`
+      // - `eventType: message`: (missing module)
+      // - `module: eventType`: (missing message)
+      // - `eventType`: (missing module and message)
+      //
+      // Returns [ data ] or [ message, data ] depending on option.messageAs.
+      // `data` is always an object containing at least a `type` and `severity` properties.
+      processLoggerArgs(...allArgs) {
```
boutell, line 159: I think this could read `moduleSelf, ...args` since it's always there.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+        const args = allArgs.slice(1);
+        const moduleSelf = allArgs[0];
+        let severity;
+        let req;
+        let eventType;
+        let message;
+        let data;
+
+        // Detect `req` argument with a simple duck type check - apos `req` object
```
boutell, line 168: That's a pretty good check. Functions are weird things to pass to loggers so I'm comfortable with checking a common property name like `t`. But if it ever becomes a problem you could check `req?.res?.redirect`

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+        // has always a translate `t` function.
+        if (args[1] && typeof args[1].t === 'function') {
+          [ severity, req, eventType, message, data ] = args;
+        } else {
+          [ severity, eventType, message, data ] = args;
+        }
+
+        if (!_.isString(eventType)) {
```
boutell, line 176: Could be typeof, no?

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,255 @@
...
+          requestId: req.requestId || self.apos.util.generateId(),
+          ...data
+        };
+      },
+
+      // Assess the module filter configuration and determine if the log
+      // should be kept or rejected.
+      //
+      // `data` argument should be an object returned from `processLogArgs(...)`.
+      shouldKeepEntry(data) {
```
boutell, line 230: Try to do this really early. It would be great if we didn't have to clone any objects before this happens.

modules/@apostrophecms/util/lib/logger.js
```diff
@@ -35,14 +37,14 @@ module.exports = function(apos) {
     // substitution strings in mind. See the
     // `console.log` documentation.
 
-    debug: function(msg) {
+    debug: function(...args) {
       // eslint-disable-next-line no-console
       if (console.debug) {
```
boutell, line 42: You can lose this check, it has existed since node 8.

boutell (changes requested):
Direction feels good. I know you're not asking for review yet but I did spot a few small things.

modules/@apostrophecms/log/index.js
myovchev: There was a case in the old implementation. Currently it's always provided. However it doesn't field good to force it at such low level.

modules/@apostrophecms/log/index.js
myovchev: Is it also added to the synthetic request (the `redirect`)?

modules/@apostrophecms/log/index.js
myovchev: Sure, part of the cleanup (too much AI).

modules/@apostrophecms/log/index.js
myovchev: Make sense. The problem is that I need the `data` object already formatted so I can assess it. But I'll look into it closely today.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
+// Structured logging for Apostrophe.
+//
+// This module is generic, low level implementation. For logging inside of a module,
+// see `logInfo`, `logError`, etc. methods in `@apostrophecms/module`.
```
boutell, line 4: clearer to say "available in every module via the base class, `@apostrophecms/module`"

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+
+        // Transform *: true - log absolutely everything.
+        if (self.filters['*'] === true) {
+          self.filters['*'] = {
+            severity: self.getDefaultSeverity(false)
+          };
+          return;
+        }
+        // Add environment specific severity levels if no severity is specified.
+        if (!self.filters['*'].severity || self.filters['*'].severity.length === 0) {
```
boutell, line 137: Empty array should not mean "everything." It's legitimate to create configuration via tools and tools are likely to emit empty arrays when they mean "nothing" rather than skipping a property.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+          self.filters['*'] = {
+            ...self.filters['*'],
+            severity: self.getDefaultSeverity(process.env.NODE_ENV === 'production')
+          };
+        }
+
+        // Handle wildcards and validate.
+        Object.keys(self.filters).forEach((module) => {
```
boutell, line 145: I think this would read better using `Object.entries`

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+          Object.keys(self.filters[module]).forEach((type) => {
+            if (self.filters[module][type] === true || self.filters[module][type] === '*') {
+              self.filters[module][type] = type === 'severity'
```
boutell, line 148: Parentheses would be good for readability here.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+                ? self.getDefaultSeverity(false)
+                : [ '*' ];
+            }
+            if (!Array.isArray(self.filters[module][type]) || self.filters[module][type].length === 0) {
```
boutell, line 152: Empty arrays are OK and don't allow anything.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+        // Keep the proper property order for better readability.
+        const message = {};
+        if (self.options.messageAs) {
+          message[self.options.messageAs] = data[self.options.messageAs];
+        }
+        const {
+          module: _module, type, severity, ...rest
+        } = data;
+
+        return {
```
boutell, line 319: Don't clone the object again, we know it is already a clone and it hurts performance. Assign the new properties to the existing object.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+        let severity;
+        let req;
+        let eventType;
+        if (args[1] && typeof args[1].t === 'function') {
+          // eslint-disable-next-line no-unused-vars
+          [ severity, req, eventType ] = args;
+        } else {
+          [ severity, eventType ] = args;
+        }
+        const module = moduleSelf.__meta?.name ?? '__unknown__';
```
boutell, line 356: We only really care about `module`, `severity` and `eventType`, and the same queries are going to be made a lot.

So after this line, check a cache based on this tuple before doing a lot of spreading and Set instantiation.

modules/@apostrophecms/log/index.js
```diff
@@ -0,0 +1,403 @@
...
+
+        return severityMatch || eventsMatch;
+      },
+
+      // Stringify object arguments. If the environment is not `production`,
+      // pretty print the objects and add a new line at the end of string arguments.
+      // This method is meant to be used from the methods of a custom `logger`.
+      // See the default logger implementation in `util/lib/logger.js` for an example.
+      formatLogByEnv(args) {
+        const formatObj = process.env.NODE_ENV === 'production'
```
boutell, line 384: Don't make new functions on every call; check NODE_ENV just once at startup and assign one of them to use going forward for the lifetime of the process.

modules/@apostrophecms/module/index.js
```diff
@@ -74,6 +75,11 @@ module.exports = {
     // Routes in their final ready-to-add-to-Express form
     self._routes = [];
 
+    // Add structured logging if we passed the util module.
```
boutell, line 78: I think you mean: "Enable structured logging if this is not the util module itself"

modules/@apostrophecms/util/index.js
```diff
@@ -811,6 +836,46 @@ module.exports = {
       },
       omit(source, keys) {
         return _.omit(source, keys);
+      },
+
+      // Internal method. Attempt to convert the log payload to an object
+      // for legacy calls and when `@apostrophecms/log` has been configured
+      // with `messageAs: 'someKey'`.
+      // This change is backwards compatible with the previous behavior because
+      // `messageAs` is newly introduced option. Custom loggers should adapt
```
boutell, line 845: is newly -> is a newly

boutell (changes requested):
Generally great. One or two functional concerns and some performance recommendations.

modules/@apostrophecms/log/index.js
myovchev: Nice one!

modules/@apostrophecms/log/index.js
myovchev: So true :)

modules/@apostrophecms/log/index.js
myovchev: Yeah, last moment change because of the prop order requirement. But there is an easy way to be done via a mutation!

modules/@apostrophecms/module/index.js
myovchev: Or `log` or `util` :)

modules/@apostrophecms/log/index.js
```diff
@@ -134,28 +145,28 @@ module.exports = {
...
 
         // Handle wildcards and validate.
-        Object.keys(self.filters).forEach((module) => {
-          Object.keys(self.filters[module]).forEach((type) => {
-            if (self.filters[module][type] === true || self.filters[module][type] === '*') {
-              self.filters[module][type] = type === 'severity'
+        for (const [ module, config ] of Object.entries(self.filters)) {
+          for (const [ type, value ] of Object.entries(config)) {
+            if (value === true || value === '*') {
+              self.filters[module][type] = (type === 'severity')
```
boutell, line 159: You can assign to `config[type]` here, no?

boutell (approved):

modules/@apostrophecms/log/index.js
myovchev: Fixed.

PR #4251, opened 2023-08-03
Pro 4257 modal operation

falkodev:
## Summary

Add the ability to have custom modals for batch operations, and align props between doc context menu custom modal and batch operation custom modal, in order to use the same component.

Test with https://github.com/apostrophecms/import-export/pull/2

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
```diff
@@ -428,13 +428,18 @@ export default {
 
     },
     async customAction(doc, operation) {
+      if (operation.label) {
```
boutell, line 431: How is this related to the label? 🤔

boutell: Is the idea that label and modal are mutually exclusive? Maybe better to check that `operation.moduleName` is not set before doing this.

boutell (changes requested):

modules/@apostrophecms/doc-type/ui/apos/components/AposDocContextMenu.vue
falkodev: `operation.moduleName` is set with the wrong module name and is used in some modal, so I guess there is a use case where it's valid. In my use case, I need the current module name.

I've change this `if` condition with a custom property I've created on the backend instead. Will it be safe @boutell ?

modules/@apostrophecms/modal/ui/apos/components/AposDocsManagerToolbar.vue
```diff
@@ -263,7 +263,18 @@ export default {
...
 
-      await this.confirmOperation(operation);
+      operation.modal ? await this.modalOperation(operation) : await this.confirmOperation(operation);
+    },
+    async modalOperation({
+      modal, ...rest
+    }) {
+      await apos.modal.execute(modal, {
+        count: this.checkedCount,
+        moduleName: this.checkedCount === 1
```
boutell, line 273: Wait, why would the label be the source of the Apostrophe module name? I'm a little lost now.

boutell (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -72,6 +72,7 @@
             :is-relationship="!!relationshipField"
             :checked-count="checked.length"
             :batch-operations="moduleOptions.batchOperations"
+            :module-name="moduleName"
```
boutell, line 75: AposMediaManager should pass it too. And the doc template library one, in a separate PR.

boutell (changes requested):

PR #4258, opened 2023-08-04
PRO-4655: fix the moduleName prop passed to context operation modals (if the developer cooperates)

boutell:
See changelog.


modules/@apostrophecms/doc/index.js
```diff
@@ -1147,40 +1147,77 @@ module.exports = {
...
+      // If all `conditions` are not met, the item is not displayed for this particular
+      // document.
+      //
+      // `conditions` may be an array containing one or multiple of these values:
+      //
       // 'canPublish', 'canEdit', 'canDismissSubmission', 'canDiscardDraft',
       // 'canLocalize', 'canArchive', 'canUnpublish', 'canCopy', 'canRestore'.
-      addContextOperation(moduleName, operation) {
+
+      addContextOperation(operation) {
```
haroun, line 1206: This will break all modules using the old syntax, e.g. `document-versions`, `doc-template-library` and all client code using `extendMethods`

haroun (changes requested):

PR #4269, opened 2023-08-14
PRO-4685: Localization sometimes fails due to incorrect stored values for lastTargetId and lastPosition in the page tree

boutell:
## Summary

Just get rid of these frequently inaccurate properties and infer them, which we can do with complete accuracy by looking at where the doc is now in the tree. Storing these properties was a legacy of A2 workflow.

* Movement in the page tree works as before.
* "Localize" works as before.
* "Publish" works as before.
* Client issue resolved (see internal notes — I tested this).

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/page/index.js
```diff
@@ -2497,6 +2491,48 @@ database.`);
             await self.apos.attachment.recomputeAllDocReferences();
           }
         });
+      },
+      async inferLastTargetIdAndPosition(doc) {
```
boutell, line 2495: This code was borrowed from an old migration that calculated these properties when missing, which was in turn removed because we no longer store these properties.

modules/@apostrophecms/page/lib/legacy-migrations.js
```diff
@@ -100,62 +99,6 @@ module.exports = (self) => {
           }
         });
       });
-    },
-    addMissingLastTargetIdAndPositionMigration() {
```
boutell, line 104: We don't store these anymore, but this code became the basis of inferring them.

test/pages.js
```diff
@@ -386,6 +386,42 @@ describe('Pages', function() {
     assert.strictEqual(page.path, `${homeId.replace(':en:published', '')}/another-parent/parent/sibling`);
   });
 
+  it('inferred page relationships are correct', async function() {
```
boutell, line 389: Various sanity checks of the inferred target id and position which should be durable if the pages created by the rest of these tests change.

test/pages.js
```diff
@@ -386,6 +386,42 @@ describe('Pages', function() {
...
+    const req = apos.task.getReq();
+    const pages = await apos.page.find(req, {}).toArray();
+    for (const page of pages) {
+      if (page.level === 0) {
```
boutell, line 393: You can't infer last target id and position on the home page because it never moves and has no parent or peer.

ValJed (changes requested):
few questions and indentation requests.
I tested it looks good (not on a michelin project though)

modules/@apostrophecms/page/lib/legacy-migrations.js
boutell: As a general rule we avoid removing data because we are fallible, make mistakes, and might regret not having it later 😄 I think that applies here.

PR #4275, opened 2023-08-22
Send the original error message to the logger

myovchev:
## Summary

Set the original error message when compiling the REST API error log and not the message meant for the browser `An error occurred`.

When there is an actual unhandled error, the original message should be logged in the CLI instead `An error occurred`. 

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/module/index.js
+++ b/modules/@apostrophecms/module/index.js
@@ -250,8 +250,12 @@ module.exports = {
         });
         function logError(req, response, error) {
           const typeTrail = response.code === 500 ? '' : `-${response.name}`;
+          // Log the actual error, not the message meant for the browser.
+          const msg = response.message !== err.message
+            ? err.message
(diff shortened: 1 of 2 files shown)


ValJed (approved):

modules/@apostrophecms/module/index.js
```diff
@@ -250,8 +250,12 @@ module.exports = {
         });
         function logError(req, response, error) {
           const typeTrail = response.code === 500 ? '' : `-${response.name}`;
+          // Log the actual error, not the message meant for the browser.
+          const msg = response.code === 500
+            ? err.message
```
boutell, line 255: Would be better to use the scoped `error` argument, but it does the same thing.

PR #4294, opened 2023-09-21
I18n fix

BoDonkey:
## Summary

Currently, if an i18n folder has any files inside besides JSON translation string files, it will break the build process. For example, this can occur if a user creates a new JSON file using the Mac OS finder. The finder will leave behind a `.DS_Store` file. This PR fixes this issue and closes PRO-4701. 

*For example:*
1. In testbed (or any project) create two files - `modules/<some-module>/i18n/test.txt` and `modules/<some-module>/i18n/testNs/test.txt`. The first will test without namespace, the second with. Testbed already has the structure in place for the `topic` module.
2. Project build should succeed with `npm run dev`

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/i18n/index.js
+++ b/modules/@apostrophecms/i18n/index.js
@@ -524,12 +524,12 @@ module.exports = {
                 (metadata[ns] && metadata[ns].browser);
               const namespaceDir = path.join(localizationsDir, ns);
               if (!fs.statSync(namespaceDir).isDirectory()) {
-                // Handle files in the namespace that aren't JSON files
+                // Skip non-directory items, such as hidden files
                 continue;
               }
               for (const localizationFile of fs.readdirSync(namespaceDir)) {
                 if (!localizationFile.endsWith('.json')) {
-                  // A JSON file for the default namespace, already handled
+                  // Exclude parsing of non-JSON files, like hidden files, in the namespace directory
                   continue;
                 }
                 const fullLocalizationFile = path.join(namespaceDir, localizationFile);
--- a/modules/@apostrophecms/i18n/index.js
+++ b/modules/@apostrophecms/i18n/index.js
@@ -523,7 +523,15 @@ module.exports = {
               self.namespaces[ns].browser = self.namespaces[ns].browser ||
                 (metadata[ns] && metadata[ns].browser);
               const namespaceDir = path.join(localizationsDir, ns);
+              if (!fs.statSync(namespaceDir).isDirectory()) {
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/i18n/index.js
```diff
@@ -523,7 +523,15 @@ module.exports = {
               self.namespaces[ns].browser = self.namespaces[ns].browser ||
                 (metadata[ns] && metadata[ns].browser);
               const namespaceDir = path.join(localizationsDir, ns);
+              if (!fs.statSync(namespaceDir).isDirectory()) {
+                // Handle files in the namespace that aren't JSON files
+                continue;
+              }
               for (const localizationFile of fs.readdirSync(namespaceDir)) {
+                if (!localizationFile.endsWith('.json')) {
+                  // A JSON file for the default namespace, already handled
```
boutell, line 532: How would a non-json file in a namespace folder be a JSON file for the default namespace? This check makes sense but the comment has me wondering if things were transposed somehow.

modules/@apostrophecms/i18n/index.js
```diff
@@ -523,7 +523,15 @@ module.exports = {
               self.namespaces[ns].browser = self.namespaces[ns].browser ||
                 (metadata[ns] && metadata[ns].browser);
               const namespaceDir = path.join(localizationsDir, ns);
+              if (!fs.statSync(namespaceDir).isDirectory()) {
+                // Handle files in the namespace that aren't JSON files
```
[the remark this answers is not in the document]
BoDonkey: Currently, the conditional at the top of `for` loop checks if the current item is a JSON file and if it is skips it because it has been handled by the `addDefaultResourcesForModule()` method. At this point, the method is assuming that everything else is a directory, which isn't always true. Therefore we need to a check to jump to the next item in the loop if it is a file that doesn't have a `.json` extension. This would include `.DS_Store` that is breaking the build.
Maybe I'm not understanding your concern.

modules/@apostrophecms/i18n/index.js
BoDonkey: Yeah - I think that copilot "helpfully" wrote that comment and I wasn't careful enough in reading it. That conditional makes sure that within the namespace folder only JSON files are being parsed. I'll change the comment.

boutell (approved):

Issue #4291, opened 2023-09-15
Return an empty string for an undefined date/time field of a piece-type is displayed in a column with component: 'AposCellDate'


chmdebeer:
## To Reproduce

Step by step instructions to reproduce the behavior:

> 1. Add a "column" for a piece-type with a date/time field and set the component to AposCellDate, for example:
```
columns: {
    add: {
      firstName: { label: 'First Name' },
      lastName: { label: 'Last Name' },
      confirmedOn: { label: 'Confirmed On', component: 'AposCellDate' }
    }
  }

```
> 2. Add a piece of that piece-type but leave the date field empty
> 3. The admin interface will display today's date since the empty field is passed to "new Date()" which returns today's date.

## Expected behavior
> Return an empty string.

## Describe the bug
> in @apostrophecms/ui/ui/apos/components/AposCellDate.vue the method formatDateColumn needs protection from an empty field.
> I did the following but it is crude and I am sure the is a more elegant way of handling it:
```
...
  methods: {
    formatDateColumn (value) {
      if (!value) {
        return "";
      }
      const format = this.$t('apostrophe:dayjsCellDateFormat');
      // endsWith allows it to still match when the i18n show debug mode
...
`

AllanKoder:
Hi, could I work on a fix for this?

BoDonkey:
Hi @AllanKoder ,
 We don't formally assign issues, so if you want to work on this it is fine. Do you have any experience with Apostrophe?
Cheers,
Bob

AllanKoder:
I've messed around with the CMS and I have the dev environment set up. I'll get started on it.

PR #4297, opened 2023-09-24
added date verification for null date and no default value

AllanKoder:
## Summary

Closes Issue number #4291. Empty dates for piece types are no longer filled with the current date after submitting nothing. 

1. create a default apos app. 
2. create a piece type: 
```module.exports = {
  options: {
    label: 'Testing Piece',
  },
  fields: {
    add: {
      date: {
        label: 'Date?',
        type: 'date'
      },
    },
    group: {
      basics: {
        label: 'Basics',
        fields: [ 'date']
      }
    }
  },
  columns: {
    add: {
      date: { label: 'Confirmed On', component: 'AposCellDate' }
    }
  }
};
```
3. create the new piece type and set date as empty
4. the date should remain as an empty string

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/lib/addFieldTypes.js
+++ b/modules/@apostrophecms/schema/lib/addFieldTypes.js
@@ -584,9 +584,9 @@ module.exports = (self) => {
         destination[field.name] = null;
         return;
       }
-      if (!newDateVal && !field.def){
+      if (!newDateVal && !field.def) {
         // If no inputted date or default date, leave as empty
-        destination[field.name] = null
+        destination[field.name] = null;
         return;
       }
       if (field.min && newDateVal && (newDateVal < field.min)) {
--- a/modules/@apostrophecms/schema/lib/addFieldTypes.js
+++ b/modules/@apostrophecms/schema/lib/addFieldTypes.js
@@ -584,17 +584,17 @@ module.exports = (self) => {
         destination[field.name] = null;
         return;
       }
+      if (!newDateVal && !field.def){
+        // If no inputted date or default date, leave as empty
+        destination[field.name] = null
+        return;
+      }
       if (field.min && newDateVal && (newDateVal < field.min)) {
         // If the min requirement isn't met, leave as-is.
         return;
       }
       if (field.max && newDateVal && (newDateVal > field.max)) {
         // If the max requirement isn't met, leave as-is.
-        return;
-      }
-      if (!newDateVal && !field.def){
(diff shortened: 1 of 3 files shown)


AllanKoder:
> Thanks for contributing!
> 
> However, I think this should be handled in the component on the front end. There's nothing wrong with `null` as a representation of "no value," and handling it on the front end will avoid the need for a data migration for existing data.
> 
> Can you take a look at handling it that way?

Hi, I edited the backend code and added a null filter for AposCellDate. However, I could not remove the backend code completely as one of the issues was that a front end user could submit an empty date and the `convert` function in addFieldTypes.js would turn the null value into a date with the line `self.apos.launder.date(newDateVal, field.def);` and the data would be saved in the mongo docs and rendered onto the frontend. Hence, I think the only way to allow empty dates is to set the value as a null.

Let me know if my line of thinking is broken somewhere. Thanks so much!

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -584,6 +584,11 @@ module.exports = (self) => {
         destination[field.name] = null;
         return;
       }
+      if (!newDateVal && !field.def){
+        // If no inputted date or default date, leave as empty
+        destination[field.name] = null
```
boutell, line 589: 👍 I'm OK with this. Technically it's a change of behavior, but I do not think this behavior served anyone on the server side, where it was very unexpected. Please add a space after `field.def)` and before the `{`

boutell: Also make sure the unit tests still pass (`npm test`) if not we might need to adjust one to account for not automatically filling in today's date.

modules/@apostrophecms/ui/ui/apos/components/AposCellDate.vue
```diff
@@ -18,7 +18,7 @@ export default {
   computed: {
     formattedDate () {
       const value = this.get(this.header.name);
-      return this.formatDateColumn(value);
+      return value ? this.formatDateColumn(value) : "";
```
boutell, line 21: Looks good.

boutell (changes requested):

boutell (approved):
Thank you!

PR #4301, opened 2023-09-26
Pro 4734 check outdated deps

ETLaurent:
image

### DIFF
--- a/.github/workflows/outdated-dependencies.yml
+++ b/.github/workflows/outdated-dependencies.yml
@@ -23,8 +23,10 @@ jobs:
         run: npm install
       - name: Check outdated dependencies
         run: |
-          echo "$(npm outdated)" > output
-          npm outdated
+          outdated_dependencies=$(npm outdated)
+          exit_status=$?
+          echo $outdated_dependencies > output
+          exit $exit_status
       - name: Report Status
         if: failure()
         run: |
--- a/.github/workflows/outdated-dependencies.yml
+++ b/.github/workflows/outdated-dependencies.yml
@@ -25,7 +25,8 @@ jobs:
         id: outdated
         # run: npm outdated | awk '{ print $0 >> "$GITHUB_ENV" }'
         run: |
-          echo 'TOTO="$(npm outdated)"' >> $GITHUB_ENV
+          TITI=$(npm outdated)
+          echo "TOTO=$TITI" >> $GITHUB_ENV
           cat $GITHUB_ENV
       - name: Report Status
         if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/outdated-dependencies.yml
```diff
@@ -1,5 +1,7 @@
 name: Check outdated dependencies
 on:
+  push:
+    branches: ["pro-4734-check-outdated-deps"]
```
ETLaurent, line 4: rm

.github/workflows/outdated-dependencies.yml
```diff
@@ -1,37 +1,43 @@
...
-          node-version: ${{ matrix.node-version }}
-      - id: run_outdated
-        run: npm outdated >> "$GITHUB_OUTPUT"
+          node-version: 18
+      - name: Install dependencies
+        run: npm install
+      - name: Check outdated dependencies
+        run: |
+          echo "$(npm outdated)" > output
+          npm outdated
```
ETLaurent, line 24: way of getting the correct exit status

.github/workflows/outdated-dependencies.yml
```diff
@@ -1,37 +1,43 @@
 name: Check outdated dependencies
 on:
-  # TODO: uncomment schedule
-  # schedule:
-  #   # Runs Mon-Fri at 0:00 UTC ( 00:00 EST )
-  #   - cron: "0 0 * * 1-5"
+  schedule:
+    # Runs Mon-Fri at 0:00 UTC ( 00:00 EST )
+    - cron: "0 0 * * 1-5"
```
boutell, line 5: Once every two weeks would be more appropriate. We will never be 100% up to date with the same major version of everything, often up to date modules require older but maintained versions of other modules, etc. If we generate this report daily we will become blind to it (:

.github/workflows/outdated-dependencies.yml
boutell: `npm outdated > output` should do this by itself 🤔 Unless your goal was just to always add a newline. Not wrong though

boutell (changes requested):

.github/workflows/outdated-dependencies.yml
ETLaurent: I wasn't able to keep the exit status when writing the output into the `output` file, so I'm running the same command again... not pretty but it works

.github/workflows/outdated-dependencies.yml
ETLaurent: Would every week do?
I found a syntax to run it every 2 week: `0 8 * * MON test $((10#$(date +\%W)\%2)) -eq 1`, but it's not compatible with GA

boutell (approved):
Once a week will have to do. 👍

PR #4309, opened 2023-10-05
add "type" property in projections by default to hide permission warnings

ETLaurent:
## Summary

Improve the `can` method of the permission module in order to hide warnings by always providing to it the `type` property of a doc or widget and getting the appropriate manager.

Log in as an editor.
Create a new page with a slug `/a`.
Create another page and type `/a` in the slug field
There should be no `type has no manager` error in the server console

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/permission/index.js
+++ b/modules/@apostrophecms/permission/index.js
@@ -61,11 +61,7 @@ module.exports = {
         }
         const type = docOrType && (docOrType.type || docOrType);
         const doc = (docOrType && docOrType._id) ? docOrType : null;
-        let manager = type && self.apos.doc.getManager(type);
-        if (doc && !manager) {
-          // Fallback to an appropriate manager module
-          manager = self.apos.util.getManagerOf(doc);
-        }
+        const manager = type && self.apos.doc.getManager(type);
         if (type && !manager) {
           self.apos.util.warn('A permission.can() call was made with a type that has no manager:', type);
           return false;
--- a/modules/@apostrophecms/permission/index.js
+++ b/modules/@apostrophecms/permission/index.js
@@ -61,7 +61,11 @@ module.exports = {
         }
         const type = docOrType && (docOrType.type || docOrType);
         const doc = (docOrType && docOrType._id) ? docOrType : null;
-        const manager = type && self.apos.doc.getManager(type);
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/permission/index.js
```diff
@@ -61,7 +61,11 @@ module.exports = {
         }
         const type = docOrType && (docOrType.type || docOrType);
         const doc = (docOrType && docOrType._id) ? docOrType : null;
-        const manager = type && self.apos.doc.getManager(type);
+        let manager = type && self.apos.doc.getManager(type);
+        if (doc && !manager) {
+          // Fallback to an appropriate manager module
+          manager = self.apos.util.getManagerOf(doc);
```
boutell, line 67: We don't support viewRole etc. for widget types. This shouldn't be necessary. We should not be calling permission.can with a widget ever actually. Are we doing that though?

boutell (changes requested):
I think maybe we got a little past the requirements here and did something that doesn't make sense, but you may have seen something I haven't seen, so let me know.

modules/@apostrophecms/permission/index.js
ETLaurent: yes, in the annotate function:

```js
      annotate(req, action, objects) {
        const property = `_${action}`;
        for (const object of objects) {
          if (self.can(req, action, object)) {
            object[property] = true;
          }
        }
      },
```

we could do something like that to exclude widgets, WDYT?

```js
      annotate(req, action, objects) {
        const property = `_${action}`;
        for (const object of objects) {
          if (object.metaType !== 'widget' && self.can(req, action, object)) {
            object[property] = true;
          }
        }
      },
```

boutell: I think we would have to check that the `metaType` is `doc`, to exclude other things that won't work, but yes I agree with this approach.

boutell (changes requested):
I think we would have to check that the metaType is doc, to exclude other things that won't work, but yes I agree with this approach.

haroun (approved):

boutell (approved):

PR #4322, opened 2023-10-23
Pro 4966 condition met required

ValJed:
[PRO-4966](https://linear.app/apostrophecms/issue/PRO-4966/replicate-condition-met-property-as-condition-met-required-property)
[PRO-4964](https://linear.app/apostrophecms/issue/PRO-4964/update-frontend-fieldrequired-on-the-fly)

## Summary

* `condition-met-required` is not passed to `AposSchema` inputs since required is computed in the `fields` computed.
* `field.required` is built on `requiredIf` if existing or `field.required`.
* New `requiredIf` works for all schema components that already used conditional fields, they all have been tested:

- [x]  AposDocEditor
- [x]  AposWidgetEditor
- [x]  AposArrayEditor
- [x]  AposTipTapAnchor
- [x]  AposImageControlDialog
- [x]  AposTipTapLink
- [x]  AposWidgetEditor
- [x]  AposSubForm
- [x]  AposInputObject
- [x]  AposInputArray

* Validate is properly emitted when a change occurs in the `conditionalFields`.
* `conditionalFields` are not re evaluated manually at component mounting and when an input changes, this is not anymore an obscure behavior (method was ran when passed as prop to the `AposSchema` and re ran for some reason at change).

Existing Cypress tests for `conditional fields` and `external conditions` should pass.
Configure schemas for different editors with `requiredIf` properties.
It should support same configurations than the `if`, the same code are evaluating both it's almost unchanged.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/schema/ui/apos/components/AposInputRadio.vue
```diff
@@ -7,31 +7,36 @@
...
           tabindex="1"
           :disabled="field.readOnly"
           @change="change($event.target.value)"
         >
         <span class="apos-input-indicator" aria-hidden="true">
           <component
-            :is="`${next === choice.value ? 'check-bold-icon' : 'span'}`"
-            :size="8" v-if="next === choice.value"
+            v-if="next === value"
+            :is="`${next === value ? 'check-bold-icon' : 'span'}`"
```
ValJed, line 30: Makes no sense since there is a v-if but whetever not the purpose of the PR.

modules/@apostrophecms/schema/index.js
```diff
@@ -1246,6 +1246,12 @@ module.exports = {
         if (field.if && field.if.$or && !Array.isArray(field.if.$or)) {
           fail(`$or conditional must be an array of conditions. Current $or configuration: ${JSON.stringify(field.if.$or)}`);
         }
+        if (field.requiredIf && field.requiredIf.$or && !Array.isArray(field.requiredIf.$or)) {
+          fail(`$or conditional must be an array of conditions. Current $or configuration: ${JSON.stringify(field.requiredIf.$or)}`);
+        }
+        if (field.if && field.if.$or && !Array.isArray(field.if.$or)) {
```
falkodev, line 1252: already present on line 1246

falkodev (changes requested):
Also, with A3 core, cypress tests are ok:
image

With your branch, like half of them fail for `fields` test:
image

myovchev (commented):
It looks to me like a major performance improvement. Few recommendations/concerns:
- `AposDocEditor` might need its own `logic` mixin, as it's sometimes used in projects to deliver custom functionality (I have a project in mind). It think this is a good time for that.
- The computed prop optimization comes with a risk of "stale" (non-reactive) data in some edge cases. I would think of combination of dynamic `choices` and related condition - we would need testbed case testing those (if we don't have currently)

ValJed:
@falkodev this is a tests problem. I fixes the fact that the tab `Templates` wasn't loaded at start so the tab `strings` was triggered. Fixed with this change: https://github.com/apostrophecms/testbed/pull/206/files
Fields tests are passing with it.

boutell (approved):

modules/@apostrophecms/schema/ui/apos/components/AposInputRadio.vue
ValJed: The if isn't new I just aligned props.

modules/@apostrophecms/schema/index.js
ValJed: Removed

PR #4324, opened 2023-10-24
Pro 4962 required if field

falkodev:
# As a developer, I want to be able to make a required field optional based on a user's selection in another field.

https://linear.app/apostrophecms/issue/PRO-4962/as-a-developer-i-want-to-be-able-to-make-a-required-field-optional

## Syntax 

- exact match on a value (boolean, string, number)
- logical AND by default
- logical OR
- not equal with exact match
- all the above with nested properties

```js
requiredIf: {
  fieldA: true,
  fieldB: 'active',
  $or: [
    { fieldC: 10 },
    { fieldD: 20 }
  ],
  fieldF: { $ne: 'default' },
  'fieldE.subFieldA': 'AAA'
}
```

## Acceptance criteria

- we have conditionally required fields on the backend
- it accepts the same syntax as if
- integration test

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -696,20 +696,6 @@ module.exports = {
               continue;
             }
 
-            if (val.min || val.max) {
-              if (val.min) {
-                if (object[key] < val.min) {
-                  result = false;
-                }
-              }
-              if (val.max) {
-                if (object[key] > val.max) {
-                  result = false;
-                }
-              }
-              break;
-            }
-
             if (conditionalFields[key] === false) {
               result = false;
               break;
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -511,7 +511,7 @@ module.exports = {
 
           if (convert) {
             try {
-              const isRequired = self.isFieldRequired({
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -465,6 +465,57 @@ module.exports = {
...
+              continue;
+            } else if (val.$ne) {
+              // eslint-disable-next-line eqeqeq
+              if (val.$ne == destinationKey) {
+                result = false;
+                break;
+              }
+            }
+
+            if (val.min) {
```
ValJed, line 494: If we support it for `requiredIf`, you should add it to `if` too?

ValJed (changes requested):

modules/@apostrophecms/schema/index.js
boutell: All features should work equally for `if` and `requiredIf`.

boutell: With as little code duplication as possible front and back end.

falkodev: done

ValJed (approved):

boutell (approved):

PR #4327, opened 2023-10-31
Update README.md

vatsalya-vyas:
the -> to use

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ To get started with Apostrophe 3, follow these steps to set up a local developme
 
 #### Prerequisites
 
-We recommend installing the following with [Homebrew](https://brew.sh/) on macOS. If you're on Linux, you should use your package manager (apt or yum). If you're on Windows, we recommend the Windows Subsystem for Linux.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ To get started with Apostrophe 3, follow these steps to set up a local developme
 
 #### Prerequisites
 
-We recommend installing the following with [Homebrew](https://brew.sh/) on macOS. If you're on Linux, you should use your package manager (apt or yum). If you're on Windows, we recommend the Windows Subsystem for Linux.
+We recommend installing the following with [Homebrew](https://brew.sh/) on macOS. If you're on Linux, you should use your package manager (apt or yum). If you're on Windows, we recommend to use Windows Subsystem for Linux.
```
boutell, line 41: The original grammar is reasonable. It could be changed to "we recommend using," but "we recommend the" is reasonable too. @BoDonkey what do you think?

boutell (changes requested):

README.md
BoDonkey: I was okay with the original. It is concise and direct.

boutell:
Yeah, I agree. Thanks for the suggestion, but we feel the language is solid here.

PR #4329, opened 2023-10-31
Pro 4962 min max if condition

falkodev:
https://linear.app/apostrophecms/issue/PRO-4962/as-a-developer-i-want-to-be-able-to-make-a-required-field-optional

Extracted `evaluate` from `isVisible` into `evaluateCondition`.
Works for both `if` and `requiredIf`
Tests updated
Run `npx mocha --watch --watch-files test/schemas.js,modules/@apostrophecms/schema/index.js test/schemas.js`


boutell (changes requested):
Please resolve conflicts

boutell (approved):

ValJed (approved):

ValJed:
Looks good, for now `$ne` as well as min and max are not supported in frontend. you might create a ticket for that since you added the backend part already.

PR #4342, opened 2023-11-14
Fix conditional fields

falkodev:
# Fix required fields not enforced when the field is not visible in Apostrophe 3.59.0

https://linear.app/apostrophecms/issue/PRO-5117/[regression]-a-required-field-is-still-enforced-when-the-field-is-not

## Summary

Fix `if` and `requiredIf` fields inside array.

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -498,10 +498,10 @@ module.exports = {
             }
             continue;
           } else if (val.$ne) {
-            // eslint-disable-next-line eqeqeq
-            if (val.$ne == destinationKey) {
+            if (val.$ne === destinationKey) {
               return false;
             }
+            continue;
           }
 
           // Handle external conditions:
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -526,23 +526,21 @@ module.exports = {
             continue;
           }
 
-          if (val.min && destinationKey < val.min) {
-            return false;
-          }
-          if (val.max && destinationKey > val.max) {
-            return false;
+          if (val.min || val.max) {
```
boutell, line 529: 0 can be a legitimate value for min or max. These need to be tested for existence.

boutell (changes requested):

modules/@apostrophecms/schema/index.js
falkodev: Yep, good idea. Done

modules/@apostrophecms/schema/index.js
```diff
@@ -526,23 +526,25 @@ module.exports = {
...
           }
 
-          if (val.min && destinationKey < val.min) {
-            return false;
-          }
-          if (val.max && destinationKey > val.max) {
-            return false;
+          // test with Object.prototype for the case val.min === 0
+          if (
+            Object.prototype.hasOwnProperty.call(val, 'min') ||
```
boutell, line 531: Use `Object.hasOwn(val, 'min')` no need for prototype etc.

boutell (changes requested):

boutell (approved):

Issue #4288, opened 2023-09-12
Line breaks don't work inside `<pre>` tags in rich text widget


tomtuv:
## To Reproduce

Step by step instructions to reproduce the behavior:

1. Add a new rich text widget
2. Select a custom `<pre>` style or create one using the Markdown shortcut <code>```</code>
3. Press <kbd>⇧</kbd> + <kbd>⏎</kbd>
4. A new paragraph is created instead of a line break

## Expected behavior
A new line break should be created inside the same `<pre>` element.

## Describe the bug
Line breaks work fine inside `<p>` tags using <kbd>⇧</kbd> + <kbd>⏎</kbd> but when inside `<pre>` tags, it creates a new paragraph.

## Details

**Version of Node.js:**
16.20.1

**Server Operating System:**
Linux

**Additional context:**
The bug is present in all browsers.

**Screenshots**
<img width="456" alt="Screenshot of the rich text widget with a `<pre>` element" src="https://github.com/apostrophecms/apostrophe/assets/16865165/eafc8f1d-5a7b-45f6-9a1f-4995a791695c">

BoDonkey:
Thanks for the bug report. The rich text widget loads the Tiptap codeblock extensions, but for some reason, it doesn't recognize a return inside. Just an FYI - Tiptap kinda implements Markdown. In this case, once you toggle a codeblock on with three backticks or tilde, you should be able to use a plain return and stay in the block. You need to use three returns or a down arrow to get out of the block (if it is working correctly).

tomtuv:
@BoDonkey Thank you for the quick response. I've tried with the Markdown way using the three backticks but there is the same problem. Pressing <kbd>⏎</kbd> creates inline whitespace and <kbd>⇧</kbd> + <kbd>⏎</kbd> creates a new paragraph.

boutell:
It's worth checking the official tiptap sandboxes to see if they also have this problem. If not we may have broken something with a custom paragraph tag.

BoDonkey:
@tomtuv Sorry - should have been clearer in my earlier comment. You are correct, something is broken. When it is fixed, you should only need to hit return to create a new line still within the code fence.
@boutell I checked the official sandbox and made sure everything worked as expected before I opened the ticket. I forgot we even had the codeblock extension loaded in addition to the code extension!

sujitmahapatra:
To solve the issue where line breaks inside a `<pre>` element are creating new paragraphs instead of new line breaks,

[...]

PR #4362, opened 2023-12-08
Fixed the <pre> tags in the rich text widget so now line breaks inside <pre> tags will work

thatsachin:
Issue #4288 which is "Line breaks don't work inside <pre> tags in rich text widget" is now resolved

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/index.js
+++ b/modules/@apostrophecms/rich-text-widget/index.js
@@ -335,8 +335,11 @@ module.exports = {
       optionsToSanitizeHtml(options) {
         return {
           ...sanitizeHtml.defaults,
-          allowedTags: self.toolbarToAllowedTags(options),
-          allowedAttributes: self.toolbarToAllowedAttributes(options),
+          allowedTags: [...self.toolbarToAllowedTags(options), 'pre'],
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -335,8 +335,11 @@ module.exports = {
       optionsToSanitizeHtml(options) {
         return {
           ...sanitizeHtml.defaults,
-          allowedTags: self.toolbarToAllowedTags(options),
-          allowedAttributes: self.toolbarToAllowedAttributes(options),
+          allowedTags: [...self.toolbarToAllowedTags(options), 'pre'],
+          allowedAttributes: {
+            ...self.toolbarToAllowedAttributes(options),
+            'pre': ['*']
```
boutell, line 341: If a change to allowed attributes is needed it should be added to `toolbarToAllowedAttributes` rather than being merged in afterward.

However allowing all attributes is a security issue, and I also don't understand why allowing all attributes would help with line breaks.

Can you clarify the reasoning behind this code?

boutell (changes requested):

boutell:
Closing due to lack of response.

PR #4372, opened 2023-12-19
Check the new "create" and "archive" permissions

falkodev:
https://linear.app/apostrophecms/issue/PRO-4793/in-core-check-the-new-create-and-archive-permissions-where-appropriate

- replace `edit` by `create` or `archive` where relevant in backend
- `annotate` function adds `_create` and `_archive` for frontend
- frontend checks new properties `_create` and `_archive`


modules/@apostrophecms/permission/index.js
```diff
@@ -175,7 +175,7 @@ module.exports = {
 
             return query;
           }
-        } else if (action === 'edit') {
+        } else if (action === 'edit' || action === 'create' || action === 'archive') {
```
boutell, line 178: This may well be correct, but instead I recommend immediately changing `action` from `create` or `archive` to `edit` right at the top of both `can` and `criteria`, to minimize the need for other changes in those methods.

boutell (changes requested):

boutell (approved):
Makes sense to me now, so go ahead and merge after assuring it does no harm in core, but please also coordinate with Val about his PR.

PR #4388, opened 2024-01-12
PRO-5341: Link HTML attributes for rich text widget 

myovchev:
## Summary

- `linkFields` top configuration property (cascade) can be configured on a project level to add additional schema for rich text Link extension.
- `htmlAttribute: 'attribute-name'` should be set on every link attribute field in order to be functional
- There are no constraints on the types, but some types won't make sense (object, array, etc). The functionality is tested against `string`, `select`, `choices` (with one only option), `boolean`. 
- ~~Schema order and overall all standard schema features are supported, including remote choices.~~ Schema order and dynamic choices are not supported.

Extend `@apostrophecms/rich-text-widget`:
```js
module.exports = {
  linkFields: {
    add: {
      ariaLabel: {
        type: 'string',
        label: 'Aria Label',
        htmlAttribute: 'aria-label'
      }
    }
  }
};
```
~~The field order can be controlled via group `link` - in this case `aria-label` will be shown before the existing `target` field.~~
The "Aria Label" field should be shown in the link management modal. Saving it will result in a link having `aria-label="value"`. The value is preserved after saving and editable on consequent link management.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Adds support for `type` query parameter for page autocomplete. This allows to filter the results by page type. Example: `/api/v1/@apostrophecms/page?autocomplete=something&type=my-page-type`.
 * Add testing for the `float` schema field query builder.
 * Add testing for the `integer` schema field query builder.
+* Add support for link HTML attributes in the rich text widget via configurable fields `linkFields`, extendable on a project level (same as it's done for `fields`). A property `htmlAttribute: true` should be set on every attribute field in order to be used as a link HTML attribute.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Adds support for `type` query parameter for page autocomplete. This allows to filter the results by page type. Example: `/api/v1/@apostrophecms/page?autocomplete=something&type=my-page-type`.
 * Add testing for the `float` schema field query builder.
 * Add testing for the `integer` schema field query builder.
-* Add support for link HTML attributes in the rich text widget via configurable fields `linkFields`, extendable on a project level (same as it's done for `fields`). A property `htmlAttribute: true` should be set on every attribute field in order to be used as a link HTML attribute.
+* Add support for link HTML attributes in the rich text widget via configurable fields `linkFields`, extendable on a project level (same as it's done for `fields`). Add an `htmlAttribute` property to the standard fields that map directly to an HTML attribute, except `href` (see special case below), and set it accordingly, even if it is the same as the field name. Setting `htmlAttribute: 'href'` is not allowed and will throw a schema validation exception (on application boot).
 
 ### Fixes
 
--- a/modules/@apostrophecms/rich-text-widget/index.js
+++ b/modules/@apostrophecms/rich-text-widget/index.js
@@ -73,23 +73,12 @@ module.exports = {
         target: {
           label: 'apostrophe:linkTarget',
           type: 'checkboxes',
-          htmlAttribute: true,
+          htmlAttribute: 'target',
           choices: [
             {
               label: 'apostrophe:openLinkInNewTab',
               value: '_blank'
             }
-          ]
-        }
-      },
-      group: {
-        link: {
-          fields: [
-            'linkTo',
-            ...linkWithType.map(type => `_${type}`),
-            'updateTitle',
-            'href',
-            'target'
           ]
         }
       }
@@ -507,7 +496,7 @@ module.exports = {
               'name',
               ...self.linkSchema
                 .filter(field => field.htmlAttribute)
-                .map(field => field.name)
+                .map(field => field.htmlAttribute)
             ]
           },
           alignLeft: {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Adds support for `type` query parameter for page autocomplete. This allows to filter the results by page type. Example: `/api/v1/@apostrophecms/page?autocomplete=something&type=my-page-type`.
 * Add testing for the `float` schema field query builder.
 * Add testing for the `integer` schema field query builder.
+* Add support for link HTML attributes in the rich text widget via configurable fields `linkFields`, extendable on a project level (same as it's done for `fields`). A property `htmlAttribute: true` should be set on every attribute field in order to be used as a link HTML attribute.
```
boutell, line 10: htmlAttribute should be set to the desired attribute name, not `true`.

"Add an htmlAttribute property to the standard fields that map directly to an HTML attribute, except href (see special case below), and set it accordingly, even if it is the same as the field name."

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -6,6 +6,95 @@ const cheerio = require('cheerio');
...
+          htmlAttribute: true,
+          choices: [
+            {
+              label: 'apostrophe:openLinkInNewTab',
+              value: '_blank'
+            }
+          ]
+        }
+      },
+      group: {
```
boutell, line 85: `group` doesn't really apply to a modal without tabs. It also looks like you already have the right default order here.

If you do need to order things in a schema without tabs, you can use `order` which takes an array of field names. But in this case I'm pretty sure the default behavior is already correct without either `group` or `order`.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -404,7 +505,9 @@ module.exports = {
               'href',
               'id',
               'name',
-              'target'
+              ...self.linkSchema
+                .filter(field => field.htmlAttribute)
+                .map(field => field.name)
```
boutell, line 510: `field.htmlAttribute` should be the actual attribute name (if any). There is no requirement that it match the schema property name, and if sometimes won't because camelCase is recommended for field names and HTML generally is kebab-base.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -675,6 +778,59 @@ module.exports = {
...
+        }
+      },
+      // Compose and register the link schema.
+      composeLinkSchema() {
+        self.linkSchema = self.apos.schema.compose({
+          addFields: self.apos.schema.fieldsToArray(`Links ${self.__meta.name}`, self.linkFields),
+          arrangeFields: self.apos.schema.groupsToArray(self.linkFieldsGroups)
+        }, self);
+        const forbiddenFields = [
+          '_id'
```
boutell, line 818: Actually not necessary here I think. _id would be a weird choice of attribute name for sure, but we're not populating an apostrophe widget, array item or document. However if you found a reason this is definitely needed I trust that.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -675,6 +778,59 @@ module.exports = {
...
+        ];
+        self.linkSchema.forEach((field) => {
+          if (forbiddenFields.includes(field.name)) {
+            throw new Error('Doc type ' + self.name + ': the field name ' + field.name + ' is forbidden');
+          }
+        });
+        self.apos.schema.validate(self.linkSchema, {
+          type: 'doc type',
```
boutell, line 826: You can set type to 'link' here. It's debugging info to provide context for error messages, and "doc type" means an apostrophe document.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -675,6 +778,59 @@ module.exports = {
...
+          subtype: self.__meta.name
```
boutell, line 827: That works.

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -675,6 +778,59 @@ module.exports = {
...
+        });
+        // Don't allow htmlAttribute on href, it's a special case.
+        if (self.linkSchema.find(field => field.name === 'href')?.htmlAttribute) {
+          throw new Error('The link field "href" can not have "htmlAttribute: true"');
```
boutell, line 831: "cannot have an explicit `htmlAttribute`"

modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -675,6 +778,59 @@ module.exports = {
...
+        }
+        self.apos.schema.register('widget', self.__meta.name, self.linkSchema);
```
boutell, line 833: I don't think you need to call this, and it could cause problems if we ever want schema fields at the level of the widget itself.

modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Link.js
```diff
@@ -8,6 +8,17 @@ export default (options) => {
...
         HTMLAttributes: {}
       };
+    },
+    addAttributes() {
+      return {
+        ...this.parent?.(),
+        ...apos.modules['@apostrophecms/rich-text-widget'].linkSchema
+          .filter(field => field.htmlAttribute)
+          .reduce((obj, field) => {
+            obj[field.name] = { default: field.def ?? null };
```
boutell, line 18: field.htmlAttribute is not a boolean, it is the attribute name to be used. Need to be able to pick up old values under that name and map them to the right schema field name and vice versa.

boutell (changes requested):

modules/@apostrophecms/rich-text-widget/index.js
myovchev: `order` does not re-order the schema fields but `filters/columns` etc. The only way to allow control of the field order when extending is a "fake" group as far as I know. 
Adding the defaults then allows fine tuning when extending (done in the testbed). E.g. "put target and my other single choice fields in the end because they don't have "field label" by design and look confusing when mixed with other fields", but I don't have to put the whole list of the default fields on a project level.

I'll remove the default group.

modules/@apostrophecms/rich-text-widget/index.js
myovchev: As far as I know, this is the only way to allow dynamic choices. The schema endpoint finds the field by ID and executes the server method provided with the schema.

I'll remove it and state that dynamic choices are not supported.

modules/@apostrophecms/rich-text-widget/index.js
myovchev: This was added because of the "registration" of the schema. Now that it's removed, it's not needed anymore.

boutell (approved):

PR #4389, opened 2024-01-12
renames archive in delete

ValJed:
[PRO-5388](https://linear.app/apostrophecms/issue/PRO-5388/were-using-archive-and-delete-interchangeably-in-advanced-permissions)

## Summary

See ticket
Also handle `discard draft` for pieces and pages:
* If user can archive but not publish, he can delete draft.
* If user cannot archive he cannot delete draft.
* If user can archive and publish he can delete draft and archive
* If user cannot edit but can archive he can delete draft (doesn't work rn because `findForEditing` checks for `edit` permission).

I also updated the frontend for the pages because we never passed the moduleOptions to the page item context menus where we need to check if we show discard draft etc..

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@
 
 * Fix the `launder` and `finalize` methods of the `float` schema field query builder.
 * Fix the `launder` and `finalize` methods of the `integer` schema field query builder.
-* Rollbacks modifications that changed the delete `permission` name in `archive`.
 
 
 ## 3.61.1 (2023-01-08)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 * Fix the `launder` and `finalize` methods of the `float` schema field query builder.
 * Fix the `launder` and `finalize` methods of the `integer` schema field query builder.
+* Rollbacks modifications that changed the delete `permission` name in `archive`.
 
 
 ## 3.61.1 (2023-01-08)
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -1502,7 +1502,7 @@ module.exports = {
           canCreate: self.apos.permission.can(req, 'create', self.name, 'draft'),
           canEdit: self.apos.permission.can(req, 'edit', self.name, 'draft'),
           canPublish: self.apos.permission.can(req, 'publish', self.name),
-          canArchive: self.apos.permission.can(req, 'archive', self.name)
+          canArchive: self.apos.permission.can(req, 'delete', self.name)
         };
         browserOptions.canLocalize = browserOptions.canEdit &&
           self.options.localized &&
@@ -1873,7 +1873,7 @@ module.exports = {
             self.apos.permission.annotate(query.req, 'create', results);
             self.apos.permission.annotate(query.req, 'edit', results);
             self.apos.permission.annotate(query.req, 'publish', results);
-            self.apos.permission.annotate(query.req, 'archive', results);
+            self.apos.permission.annotate(query.req, 'delete', results);
           }
         },
 
--- a/modules/@apostrophecms/doc-type/ui/apos/logic/AposDocContextMenu.js
+++ b/modules/@apostrophecms/doc-type/ui/apos/logic/AposDocContextMenu.js
@@ -262,7 +262,7 @@ export default {
     },
     canArchive() {
       return (
-        this.context._archive &&
+        this.context._delete &&
         this.context._id &&
         !this.moduleOptions.singleton &&
         !this.context.archived &&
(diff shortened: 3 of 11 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -1502,7 +1502,7 @@ module.exports = {
           canCreate: self.apos.permission.can(req, 'create', self.name, 'draft'),
           canEdit: self.apos.permission.can(req, 'edit', self.name, 'draft'),
           canPublish: self.apos.permission.can(req, 'publish', self.name),
-          canArchive: self.apos.permission.can(req, 'archive', self.name)
+          canArchive: self.apos.permission.can(req, 'delete', self.name)
```
ValJed, line 1505: I suppose we can keep the canArchive method name. But I didn't find where this one is used (there are others canArchive in the frontend but not this one)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 * Fix the `launder` and `finalize` methods of the `float` schema field query builder.
 * Fix the `launder` and `finalize` methods of the `integer` schema field query builder.
+* Rollbacks modifications that changed the delete `permission` name in `archive`.
```
boutell, line 16: If this was never released then we do not need to mention in the changelog when we undo it.

modules/@apostrophecms/doc-type/index.js
boutell: Agreed I don't see it used but we can have it here for bc in case someone else uses it

boutell (changes requested):
Minor note on the changelog. I see it's a draft as well

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -29,8 +29,9 @@
         :disabled="!!relationshipErrors"
         @click="saveRelationship"
       />
+      <!-- Currently we cannot create a document if we can't edit them -->
       <AposButton
-        v-else-if="moduleOptions.canEdit && moduleOptions.showCreate"
+        v-else-if="moduleOptions.canCreate && moduleOptions.canEdit && moduleOptions.showCreate"
```
boutell, line 34: This button has to be available if you `canCreate`, even if you can't edit. Otherwise we are unable to take advantage of our create permission, right?

boutell: (That scenario does not arise in core but it does arise in AP so this logic needs to be looking for canCreate and not canEdit)

boutell (changes requested):

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
ValJed: Yes but currently we cannot create a document if we cannot edit this one.. so there is a missing part somewhere in the core I suppose.

CHANGELOG.md
ValJed: Removed

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
ValJed: I remove the  `canEdit` here but we'll need to fix the fact that we cannot create a piece without the edit permission.

modules/@apostrophecms/doc-type/ui/apos/logic/AposDocContextMenu.js
```diff
@@ -245,6 +245,7 @@ export default {
       if (!this.context._id) {
         return false;
       }
+      // TODO: Should we remove this check since done backend?
       if (!this.canEdit) {
         return false;
       }
```
ValJed, line 251: Should we remove this check since it's done on backend?

modules/@apostrophecms/doc-type/ui/apos/logic/AposDocContextMenu.js
```diff
@@ -245,9 +245,6 @@ export default {
       if (!this.context._id) {
         return false;
       }
-      if (!this.canEdit) {
-        return false;
-      }
```
ValJed, line 250: This check is done on backend now, even if some cases are not working rn.
eg. user can archive but cannot edit, the delete method perform a `findForEditing` that checks if user has the `edit` permission for that document.

modules/@apostrophecms/page/index.js
```diff
@@ -408,6 +408,10 @@ module.exports = {
         const page = await self.findOneForEditing(req, {
           _id
         });
+
+        if (!page) {
+          throw self.apos.error('notfound');
+        }
```
ValJed, line 414: Just throwing the right error here.

boutell (approved):

boutell (approved):

PR #4390, opened 2024-01-16
Pro 4801 is indeterminate

haroun:
## Summary

Remove annotate permission with `create`. If it's a new document, it will not be present in the DB and the annotate `create` won't be available.

Is indeterminate on checkboxes is needed to be able to tell is all checkboxes on the row are checked.

*For example:*

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -1870,6 +1870,7 @@ module.exports = {
           after(results) {
             // In all cases we mark the docs with ._edit and ._publish if
             // the req is permitted to do those things
+            self.apos.permission.annotate(query.req, 'create', results);
             self.apos.permission.annotate(query.req, 'edit', results);
             self.apos.permission.annotate(query.req, 'publish', results);
             self.apos.permission.annotate(query.req, 'delete', results);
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -1870,7 +1870,6 @@ module.exports = {
           after(results) {
             // In all cases we mark the docs with ._edit and ._publish if
             // the req is permitted to do those things
-            self.apos.permission.annotate(query.req, 'create', results);
```
boutell, line 1873: What about pages? For pages, `create` permission on the parent page is requires to create a page.

boutell: Unless that is handled outside of this PR?

boutell (changes requested):

modules/@apostrophecms/doc-type/index.js
haroun: I'm checking with Val.

boutell:
There shouldn't be any harm in keeping this annotation, it'll make it easier if we get around to folders for general purpose use in pieces, but do what works as long as pages behave sensibly

boutell (dismissed):

ValJed (approved):

PR #4394, opened 2024-01-18
Changes support for image upload from `canEdit` to `canCreate`.

falkodev:
https://linear.app/apostrophecms/issue/PRO-5419/upload-input-still-displayed-when-you-cant-create

This is in the Manage view of images -- only in the manage view, because when you edit an image you can also upload an image to change it and this one is not part of the issue.


boutell (approved):
Correct — this is permission to make a new image piece — "replace image" replaces just the attachment within the existing piece, so it doesn't need any modification.

ValJed (approved):

PR #4397, opened 2024-01-19
Archive images: remove menu for users without permission `archive`

falkodev:
https://linear.app/apostrophecms/issue/PRO-5418/user-can-archive-images-without-having-delete-permissions-for-images

Needs to be fixed, as delete is supposed to be required to archive anything, it sounds like something isn’t checking for delete permission.

Might be same logic as hiding input when you can create, just passing the right permissions info to the modal in order to show the right data.


boutell (changes requested):
Hi Anthony, this is probably unnecessary now, see:

https://github.com/apostrophecms/apostrophe/pull/4396/files

I had to make a similar change, but with support for per-doc permissions, because per-doc permissions need to work (at least under the hood) for images. Sorry for the redundancy.

PR #4408, opened 2024-01-25
removes related reverse IDs when deleting a draft document

ValJed:
[PRO-5489](https://linear.app/apostrophecms/issue/PRO-5489/[advanced-permission-3]-deleting-draft-does-not-remove-per-doc)

## Summary

* When deleting a draft document, we remove related reverse IDs of documents having a relation to the deleted one.
* Updates docs.js tests to make them independants.
* Adding two test for the fix (one for delete draft, one for delete unlocalized docs).

Deleting a draft document should delete the related reverse IDs of all document having a relation to the deleted document.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -381,10 +381,12 @@ module.exports = {
   methods(self) {
     return {
       async deleteRelatedReverseId(doc, deleting = false) {
-        const locales = doc.aposLocale && deleting ? [
-          doc.aposLocale.replace(':draft', ':published'),
-          doc.aposLocale.replace(':published', ':draft')
-        ] : [ doc.aposLocale ];
+        const locales = doc.aposLocale && deleting
+          ? [
+            doc.aposLocale.replace(':draft', ':published'),
+            doc.aposLocale.replace(':published', ':draft')
+          ]
+          : [ doc.aposLocale ];
         return self.apos.doc.db.updateMany({
           relatedReverseIds: { $in: [ doc.aposDocId ] },
           aposLocale: { $in: [ ...locales, null ] }
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -330,9 +330,9 @@ module.exports = {
       },
       afterDelete: {
         async deleteRelatedReverseId(req, doc) {
-          // When deleting a draft document, we remove related reverse IDs
-          // of documents having a relation to the deleted one
-          if (doc.aposMode === 'draft') {
+          // When deleting an unlocalized or draft document,
+          // we remove related reverse IDs of documents having a relation to the deleted one
+          if (!doc.aposMode || doc.aposMode === 'draft') {
             await self.deleteRelatedReverseId(doc);
           }
         }
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -328,6 +328,15 @@ module.exports = {
           }
         }
       },
+      afterDelete: {
+        async deleteRelatedReverseId(req, doc) {
+          // When deleting a draft document, we remove related reverse IDs
+          // of documents having a relation to the deleted one
+          if (doc.aposMode === 'draft') {
+            await self.deleteRelatedReverseId(doc);
+          }
+        }
+      },
       afterRescue: {
         async revertDeduplication(req, doc) {
           const $set = await self.getRevertDeduplicationSet(req, doc);
@@ -371,18 +380,21 @@ module.exports = {
 
   methods(self) {
     return {
-      async updateCacheField(req, doc) {
-        const relatedDocsIds = self.getRelatedDocsIds(req, doc);
-
-        // - Remove current doc reference from docs that include it
-        // - Update these docs' cache field
-        await self.apos.doc.db.updateMany({
+      async deleteRelatedReverseId(doc) {
+        return self.apos.doc.db.updateMany({
           relatedReverseIds: { $in: [ doc.aposDocId ] },
           aposLocale: { $in: [ doc.aposLocale, null ] }
         }, {
           $pull: { relatedReverseIds: doc.aposDocId },
           $set: { cacheInvalidatedAt: doc.updatedAt }
         });
+      },
+      async updateCacheField(req, doc) {
+        const relatedDocsIds = self.getRelatedDocsIds(req, doc);
+
+        // - Remove current doc reference from docs that include it
+        // - Update these docs' cache field
+        await this.deleteRelatedReverseId(doc);
 
         if (relatedDocsIds.length) {
           // - Add current doc reference to related docs
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -333,7 +333,7 @@ module.exports = {
           // When deleting an unlocalized or draft document,
           // we remove related reverse IDs of documents having a relation to the deleted one
           if (!doc.aposMode || doc.aposMode === 'draft') {
-            await self.deleteRelatedReverseId(doc);
+            await self.deleteRelatedReverseId(doc, true);
           }
         }
       },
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -328,6 +328,15 @@ module.exports = {
           }
         }
       },
+      afterDelete: {
+        async deleteRelatedReverseId(req, doc) {
+          // When deleting a draft document, we remove related reverse IDs
+          // of documents having a relation to the deleted one
+          if (doc.aposMode === 'draft') {
```
boutell, line 335: Hmm. What about localized: false types?

boutell (changes requested):

modules/@apostrophecms/doc-type/index.js
ValJed: Still a draft PR ahah but `localized: false` documents are never deleted, they can only be archived.

boutell: True right now via the UI, but the delete backend API exists and someone may invoke it.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -328,6 +328,15 @@ module.exports = {
           }
         }
       },
+      afterDelete: {
+        async deleteRelatedReverseId(req, doc) {
+          // When deleting an unlocalized or draft document,
+          // we remove related reverse IDs of documents having a relation to the deleted one
+          if (!doc.aposMode || doc.aposMode === 'draft') {
```
ValJed, line 335: unlocalized docs have no `doc.aposMode`, should be enough? I also added a test for unlocalized pieces to be sure it works.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -371,18 +380,25 @@ module.exports = {
...
-        const relatedDocsIds = self.getRelatedDocsIds(req, doc);
-
-        // - Remove current doc reference from docs that include it
-        // - Update these docs' cache field
-        await self.apos.doc.db.updateMany({
+      async deleteRelatedReverseId(doc, del = false) {
+        const locales = doc.aposLocale && del ? [
+          doc.aposLocale.replace(':draft', ':published'),
+          doc.aposLocale.replace(':published', ':draft')
+        ] : [ doc.aposLocale ];
```
ValJed, line 387: When deleting a document, we want to remove related reverse IDS to draft and published docs.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -371,18 +380,25 @@ module.exports = {
...
           $pull: { relatedReverseIds: doc.aposDocId },
           $set: { cacheInvalidatedAt: doc.updatedAt }
         });
+      },
+      async updateCacheField(req, doc) {
+        const relatedDocsIds = self.getRelatedDocsIds(req, doc);
+
+        // - Remove current doc reference from docs that include it
+        // - Update these docs' cache field
+        await this.deleteRelatedReverseId(doc);
```
ValJed, line 401: Here I kept the behavior exactly as it was.

modules/@apostrophecms/doc/index.js
```diff
@@ -717,7 +717,6 @@ module.exports = {
       // This operation ignores the locale and mode of `req`
       // in favor of the actual document's locale and mode.
       async delete(req, doc, options = {}) {
-        options = options || {};
```
ValJed, line 720: Useless

modules/@apostrophecms/doc/index.js
```diff
@@ -1057,17 +1056,17 @@ module.exports = {
...
           updatedAt: new Date()
         };
-        const result = await self.db.updateOne(criteria, {
+        const { result } = await self.db.updateOne(criteria, {
           $set: {
             advisoryLock: doc.advisoryLock
           }
         });
-        if (!result.result.nModified) {
+        if (!result.nModified) {
```
ValJed, line 1069: I made these tiny changes because I had a flaky test but it doesn't seem to fix it.

test/docs.js
```diff
@@ -2,19 +2,17 @@ const t = require('../test-lib/test.js');
 const assert = require('assert');
 const _ = require('lodash');
 
-let apos;
-
 describe('Docs', function() {
-
+  let apos;
```
ValJed, line 6: Made a bunch of changes to make these tests independants and idempotents.

test/docs.js
```diff
@@ -443,6 +377,103 @@ describe('Docs', function() {
     assert(larryDoc.relatedReverseIds.length === 1);
     assert(larryDoc.relatedReverseIds[0] === 'paul');
     assert(larryDoc.cacheInvalidatedAt.getTime() === response.updatedAt.getTime());
+
+    apos.doc.db.deleteMany({ slug: { $in: [ 'paul', 'carl', 'larry' ] } });
+  });
+
+  it('should remove the related reverse IDs when you delete a draft document', async function () {
```
ValJed, line 384: Testing that deleting a draft properly remove related reverse IDs to draft and published docs.

test/docs.js
```diff
@@ -443,6 +377,103 @@ describe('Docs', function() {
...
+      larryHasRelatedBeforeDelete: true,
+      carlHasRelatedBeforeDelete: true,
+      larryHasRelatedAfterDelete: false,
+      carlHasRelatedAfterDelete: false
+    };
+
+    assert.deepEqual(actual, expected);
+  });
+
+  it('should remove the related reverse IDs when you delete an unlocalized document', async function () {
```
ValJed, line 429: Testing that deleting an unlocalized doc properly removes related reverse IDs.

test/docs.js
```diff
@@ -805,20 +878,25 @@ describe('Docs', function() {
 
   it('should be able to refresh the lock with the same tabId', async function() {
```
ValJed, line 879: This test is flaky I didn't find why, you might have an idea.

test/docs.js
```diff
@@ -805,20 +878,29 @@ describe('Docs', function() {
...
+    await insertPeople(apos);
+    const doc = await apos.doc.db.findOne({ _id: 'carl:en:published' });
+    const wait = (time) => new Promise((resolve) => setTimeout(() => {
+      resolve();
+    }, time));
 
     try {
       await apos.doc.lock(req, doc, 'abc');
+      const locked = await apos.doc.db.findOne({ _id: 'carl:en:published' });
+      await wait(500);
```
ValJed, line 890: I needed to wait here otherwise the test is flaky. If you lock again with the same `updatedAt` the update isn't considered because the data is the same and the test fail.

modules/@apostrophecms/doc-type/index.js
ValJed: Right I added support for doc not having aposMode (aka unlocalized docs)

modules/@apostrophecms/doc-type/index.js
boutell: Yes.

modules/@apostrophecms/doc-type/index.js
boutell: Either this is missing parens so I can follow it better, or it always results in an array containing null in the "else" case...

modules/@apostrophecms/doc/index.js
boutell: OK yes should be equivalent

boutell (changes requested):

modules/@apostrophecms/doc-type/index.js
ValJed: No the deleting option is only passed when we perform a delete draft.
During an update I kept the exact same behavior as before which was just passing `doc.aposLocale` as it is.
if I have `doc.aposLocal` AND the `deleting` option that is passed only when deleting draft or unlocalized doc, we want to remove related reverse ids from published and draft docs.
In the case of the update I wasn't totally sure so I kept the behavior since already working.

boutell (dismissed):
Looks good.

boutell (dismissed):

ValJed:
@boutell just resolved changelog conflict.

boutell (approved):

PR #4413, opened 2024-01-28
PRO-5513: Fix publish after draft

myovchev:
## Summary

Also closes PRO-5151

Publish and move published page just after a (same level, sibling) draft page works as expected. 
> It also fixes eslint warning that annoys me from some time!

Create 3 draft pages on a same tree level. Publish the last one (no errors). Moved it after the 1st one (no errors). Should work for any tree level (all pages above should be the same level).

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/page/index.js
+++ b/modules/@apostrophecms/page/index.js
@@ -988,6 +988,15 @@ database.`);
           page.path = self.apos.util.addSlashIfNeeded(parent.path) + page.aposDocId;
           page.level = parent.level + 1;
           await self.apos.doc.insert(req, page, options);
+          // Prevent a published page from being inserted as a child of a draft page.
+          // In effect when this method is called again from the `afterInsert` event
+          // of the `doc` module. This can happen when we are inserting a page
+          // in req.mode == 'published', which results in insertDrafOf being called
+          // (`page-type` module).
+          if (page.lastPublishedAt && !parent.lastPublishedAt) {
+            await self.unpublish(req, page);
+            throw self.apos.error('forbidden', 'Publish the parent page first.');
+          }
           return page;
         });
       },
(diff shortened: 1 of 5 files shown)


test/pages.js
```diff
@@ -172,9 +172,11 @@ describe('Pages', function() {
       }
     ];
     // Insert draft versions too to match the A3 data model
+    const lastPublishedAt = new Date();
     const draftItems = await apos.doc.db.insertMany(testItems.map(item => ({
       ...item,
       aposLocale: item.aposLocale.replace(':published', ':draft'),
+      lastPublishedAt,
```
myovchev, line 179: Now without this, the old tests fail with "publish parent page first", which is a good sign.

modules/@apostrophecms/page/index.js
```diff
@@ -988,6 +988,15 @@ database.`);
           page.path = self.apos.util.addSlashIfNeeded(parent.path) + page.aposDocId;
           page.level = parent.level + 1;
           await self.apos.doc.insert(req, page, options);
+          // Prevent a published page from being inserted as a child of a draft page.
```
boutell, line 991: This is this check not happening until after the insert? We can tell if this is going to be a published page or not right away based on its mode. Then we would not have to publish and unpublish it, momentarily breaking the rule.

modules/@apostrophecms/page/index.js
```diff
@@ -1252,23 +1276,32 @@ database.`);
       // value. `position` is used to prevent attempts to move after the archive
       // "page."
       async getTarget(req, targetId, position) {
-        const criteria = self.getIdCriteria(self.inferIdLocaleAndMode(req, targetId));
+        // self.inferIdLocaleAndMode (see i18n module)
+        // is mutating the req object. This leads to various issues. This
+        // handler is shared among different routines (publish, insert, etc)
+        // and should NOT alter the request state.
+        const _req = req.clone({});
```
boutell, line 1283: 👍

boutell (changes requested):

PR #4417, opened 2024-02-01
Display a more informative error message when publishing a page

falkodev:
https://linear.app/apostrophecms/issue/PRO-5479/improve-the-experience-for-per-document-editors-who-think-they-can

## How to check

With advanced-permission and given a parent page where the current user has no permission

image

And a child page where the user has the "publish" permission

image

If they try to publish the child page, they get this message (which has not changed in this PR)

image

And after publishing the parent page fails (which is normal because they don't have the publish permission on the parent page), they now see this more meaningful message

image

**Note**
Also checked without advanced-permission where no error message occurs as expected

### DIFF
--- a/modules/@apostrophecms/i18n/i18n/en.json
+++ b/modules/@apostrophecms/i18n/i18n/en.json
@@ -165,7 +165,7 @@
   "errorWhilePublishing": "An Error Occurred While Publishing",
   "errorWhilePublishingDocument": "An error occurred while publishing the document.",
   "errorWhilePublishingParentPage": "An error occurred while publishing a parent page.",
-  "errorWhilePublishingParentPageForbidden": "Because you do not have permission to publish the parent page, we are saving and submitting your draft. Please let your manager know that the parent page needs to be published",
+  "errorWhilePublishingParentPageForbidden": "Because you do not have permission to publish the parent page, we are saving and submitting your draft. Please let your manager know that the parent page needs to be published.",
   "errorWhileRestoring": "An error occurred while restoring the previously published version.",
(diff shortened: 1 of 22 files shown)


modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -165,6 +165,7 @@
   "errorWhilePublishing": "An Error Occurred While Publishing",
   "errorWhilePublishingDocument": "An error occurred while publishing the document.",
   "errorWhilePublishingParentPage": "An error occurred while publishing a parent page.",
+  "errorWhilePublishingParentPageForbidden": "Because you do not have permission to publish the parent page, we are saving and submitting your draft. Please let your manager know that the parent page needs to be published",
```
boutell, line 168: Please add a period at the end.

modules/@apostrophecms/ui/ui/apos/mixins/AposPublishMixin.js
```diff
@@ -65,9 +65,10 @@ export default {
               // Retry now that ancestors are published
               return this.publish(doc);
             } catch (e) {
+              const errorMessage = e.message === 'forbidden' ? 'apostrophe:errorWhilePublishingParentPageForbidden' : e.message;
```
boutell, line 68: Use `e.name`, should also be `forbidden`, while the message could be different if someone takes it upon themselves to throw a more informative one on the server side so that could break one day

boutell (changes requested):
Solid, minor notes

boutell (approved):

boutell (approved):

boutell (approved):

PR #4421, opened 2024-02-06
Remove blocking page slug

falkodev:
https://linear.app/apostrophecms/issue/PRO-5490/crash-due-to-adding-a-slash-in-a-page-slug-field

Tested in dev and prod modes, works correctly: slashes can be added in a slug without problem.


ETLaurent (approved):
Seems like a random issue

PR #4423, opened 2024-02-09
Option "apiFallback" in i18n

falkodev:
Add option `apiFallback` in i18n, forcing `req.locale` when no prefix in request but existing hostname matching the request.

https://linear.app/apostrophecms/issue/PRO-5578/as-a-developer-i-want-to-be-able-to-redirect-all-unprefixed-api

### DIFF
--- a/test/i18n.js
+++ b/test/i18n.js
@@ -477,7 +477,7 @@ describe('no redirection to first locale', function() {
   });
 });
 
-describe('apiFallback option', function() {
+describe.only('apiFallback option', function() {
   this.timeout(t.timeout);
 
   let apos;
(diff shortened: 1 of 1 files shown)


boutell (changes requested):
This does not meet the requirements if the URL falls through all the way to the page route. In that situation it is still supposed to 404.

See my earlier linear comment and implementation suggestion below. Test coverage must address this case as well:

For any given hostname,

If /robots.txt (or any other route except the page route) is accessed with no prefix,

And no locale for that hostname is configured without a prefix,

But one of the locales associated with that hostname has apiFallback: true set,

Then the request should be mapped to that locale,

**Except that if it makes it all the way to the page route it should be considered “not found.”**

The easiest way to achieve that last requirement is probably to set an `aposApiFallback: true` flag on req when you set the locale in the middleware based on apiFallback, so that the page route can spot that and say “thanks, but no thanks.”

PR #4426, opened 2024-02-14
Pro 5355 localization translation

ValJed:
[PRO-5355](https://linear.app/apostrophecms/issue/PRO-5355/as-a-user-i-want-to-automatically-translate-content-when-localizing-it)

## Summary

* Create the new chip for translation
* Adds UI in the localization manager (with and without related docs)
* Handle different errors scenarios (mocked since no provider available)

Try to localize a document or page without related docs, just before to localize you should see this new section:
image

If you check the translate checkbox, you should see an error:
image

Errors for unavailable targets are already working but not functional since no provider right now.

You should be able to do the same with piece with related docs:
image
 

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -45,6 +45,11 @@
   "assetWebpackCacheCleared": "Build cache cleared.",
   "assetWebpackConfigWarning": "⚠️ In the module {{ module }}, your webpack config is incorrect. It must be an object and should contain only the properties: {{ properties }}.",
   "at": "at",
+  "automaticTranslationCheckbox": "Translate text content",
+  "automaticTranslationSettings": "Automatic translation settings",
+  "automaticTranslationSourceErrMsg": "The current locale <strong>{{ source }}</strong> is not suitable for translation. Translation cannot happen.",
```
myovchev, line 50: "Translation will be skipped" probably? Not sure though, native English speakers have to advise.

modules/@apostrophecms/i18n/i18n/en.json
```diff
@@ -45,6 +45,11 @@
...
+  "automaticTranslationTargetErrMsg": "Could not find a suitable language for the locale <strong>{{ targets }}</strong>. Translation will be skipped when localizing.",
+  "automaticTranslationTargetErrMsg_plural": "Could not find a suitable language for the locales <strong>{{ targets }}</strong>. Translation will be skipped when localizing.",
```
myovchev, line 52: What bothers me with those messages (coming from the prototype) is that it's really not clear that the translation will be skipped for the given locales only. At  least this is our plan. Again I think this should be synced with Stu and Tom probably.

modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -370,7 +395,9 @@ export default {
           value: 'relatedDocsOnly',
           label: 'apostrophe:relatedDocsOnly'
         }
-      ]
+      ],
+      translationEnabled: apos.modules['@apostrophecms/translation']?.enabled,
```
myovchev, line 399: I think optional chaining not needed anymore.

myovchev (changes requested):
Tiny stuff.

modules/@apostrophecms/i18n/i18n/en.json
ValJed: Updated based on Tom and Stu feedback.

modules/@apostrophecms/i18n/i18n/en.json
ValJed: Right, I updated that.

modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
ValJed: Removed

myovchev (approved):

PR #4429, opened 2024-02-19
Pro 5643 manager indicator

ValJed:
[PRO-5643](https://linear.app/apostrophecms/issue/PRO-5643/ui-manager-indicators)

## Summary

Handles the new  `Pending Translations` on pieces and pages managers.

⚠️ Must be tested with [this PR](https://github.com/apostrophecms/automatic-translation/pull/1) in `automatic-translation`.

Since the conditions about the meta isn't existing for now you won't see the new indicator.
You can change the condition in `automatic-translation` to see the indicator being showed or you can create a new one.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
+++ b/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
@@ -495,7 +495,6 @@ export default {
     },
     async 'wizard.values.translateContent.data'(value) {
       await this.checkAvailableTranslations(value);
-      console.log('this.wizard.values.translateTargets', this.wizard.values.translateTargets.data);
     },
     selectedLocales() {
       this.updateRelatedDocs();
--- a/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
+++ b/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
@@ -218,7 +218,7 @@
               </div>
               <div v-if="translationEnabled" class="apos-wizard__translation">
                 <p class="apos-wizard__translation-title">
-                  <AposTranslationChip />
+                  <AposTranslationIndicator :size="18" />
                   <span class="apos-wizard__translation-title-text">
                     {{ $t('apostrophe:automaticTranslationSettings') }}
                   </span>
--- a/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
+++ b/modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
@@ -364,7 +364,9 @@ export default {
           toLocales: { data: this.locale ? [ this.locale ] : [] },
           relatedDocSettings: { data: 'localizeNewRelated' },
           relatedDocTypesToLocalize: { data: [] },
-          translateContent: { data: false }
+          translateContent: { data: false },
+          translateTargets: { data: [] },
+          translateProvider: { data: apos.translation.providers[0]?.name || null }
         }
       },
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -38,7 +49,33 @@ import AposCellMixin from 'Modules/@apostrophecms/ui/mixins/AposCellMixin';
...
+      try {
+        return Object.entries(conditions).every(([ property, expected ]) => {
+          const properties = property.split('.');
+
+          const draftProp = properties.reduce((acc, cur) => {
+            if (Object.hasOwn(acc, cur)) {
+              return acc[cur];
+            }
+
+            throw new Error(`Property not found in draft document: ${property}`);
```
ValJed, line 69: throwing here to be sure we won't have false positive.

modules/@apostrophecms/i18n/ui/apos/components/AposI18nLocalize.vue
```diff
@@ -491,8 +493,9 @@ export default {
     'wizard.values.toLocalize.data'() {
       this.updateRelatedDocs();
     },
-    'wizard.values.translateContent.data'(value) {
-      this.checkAvailableTranslations(value);
+    async 'wizard.values.translateContent.data'(value) {
+      await this.checkAvailableTranslations(value);
+      console.log('this.wizard.values.translateTargets', this.wizard.values.translateTargets.data);
```
myovchev, line 498: Leftover or there in purpose?

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
```diff
@@ -30,6 +30,17 @@
         :modifiers="[ 'apos-is-filled', 'apos-is-danger' ]"
       />
     </span>
+    <span
+      v-for="({component, props, if: conditions}, index) in customCellIndicators"
+      :key="index"
```
myovchev, line 35: I would do `component` here instead of the index.

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
myovchev: I think you can import lodash/get and use it `_.get(this.draft, property)`. If the draft prop doesn't exist, lodash will just return undefined, which is fine. 
No problem with loading lodash/get because fields already use it (array retrieve the tab label via lodash.get).

myovchev (changes requested):

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
ValJed: I prefer to avoid lodash when it's not complex stuff most of the time.
What bother you is the use of try catch and error to get out of the loop?

modules/@apostrophecms/ui/ui/apos/components/AposCellLabels.vue
ValJed: What if we have multiple time the same component with different props like label? (Like it's done with `AposLabel`)
Not sure it would happen but might.

myovchev: Yes, makes sense. With different prosp. Ignore.

myovchev (approved):

PR #4443, opened 2024-02-27
Bump he to v1

myovchev:
## Summary

We are using an ancient version without a good reason.


boutell (changes requested):
Please update the changelog.

ValJed (approved):

PR #4452, opened 2024-03-05
Highlight differences when comparing

falkodev:
https://linear.app/apostrophecms/issue/PRO-5465/as-an-editor-i-can-see-the-differences-between-the-two-document

Goes with         https://github.com/apostrophecms/document-versions/pull/46

### DIFF
--- a/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
+++ b/modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
@@ -40,7 +40,7 @@
       <component
         v-show="displayComponent(field)"
         v-model="fieldState[field.name]"
-        :class="{ 'apos-field__wrapper--highlight': meta[field.name]?.['@apostrophecms/schema:highlight'] }"
+        :class="{ 'apos-field__wrapper--highlight': highlight(field.name) }"
         :is="fieldComponentMap[field.type]"
         :following-values="followingValues[field.name]"
         :condition-met="conditionalFields?.if[field.name]"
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -40,6 +40,7 @@
       <component
         v-show="displayComponent(field)"
         v-model="fieldState[field.name]"
+        :class="{ 'apos-field__wrapper--highlight': meta[field.name]?.['@apostrophecms/schema:highlight'] }"
```
haroun, line 43: we should make this line smaller than 90 characters, the linter will complain otherwise. 
Maybe using a computed prop.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -139,4 +141,9 @@ export default {
       }
     }
   }
+
+  :deep(.apos-field__wrapper--highlight > .apos-field) {
+    padding: 10px;
+    background: #FFF6E8;
```
haroun, line 147: IMHO we should create a new variable to store the color in `_theme.scss`

haroun (changes requested):
Also changelog is missing

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
falkodev: Done with a method (since there is a parameter to pass)

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
falkodev: Done

modules/@apostrophecms/ui/ui/apos/scss/global/_theme.scss
```diff
@@ -8,6 +8,7 @@
   --a-warning-dark: #a75c07;
   --a-warning-fade: #ffce0030;
   --a-progress-bg: #2c354d;
+  --a--highlight: #fff6e8;
```
haroun, line 11: why double dash?

falkodev: Well spotted

haroun (approved):

PR #4453, opened 2024-03-06
feat(PRO-5621): fix console error when palette is toggled

itsrogil:
# Fix Console Error on Palette Toggle

## Overview
This PR addresses the console error observed when opening and closing the palette across various projects, as detailed in [PRO-5621](https://linear.app/apostrophecms/issue/PRO-5621/console-error-when-opening-and-closing-palette).

## Issue Description
An error was initially spotted in the [starter-kit-assembly-hospitality](https://github.com/apostrophecms/starter-kit-assembly-hospitality) and has since been identified in multiple repositories, including:

- starter-kit-assembly-essentials
- a3-assembly-boilerplate
- a3-assembly-client-demo
- testbed (upon activating the palette module in app.js)

## Fix Description
The implemented fix resolves the console error by ensuring proper initialization and cleanup processes for the palette module across all affected repositories.

## Testing
The fix has been tested across all mentioned projects, confirming the elimination of the console error when the palette is toggled.

## Related Issue
[PRO-5621](https://linear.app/apostrophecms/issue/PRO-5621/console-error-when-opening-and-closing-palette)

Please review the changes and provide feedback or approval for merging.

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
@@ -382,9 +382,9 @@ export default {
       locale = locale || apos.i18n.locale;
       doc = doc || this.context;
       if ((mode === this.draftMode) && (locale === apos.i18n.locale)) {
-        if ((this.context._id === doc._id) && (!this.urlDiffers(doc._url))) {
+        if ((this.context._id === doc._id) && (doc && !this.urlDiffers(doc._url))) {
           return;
-        } else if (navigate && this.urlDiffers(doc._url)) {
+        } else if (navigate && doc && this.urlDiffers(doc._url)) {
           await this.unlock();
           return window.location.assign(doc._url);
         } else {
@@ -536,7 +536,7 @@ export default {
         }
       });
 
-      if (this.urlDiffers(doc._url)) {
+      if (doc && this.urlDiffers(doc._url)) {
         // Slug changed, change browser URL to reflect the actual url of the doc
         history.replaceState(null, '', doc._url);
       }
@@ -685,12 +685,15 @@ export default {
     },
     urlDiffers(url) {
       // URL might or might not include hostname part
-      url = url.replace(/^https?:\/\/.*?\//, '/');
-      if (url === (window.location.pathname + (window.location.search || ''))) {
-        return false;
-      } else {
-        return true;
+      if (url) {
+        url = url.replace(/^https?:\/\/.*?\//, '/');
+        if (url === (window.location.pathname + (window.location.search || ''))) {
+          return false;
+        } else {
+          return true;
+        }
       }
+      return false
     },
     lockNotAvailable() {
       if (this.contextStack.length) {
--- a/modules/@apostrophecms/util/ui/src/http.js
+++ b/modules/@apostrophecms/util/ui/src/http.js
@@ -136,7 +136,8 @@ export default () => {
           query = {};
         }
         query.aposMode = options.draft ? 'draft' : 'published';
-        url = apos.http.addQueryToUrl(url, query);
+        
+        url = url ? apos.http.addQueryToUrl(url, query) : '/';
       }
     }
 
@@ -147,7 +148,7 @@ export default () => {
     let i;
 
     if (options.qs) {
-      url = apos.http.addQueryToUrl(url, options.qs);
+      url = url ? apos.http.addQueryToUrl(url, options.qs) : '/';
     }
     if (options.busy) {
       if (!busyActive[busyName]) {
@@ -344,6 +345,7 @@ export default () => {
   // URL remain unchanged.
 
   apos.http.addQueryToUrl = function(url, data) {
+    // url = url ? url : '/';
     let hash = '';
     const hashAt = url.indexOf('#');
     if (hashAt !== -1) {
--- a/modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposButton.vue
@@ -316,7 +316,7 @@ export default {
   }
   .apos-button__color-preview__checkerboard {
     z-index: $z-index-base;
-    background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAAMElEQVQ4T2N89uzZfwY8QFJSEp80A+OoAcMiDP7//483HTx//hx/Ohg1gIFx6IcBALl+VXknOCvFAAAAAElFTkSuQmCC');
+    // background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAAMElEQVQ4T2N89uzZfwY8QFJSEp80A+OoAcMiDP7//483HTx//hx/Ohg1gIFx6IcBALl+VXknOCvFAAAAAElFTkSuQmCC');
   }
 
   .apos-button--small {


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -693,7 +693,7 @@ export default {
           return true;
         }
       }
-      return false
+      return false;
```
boutell, line 696: Go with the filter pattern — do `if (!url)...` up front so that most of the function does not have to be indented.

boutell (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
itsrogil: I updated the entire function since I was at it, including your suggestion 👍

boutell (dismissed):
Oh well ain't that purty as a picture

itsrogil:
I only added a commit that removed leftovers of tests and whatnots but I re-requested a review, @boutell just to be on the clear.

itsrogil:
@boutell Nvm, seems it breaks now that I rebased the whole thing on main.

boutell (approved):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -382,9 +382,9 @@ export default {
       locale = locale || apos.i18n.locale;
       doc = doc || this.context;
       if ((mode === this.draftMode) && (locale === apos.i18n.locale)) {
-        if ((this.context._id === doc._id) && (!this.urlDiffers(doc._url))) {
+        if ((this.context._id === doc._id) && (doc && !this.urlDiffers(doc._url))) {
           return;
-        } else if (navigate && this.urlDiffers(doc._url)) {
+        } else if (navigate && doc && this.urlDiffers(doc._url)) {
```
itsrogil, line 387: Remove the doc verification as it's checked before

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -382,9 +382,9 @@ export default {
       locale = locale || apos.i18n.locale;
       doc = doc || this.context;
       if ((mode === this.draftMode) && (locale === apos.i18n.locale)) {
-        if ((this.context._id === doc._id) && (!this.urlDiffers(doc._url))) {
+        if ((this.context._id === doc._id) && (doc && !this.urlDiffers(doc._url))) {
```
haroun, line 385: early return after setting doc if doc is empty

modules/@apostrophecms/util/ui/src/http.js
```diff
@@ -147,7 +148,7 @@ export default () => {
     let i;
 
     if (options.qs) {
-      url = apos.http.addQueryToUrl(url, options.qs);
+      url = url ? apos.http.addQueryToUrl(url, options.qs) : '/';
```
haroun, line 151: we should set the `url` at the beginning of the function to avoid weird edge cases

haroun (changes requested):

haroun (dismissed):

modules/@apostrophecms/util/ui/src/http.js
```diff
@@ -94,6 +94,7 @@ export default () => {
   // on all lean requests, etc.
 
   apos.http.remote = function(method, url, options, callback) {
+    url = url || '/';
```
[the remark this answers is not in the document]
itsrogil: Yep. I can confirm. It gives me the following error:

`AposTiptapTable.vue:8 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'indexOf')` which was tied (I did this code a little while back and couldn't remember exactly).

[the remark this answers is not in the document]
boutell: OK, so we need to detect how that happens and fix it in the caller (as noted in slack)

ETLaurent (changes requested):
See https://apostrophe-inc.slack.com/archives/CS661KRA5/p1711472423084959
We can pair debug it

modules/@apostrophecms/util/ui/src/http.js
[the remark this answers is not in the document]
ETLaurent: @itsrogil this might be what causing the issue.
should we maybe early return if url is not defined?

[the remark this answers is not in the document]
ETLaurent: early returns does nothing good.

but the issue is there, for instance if we set a specific page by default (ex: a page `/palette`), the loaded content will be the the one of `/palette` on any page, when opening the palette:

image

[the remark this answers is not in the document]
ETLaurent: this does the trick, but the error might be caught somewhere and the trace is shit...
would be good to trace the origin and figure out why the url is not defined in the first place

image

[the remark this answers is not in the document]
ETLaurent: comes from `modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue` (onPushContext > refresh)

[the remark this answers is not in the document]
ETLaurent: Ok, palette doc just does not have a `_url`.
I think it's safe to early return before making the request with an undefined url:

in modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue#refresh
image

@boutell wdyt?

btw, the called is not wrapped into a try..catch, weird...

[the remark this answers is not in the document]
ETLaurent: pushed the 2 last commits
cc @itsrogil

ETLaurent (dismissed):

modules/@apostrophecms/schema/ui/apos/components/AposInputColor.vue
```diff
@@ -20,7 +20,7 @@
           >
             <Picker
               v-bind="pickerOptions"
-              :model-value="next"
+              :model-value="pickerValue"
```
boutell, line 23: Unrelated to this PR.

modules/@apostrophecms/util/ui/src/http.js
```diff
@@ -99,6 +99,9 @@ export default () => {
         throw new Error('If you wish to receive a promise from apos.http methods in older browsers you must have a Promise polyfill. If you do not want to provide one, pass a callback instead.');
       }
       return new window.Promise(function(resolve, reject) {
+        if (!url) {
+          return reject(new Error('url is not defined'));
```
boutell, line 103: 👍 Much more comfortable flagging this for easier debugging rather than just tolerating it

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -691,13 +695,14 @@ export default {
       }
     },
     urlDiffers(url) {
-      // URL might or might not include hostname part
-      url = url.replace(/^https?:\/\/.*?\//, '/');
-      if (url === (window.location.pathname + (window.location.search || ''))) {
+      if (!url) {
```
boutell, line 698: WAT... but, scoped like this inside TheAposContextBar I'm OK with it.

boutell (changes requested):

modules/@apostrophecms/schema/ui/apos/components/AposInputColor.vue
ETLaurent: I think that was what @itsrogil and @ValJed found to fix the new warning.
Not directly related to the original ticket, but it still removes a console warning and fixes the color picker

boutell (approved):

ETLaurent (approved):

ETLaurent (approved):

PR #4466, opened 2024-03-15
eslint, including one missing emit

boutell:
just eslint.


modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -190,7 +190,7 @@ export default {
       }
     }
   },
-  emits: [ 'update', 'update:checked' ],
+  emits: [ 'update', 'update:checked', 'change' ],
```
boutell, line 193: This was my only manual change. Everything else is the result of eslint --fix

BoDonkey (dismissed):

BoDonkey (approved):

PR #4468, opened 2024-03-15
PRO-5711 refactor login schema

triskuit:
## Summary
Update the login form to draw its schema from the user piece config instead of being hardcoded, standardizing the form verbiage. [PRO-5711](https://linear.app/apostrophecms/issue/PRO-5711/the-apostrophe-login-form-cannot-be-localized-or-customized-through)

**Acceptance Criteria**
As a developer, you can restate the username field with a different label at project level in project-level configuration of @apostrophecms/user and see that updated label in the login form, not just “manage users”

1. `npm link` to a test project
2. Override the label of the username field in `modules/@apostrophecms/user/index.js` (example below)
3. `npm run dev` => http://localhost:3000/login
4. Your test phrase is displayed in the username field

```
module.exports = {
  fields(self, options) {
    return {
      add: {
        username: {
          type: 'string',
          label: 'THIS IS A TEST',
          required: true,
          following: 'title'
        }
      }
    };
  }
};
```

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/login/index.js
```diff
@@ -909,8 +910,20 @@ module.exports = {
...
-      }
+      },
 
+      getSchema() {
+        return self.apos.user.schema
+          .filter(({ name }) => [ 'username', 'password' ].includes(name))
+          .map(field => ({
+            name: field.name,
+            label: field.label,
+            placeholder: `Enter ${field.label.replace('apostrophe:', '').toLowerCase()}`,
```
triskuit, line 921: We're passing in a phrase which usually gets parsed. I just stripped it but if there's a better way lemme know.

boutell: Hmm. This is messy, isn't it. Our original goal here was to make this placeholder easy to override and localize server side, and now it it's server side, but it's still not possible to localize fully 😄 

So I suggest:

* Add `usernamePlaceholder` and `passwordPlaceholder` options to this module
* Default those to `apostrophe:enterUsername` and `apostrophe:enterPassword` and add the appropriate phrases to `en.json`
* Utilize those options as needed here as part of your massaging of the schema array

boutell (changes requested):

modules/@apostrophecms/login/index.js
```diff
@@ -66,6 +70,7 @@ module.exports = {
     }
   },
   async init(self) {
+    console.log({ opt: self.options });
```
boutell, line 73: Stray log

modules/@apostrophecms/login/index.js
```diff
@@ -909,8 +915,20 @@ module.exports = {
...
-      }
+      },
 
+      getSchema() {
+        return self.apos.user.schema
+          .filter(({ name }) => [ 'username', 'password' ].includes(name))
+          .map(field => ({
+            name: field.name,
+            label: field.label,
+            placeholder: field.placeholder ?? self.options.placeholder[field.name],
```
boutell, line 926: Why the check for an existing placeholder from the user module? We definitely want one on the login form, so I think you can simplify.

boutell (changes requested):

modules/@apostrophecms/login/index.js
triskuit: It feels confusing to be able to override the label from the user module but not the placeholder, yeah? Or am I confusing something?

modules/@apostrophecms/login/index.js
triskuit: 🤦‍♀️

boutell (approved):

boutell (approved):

PR #4478, opened 2024-03-21
PRO-5793 hotfix for rich text styles option bug

boutell:
TL;DR: mutating things, even subproperties of things, causes Vue 3 to consider them to have changed and can lead to unexpected behavior. Methods that return a result usually don't alter their arguments. This is a safe workaround for now


boutell:
Vue 2 didn't watch props as deeply so it didn't trigger on this.

modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -140,6 +140,8 @@ import TableHeader from '@tiptap/extension-table-header';
 import TableRow from '@tiptap/extension-table-row';
 import Placeholder from '@tiptap/extension-placeholder';
 
+import { klona } from 'klona';
```
boutell, line 143: I tried `structuredClone` but it was too strict about not cloning Vue's proxy objects.

BoDonkey (approved):
A PR to study and ponder. LGTM

PR #4480, opened 2024-03-26
PRO-5141: Just bring back what we previously reverted re: Stu's improvements to marks and nodes in the rich text editor

boutell:
I hand-merged a tiny bit of logic where we introduced klona to prevent the Vue 3 crash with style attributes.


BoDonkey (approved):
I'm not sure if this includes changes that Stu made after I found that when there was a style added to a node it would cause the style dropdown to not display the correct label.

boutell:
That's in a separate PR which I'm about to de-draft.

PR #4489, opened 2024-04-04
[A4] update mongodb and connect-mongo dependencies

haroun:
## Summary

Use emulate-mongo-3-driver

1. `npm install && npm test` should be green

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


package.json
```diff
@@ -31,6 +31,7 @@
   "author": "Apostrophe Technologies, Inc.",
   "license": "MIT",
   "dependencies": {
+    "@apostrophecms/emulate-mongo-3-driver": "github:apostrophecms/emulate-mongo-3-driver#pro-5766-emulate-mongo-3",
```
haroun, line 34: Must be switch to version number once released and before merging

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 There is no change in how the `styles` option is configured.
 * `i18n` module now uses the regular `req.redirect` instead of a direct `res.redirect` to ensure redirection, enabling more possibilities for `@apostrophecms/redirect` module
 * Refactors `AposModal` component with composition api to get rid of duplicated code in `AposFocusMixin` and `AposFocus`.
+* `APOS_MONGODB_LOG_LEVEL` has been removed. According to [mongodb documentation](https://github.com/mongodb/node-mongodb-native/blob/main/etc/notes/CHANGES_5.0.0.md#mongoclientoptionslogger-and-mongoclientoptionsloglevel-removed) "Both the logger and the logLevel options had no effect and have been removed."
+* Update `mongodb` to `6.x` and `connect-mongo` to `5.x`. Add `@apostrophecms/emulate-mongo-3-driver` dependency to keep supporting `mongodb@3.x` queries.
```
boutell, line 12: * Why do we still need a dependency on `mongodb` at all?
* We're forcing the upgrade, originally we said this would be optional. But I think forcing it will be OK because it looks like you provided good coverage for the 3.x expectations people are likely to have and we are doing it to address compatibility with a supported driver, which has historically been acceptable. Also we did it this way for the 2.x emulator and nobody had a problem with that. So let's go ahead.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -2703,8 +2703,9 @@ module.exports = {
               subquery.limit(undefined);
               subquery.page(undefined);
               subquery.perPage(undefined);
-              const mongo = await subquery.toMongo();
-              const count = await mongo.count();
+              // TODO is toMongo needed here?
+              // const mongo = await subquery.toMongo();
```
boutell, line 2707: It is, because of the lateCriteria feature which is important for certain edge cases, and also for consistency with any extensions that are created. More importantly, why did you need to change it out — is a feature missing?

modules/@apostrophecms/doc/lib/migrations.js
```diff
@@ -25,7 +25,7 @@ module.exports = (self) => {
 
     addSetDocumentModesMigration() {
       self.apos.migration.add('set-document-modes', async () => {
-        return self.apos.migration.eachDoc(5, async (doc) => {
```
boutell, line 28: Huh, old one.

modules/@apostrophecms/express/index.js
```diff
@@ -554,12 +554,13 @@ module.exports = {
           }
           if (!sessionOptions.store.name) {
             // require from this module's dependencies
-            Store = require('connect-mongo')(expressSession);
+            const MongoStore = require('connect-mongo');
+            sessionOptions.store = MongoStore.create(sessionOptions.store.options);
```
boutell, line 558: Good catch.

package.json
```diff
@@ -91,7 +92,7 @@
     "mini-css-extract-plugin": "^1.6.0",
     "minimatch": "^3.0.4",
     "mkdirp": "^0.5.5",
-    "mongodb": "^3.6.6",
+    "mongodb": "^6.5.0",
```
boutell, line 95: Should probably go away here since the emulator is used?

boutell (changes requested):
Couple things

modules/@apostrophecms/db/index.js
```diff
@@ -96,9 +97,10 @@ module.exports = {
           self.connectionReused = true;
           return;
         }
+        // TODO should be removed since it does nothing according to mongodb documentation
```
haroun, line 100: @boutell should I remove this block?

modules/@apostrophecms/doc-type/index.js
boutell: You can't add a reference to the collection to the cursor objects as they are created?

modules/@apostrophecms/express/index.js
boutell: Per my notes on the A3 PR this can be factored back out of the if statement I think.

boutell (changes requested):

modules/@apostrophecms/express/index.js
haroun: cf https://github.com/apostrophecms/apostrophe/pull/4490#discussion_r1561350790

boutell (approved):

boutell (approved):

PR #4490, opened 2024-04-04
[A3] use @apostrophecms/emulate-mongo-3-driver

haroun:
## Summary

⚠️ this is for A3

Use `@apostrophecms/emulate-mongo-3-driver` in A3

1. `npm install && npm test` must be green

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/doc-type/index.js
```diff
@@ -2682,8 +2682,11 @@ module.exports = {
               subquery.limit(undefined);
               subquery.page(undefined);
               subquery.perPage(undefined);
-              const mongo = await subquery.toMongo();
-              const count = await mongo.count();
```
boutell, line 2686: We should try to find a way to make `count()` work on cursors for better bc. Can you extend cursor objects the same way you extend the others? The methods of a collection that return a cursor could be wrapped to attach a collection property to that cursor so that your added `count()` method can work, unless I am missing something. Possibly the criteria used to create the cursor are hard to access when extending cursor?

boutell (changes requested):

modules/@apostrophecms/doc-type/index.js
haroun: I've restored the `count` method. It was a journey to implement 😄

modules/@apostrophecms/express/index.js
```diff
@@ -554,12 +554,13 @@ module.exports = {
           }
           if (!sessionOptions.store.name) {
             // require from this module's dependencies
-            Store = require('connect-mongo')(expressSession);
+            const MongoStore = require('connect-mongo');
+            sessionOptions.store = MongoStore.create(sessionOptions.store.options);
```
boutell, line 558: Looks like this line is now unnecessarily duplicated and the original version of it would be OK again.

boutell (changes requested):

boutell (approved):

modules/@apostrophecms/express/index.js
haroun: Actually no because connect-mongo contains a breaking change.

> Create the store using `MongoStore.create(options)` instead of `new MongoStore(options)`

Previously `MongoStore` the class expected a reference to the express session object, now it is no longer a class and we must call the `create` method and pass the store options instead.

boutell (dismissed):

ValJed (approved):

PR #4510, opened 2024-04-18
Allow to disable page refresh

ValJed:
## Summary

Allows to disable page refresh on content changed for page types when added to options.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
@@ -495,7 +495,9 @@ export default {
       }
 
       // Check that refresh hasn't been disbled for this page type
-      const contextOptions = apos.modules[this.context.type];
+      const contextOptions = this.context
+        ? apos.modules[this.context.type]
+        : { contentChangedRefresh: true };
       if (contextOptions.contentChangedRefresh) {
         await this.refresh({
           scrollcheck: e.action === 'history'
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
@@ -493,11 +493,10 @@ export default {
           });
         }
       }
-      const refreshOptions = {
-        refresh: true
-      };
-      apos.bus.$emit('apos-refreshing', refreshOptions);
-      if (refreshOptions.refresh) {
+
+      // Check that refresh hasn't been disbled for this page type
+      const contextOptions = apos.modules[this.context.type];
+      if (contextOptions.contentChangedRefresh) {
         await this.refresh({
           scrollcheck: e.action === 'history'
         });
--- a/modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
+++ b/modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
@@ -55,6 +55,11 @@ export default {
     serverError: {
       type: Object,
       required: false
+    },
+
+    noBlurEmit: {
+      type: Boolean,
+      default: false
     }
   },
   data () {
@@ -122,7 +127,7 @@ export default {
       }
     },
     focus(value) {
-      if (!value) {
+      if (!this.noBlurEmit && !value) {
         this.validateAndEmit();
       }
     }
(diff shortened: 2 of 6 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -493,11 +493,10 @@ export default {
...
         }
       }
-      const refreshOptions = {
-        refresh: true
-      };
-      apos.bus.$emit('apos-refreshing', refreshOptions);
-      if (refreshOptions.refresh) {
+
+      // Check that refresh hasn't been disbled for this page type
+      const contextOptions = apos.modules[this.context.type];
```
ValJed, line 498: `this.context` always exist right?

boutell (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
ValJed: Alright fixed!

modules/@apostrophecms/ui/ui/apos/utils/index.js
```diff
@@ -1,9 +1,15 @@
...
+    return new Promise((resolve, reject) => {
+      clearTimeout(timer);
+      timer = setTimeout(() => {
+        try {
+          resolve(func.apply(this, args));
+        } catch (err) {
+          reject(err);
+        }
+      }, timeout);
+    });
```
ValJed, line 13: @boutell Also made this modification, it allows to await a debounced function.
It works for an async function and will await it to be resolved.

boutell (changes requested):

modules/@apostrophecms/ui/ui/apos/utils/index.js
```diff
@@ -1,9 +1,15 @@
...
   return (...args) => {
-    clearTimeout(timer);
-    timer = setTimeout(() => {
-      func.apply(this, args);
-    }, timeout);
+    return new Promise((resolve, reject) => {
+      clearTimeout(timer);
+      timer = setTimeout(() => {
+        try {
+          resolve(func.apply(this, args));
```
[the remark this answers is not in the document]
ValJed: The only purpose of this change is to be able to await the debounced function and to get its result without changing the behavior of existing ones.
I agree for the try catch I didn't realized it had to await.
But the `resolve(func.apply(this, args))` works perfectly even if `func` is async.
I updated the code to make it work in case of errors too.

modules/@apostrophecms/schema/ui/apos/mixins/AposInputMixin.js
```diff
@@ -55,6 +55,11 @@ export default {
     serverError: {
       type: Object,
       required: false
+    },
+
+    noBlurEmit: {
+      type: Boolean,
+      default: false
```
ValJed, line 62: New change to allow to cancel blur emit because it brings in the multisite dashboard case.

ValJed:
@boutell 
* Added throttle has asked
* Switched the file to commonjs to be testable
* Added tests for debounce and throttle
* Added a prop to `AposInputMixin` to be able to cancel the blur event emit (needed for multisite dashboard and globally any use out of `AposSchema`).

boutell (changes requested):
I updated the tests to verify that the function invocations not only start in the right order, the end in the right order: they should never overlap. We have to wait for a current invocation's promise to settle before we start another one. Making this test pass should complete the work.

ValJed:
@boutell Tests are red and I think you updated the wrong test. You updated the debounce test, there is no overlap issue with debounce since only the last call is effective.
You wanted to update the throttle test I suppose since you mentioned overlapping.

And I think the throttle test is ok because it already tests overlapping. For example in this sequence:
```javacscript
      throttledNormal(1);
      throttledNormal(2);
      throttledNormal(3);
      throttledNormal(4);

      await wait(100);
      await throttledNormal(5);

      // res -> [1, 2, 5]
```
The 1 and 2 are called (because 2 arrives after the throttle delay). But the 3 and 4 aren't and finally the 5 is ok because of the delay.
As we can see there is no overlapping

ValJed:
@boutell `debounce` function improved to make tests pass.

PR #4513, opened 2024-04-19
poc of live access on demand (not reactive, but up to date on demand)

boutell:
This works great from a top level field, and seo fields are top level fields. Any changes in the fields are available immediately when the button is clicked.

What's left from here is figuring out the best place to put the buttons themselves in the code. I haven't seen the design for this. We could certainly make them custom fields and it would definitely work that way.

Screenshot from 2024-04-19 10-32-25


boutell:
Closing because @myovchev already incorporated the technique.

PR #4519, opened 2024-04-29
hotfix 4.2.1: drag and drop in page tree

boutell:
Just a hotfix release


modules/@apostrophecms/ui/ui/apos/components/AposTreeRows.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <draggable
-    v-bind="dragOptions"
```
boutell, line 3: This is all previously approved code from Greg. Tests pass.

BoDonkey (approved):

PR #4521, opened 2024-04-30
PRO-5822: documents autocrop on image replace

myovchev:
## Summary

Fixes broken, previously cropped, images, when an attachment is `replaced` in the Media Manager. The change ensures that the replaced image will be autocropped in all relations that it's used (when appropriate). 

Replace an image in Media Manager. Ensure that widgets using that image and having configured aspect ratio will be re-cropped and not broken on the page view. 

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/image/index.js
+++ b/modules/@apostrophecms/image/index.js
@@ -122,6 +122,18 @@ module.exports = {
       }
     }
   },
+  handlers(self) {
+    return {
+      beforeUpdate: {
+        async autoCropRelations(req, piece, options) {
+          if (piece.aposMode !== 'draft') {
+            return;
+          }
+          await self.updateImageCropRelationships(req, piece);
+        }
+      }
+    };
+  },
   commands(self) {
     return {
       remove: [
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/image/index.js
```diff
@@ -122,6 +122,19 @@ module.exports = {
       }
     }
   },
+  handlers(self) {
+    return {
+      beforeUpdate: {
+        // Ensure the crop fields are updated on publishing an image
+        async autoCropRelations(req, piece, options) {
+          if (piece.aposMode !== 'published') {
```
boutell, line 130: Hmm. Does the logic wind up adjusting references from the drafts too? (Images are autopublished but do have draft and published modes under the hood)

modules/@apostrophecms/image/index.js
```diff
@@ -457,7 +470,156 @@ module.exports = {
           name: 'dummy',
           width: 0
         }).name;
+      },
+      async updateImageCropRelationships(req, piece) {
+        if (!piece.relatedReverseIds?.length) {
+          return;
+        }
+        if (piece.__prevAttachmentId === piece.attachment?._id) {
```
boutell, line 478: One `_` is convention.

modules/@apostrophecms/image/index.js
```diff
@@ -457,7 +470,156 @@ module.exports = {
...
+          return;
+        }
+        const corppedIndex = {};
```
boutell, line 481: corpped -> cropped

modules/@apostrophecms/image/index.js
```diff
@@ -457,7 +470,156 @@ module.exports = {
...
+          await self.updateImageCropsForRelationship(req, docId, piece, corppedIndex);
+        }
+      },
+      // This handler operates on all documents of a given aposDocId. The `piece`
+      // argument is the image piece that has been updated.
+      // - Auto re-crop the image, using the same width/height ratio if the
+      // image has been cropped.
+      // - Remove any existing focal point data.
+      //
+      // `corppedIndex` is used to avoid re-cropping the same image when updating multiple
```
boutell, line 492: corpped -> cropped etc.

modules/@apostrophecms/image/index.js
```diff
@@ -457,7 +470,156 @@ module.exports = {
...
+          aposDocId
+        }).toArray();
+        const changeSets = dbDocs.flatMap(doc => getDocRelations(doc, piece));
+        for (const changeSet of changeSets) {
+          try {
+            const cropFields = await autocrop(changeSet.image, changeSet.cropFields, corppedIndex);
+            const $set = {
+              [changeSet.docDotPath]: cropFields
+            };
+            self.logDebug(req, 'replace-autocrop', {
```
boutell, line 505: Nice

test/images.js
```diff
@@ -1,9 +1,11 @@
 const t = require('../test-lib/test.js');
-const assert = require('assert');
-const fs = require('fs');
+const assert = require('assert/strict');
+const fs = require('fs-extra');
```
boutell, line 3: Do we need this dependency, modern fs has a lot of cool toys:

https://nodejs.org/api/fs.html#fsrmpath-options-callback

(Available at least back to node 18)

boutell (changes requested):
This is definitely the way! A few minor things and one question about drafts (even though there is autopublish, images do have a draft mode under the hood)

modules/@apostrophecms/image/index.js
myovchev: My thoughts - "relations" should be valid published version of a document. Even if the original document is a draft, it should reference a published doc (image piece in this case). 

We update all "mode" versions of the referenced document. E.g. a page using the image as a widget will update cropped data for "draft", "publised" and "previous" if crop data is available in the document.

modules/@apostrophecms/image/index.js
myovchev: A typo.

modules/@apostrophecms/image/index.js
```diff
@@ -457,7 +470,160 @@ module.exports = {
...
+          return results;
+        }
+
+        async function autocrop(image, oldFields, croppedIndex) {
+          let crop = {
+            ...oldFields,
+            x: null,
+            y: null
+          };
+          if (oldFields.width) {
```
boutell, line 591: I think you got this bit from the autocrop async function in the autocrop route; could it be factored to a method and called by both? If it's too different don't worry about it.

modules/@apostrophecms/image/index.js
boutell: OK. As long as a user who has set up a crop in a draft page that's never been published, let's say, will not find a broken image on their page when they use "replace" on that image then I'm fine.

boutell (changes requested):

modules/@apostrophecms/image/index.js
myovchev: This is the case. Once an image is "published", our handler will search for all documents that reference it, no matter what mode are they.

modules/@apostrophecms/image/index.js
myovchev: It's not too different, but needs some logic changes.

boutell (approved):

PR #4530, opened 2024-05-09
Pro 5937 move peer page

haroun:
## Summary

Recognize when the slug of a new child already contains the new parent's slug and not double it

1. `npx mocha test/pages.js` must be green.
2. the [Checks](https://github.com/apostrophecms/apostrophe/pull/4530/checks) tab is green

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ Thanks to Michelin for contributing this feature.
 
 * Improves `debounce` function to handle async properly (waiting for previous async call to finish before to trigger new one).
 * Adds the `copyOfId` property to the `edit()` method while still allowing the entire `copyOf` object for BC
+* When moving a page, recognize when the slug of a new child
+already contains the new parent's slug and not double it.
+For example, given we have two pages as children of the home page, page A and page B.
+Page A and page B are siblings.
+Page A has the slug `/peer` and page B has the slug `/peer/page`.
+Now we want page B to be the child of page A.
+We will now end up with page B slug as `/peer/page` and not `/peer/peer/page` as before.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,46 +33,8 @@ error on the page. A "required" field in a widget, for instance, once again bloc
 
 ### Fixes
 
-* Rich Text editor properly unsets marks on heading close
-
-### Changes
-
-* When moving a page, recognize when the slug of a new child
-already contains the new parent's slug and not double it.
-For example, given we have two pages as children of the home page, page A and page B.
-Page A and page B are siblings.
-Page A has the slug `/peer` and page B has the slug `/peer/page`.
-Now we want page B to be the child of page A.
-We will now end up with page B slug as `/peer/page` and not `/peer/peer/page` as before.
-=======
-
-### Changes
-
-* Improves widget tabs for the hidden entries, improves UX when validation errors are present in non-focused tabs.
-
-### Fixes
-
-* Rich Text editor properly unsets marks on heading close.
-* Widget client side schema validation.
-
-## 4.3.2 (2024-05-18)
-
-### Fixes
-
-* Corrects a regression introduced in version 4.3.0 that broke the validation of widget modals, resulting in a confusing
-error on the page. A "required" field in a widget, for instance, once again blocks the save operation properly.
-
-### Changes
-
-* Improves widget tab UI for the hidden entries, improves UX when validation errors are present in non-focused tabs.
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/page/index.js
```diff
@@ -1276,11 +1276,11 @@ database.`);
             if (parent._id !== oldParent._id) {
               const matchOldParentSlugPrefix = new RegExp('^' + self.apos.util.regExpQuote(self.apos.util.addSlashIfNeeded(oldParent.slug)));
               if (moved.slug.match(matchOldParentSlugPrefix)) {
-                let slugStem = parent.slug;
-                if (slugStem !== '/') {
-                  slugStem += '/';
-                }
-                moved.slug = moved.slug.replace(matchOldParentSlugPrefix, self.apos.util.addSlashIfNeeded(parent.slug));
+                const relativeSlug = path.relative(parent.slug, moved.slug);
```
boutell, line 1279: Not suitable for page slugs because on some platforms it may return a platform-dependent path separator.

test/pages.js
```diff
@@ -457,6 +457,79 @@ describe('Pages', function() {
     assert.strictEqual(page.path, `${homeId.replace(':en:published', '')}/another-parent/parent/sibling`);
   });
 
+  it('moving peer /parent/peer-page into /parent should ends with /parent/peer-page', async function() {
```
boutell, line 460: Please add coverage of the more common case as well, e.g. where the peers have slugs `/a` and `/b` and the outcome should be `/a/b`. Looking quickly over the existing tests I am not convinced we have that covered right now.

modules/@apostrophecms/page/index.js
boutell: I know right now our documentation says native Windows is not supported, but it's not best practice to use `path` contrary to its intended purpose, and the way things are going with feedback from customers I think we will have to ease up and support native Node.js on Windows soon so let's not dig the hole deeper on that

boutell (changes requested):
I might be missing something but I still don't see this test:

/foo and /bar are peers
Move /bar under /foo
Should wind up with the slug /foo/bar

Just to cover that we didn't cause any regression in the more typical case.

modules/@apostrophecms/page/index.js
```diff
@@ -1276,11 +1276,14 @@ database.`);
...
-                if (slugStem !== '/') {
-                  slugStem += '/';
-                }
-                moved.slug = moved.slug.replace(matchOldParentSlugPrefix, self.apos.util.addSlashIfNeeded(parent.slug));
+                const movedSlugCandidate = moved.slug
+                  .split('/')
+                  .slice(0, -1)
+                  .join('/');
+
+                moved.slug = parent.slug.endsWith(movedSlugCandidate)
```
boutell, line 1284: Also I think this could be confused in the following scenario:

These are peers:

```
/foo
/foobar
```

(Note no second slash)

And the second one gets moved under the first.

That should result in:

/foo/foobar

(Because no slash)

Tests should cover this case.

boutell: This is in addition to the concern about testing plain old:

/foo
/bar

To make sure that when /bar is moved under /foo we still get /foo/bar in that typical case.

boutell (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,35 @@
 
 ## UNRELEASED
 
+### Changes
+
+* Improves widget tabs for the hidden entries, improves UX when validation errors are present in non-focused tabs.
```
boutell, line 7: Unrelated stuff in changelog?

CHANGELOG.md
```diff
@@ -15,6 +44,35 @@ Page A and page B are siblings.
...
+
+* Improves widget tab UI for the hidden entries, improves UX when validation errors are present in non-focused tabs.
+
+## 4.3.1 (2024-05-17)
+
+### Fixes
+
+* Databases containing documents that no longer correspond to any module no longer cause the migration that adds missing mode properties
+to fail (an issue introduced in version 4.2.0). Databases with no such "orphaned" documents were not affected.
+>>>>>>> main
```
boutell, line 75: Conflict not fully resolved?

boutell (changes requested):
All looks great, except something funky happened to the changelog.

boutell (approved):

PR #4535, opened 2024-05-13
Solves development multiple vue instances being loaded with npm links

ValJed:
Adds function to set vue alias targetting a unique instance based on where it's installed, for development purpose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Thanks to Michelin for contributing this feature.
 
 ### Fixes
 * Do not show widget editor tabs when the developer hasn't created any groups.
+* Adds function to set vue alias targetting a unique instance based on where it's installed, for development purpose.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Thanks to Michelin for contributing this feature.
 
 ### Fixes
 * Do not show widget editor tabs when the developer hasn't created any groups.
-* Adds function to set vue alias targetting a unique instance based on where it's installed, for development purpose.
+* Adds function to set vue alias targeting a unique instance based on where it's installed, for development purpose.
 
 ### Changes
 
--- a/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
+++ b/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const path = require('path');
 const merge = require('webpack-merge').merge;
 const scss = require('./webpack.scss');
@@ -64,7 +65,7 @@ module.exports = ({
     resolve: {
       extensions: [ '.*', '.js', '.vue', '.json' ],
       alias: {
-        vue$: '@vue/runtime-dom',
+        vue$: getVueAlias(),
         // resolve apostrophe modules
         Modules: path.resolve(modulesDir)
       },
@@ -86,3 +87,16 @@ module.exports = ({
 
   return merge(config, ...tasks);
 };
+
+function getVueAlias() {
+  if (!process.env.APOS_DEV) {
+    return '@vue/runtime-dom';
+  }
+
+  const vueLocalPath = path.resolve(__dirname, '../../../../../../node_modules/@vue/runtime-dom');
+  const vueLocallyInstalled = fs.existsSync(vueLocalPath);
+
+  return vueLocallyInstalled
+    ? vueLocalPath
+    : path.join(process.cwd(), 'node_modules/@vue/runtime-dom');
+}
--- a/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
+++ b/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
@@ -93,7 +93,7 @@ function getVueAlias() {
     return '@vue/runtime-dom';
   }
 
-  const vueProjectLevelPath = path.join(process.cwd(), 'node_modules/@vue/runtime-dom');
+  const vueProjectLevelPath = path.resolve('./node_modules/@vue/runtime-dom');
   const vueProjectLevelInstalled = fs.existsSync(vueProjectLevelPath);
 
   return vueProjectLevelInstalled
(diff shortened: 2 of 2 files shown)


modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -86,3 +87,16 @@ module.exports = ({
...
+  if (!process.env.APOS_DEV) {
+    return '@vue/runtime-dom';
+  }
+
+  const vueProjectLevelPath = path.join(process.cwd(), 'node_modules/@vue/runtime-dom');
+  const vueProjectLevelInstalled = fs.existsSync(vueProjectLevelPath);
+
+  return vueProjectLevelInstalled
+    ? vueProjectLevelPath
+    : path.resolve(__dirname, '../../../../../../node_modules/@vue/runtime-dom');
```
ETLaurent, line 101:
```suggestion
    : path.resolve(./node_modules/@vue/runtime-dom');
```

this should work, and be safe if somehow we move `webpack.config.js`

ETLaurent (changes requested):

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -86,3 +87,16 @@ module.exports = ({
 
   return merge(config, ...tasks);
 };
+
+function getVueAlias() {
+  if (!process.env.APOS_DEV) {
+    return '@vue/runtime-dom';
+  }
+
+  const vueProjectLevelPath = path.join(process.cwd(), 'node_modules/@vue/runtime-dom');
```
ETLaurent, line 96:
```suggestion
  const vueProjectLevelPath = path.resolve('./node_modules/@vue/runtime-dom');
```

☝️ does it work with npm links?

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -64,7 +65,7 @@ module.exports = ({
     resolve: {
       extensions: [ '.*', '.js', '.vue', '.json' ],
       alias: {
-        vue$: '@vue/runtime-dom',
+        vue$: getVueAlias(),
```
ETLaurent, line 68:
```suggestion
        vue$: getVueAlias(config),
```

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -86,3 +87,16 @@ module.exports = ({
 
   return merge(config, ...tasks);
 };
+
+function getVueAlias() {
+  if (!process.env.APOS_DEV) {
+    return '@vue/runtime-dom';
+  }
```
ETLaurent, line 94:
```suggestion
function getVueAlias(config) {
  if (config.mode !== 'development') {
    return '@vue/runtime-dom';
  }
```

this uses `NODE_ENV`, not `APOS_DEV`

ETLaurent (changes requested):

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
ValJed: It won't work because for example if my project is called apos-wp, the path that you provide will give:
`/home/user/Documents/workspace/apo/a3-wp/node_modules/@vue/runtime-dom`
While mine will give
`/home/user/Documents/workspace/apo/apostrophe/node_modules/@vue/runtime-dom`

I tried with a path to the `apostrophe` module:
`path.resolve('./apostrophe/node_modules/@vue/runtime-dom')`

but unfortunately it doesn't work with links...

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
ValJed: I took this line, it should work because it target the current running project `node_modules` folder.

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
ValJed: We're already in the config definition, don't see the point here.

boutell:
Why did this work with Vue 2?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Thanks to Michelin for contributing this feature.
 
 ### Fixes
 * Do not show widget editor tabs when the developer hasn't created any groups.
+* Adds function to set vue alias targetting a unique instance based on where it's installed, for development purpose.
```
ETLaurent, line 17:
```suggestion
* Adds function to set vue alias targeting a unique instance based on where it's installed, for development purpose.
```

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -22,6 +23,7 @@ module.exports = ({
     )
   );
 
+  const mode = process.env.NODE_ENV || 'development';
```
ETLaurent, line 26: wouldn't it be safer to keep it in the function?

ETLaurent (changes requested):
small typo

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
ValJed: It is still in the function.. Just moved it out of the config const to pass it to the `getVueAlias` function.

ValJed:
@boutell I could not tell exactly but vue 3 seems to work quite differently than vue 2.
It does not like to have multiple instances of vue loaded.
For example if it's installed in a module npm linked in my project + at the root project and loaded twice in the frontend it just breaks so we need to make sure in dev that only one instance is loaded.

modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
ETLaurent: ok, my bad!

ETLaurent (approved):

ValJed:
@boutell good points, I'm not sure it would since pnpm has a specific modules structures.
I'll check with @myovchev and test that case.

myovchev (approved):
Approving as we need an urgent solution. The last time I did a core work, I wasn't able to rebuild the testbed project with linked in any possible way core (including a fresh clone in node_modules). 
I'm a bit concerned about pnpm in development. We should probably have a check and fallback to the old behavior (which is supposed to work). I'm talking only about running correctly with pnpm, not developing a linked core under pnpm.

PR #4539, opened 2024-05-15
release 3.66.0

boutell:
release only

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apostrophe",
-  "version": "3.64.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "apostrophe",
-  "version": "3.64.0",
+  "version": "3.66.0",
```
boutell, line 3: 3.65.0 was a hotfix release, the code from it did reach main if you're wondering, just not the version number.

BoDonkey (approved):

PR #4537, opened 2024-05-14
fix a tiptap bug when a pasted text takes the style of the first configured text style element

ETLaurent:
## Summary

> When pasting content from your clipboard to our rich text editor, tiptap goes through the process of forcing your paste into appropriate markup based on your RT config. This often ends with wrapping something in a span. If the RT config has a span without a class defined this will appear as expected, but if the config does not have a plain span the span will get converted to the first available span configuration, which can be an unexpected stylization.

**Acceptance criteria**

- If a RT configuration has spans, ensure an unclassed span is added to the configuration.
- Do not add this unclassed span style to the Inline Styles menu
- Highlighting the unclassed span will not show classed spans as the active style. It will accurately reflect that the user has made no selection from the inline style menu.

## What are the specific steps to test this change?

See ticket 5922

## What kind of change does this PR introduce?
*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

## Make sure the PR fulfills these requirements:

- [x] It includes a) the existing issue ID being resolved, b) a convincing reason for adding this feature, or c) a clear description of the bug it resolves
- [x] The changelog is updated
- [ ] Related documentation has been updated
- [ ] Related tests have been updated

If adding a new feature without an already open issue, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

PR #4544, opened 2024-05-16
when pasting content into RT don't force fit a valid class

stuartromanek:
## Summary

An alternate fix to PRO-5922

In our current provision for ensuring only configured classes remain in tact when pasting content into a RT editor we convert a valid tag with an invalid class to the valid tag with the first-found valid class of its tag type. This is causing unexpected styling, often with spans, where any span pasted is now converted to (what we encourage people to configure) is a stylized mark. 

This change removes that first-found default class behavior and allows the tag to be returned **without** any class, which I think is the more correct behavior (now). This gets us a similar effect to #4537 but more broadly across tags and without adding hacky hidden elements to our rich text config.

@ETLaurent 

*(Check at least one)*

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Classes.js
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Classes.js
@@ -64,6 +64,7 @@ export default (options) => {
                   });
                 }
               } else {
+                console.log('toggleMark');
                 commands.toggleMark('textStyle', { class: options.class });
               }
             }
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Classes.js
```diff
@@ -64,6 +64,7 @@ export default (options) => {
                   });
                 }
               } else {
+                console.log('toggleMark');
```
boutell, line 67: stray log

boutell (changes requested):
It looks like this would accept invalid classes, which we shouldn't do either. I recognize it may be a while before we can chat about it.

stuartromanek:
> It looks like this would accept invalid classes, which we shouldn't do either. I recognize it may be a while before we can chat about it.

invalid classes are filtered out here https://github.com/apostrophecms/apostrophe/blob/2ed7bf0e1e0b651630a678c1dd60b5ffe64efe6c/modules/%40apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Classes.js#L93

PR #4550, opened 2024-05-20
Pro 6049 tab indicator

boutell:
Hopefully this time github will be kind enough to show that most of this code is already merged! I didn't change anything, just fighting with github.


myovchev (approved):

myovchev:
Looks like it's the missing test attributes only indeed.

PR #4554, opened 2024-05-27
PRO-6110: Improve menu, fix spinner, allow string replace value

myovchev:
## Summary

New Features:
- Expose Context Menu owned method for re-calculation of the content position
- Field Meta components of `slug` and `string` types can now fire `@replace-field-value` events with text value, which will replace the current field value
- `rows` prop for controlling textarea rows in `AposInputString`

Improvements
- `AposSpinner` now respects the colors for `heavy` weight mode and also accepts second, "light" color in this mode. Props JSDoc blocks are added.
- `AposContextMenu` now respects the `menuOffset` component property.

- Spinner color can be controlled when "weight: heavy` prop is sent.
- Passing `menuOffset` to `AposContextMenu`  modifies the content position offset.
- Emitting  `replace-field-value` from within e Field Meta component replaces the current `string` or `slug` field.

*(Check at least one)*

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
```diff
@@ -118,7 +118,8 @@ const dropdownContentStyle = ref({});
 const arrowEl = ref();
 
 defineExpose({
-  hide
+  hide,
+  setDropdownPosition
```
ValJed, line 122: The fact that you need to expose and reuse this method `setDropDownPosition` inside you implementation of `AposContext` is weird.
This should be run automatically when the context menu is opened.

From what I see in `seo-assistant` you try to manager the menu opening with `isModalVisible` and a `v-if`. I think this is where the mistake is. The menu opening should be an internal behavior.
Basically it generates the button, and when this button is clicked it opens the menu and compute the position.
It allows you to handle logic by providing some events:
* `@open`
* `@close`
* `@item-clicked`

ValJed (changes requested):
Feedback on `AposContextMenu`

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
myovchev: Menu is auto closed on cclick. This behaviour is deep into the component and it's children and can't be easily changed. We need to stay opened and show yet another modal. What we do is allow the slot content to be controlled from outside. Changing the slot content requires position recalculations, which is a more elegant solution.

ValJed (dismissed):

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
myovchev: Nope, the general use case uses it internally to position the content initially and to listen for viewport changes (resize).

modules/@apostrophecms/ui/ui/apos/components/AposSpinner.vue
```diff
@@ -57,10 +57,25 @@
...
       type: String,
       default: '--a-primary'
     },
+    /**
+     * In `heavy` mode, this would be the "lighter" color.
+     * Ignored in the normal `weight` mode.
+     */
+    colorLight: {
+      type: String,
+      default: '#cca9ff'
```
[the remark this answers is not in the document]
myovchev: I just used the old values. If you are confident that this corresponds to it I'll do it. You might want look at the other hardcoded color, I'll mark it for you in the code.

modules/@apostrophecms/ui/ui/apos/components/AposSpinner.vue
```diff
@@ -74,6 +89,14 @@ module.exports = {
 
       return 'color: var(--a-primary);';
     },
+    stopColorLight() {
+      const color = this.colorLight || this.color;
+      return color?.startsWith('--') ? `var(${color})` : color;
+    },
+    stopColorDark() {
+      const color = this.color === '--a-primary' ? '#6516dd' : this.color;
```
myovchev, line 97: @stuartromanek This is the other "old" color I'm directly using, does it have a CSS var analogue?

stuartromanek (approved):

PR #4555, opened 2024-05-29
fixes selectAll of relationship items

ValJed:
[PRO-6071](https://linear.app/apostrophecms/issue/PRO-6071/the-select-all-checkbox-doesnt-work-when-trying-to-give-a-user)

## Summary

Fix select all broken for all relationships + for manually published relationships.

Update code to remove the `updateCheckedDocs` watcher. When checking a box it use a proxy to properly updated `checked` and `checkedDocs` together. And we have proper methods to set, add, remove checked docs (that will updated both).
The reason behind this change is code predictability improvement, it's easier to understand what code does, also to avoid side effects which can be really hard to debug.
Watchers are great but the often can be avoided with more straightforward implementations.
Also the bug we had here was linked to the watcher not being triggered when doing a `checked.push`. We have now the right methods to use from the mixins.

See [this PR](https://github.com/apostrophecms/doc-template-library/pull/60) too (that probably had bug because using `this.checked.push` which is not watched).

You can now properly select relationships, docs or pages, the AposSlatList is updated and doc saved. Also it fixes relationships of non localized documents like users.

Cypress tests running [here](https://github.com/apostrophecms/testbed/actions/runs/9368290458) 🟢 

- [X] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
+++ b/modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
@@ -111,27 +111,32 @@ export default {
           doc._fields = this.subfields[doc._id];
         }
       }
-    },
-    checked() {
-      this.updateCheckedDocs();
     }
   },
   methods: {
-    addCheckedDoc(docOdId) {
-      if (docOdId._id) {
-        this.checked.push(docOdId._id);
-        this.checkedDocs.push(docOdId);
-      } else if (typeof docOdId === 'string') {
-        const found = this.items.find(item => item._id === docOdId);
-        if (found) {
-          this.checked.push(found._id);
-          this.checkedDocs.push(found);
-        }
+    addCheckedDoc(docOrId) {
+      const [ docId, doc ] = docOrId._id
+        ? [ docOrId._id, docOrId ]
+        : [ docOrId, this.items.find(item => item._id === docOrId) ];
+
+      if (!doc) {
+        return;
       }
+
+      this.checked = [ ...this.checked, docId ];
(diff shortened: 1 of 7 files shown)


modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -108,12 +111,40 @@ export default {
           doc._fields = this.subfields[doc._id];
         }
       }
-    },
-    checked() {
-      this.updateCheckedDocs();
```
ValJed, line 113: This method was overriden in AposPagesManager, it's not needed anymore. It's also unused in our modules.
Wondering if it's considered a BC

modules/@apostrophecms/modal/ui/apos/mixins/AposDocsManagerMixin.js
```diff
@@ -122,20 +153,19 @@ export default {
...
-          if (relationshipsMaxedOrUnpublished) {
+        this.items.forEach((item) => {
+          const notPublished = this.manuallyPublished && !item.lastPublishedAt;
+          if (this.relationshipField && (this.maxReached() || notPublished)) {
             return;
           }
 
-          this.checked.push(item._id);
+          this.addCheckedDoc(item);
         });
```
ValJed, line 169: The main fix is in this `selectAll` method. Before manually published doc were impossible to check. 
And `checkedDocs` as no updated because the `checked.push` wasn't triggering the watcher..

boutell (approved):

PR #4569, opened 2024-06-06
Pro 6209 vue3 satan bug

ValJed:
[PRO-6209](https://linear.app/apostrophecms/issue/PRO-6209/production-build-array-editing-still-has-a-bug)

## Summary

Temporary but simple fix to avoid vue 3 bug race condition that happens when inserting a node and not finding the parent `insertBefore` vue 3 method having `parent` param `null`.
In our case happened in array editor when triggering a field error before to delete the element.

Add a basic item in an array with a required field, make the file showing the error, then make the error disappear and finally delete the item, you shouldn't have a big obscure console error.

Cypress tests running [here](https://github.com/apostrophecms/testbed/actions/runs/9403570776) 🟠 

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


boutell (approved):
Haa I like it simple and robust hopefully

PR #4570, opened 2024-06-07
Pro 6140 no def in convert

haroun:
## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/lib/addFieldTypes.js
+++ b/modules/@apostrophecms/schema/lib/addFieldTypes.js
@@ -96,7 +96,7 @@ module.exports = (self) => {
   self.addFieldType({
     name: 'string',
     convert(req, field, data, destination) {
-      destination[field.name] = self.apos.launder.string(data[field.name], destination[field.name] || undefined);
+      destination[field.name] = self.apos.launder.string(data[field.name]);
       destination[field.name] = checkStringLength(destination[field.name], field.min, field.max);
       // If field is required but empty (and client side didn't catch that)
       // This is new and until now if JS client side failed, then it would
@@ -205,7 +205,7 @@ module.exports = (self) => {
   self.addFieldType({
     name: 'boolean',
     convert: function (req, field, data, destination) {
-      destination[field.name] = self.apos.launder.boolean(data[field.name], destination[field.name] || undefined);
+      destination[field.name] = self.apos.launder.boolean(data[field.name]);
     },
     isEmpty: function (field, value) {
       return !value && value !== false;
@@ -256,7 +256,7 @@ module.exports = (self) => {
   self.addFieldType({
     name: 'color',
     async convert(req, field, data, destination) {
-      destination[field.name] = self.apos.launder.string(data[field.name], destination[field.name] || undefined);
+      destination[field.name] = self.apos.launder.string(data[field.name]);
 
       if (field.required && (_.isUndefined(destination[field.name]) || !destination[field.name].toString().length)) {
         throw self.apos.error('required');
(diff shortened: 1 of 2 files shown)


haroun:
@boutell I've run the Cypress tests (https://github.com/apostrophecms/testbed/actions/runs/9418474402), it's green.
But we kind of change the behavior of the `convert` method.
I haven't checked what we say in the documentation yet.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -96,7 +96,7 @@ module.exports = (self) => {
   self.addFieldType({
     name: 'string',
     convert(req, field, data, destination) {
-      destination[field.name] = self.apos.launder.string(data[field.name], field.def);
+      destination[field.name] = self.apos.launder.string(data[field.name], destination[field.name] || undefined);
```
boutell, line 99: I see we need some nuance here. For the basic form input fields, we don't want to go all the way back to undefined, because the empty string is a legitimate value for a string.

So we don't want undefined here, in fact we don't need an `||` clause at all. But we still want to get rid of passing `field.def` here because it is not our job (it is the job of `newInstance` and similar code).

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -165,7 +165,7 @@ module.exports = (self) => {
     convert (req, field, data, destination) {
       const options = self.getSlugFieldOptions(field, data);
 
-      destination[field.name] = self.apos.util.slugify(self.apos.launder.string(data[field.name], field.def), options);
+      destination[field.name] = self.apos.util.slugify(self.apos.launder.string(data[field.name]), options);
```
boutell, line 168: Looks good.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -205,7 +205,7 @@ module.exports = (self) => {
   self.addFieldType({
     name: 'boolean',
     convert: function (req, field, data, destination) {
-      destination[field.name] = self.apos.launder.boolean(data[field.name], field.def);
```
boutell, line 208: We just want:

```
destination[field.name] = self.apos.launder.boolean(data[field.name]);
```

`false` is a legitimate value and the destination's current value should not be consulted in convert.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -256,7 +256,7 @@ module.exports = (self) => {
   self.addFieldType({
     name: 'color',
     async convert(req, field, data, destination) {
-      destination[field.name] = self.apos.launder.string(data[field.name], field.def);
+      destination[field.name] = self.apos.launder.string(data[field.name], destination[field.name] || undefined);
```
boutell, line 259: In this case, `undefined` is a sensible value. But you shouldn't need `destination[field.name]` here.

boutell (changes requested):
I'm sorry I didn't see the need to specifically address simple form input field types like string and boolean and range differently. Specifically, they should never be set back to undefined, because they have a sensible falsy value (empty string, or `false`, or `0`, etc). We don't want that, we just want to leave `field.def` out of the convert call because it's not our job, it's the job of `newInstance` and similar code to handle `def` when something is actually new. And only when it is new.

I am not sure how checking the existing value in `destination` came into play here.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -698,7 +698,7 @@ module.exports = (self) => {
     name: 'time',
     vueComponent: 'AposInputString',
     async convert(req, field, data, destination) {
-      destination[field.name] = self.apos.launder.time(data[field.name], field.def);
+      destination[field.name] = self.apos.launder.time(data[field.name], destination[field.name] || undefined);
```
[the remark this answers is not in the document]
boutell: Yes that was a bad choice on my part long ago in the launder module. Ideally we will need to explicitly accept blank, null and undefined and keep the value blank. Unfortunately we can't rely on launder to do it (unless we bump the major version of launder to change its policy).

boutell (approved):

PR #4576, opened 2024-06-12
PRO-6050 fix all deprecation warnings that can reasonably be fixed at this time without bc issues

boutell:
With this change:

```
rm -rf node_modules
rm package-lock.json
npm install --omit=dev
```

... Yields only one deprecation warning about `querystring` not being ideal in new code, and it has been traced to AWS SDK v2 in uploadfs. Upgrading from that library, which is still supported with security and bug fixes, would cause bc breaks for anyone passing an AWS.Credentials object directly to uploadfs so it is not acceptable at least to do without more thought.


boutell:
The tests are going to fail until `uploadfs` 1.22.4 gets published later today.

test/pieces.js
```diff
@@ -1053,7 +1053,9 @@ describe('Pieces', function() {
     assert(response.choices._articles);
     assert(response.choices._articles[0].label === 'First Article');
     // an _id
-    assert(response.choices._articles[0].value.match(/^c/));
+    console.log(response.choices._articles);
+    console.log(response.choices._articles[0].value);
```
ValJed, line 1057: Forgotten logs?

ValJed (changes requested):
LGTM, just some missed logs.
Do we have a PR to do it in other modules? I know we use `cuid` in a lot of places.
Also, might be a good idea to run cypress tests ?

boutell:
Val is out, his concern was addressed, and the failing tests also fail in `main`, they are an unrelated problem that can't be reproduced locally 🤔 So I'm merging this.

PR #4595, opened 2024-06-19
initial commit

BoDonkey:
## Summary

This PR makes a color picker tool available for the rich-text-widget toolbar. This tool uses the same picker that is used in the `color` schema field and has the same swatches. Additionally, this tool can be configured at the project or per-area level with the same options (`colorPicker` and `format`) that can be passed to the `color` schema field. Closes PRO-6126.

*For example:*
1. At the project or per-area level add the `color` tool to the rich-text-widget `toolbar` configuration.
2. Optionally add `colorPicker` and `format` configuration properties. These should be passed at the same level as the `toolbar` and `styles` configuration properties. For a full list of configuration properties see the [`color` schema field documentation](https://docs.apostrophecms.org/reference/field-types/color.html#options).
3. Check that you can add colors from the picker square or swatches.
4. Check that the expected swatches can be passed from the project or per-area level.
5. Check that the button label changes color correctly.
6. Check that selecting text with color added changes the label color.
7. Check with the devTools that the color value added to the style element is the correct format.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/rich-text-widget/ui/apos/tiptap-extensions/Color.js
```diff
@@ -0,0 +1,5 @@
+// imports the tiptap extension from node_modules
```
boutell, line 1: Interesting that we do this but don't extend it in any way, there are tiptap extensions we just import directly I think, we don't map one to one to `tiptap-extensions` for everything. But this provides a nice point for future extension.

boutell (approved):
Good stuff

stuartromanek (dismissed):
i am biased, but i was asked

stuartromanek (approved):

PR #4599, opened 2024-06-24
Pro 6252 fix context menu position

ValJed:
[PRO-6252](https://linear.app/apostrophecms/issue/PRO-6252/page-context-menu-is-badly-displayed)

## Summary

Fixes display of context menu by removing unnecessary offsets.
Update Context menu to accept array of offset or single number.

Where the offset have been removed the `AposContextMenu` is a good position.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
+++ b/modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
@@ -31,7 +31,6 @@
           :button="draftButton"
           :menu="draftMenu"
           :disabled="hasCustomUi || isUnpublished"
-          menu-offset="13, 10"
           menu-placement="bottom-end"
           @item-clicked="switchDraftMode"
         />
--- a/modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
@@ -116,9 +116,7 @@ const dropdown = ref();
 const dropdownContent = ref();
 const dropdownContentStyle = ref({});
 const arrowEl = ref();
-const menuOffset = ref(
-  Array.isArray(props.menuOffset) ? props.menuOffset : [ props.menuOffset, 0 ]
-);
+const menuOffset = getMenuOffset();
 
 defineExpose({
   hide,
@@ -176,6 +174,13 @@ onBeforeUnmount(() => {
   apos.bus.$off('widget-focus', hide);
 });
 
+function getMenuOffset() {
+  return {
+    mainAxis: Array.isArray(props.menuOffset) ? props.menuOffset[0] : props.menuOffset,
+    crossAxis: Array.isArray(props.menuOffset) ? (props.menuOffset[1] ?? 0) : 0
+  };
+}
+
 function hideWhenOtherOpen(id) {
   if (menuId.value !== id) {
     hide();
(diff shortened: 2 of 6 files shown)


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
```diff
@@ -31,7 +31,6 @@
           :button="draftButton"
           :menu="draftMenu"
           :disabled="hasCustomUi || isUnpublished"
-          menu-offset="13, 10"
```
ValJed, line 34: I removed all of them because they were messing up the positions (it's nicely computed and should not need this).

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
```diff
@@ -87,7 +87,7 @@ const props = defineProps({
     default: 'bottom'
   },
   menuOffset: {
-    type: [ Number, String ],
+    type: [ Number, Array ],
```
ValJed, line 90: Allows to pass a number liker before, or an array with two values:
```javascript
:menu-offset="15"
:menu-offset="[15, 10]"
```

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
stuartromanek: This UI is misplaced. We are setting `menu-placement="bottom-end"` which should align the popup underneath the element ('bottom', working) and at the end of the x-axis (not working)
image

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
```diff
@@ -116,6 +116,9 @@ const dropdown = ref();
 const dropdownContent = ref();
 const dropdownContentStyle = ref({});
 const arrowEl = ref();
+const menuOffset = ref(
+  Array.isArray(props.menuOffset) ? props.menuOffset : [ props.menuOffset, 0 ]
```
stuartromanek, line 120: This doesn't seem like enough. If they pass `[10]` then their element goes all the way to the left edge of the screen.

If they pass `[10]` the resulting offset should become `[10, 0]`

stuartromanek (changes requested):

modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextTitle.vue
ValJed: It's actually the center of the button that triggers it some I'm not shocked.
image

ValJed: Also `bottom-end` means that the arrow is positioned `bottom-end` which is the case here.

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
stuartromanek: This case straight up breaks the UI but is a valid value, so it needs to be handled

ValJed: Done

stuartromanek (approved):

PR #4630, opened 2024-07-11
Pro 6293 updates modal store for locales

ValJed:
[PRO-6293](https://linear.app/apostrophecms/issue/PRO-6293/update-the-pinia-modal-store-and-vue-to-pass-modal-data-to-each-editor)
[PRO-6294](https://linear.app/apostrophecms/issue/PRO-6294/create-a-new-getter-method-that-makes-it-possible-to-get-the-active)

## Summary

Sorry the PR is large and is for two tickets..
In my defense the last one is just few lines of codes and was needed to bind things together and make them working.

See [Tech Design](https://github.com/apostrophecms/tech-designs/tree/main/switch-locales-in-doc-editor).
* Should see the locale switcher of editor modal for pieces and pages that are `localized` (not displayed for user and other non localized docs). Should also work on `autopublished` docs.
* You should see in the locale switcher already localized docs in other locales (like for pages in the normal locale switcher). 
* You can switch to localized or non existing docs, depending, it will update existing doc or create a new one (with the same ID of the initial doc editor).
* Context menu options should be available and work, they should not redirect to the document locale.
* locale switcher is available when creating pieces, also for archived docs (and will switch between docs whenever archived or not).
* When switching locale when doc has changes it ask if you want to discard and switch or save and switch, you can save draft of current doc before switching.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/admin-bar/ui/apos/components/TheAposContextBar.vue
```diff
@@ -498,7 +498,7 @@ export default {
       const contextOptions = this.context
         ? apos.modules[this.context.type]
         : { contentChangedRefresh: true };
-      if (contextOptions.contentChangedRefresh) {
+      if (!e.isLocalizing && contextOptions.contentChangedRefresh) {
```
ValJed, line 501: When localizing a doc (different url than current page) we don't want the event listener to rewrite the url...
Don't really know why we do that in first instance.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -784,6 +786,10 @@ module.exports = {
 
         await self.apos.schema.convert(req, schema, input, doc);
 
+        if (options.createId) {
+          doc.aposDocId = options.createId;
+        }
```
ValJed, line 791: We can now specify an `aposDocId` when creating a document. It allows us to create a doc with the ID of the original. We don't want to localize since it's not a copy of an existing doc.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -3,6 +3,7 @@
     class="apos-doc-editor"
     :modal="modal"
     :modal-title="modalTitle"
+    :modal-data="modalData"
```
ValJed, line 6: Prop comes from `TheAposModals.vue` we need it here to get the current modal locale.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -191,12 +198,14 @@ export default {
       readOnly: false,
       restoreOnly: false,
       saveMenu: null,
-      generation: 0
+      generation: 0,
+      isLocalizing: false,
+      currentId: this.docId
```
ValJed, line 203: This new currentId, allows us to switch it when switching the locale (passed to `AposSchema` for ex), we still keep `docId` as a reference to the original doc id.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -342,6 +351,9 @@ export default {
         pref = null;
       }
       return pref;
+    },
+    showLocalePicker() {
+      return this.moduleOptions.localized !== false && !this.modalData.hasContextLocale;
```
ValJed, line 356: If the module isn't localized or we are already in a localized version of a doc we do not show the locale picker.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -375,6 +387,21 @@ export default {
       type: this.$t(this.moduleOptions.label)
     };
     if (this.docId) {
+      await this.instantiateExistingDoc();
+    } else if (this.copyOfId) {
+      this.instantiateCopiedDoc();
+    } else {
+      await this.$nextTick();
+      await this.instantiateNewInstance();
+    }
```
ValJed, line 396: Messy code here has been abstracted.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -431,37 +459,44 @@ export default {
...
+        await this[action](saveOpts);
+      } else {
+        this.triggerValidation = false;
+        await apos.notify('apostrophe:resolveErrorsBeforeSaving', {
+          type: 'warning',
+          icon: 'alert-circle-icon',
+          dismiss: true
+        });
+        this.focusNextError();
+      }
```
ValJed, line 499: I updated this `saveHandler` to make it awaitable. otherwise when saving and switching locale and switch happened before the save...

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -502,8 +537,14 @@ export default {
...
           };
+          // TODO: Is this block even useful since published is fetched after loadDoc?
           if (this.published) {
-            this.changed = detectDocChange(this.schema, this.original, this.published, { differences: true });
+            this.changed = detectDocChange(
+              this.schema,
+              this.original,
+              this.published,
+              { differences: true }
+            );
```
ValJed, line 547: Linting.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -602,6 +648,8 @@ export default {
         }
         if (this.copyOfId) {
           body._copyingId = this.copyOfId;
+        } else if (this.isLocalizing && this.docId) {
+          body._createId = this.docId.split(':')[0];
```
ValJed, line 652: New body data to give the `aposDocId` we want to create the doc with.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -619,7 +667,8 @@ export default {
         }
         apos.bus.$emit('content-changed', {
           doc,
-          action: (requestMethod === apos.http.put) ? 'update' : 'insert'
+          action: (requestMethod === apos.http.put) ? 'update' : 'insert',
+          isLocalizing: this.isLocalizing
```
ValJed, line 671: I need to pass this new data to all docs `content-changed` emitted events to avoid `TheAposContextBar.vue` updating the window locale....

modules/@apostrophecms/i18n/ui/src/index.js
```diff
@@ -0,0 +1,3 @@
+export default () => {
+  apos.getActiveLocale = () => apos.modal?.getActiveLocale?.() || apos.locale;
+};
```
ValJed, line 3: This method is used for logged in and logged out users. so needs to work in all contexts.

modules/@apostrophecms/modal/ui/apos/components/TheAposModals.vue
```diff
@@ -5,58 +5,68 @@
       :is="modal.componentName"
       v-for="modal in store.stack"
       :key="modal.id"
-      :modal-id="modal.id"
+      :modal-data="modal"
```
ValJed, line 8: Only real change

modules/@apostrophecms/modal/ui/apos/components/TheAposModals.vue
```diff
@@ -5,58 +5,68 @@
...
       @modal-result="store.setModalResult(modal.id, $event)"
     />
   </div>
 </template>
```
ValJed, line 13: Updated the component to use `setup () {}` method instead of the `script setup` because it's a root component and we don't have the name of the component with the script setup.

modules/@apostrophecms/util/ui/src/http.js
```diff
@@ -115,10 +115,10 @@ export default () => {
       return callback(new Error('url is not defined'));
     }
 
-    if (apos.prefix && options.prefix !== false) {
+    if (apos.sitePrefix && options.prefix !== false) {
```
haroun, line 118: Not sure about the impact of this change.
We can have setups with 
* multiple locales but no prefix (DNS only)
* multiple locales some with prefix and some without
* multiple locales all with prefix

Can you check in all these contexts that the change is ok?

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -284,17 +293,10 @@ export default {
...
-          key: 'apostrophe:editType',
-          type: this.$t(this.moduleOptions.label)
-        };
-      } else {
-        return {
-          key: 'apostrophe:newDocType',
-          type: this.$t(this.moduleOptions.label)
-        };
-      }
+      return {
```
haroun, line 296: We now have the same title when we create a new piece and edit an existing document?

haroun (changes requested):

modules/@apostrophecms/modal/ui/apos/components/AposModalConfirm.vue
```diff
@@ -3,7 +3,7 @@
     :modal="modal"
     class="apos-confirm"
     :class="{ 'apos-confirm--tiny': options.tiny }"
-    v-on="mode !== 'alert' ? { esc: cancel } : null"
+    v-on="mode !== 'alert' ? { esc: close } : null"
```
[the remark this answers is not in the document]
ValJed: Because in the new modal I need three results possible:
* True
* False
* Null (when the user  clicks on the cross..)
Before we had only two steps. Since null is false it won't break any existing behavior.
So the new close method return null instead of false.

modules/@apostrophecms/piece-type/ui/apos/components/AposDocsManager.vue
```diff
@@ -486,8 +491,14 @@ export default {
     },
 
     async onContentChanged({ doc, action }) {
-      await this.getPieces();
-      this.getAllPiecesTotal();
+      if (
+        !doc ||
```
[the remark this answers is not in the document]
ValJed: Yeah thats the point, for batch operations we keep the intended behavior.

modules/@apostrophecms/page/ui/apos/logic/AposPagesManager.js
```diff
@@ -164,6 +169,15 @@ export default {
         this.create();
       }
     },
+    onContentChanged({ doc }) {
+      if (
+        !doc ||
```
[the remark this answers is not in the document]
ValJed: Same here, batch operations keep the intended behavior.

modules/@apostrophecms/util/ui/src/http.js
```diff
@@ -153,9 +153,16 @@ export default () => {
...
-    }
+    qat = url.indexOf('?');
+    const qs = {
+      ...qat !== -1 && apos.http.parseQuery(url.substring(qat)),
+      ...options.qs || {},
+      ...options.qs?.aposLocale
+        ? { aposLocale: options.qs.aposLocale }
+        : { aposLocale: apos.getActiveLocale() }
+    };
+    url = apos.http.addQueryToUrl(url, qs);
```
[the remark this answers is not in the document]
ValJed: This is part of the tech design.
It should be since when not in a localization mode we simply pass the current locale which does not change the behavior.

modules/@apostrophecms/util/ui/src/http.js
ValJed: The purpose of using `sitePrefix` is that it does not contain the locale in the prefix.
Otherwise it conflicts backend side with the one passed through `aposLocale`.

modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
ValJed: Hum nope the opposite, if there is a `currentId` means we're editing otherwise creating.

ETLaurent (approved):

PR #4639, opened 2024-07-13
PRO-4659, big uploads: works! needs lots of cleanup

boutell:
draft for now because it needs a lot of cleanup

see matching PR in testbed which will eventually be a polite way to test

I might need a way to defer the real work of the route so we also avoid the proxy timeout while copying all these chunks back out of s3


boutell:
You can change `bigUploadChunkSize` to a smaller number to guarantee the file gets split into chunks and reassembed successfully. In practice fewer chunks are better so 4MB is a nice figure that is also safe for any reasonable proxy setup that allows uploads at all.

PR #4653, opened 2024-07-22
fix search request timeouts by adding a default projection

ETLaurent:
See #4652


ETLaurent:
Cypress run: https://github.com/apostrophecms/testbed/pull/274

ETLaurent:
The automatic-translation test that failed is actually flaky and already fails sometime on `main`. I.e: not related to the hotfix made in this PR.

Issue #4505, opened 2024-04-18
tooltip for the Update Button


harshmehta813:
## The problem to solve

Is your feature request related to a problem you are experiencing that is *not* a bug? Please describe.

> When an admin or editor has made some changes and they hover over the `PREVIEW` button, they see the tooltip, but the `UPDATE` button does not have any tooltip.

## Proposed solution

Describe the solution you'd like

> Can add the tooltip for the `UPDATE` button the same as `PREVIEW`

![image](https://github.com/apostrophecms/apostrophe/assets/87087018/4f4085dc-74a2-46b8-8ad2-67b8a9cf885d)

![image](https://github.com/apostrophecms/apostrophe/assets/87087018/2b70bb91-3288-44e9-acf5-0d849c7db715)

stuartromanek:
Hi @harshmehta813 .. what text do you think would be helpful here? Would it be clearer if the button label was something like 'Publish Updates' ?

harshmehta813:
Hi @stuartromanek 
apologies for the late reply.
"Publish your changes to make them live" should be more clear to the users.

PR #4660, opened 2024-07-28
added tooltip for update button

gkumar9891:
Enhancement #4505


BoDonkey:
Hi,
Thank you for your work on this. I've reviewed the publishTooltip method, and I think we can simplify the logic to make it more concise and readable. Instead of nesting your conditionals that way, you can combine them:
` if (this.canPublish && this.context.lastPublishedAt && !this.hasBeenPublishedThisPageload) `
Can you also update the CHANGELOG, please. I'm going to ask another colleague to look at this to make sure they agree with the text of the tooltip.
Thanks!

PR #4661, opened 2024-07-29
if we cannot parse a mongodb URL ourselves just give it to the driver

boutell:
Just as it says.

I tested this by throwing an error instead of setting `parsed` and it did allow the URL to fall through successfully to the mongodb driver. (On a Mac that could cause a failure because of localhost not resolving to ip4 etc., which is why we parse the mongodb URI at all. On Linux it was easier to test.)

I didn't test this with an actual replica set but passing the URL as-is to the driver is pretty much all we can do here.


BoDonkey (approved):
LGTM

PR #4670, opened 2024-08-06
Pro 6387 images errors

ValJed:
[PRO-6387](https://linear.app/apostrophecms/issue/PRO-6387/[import-export]-images-not-displayed-after-import-in)
[PRO-6420](https://linear.app/apostrophecms/issue/PRO-6420/convert-call-made-by-importer-should-be-altered-not-to-actually-fetch)

## Summary

Rework in core to allow running `convert` method on data from DB.
To do so, add a method `simulateRelationshipsFromStorage` that does the opposite than `prepareForStorage` partially.
It adds a property to relationships that can be converted.

Also add an option to convert to be able to not check if relationships already exist. In the case of imports we cannot verify the order of insertion, especially when a document is a relation to different documents of the same import.

Update convert methods for different fields where needed like: `area` (as well as the sanitize method that call convert), `object`, `array`, `relationship`.

Add tests for method `simulateRelationshipsFromStorage` as well as the new convert option. Test relationships in areas, objects, arrays, or relationship fields.

Tests 🟢 

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Add a locale switcher in pieces and pages editor modals. This is available for localized documents only, and allows you to switch between locales for the same document.
   The locale can be switche at only one level, meaning that sub documents of a document that already switched locale will not be able to switch locale itself.
 * Adds visual focus states and keyboard handlers for engaging with areas and widgets in-context
+* Adds method `simulateRelationshipsFromStorage` method in schema module, for generate relationships in documents from database, doing the opposite than `prepareForStorage`.
+* Adds new option `fetchRelationships` to fields `convert` to avoid to fetch relationships and to remove data when not found. 
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,11 @@
 * Add a locale switcher in pieces and pages editor modals. This is available for localized documents only, and allows you to switch between locales for the same document.
   The locale can be switche at only one level, meaning that sub documents of a document that already switched locale will not be able to switch locale itself.
 * Adds visual focus states and keyboard handlers for engaging with areas and widgets in-context
-* Adds method `simulateRelationshipsFromStorage` method in schema module, for generate relationships in documents from database, doing the opposite than `prepareForStorage`.
-* Adds new option `fetchRelationships` to fields `convert` to avoid to fetch relationships and to remove data when not found. 
+* Adds method `simulateRelationshipsFromStorage` method in schema module. 
+This method populates the relationship field with just enough information to allow convert to accept it. It does not fully fetch the related documents. It does the opposite of prepareForStorage.
+* A new options object has been added to the convert method. 
+Setting the `fetchRelationships` option to false will prevent convert from actually fetching relationships to check which related documents currently exist. 
+The shape of the relationship field is still validated.
 
 ### Changes
 
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -785,7 +785,7 @@ module.exports = {
           };
         }
         const convertOptions = {
-          fetchRelationships: options.fetchRelationships
+          fetchRelationships: options.fetchRelationships !== false
         };
         await self.apos.schema.convert(req, schema, input, doc, convertOptions);
 
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -759,6 +759,7 @@ module.exports = {
 
       async convert(req, input, doc, options = {
         presentFieldsOnly: false,
+        fetchRelationships: true,
         type: null,
         copyingId: null,
         createId: null
@@ -783,8 +784,10 @@ module.exports = {
             ...input
           };
         }
-
-        await self.apos.schema.convert(req, schema, input, doc);
+        const convertOptions = {
+          fetchRelationships: options.fetchRelationships
+        };
+        await self.apos.schema.convert(req, schema, input, doc, convertOptions);
 
         if (options.createId) {
           doc.aposDocId = options.createId;
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -1734,14 +1734,13 @@ module.exports = {
         }
 
         function forSchema(schema, doc) {
+          if (!doc) {
+            return;
+          }
           for (const field of schema) {
             if (field.type === 'area' && doc[field.name] && doc[field.name].items) {
               for (const widget of doc[field.name].items) {
-                self.walkByMetaType(widget, {
-                  arrayItem: handlers.arrayItem,
-                  object: handlers.object,
-                  relationship: handlers.relationship
-                });
+                self.walkByMetaType(widget, handlers);
               }
             } else if (field.type === 'array') {
               if (doc[field.name]) {
@@ -1752,14 +1751,10 @@ module.exports = {
               }
             } else if (field.type === 'object') {
               const value = doc[field.name];
-              if (value) {
-                handlers.object(field, value);
-                forSchema(field.schema, value);
-              }
+              handlers.object(field, value);
+              forSchema(field.schema, value);
             } else if (field.type === 'relationship') {
-              if (Array.isArray(doc[field.name])) {
-                handlers.relationship(field, doc);
-              }
+              handlers.relationship(field, doc, true);
             }
           }
         }
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -1754,7 +1754,7 @@ module.exports = {
               handlers.object(field, value);
               forSchema(field.schema, value);
             } else if (field.type === 'relationship') {
-              handlers.relationship(field, doc, true);
+              handlers.relationship(field, doc);
             }
           }
         }
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -599,7 +599,7 @@ module.exports = {
       // set error class names, etc. If the error is not a string, it is a
       // database error etc. and should not be displayed in the browser directly.
 
-      async convert(req, schema, data, destination) {
+      async convert(req, schema, data, destination, options) {
         if (Array.isArray(req)) {
           throw new Error('convert invoked without a req, do you have one in your context?');
         }
@@ -629,7 +629,8 @@ module.exports = {
                   required: isRequired
                 },
                 data,
-                destination
+                destination,
+                options
               );
             } catch (error) {
               if (Array.isArray(error)) {
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -1039,7 +1039,7 @@ module.exports = {
 
         const handlers = {
           arrayItem: (field, object) => {
-            if (!can(field)) {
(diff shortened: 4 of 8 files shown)


modules/@apostrophecms/doc/index.js
```diff
@@ -1752,14 +1751,10 @@ module.exports = {
...
-                handlers.object(field, value);
-                forSchema(field.schema, value);
-              }
+              handlers.object(field, value);
+              forSchema(field.schema, value);
             } else if (field.type === 'relationship') {
-              if (Array.isArray(doc[field.name])) {
-                handlers.relationship(field, doc);
-              }
+              handlers.relationship(field, doc, true);
```
ValJed, line 1757: Since for relationships I need to pass inside the handler when I have no `doc[field.name]` in my document. I decided to move the logic that checks if the value exists inside the handlers of `prepareForStorage`.
This way I can reuse this method for `simulateRelationshipsFromStorage`.

boutell:
Happy with the general direction here.

modules/@apostrophecms/schema/index.js
```diff
@@ -1048,15 +1049,15 @@ module.exports = {
...
+            if (!object || !can(field)) {
               return;
             }
 
             object.metaType = 'object';
             object.scopedObjectName = field.scopedObjectName;
           },
           relationship: (field, doc) => {
-            if (!can(field)) {
+            if (!Array.isArray(doc[field.name]) || !can(field)) {
```
ValJed, line 1060: I handle empty values directly in `prepareForStorage` because in the simulate case I want to be able to enter in the handler when there is no value.

modules/@apostrophecms/doc/index.js
```diff
@@ -1734,14 +1734,13 @@ module.exports = {
         }
 
         function forSchema(schema, doc) {
+          if (!doc) {
+            return;
+          }
```
ValJed, line 1739: To handle empty values that aren't catched in `walkByMetaType` anymore.

modules/@apostrophecms/doc/index.js
```diff
@@ -1752,14 +1751,10 @@ module.exports = {
               }
             } else if (field.type === 'object') {
               const value = doc[field.name];
-              if (value) {
-                handlers.object(field, value);
-                forSchema(field.schema, value);
-              }
+              handlers.object(field, value);
+              forSchema(field.schema, value);
```
ValJed, line 1755: Same as below for consistency

ETLaurent (changes requested):
`console.log` to be removed, otherwise few questions

modules/@apostrophecms/doc/index.js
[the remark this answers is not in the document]
ValJed: That's right, I forgot it, it was for tests purposes. Thanks

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -1022,6 +1027,7 @@ module.exports = (self) => {
         return;
       }
       const results = await manager.find(req, { $or: clauses }).relationships(false).toArray();
+      console.log('results', results);
```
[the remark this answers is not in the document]
ValJed: Removed

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -991,6 +991,11 @@ module.exports = (self) => {
       if (field.max && field.max < input.length) {
         throw self.apos.error('max', `Maximum ${field.withType} required reached.`);
       }
+      if (options.fetchRelationships === false) {
+        destination[field.name] = input;
+        return;
+      }
+
```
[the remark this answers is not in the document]
ValJed: I'm not convinced with the naming here neither but input is just the value of the current field  in doc, that we set in destination.
If the value in doc is null it is set to empty array:
```javascript
      let input = data[field.name];
      if (input == null) {
        input = [];
      }
```
So I use input here to be sure I have at leastan empty array.

ETLaurent (approved):

modules/@apostrophecms/doc/index.js
ETLaurent: @ValJed should we do the same also for arrays above?

ETLaurent: well arrays are different

ValJed: I wanted yes but the problem is that for arrays we call `arrayItem` and not `array` handler. It is called on each array item and not on the array itself.
I could have change that maybe and passed directly the array?
But I wanted to bring the less amount of changes possible.

modules/@apostrophecms/doc-type/index.js
```diff
@@ -783,8 +784,10 @@ module.exports = {
             ...input
           };
         }
-
-        await self.apos.schema.convert(req, schema, input, doc);
+        const convertOptions = {
+          fetchRelationships: options.fetchRelationships
```
haroun, line 788: we should enforce it to always be a boolean. If you pass `options` without `fetchRelationships` the value will be undefined and code later in this PR will not except this

modules/@apostrophecms/doc/index.js
```diff
@@ -1734,14 +1734,13 @@ module.exports = {
...
+          }
           for (const field of schema) {
             if (field.type === 'area' && doc[field.name] && doc[field.name].items) {
               for (const widget of doc[field.name].items) {
-                self.walkByMetaType(widget, {
-                  arrayItem: handlers.arrayItem,
-                  object: handlers.object,
-                  relationship: handlers.relationship
-                });
+                self.walkByMetaType(widget, handlers);
```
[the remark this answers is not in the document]
haroun: Please note we are mutating `handlers` in `walkByMetaType` and use recursive calls to that method. I don't think this is a good idea, while traversing the schema, hypothetically you can end up with an unrelated `handlers` object.

We should change line 1717 before I think.

haroun (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 * Add a locale switcher in pieces and pages editor modals. This is available for localized documents only, and allows you to switch between locales for the same document.
   The locale can be switche at only one level, meaning that sub documents of a document that already switched locale will not be able to switch locale itself.
 * Adds visual focus states and keyboard handlers for engaging with areas and widgets in-context
+* Adds method `simulateRelationshipsFromStorage` method in schema module, for generate relationships in documents from database, doing the opposite than `prepareForStorage`.
```
boutell, line 10: Clearer explanation:

"This method populates the relationship field with just enough information to allow `convert` to accept it. It does not fully fetch the related documents. It does the opposite of `prepareForStorage`."

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -991,6 +991,11 @@ module.exports = (self) => {
       if (field.max && field.max < input.length) {
         throw self.apos.error('max', `Maximum ${field.withType} required reached.`);
       }
+      if (options.fetchRelationships === false) {
+        destination[field.name] = input;
+        return;
```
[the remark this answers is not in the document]
boutell: I shouldn't say "etc" here. The parts that matter are:

`_id` must be a string.
`_fields` must pass convert, where applicable.

That is the sum of the relationship, so we don't have to validate other properties of each document in `input` in any way.

boutell (changes requested):

modules/@apostrophecms/doc/index.js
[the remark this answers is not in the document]
ValJed: Yes, this is the same, `walkByMetaType` simply creates a new object with default empty methods for those not existing.
The handler is always the same

modules/@apostrophecms/doc-type/index.js
ValJed: I think the cleaner would have been to choose an falsy option like: `noRelationshipFetch`
Because in the current state, the option can be null, so we have to check explicitly if the option passed is `false`. This is what I do when using the option. That's why I did not typed this one, because it might generates bugs.

modules/@apostrophecms/schema/lib/addFieldTypes.js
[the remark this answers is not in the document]
ValJed: Ok right, I added a convert step here for `_id` and `fields` props, and updated tests accordingly.

modules/@apostrophecms/schema/index.js
```diff
@@ -599,7 +599,7 @@ module.exports = {
       // set error class names, etc. If the error is not a string, it is a
       // database error etc. and should not be displayed in the browser directly.
 
-      async convert(req, schema, data, destination) {
+      async convert(req, schema, data, destination, options) {
```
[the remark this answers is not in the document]
ValJed: You mean for readability?
Because for me the safer is to handle the default value in the field convert itself.
Since convert methods are called from a lot  of places without options most of the time it feels safer and less verbose to have one true default location where the option is actually used. WDYT?

modules/@apostrophecms/doc-type/index.js
ValJed: Updated to:
```javasccript
          fetchRelationships: options.fetchRelationships !== false
```

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -972,7 +972,7 @@ module.exports = (self) => {
     // properties is handled at a lower level in a beforeSave
     // handler of the doc-type module.
 
-    async convert(req, field, data, destination) {
+    async convert(req, field, data, destination, options = { fetchRelationships: true }) {
```
boutell, line 975: As we were talking about this morning, use the `{ fetchRelationships: true } = {}` syntax to gracefully handle both cases where no object is passed at all and cases where the object is passed but this particular option has no specified value.

boutell (changes requested):

modules/@apostrophecms/schema/lib/addFieldTypes.js
ValJed: Yes I did but options is reused below. I'll create a constant for that.

ValJed: Done

boutell (changes requested):
As documentation for our own team and others, please add the options object with its defaults at the two common entry points: `doc-type.convert` and `schema.convert`. I understand that it is not technically necessary to get the right result but it will help developers. If this were typescript we would have to do it.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -1034,9 +1060,13 @@ module.exports = (self) => {
           const result = results.find(doc => (doc._id === item._id));
           if (result) {
             if (field.schema) {
-              result._fields = { ...(destination[field.name]?.find?.(doc => doc._id === item._id)?._fields || {}) };
+              result._fields = {
+                ...(destination[field.name]
+                  ?.find?.(doc => doc._id === item._id)
+                  ?._fields || {})
+              };
```
ValJed, line 1067: Indentation

boutell (approved):

haroun (changes requested):

haroun (approved):

PR #4682, opened 2024-08-08
PRO-4659: new big upload middleware & corresponding client implementation

boutell:
See the mocha test. Note that when calling from the browser you don't have to create an `http` object, the standard `apos.http` is automatically used. Also the files in your `files` option should be actual web browser `File` objects.

To utilize this from the import-export module you can import it into your front end code with:

```
import bigUpload from 'apostrophe/lib/big-upload.js';
```

As seen in the tests, the route on the receiving end must use the matching middleware, in lieu of the normal `multiparty` middleware. And then it "just works," with `req.files` populating as normal with the final result.

I intentionally did not add this to the standard `apos.http` implementation to reduce frontend bloat in projects that don't use it and simply to keep the code more modular. And of course it made it more testable.

When calling the middleware be sure to provide an `authorize` function to throw an exception and stop the show if the user has no business doing what they are doing. This will save major resources (DOS risk).


test/big-upload.js
```diff
@@ -0,0 +1,111 @@
...
+  let apos;
+  let received = false;
+
+  after(async function () {
+    return t.destroy(apos);
+  });
+
+  this.timeout(t.timeout);
+
+  it('init apos', async function() {
```
ETLaurent, line 22: apos initialization should be made from a `before` function in order to make tests independent ("onlyable")

ETLaurent (approved):

PR #4693, opened 2024-08-19
Array modal's fields now keyboard accessible

stuartromanek:
## Summary

- Ensures array modals' existing fields are keyboard accessible when modal is opened.
- Ensures new array items' fields are keyboard accessible when added
- Change the array editor's `isModified` method to a computed prop for consistency with other components

1. Check these two scenarios
https://github.com/user-attachments/assets/b1e1f42d-982c-4b13-ba12-4552dae470b3

## Tests
https://github.com/apostrophecms/testbed/actions/runs/12794762166

*(Check at least one)*

- [x] Bug fix


modules/@apostrophecms/ui/ui/apos/mixins/AposModifiedMixin.js
```diff
@@ -25,7 +25,8 @@ export default {
     // Returns true if the cancellation does occur
     async confirmAndCancel() {
       let dismiss;
-      if (this.isModified) {
+      const modified = typeof this.isModified === 'function' ? this.isModified() : this.isModified;
+      if (modified) {
```
stuartromanek, line 29: @ValJed Unrelated to main pr, but i noticed array's `isModified` is a function, not a boolean. The bug it was causing was that array modals always acted as if they had been modified and threw a confirm when exiting a clean modal.

ValJed: oh wow nicely spotted, wouldn't it be a better solution to use a computed everywhere or a function ?

ValJed (changes requested):
Question about strat.

stuartromanek:
> wouldn't it be a better solution to use a computed everywhere or a function ?

That sounds right to me. 

I'm already doing too much in some of these PRs .. how would you want to move on a computed? Should i back out my fix and make a ticket for a better a solution? Or leave this and make the ticket anyway, knowing that this could change?

stuartromanek:
@ValJed It wasn't as involved to make the upstream change as I thought, so I've converted the array modal's `isModified` method to a computed property and reverted the double check I initially had here. Running updated tests now but I think this is a solid fix.

ValJed (approved):

PR #4695, opened 2024-08-20
Keep widget focused when menu is open

stuartromanek:
## Summary

The idea for this fix is that when a widget's context menu is open that widget should assume focus until something else takes it. With this we can ensure all the proper UI stays on the screen without exact pointer interactions

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
@@ -179,13 +179,11 @@ const { themeClass } = useAposTheme();
 onMounted(() => {
   apos.bus.$on('context-menu-toggled', hideWhenOtherOpen);
   apos.bus.$on('close-context-menu', hideMe);
-  // apos.bus.$on('widget-focus', hide);
 });
 
 onBeforeUnmount(() => {
   apos.bus.$off('context-menu-toggled', hideWhenOtherOpen);
   apos.bus.$off('close-context-menu', hideMe);
-  // apos.bus.$off('widget-focus', hide);
 });
 
 function getMenuOffset() {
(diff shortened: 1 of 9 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
```diff
@@ -173,12 +178,12 @@ const { themeClass } = useAposTheme();
 
 onMounted(() => {
   apos.bus.$on('context-menu-opened', hideWhenOtherOpen);
-  apos.bus.$on('widget-focus', hide);
+  // apos.bus.$on('widget-focus', hide);
```
stuartromanek, line 181: @ValJed seeing that you added this .. was there a specific scenario this was cleaning up?

stuartromanek: (to clarify, I commented the line out)

boutell (changes requested):
This solves the main set of problems well, but the "Browse" button for the relationship field beneath the area still winds up on top of the menu.

boutell (dismissed):
With this change it's happy kittens all the way down as far as I'm concerned.

Since this isn't a simple rollback I'm good with waiting for Val to review the frontend code in the morning, and then I'll do a regression test of this branch, and then we can ship.

modules/@apostrophecms/ui/ui/apos/components/AposContextMenu.vue
ValJed: I posted a video on slack for you to understand why I added this code. Since I'm not sure about the current behavior I let you see if it can be safely removed.

stuartromanek: Nice catch! I added a provision to close context menus by name and fire that specifically when the Rich Text editor gets focused

boutell (approved):

PR #4705, opened 2024-08-26
WIP on if not canceled approach

boutell:
This is incomplete, I have to finish making the remaining functions stateless and test the whole thing. You can see what I intend to do in the documentation comment and in the media manager.

### DIFF
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
@@ -377,13 +377,10 @@ export default {
       this.isLoading = true;
       this.currentPage = 1;
       this.items = [];
-      try {
-        await this.debouncedGetMedia.skipDelay(opts);
-      } finally {
-        this.isLoading = false;
-        this.modified = false;
-        this.updateEditing(null);
-      }
+      await this.debouncedGetMedia.skipDelay(opts);
+      this.isLoading = false;
+      this.modified = false;
+      this.updateEditing(null);
     },
     async filter(name, value) {
       this.filterValues[name] = value;
--- a/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
+++ b/modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
@@ -274,30 +274,19 @@ export default {
   },
   created() {
     this.setDefaultFilters();
-    const debouncedGetMedia = debounceAsync(this.getMedia, DEBOUNCE_TIMEOUT);
-    this.debouncedGetMedia = async function (...args) {
-      try {
-        const result = await debouncedGetMedia(...args);
-        return result;
-      } catch (e) {
-        if (e.message !== 'debounce:canceled') {
-          throw e;
-        }
-      }
-    };
-    this.debouncedGetMedia.cancel = debouncedGetMedia.cancel;
+    this.debouncedGetMedia = debounceAsync(this.getMedia, DEBOUNCE_TIMEOUT, {
+      ifNotCanceled: this.appendMedia
+    });
   },
   async mounted() {
-    this.mounted = true;
     this.modal.active = true;
-    await this.getMedia({ tags: true });
-    this.isFirstLoading = false;
-
+    // Do these before any async work or they might get added after they are "removed"
     apos.bus.$on('content-changed', this.onContentChanged);
     apos.bus.$on('command-menu-manager-close', this.confirmAndCancel);
+    await this.debouncedGetMedia.skipDelay({ tags: true });
+    this.isFirstLoading = false;
   },
   beforeUnmount() {
-    this.mounted = false;
     this.debouncedGetMedia.cancel();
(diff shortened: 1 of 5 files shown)


modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -349,9 +336,6 @@ export default {
           draft: true
         }
       ));
-      if (!this.mounted) {
```
boutell, line 352: Continuing here will do no harm.

boutell: moduleOptions is immutable, so should be harmless to call here even if unmounted.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -367,29 +351,39 @@ export default {
...
+      result.currentPage = apiResponse.currentPage;
+      result.totalPages = apiResponse.pages;
+      result.items = [];
       for (const image of apiResponse.results) {
-        this.items.push(image);
+        result.items.push(image);
+      }
+      return result;
+    },
+    async appendMedia({ tagList, currentPage, totalPages, items }) {
```
boutell, line 368: All the state modification occurs at the end in the ifNotCanceled function.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -367,29 +351,39 @@ export default {
...
       this.currentPage = 1;
       this.items = [];
-      await this.getMedia(opts);
-      this.isLoading = false;
-      this.modified = false;
-      this.updateEditing(null);
+      try {
+        await this.debouncedGetMedia.skipDelay(opts);
+      } finally {
+        this.isLoading = false;
```
boutell, line 383: We can still handle flags this way because it is still awaitable and waits to settle before returning, although it returns nothing of interest, not even errors. Error reporting (if we had any) would be the responsibility of `getMedia` itself.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -405,8 +399,7 @@ export default {
     async completeUploading (imgIds) {
       this.currentPage = 1;
       this.items = [];
-      await this.getMedia();
-
+      await this.debouncedGetMedia.skipDelay();
```
boutell, line 402: Invoking the debounced version but skipping the delay keeps all the benefits of race condition prevention.

modules/@apostrophecms/ui/ui/apos/utils/index.js
```diff
@@ -40,25 +76,46 @@ function debounceAsync(fn, delay, options = {}) {
...
+        setTimer(resolve, reject, args, false);
+      }
       clearTimeout(timer);
       timer = setTimer(resolve, reject, args, delay);
     });
   };
 
+  if (options.ifNotCanceled) {
+    wrapper = wrapper().then(...args => {
+      options.ifNotCanceled(...args);
```
boutell, line 103: Must be synchronous, has no job except side effects, so we intentionally pay no attention to its return value.

modules/@apostrophecms/image/ui/apos/components/AposMediaManager.vue
```diff
@@ -367,26 +351,33 @@ export default {
...
+        this.items.push(item);
       }
     },
     async refetchMedia(opts) {
       this.isLoading = true;
       this.currentPage = 1;
       this.items = [];
-      await this.getMedia(opts);
+      await this.debouncedGetMedia.skipDelay(opts);
       this.isLoading = false;
```
boutell, line 381: No need for try/catch because in the presence of `ifNotCanceled` the wrapper always resolves to null regardless of which way "fn" settles (if it even gets called). However it does wait for "fn" to settle if it is called so we can do this kind of cleanup easily.

(I don't know if setting `isLoading` only some of the time is a bug or a feature, I assume it's a feature. Otherwise it would have been moved inside getMedia itself by now)

ValJed:
Strategy looks good, it's understandable (still complex but the feature is a bit). We need to remember to call the `skipDelay` also that the passed function must be synchronous.
Maybe the `skipDelay` could be called `init` since we gonna call it eveytime we init a debounce, not sure about that one since it describes less the actual behavior than `skipDelay`, or `immediate`? I think we use it already elsewhere.
Also the `ifNotCanceled` could be maybe clearer about what is does, like `handleData`, `handleResponse` or something like that?

boutell:
@ValJed "skipDelay" or "immediate" is more appropriate because it isn't just used at init time, it is sometimes used later.

boutell:
`handleResponse` is a good name, I like that name.

myovchev:
I like skipDelay (and it's definitely not `immediate`, it's misleading - it tries to skip the delay). I would say `onSuccess` is better. It sounds like callback (because it is) and it says that will be executed in the right time if ever.

PR #4716, opened 2024-09-05
set def field value for visible fields

ETLaurent:
## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -650,28 +650,17 @@ module.exports = {
 
         for (const error of errors) {
           if (error.path) {
-            // `self.isVisible` will only throw for required fields that have
-            // an external condition containing an unknown module or method:
-            const isVisible = await self.isVisible(req, schema, destination, error.path);
+            const name = error.path;
+            const field = schema.find(field => field.name === name);
+            const def = field && klona((field.def !== undefined) ? field.def : self.fieldTypes[field.type]?.def);
+            if (def) {
+              destination[field.name] = def;
+              continue;
+            }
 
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -650,28 +650,17 @@ module.exports = {
...
         for (const error of errors) {
           if (error.path) {
-            // `self.isVisible` will only throw for required fields that have
-            // an external condition containing an unknown module or method:
-            const isVisible = await self.isVisible(req, schema, destination, error.path);
+            const name = error.path;
+            const field = schema.find(field => field.name === name);
+            const def = field && klona((field.def !== undefined) ? field.def : self.fieldTypes[field.type]?.def);
+            if (def) {
+              destination[field.name] = def;
```
boutell, line 657: This is not the place to do it. The front end should be honoring the defaults it already has and is already showing. Trying to force defaults at save time will be a regression to something we already agreed is bad: forcing the use of defaults when people intentionally clear fields.

boutell (changes requested):

modules/@apostrophecms/schema/index.js
boutell: To be clear I am saying this a front end bug that needs a front end fix.

boutell: To be clear: the front end should actually send the values it is initially displaying, whether those specific fields get touched or not. Not just those for fields the user moves the focus to.

PR #4721, opened 2024-09-10
PRO-6472: at startup, automatically supply a value for any new schema property that has a def or fallback def

boutell:
## Summary

See tests.

*(Check at least one)*

- [X] Bug fix
- [X] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


index.js
```diff
@@ -300,7 +300,7 @@ async function apostrophe(options, telemetry, rootSpan) {
     self.apos.schema.validateAllSchemas();
     self.apos.schema.registerAllSchemas();
     await self.apos.lock.withLock('@apostrophecms/migration:migrate', async () => {
-      await self.apos.migration.migrate(); // emits before and after events, inside the lock
+      await self.apos.migration.migrate(self.argv);
```
boutell, line 303: For consistency, it was passed in the other, redundant invocation.

modules/@apostrophecms/migration/index.js
```diff
@@ -38,15 +39,11 @@ module.exports = {
               manager.addSortifyMigration(field.name);
             });
           });
-        },
-        async executeMigrations() {
```
boutell, line 42: D'oh. This was completely redundant in the current setup.

modules/@apostrophecms/migration/index.js
```diff
@@ -38,15 +39,11 @@ module.exports = {
...
-          if ((process.env.NODE_ENV !== 'production') || self.apos.isNew) {
-            // Run migrations at dev startup (low friction).
-            // Also always run migrations at first startup, so even
-            // in prod with a brand new database the after event always fires
-            // and we get a chance to mark the migrations as skipped
-            await self.migrate(self.apos.argv);
-          }
+        }
+      },
+      before: {
```
boutell, line 44: Invoke the new capability, which is not a registered, named migration, but rather an independent part of the migration phase that does its own checks to determine if work is needed.

modules/@apostrophecms/migration/index.js
```diff
@@ -245,8 +242,13 @@ module.exports = {
...
           // start to run migrations on a previous attempt,
           // which causes an occasional unique key error if not
-          // corrected for here
-          await self.db.removeMany({});
+          // corrected for here.
+          //
+          // Other migration-related facts that are not migration
+          // names are stored with a leading *, leave them alone
+          await self.db.removeMany({
+            _id: /^[^*]/
```
boutell, line 250: Somewhere to keep memos like the list of existing schema fields. Otherwise we have a proliferation of tiny mongo collections.

modules/@apostrophecms/migration/index.js
```diff
@@ -286,14 +288,18 @@ module.exports = {
...
   },
   tasks(self) {
     return {
       migrate: {
         usage: 'Apply any necessary migrations to the database.',
-        task: self.migrate
+        // Migrations actually run on every invocation
+        // and automatically detect whether any work
+        // must be done
+        task: () => {}
```
boutell, line 302: Previously this task invoked the migrate code for a third time 🪦

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -0,0 +1,141 @@
+const _ = require('lodash');
+const { klona } = require('klona');
+
+module.exports = (self) => {
+  return {
+    async addMissingSchemaFields() {
+      let scans = 0; let updates = 0;
+      const lastPropLists = await self.getLastPropLists();
```
boutell, line 8: A "prop list" is a list of all schema properties that existed, for comparison with the new list for this startup to determine if any property additions might potentially be needed. We make this determination quickly and move on if new work is not needed.

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -0,0 +1,141 @@
...
+          continue;
+        }
+        schema[name] = [];
+        self.expandPropList(module.schema, schema[name], '');
+      }
+      return schema;
+    },
+    expandPropList(schema, propList, dotPath) {
+      for (const field of schema) {
+        const newDotPath = dotPath ? `${dotPath}.${field.name}` : field.name;
```
boutell, line 116: We're not coming up with a real dot path to patch anything at this stage, just a conceptual dot path that namespaces these properties enough to determine if there are differences from one generation of code to the next by comparing two lists of paths.

Real dot paths get figured out in the guts of addMissingSchemaFieldsFor, which does the hard work if a difference does exist.

boutell:
Please note I broke document versions out separately because it can be done at any point, ideally also this cycle, but nothing gets worse if we don't get to it this cycle:

https://linear.app/apostrophecms/issue/PRO-6577/improve-handling-of-new-defaults-for-properties-of-old-document

modules/@apostrophecms/migration/index.js
ETLaurent: We could have just one new, separate collection like `aposModuleSchema` or `aposRegisteredSchema` so we don't store it into the `aposMigration` collection, which is a bit tied to migrations, but not really...

Inside that dedicated collection, we could have the schema stored in documents: 1 document in that collection = 1 module schema.

That would require changing a lot of code, and since you've already finished the feature, maybe it's not worth it.
Just wanted to have your thoughts on that, to have the schemas stored in a dedicated collection for clarity

modules/@apostrophecms/migration/index.js
boutell: The task must exist otherwise users will get an error when launching the migration task according to our documentation. The fact that it is implicit is an implementation detail that should not concern users. Later we may eliminate the implicit invocation except when the task is being invoked, in production.

ETLaurent (approved):
overall logic lgtm

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -0,0 +1,141 @@
...
+    async addMissingSchemaFields() {
+      let scans = 0; let updates = 0;
+      const lastPropLists = await self.getLastPropLists();
+      const propLists = self.getPropLists();
+      let changesToPropLists = false;
+      for (const name of Object.keys(propLists)) {
+        if (!_.isEqual(lastPropLists?.[name], propLists[name])) {
+          changesToPropLists = true;
+          scans++;
+          updates += await self.addMissingSchemaFieldsForDocType(name, propLists[name]);
```
haroun, line 15: This method accepts only 1 parameter below

haroun (changes requested):
Also, we need to come up with a solution for the document-versions module.
Versions are stored in another collections.
If the field is required, we might want to add the def values or something (like for required fields without default value) during the restore method call.

modules/@apostrophecms/migration/index.js
boutell: @haroun it has worked like this for about as long as A3 has existed. Plus extra, redundant invocations in two other places. Since we have this automatic invocation at every startup I am fixing the bug of redundant invocations. As I mentioned we might reconsider this behavior later for production, but in dev it never makes sense to start up without migrating, if your code does not behave after migrations then that is a bug in the migrations that needs attention and something that will break in production if bypassed, etc.

modules/@apostrophecms/migration/index.js
```diff
@@ -286,14 +288,18 @@ module.exports = {
             throw err;
           }
         }
-      }
+      },
+      ...require('./lib/addMissingSchemaFields.js')(self)
```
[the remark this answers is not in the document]
boutell: Done

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
boutell: The migration phase **will run** on any startup. **Including** an empty task. So omitting the call from the task is just eliminating redundancy.

At a future time we might reintroduce an optional behavior where the migration does not run in production except in the task, but that behavior has effectively not been possible before, so I don't have to introduce that feature to fix the redundancy bug.

haroun (approved):

haroun (approved):

PR #4727, opened 2024-09-16
add <use> tag to DOMPurify allow list when sanitizing uploaded SVGS

stuartromanek:
## Summary

- Adds `<use>` tag to DOMPurify allow list config on svg upload

Test file
illustration_gift

### What it looks like when broken
image

### What it looks like when `<use>` is allowed
image

## Notes
There is a configuration profile specifically for SVGs, I would have felt better of using that just *worked*, but it did not. That would look like.
```
// allow all safe SVG elements and SVG Filters, no HTML or MathML
const clean = DOMPurify.sanitize(dirty, {USE_PROFILES: {svg: true, svgFilters: true}});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixes
 
+* Uploaded SVGs now permit `<use>` tags granted their `xlink:href` property is a local reference and begins with the `#` character. This improves SVG support while mitgating XSS vulnerabilities.
 * Default properties of object fields present in a widget now populate correctly even if never focused in the editor.
 
 ## 4.7.0 (2024-09-05)
--- a/modules/@apostrophecms/attachment/index.js
+++ b/modules/@apostrophecms/attachment/index.js
@@ -519,6 +519,11 @@ module.exports = {
         const writeFile = require('util').promisify(fs.writeFile);
         const window = new JSDOM('').window;
         const DOMPurify = createDOMPurify(window);
+        DOMPurify.addHook('afterSanitizeAttributes', node => {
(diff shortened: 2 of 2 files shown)


BoDonkey (approved):

boutell (changes requested):
I can't figure out why this is necessary, it looks like DOMPurify did some work a while ago to ensure `use` is permitted in just this situation:

https://github.com/cure53/DOMPurify/issues/233

Do you know whether `ADD_TAGS` will let it through as-is 100% of the time (this would be bad, XSS vector) or just with this sort of validation mentioned in the github issue?

stuartromanek:
@boutell DOMPurify is gobbling up `<use>` tags, a semi-common technique in splitting the definition from implementation of (SVGs || HTML) markup with some benefit around 'load once, use many times with no overhead' .. These get stripped automatically because its been flagged as an XSS vuln.

As per a solution within DOMPurify's issues, we re evaluate the `<use>` after initial sanitization to ensure the `xlink:href` begins with a `#` and is safe as an XSS vuln
https://github.com/cure53/DOMPurify/issues/233

Seem okay?

stuartromanek:
> I can't figure out why this is necessary, it looks like DOMPurify did some work a while ago to ensure `use` is permitted in just this situation:
> 
> [cure53/DOMPurify#233](https://github.com/cure53/DOMPurify/issues/233)
> 
> Do you know whether `ADD_TAGS` will let it through as-is 100% of the time (this would be bad, XSS vector) or just with this sort of validation mentioned in the github issue?

This issue illustrates the vulnerability and how it can be mitigated, not that a bug was fixed

BoDonkey (approved):

boutell (approved):

boutell:
Approved!

PR #4078, opened 2023-03-07
dynamic choices syntax accepts module and parenthesis

ETLaurent:
## Summary

- adapt dynamic choices to accept a module name at the beginning and parenthesis at the end
- adapt and use method shared with external conditions
- adapt and add tests for `/choices` api

## What are the specific steps to test this change?

In the example below, we're defining methods to get choices dynamically, from the same module, from another module, with parenthesis.

Test that it still works without parenthesis,  and that an unknown module or method name will result in clean 400 errors (in backend, and clean message given to the frontend and displayed in the modal).
Defining a method with parenthesis with an argument should send a warning on server-side.

```js
  // in "article" piece
  fields: {
    add: {
      select: {
        label: 'Dynamic select',
        type: 'select',
        choices: 'getChoices()'
      },
      checkboxes: {
        label: 'Dynamic checkboxes',
        type: 'checkboxes',
        choices: 'topics:getChoices()'
      },
    }
  },
  methods(self) {
    return {
      async getChoices(req, { docId }) {
        return new Promise(resolve => {
          setTimeout(() => {
            const res = [
              {
                label: 'One',
                value: 'one'
              },
              {
                label: 'Two',
                value: 'two'
              },
              {
                label: 'Three',
                value: 'three'
              }
            ];
            console.log('🚀 ~ file: index.js:126 ~ b ~ docId:', docId);
            resolve(res);
          }, 2000);
        });
      }
    };
  }
```

```js
  // in "topic" piece
 methods(self) {
    return {
      async getChoices(req, { docId }) {
        return new Promise(resolve => {
          setTimeout(() => {
            const res = [
              {
                label: 'Un',
                value: 'un'
              },
              {
                label: 'Deux',
                value: 'deux'
              },
              {
                label: 'Trois',
                value: 'trois'
              }
            ];
            console.log('🚀 ~ file: index.js:126 ~ b ~ docId:', docId);
            resolve(res);
          }, 2000);
        });
      }
    };
```

## What kind of change does this PR introduce?
*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

## Make sure the PR fulfills these requirements:

- [x] It includes a) the existing issue ID being resolved, b) a convincing reason for adding this feature, or c) a clear description of the bug it resolves
- [x] The changelog is updated
- [ ] Related documentation has been updated
- [x] Related tests have been updated

[...]

PR #4731, opened 2024-09-18
fix piecesFilters with dynamic choices

ETLaurent:
See https://github.com/apostrophecms/starter-kit-essentials/pull/63

### DIFF
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -1706,6 +1706,24 @@ module.exports = {
         });
       },
 
+      async getChoicesForQueryBuilder(field, query) {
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -1706,6 +1706,24 @@ module.exports = {
         });
       },
 
+      async getChoicesForQueryBuilder(field, query) {
+        const req = self.apos.task.getReq();
+        const allChoices = await self.getChoices(req, field);
```
ETLaurent, line 1711: `self.getChoices` was already there and previously added in #4078, but never used in the builders

boutell (approved):

PR #4736, opened 2024-09-24
fix oversights in add missing schema fields

boutell:
Fix issues that came up in testbed.

### DIFF
--- a/modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
+++ b/modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
@@ -35,12 +35,48 @@ module.exports = (self) => {
       }, async doc => {
         const changes = {};
         await self.addMissingSchemaFieldsFor(doc, schema, '', changes);
+        if (doc.type === '@apostrophecms/user') {
+          // In an abundance of caution we should leave permissions-related
+          // properties alone. The role will only be missing if transitioning
+          // from advanced permissions back to regular and there is a right
+          // way to do that (via a task) so we should not interfere. There
+          // should be no default for the password, but don't chance it;
+          // also leave advanced permission group data alone just for good measure
+          delete changes.password;
+          delete changes.role;
+          delete changes.groupsIds;
+          delete changes.groupsFields;
+        }
         if (Object.keys(changes).length > 0) {
           updates++;
+          const seen = new Set();
+          const keep = {};
+          for (const key of Object.keys(changes)) {
+            // MongoDB won't tolerate $set for both
+            // a parent key and a subkey, which makes sense
+            // because we set the entire value for the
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -35,12 +35,48 @@ module.exports = (self) => {
...
+          delete changes.role;
+          delete changes.groupsIds;
+          delete changes.groupsFields;
+        }
         if (Object.keys(changes).length > 0) {
           updates++;
+          const seen = new Set();
+          const keep = {};
+          for (const key of Object.keys(changes)) {
+            // MongoDB won't tolerate $set for both
```
boutell, line 55: It seems like the existing unit tests should have caught this, so I don't know of a good way to test it better, but in testbed this did come up and required the fix I have implemented here.

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -54,8 +90,10 @@ module.exports = (self) => {
     addMissingSchemaFieldsFor(doc, schema, dotPath, changes) {
       for (const field of schema) {
         const newDotPath = dotPath ? `${dotPath}.${field.name}` : field.name;
-        // Supply the default
-        if (doc[field.name] === undefined) {
+        // Supply the default if a field is undefined, and also in the
```
boutell, line 93: Not really our fault per se if the object field had a weird non-object value rather than being undefined, but it happened "in the field" with testbed, so fix it!

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -91,7 +129,10 @@ module.exports = (self) => {
           }
         } else if (field.type === 'relationship') {
           for (const [ key, item ] of Object.entries(doc[field.fieldsStorage] || {})) {
-            const itemPath = `${newDotPath}.${field.fieldsStorage}.${key}`;
```
boutell, line 94: This was the wrong path.

modules/@apostrophecms/migration/lib/addMissingSchemaFields.js
```diff
@@ -101,7 +142,7 @@ module.exports = (self) => {
     getPropLists() {
       const schema = {};
       for (const [ name, module ] of Object.entries(self.apos.doc.managers)) {
-        if (!module.__meta.name) {
+        if (!module.__meta?.name) {
```
boutell, line 145: Fake managers have no `__meta` so this wasn't tolerant enough.

haroun (approved):

boutell:
Good point, we should document this in the migrations docs.

PR #4741, opened 2024-10-02
Enable by default, other tweaks

boutell:
* enable by default
* adjust the breakpoints to give them a basis in popular device sizes
* change the option name and translation key names (we can do the rest later) to match the way we talk about the feature elsewhere before it is too late to harmonize them


modules/@apostrophecms/asset/index.js
```diff
@@ -89,6 +89,7 @@ module.exports = {
 
   async init(self) {
 
+    self.options.devicePreviewMode = self.options.breakpointPreviewMode || self.options.devicePreviewMode;
```
boutell, line 92: This is temporary until the rest of the logic can be changed over, I didn't want to cowboy that and risk bugs, the idea is to document `breakpointPreviewMode` and change the rest of the code when we have time

modules/@apostrophecms/asset/index.js
```diff
@@ -46,12 +46,12 @@ module.exports = {
...
-    // to the project for override to work properly.
+    // NOTE: the whole breakpointPreviewMode option must be carried over
+    // to the project for overrides to work properly.
     // Nested object options are not deep merged in Apostrophe.
-    devicePreviewMode: {
-      // Enable device preview mode
-      enable: false,
+    breakpointPreviewMode: {
+      // Enable breakpoint preview mode
+      enable: true,
```
boutell, line 54: Please tell me if we're missing something, e.g. we think this feature is highly likely to break somebody's build completely if we default it to true. Otherwise we really want people to have this experience out of the box

haroun: We do some css transformations on css declarations inside media queries.
If you have 

```css
@media (max-width: 700px) {
  main {
    width: 100%;
  }
}
```

you will end up with

```css
@media (max-width: 700px) {
  :where(:not(body[data-breakpoint-preview-mode])) main {
    width: 100%;
  }
}
@container (max-width: 700px) {
  main {
    width: 100%;
  }
}
```

`:where` is widely supported https://developer.mozilla.org/en-US/docs/Web/CSS/:where

I should be safe to use.

haroun: that was the main driver for setting `enabled: false` by default.

modules/@apostrophecms/asset/index.js
haroun: I'm doing the changes, releasing with the two names will only make our life harder.

BoDonkey (approved):
LGTM

PR #4746, opened 2024-10-06
warnDev now prefixes in the expected way

boutell:
Before:

image

After:

image

This regressed when we introduced structured logging and has looked kind of doofy ever since.


boutell:
(In my second screenshot I had one extra warnDev call if you're wondering)

boutell:
Fixed tests, documented behavior

PR #4794, opened 2024-11-01
HMR condition argument and widget player fix

myovchev:
## Summary

* Widget players are now properly invoked when they arrive later in the page load process.
* It's possible now to target the HMR build when registering via `template.append` and `template.prepend`. Use `when: 'hmr:public'` or `when: 'hmr:apos'` that will be evaluated against the current asset `options.hmr` configuration.

Players should always run during the initial page load.

The following injection:
```js
self.apos.template.prepend({
  where: 'head',
  when: 'hmr:public',
  bundler: 'vite',
  component: 'vite-react:reactRefresh'
});
```
should inject the component only when HMR is on and the asset module option `hmr` is set to `public`. The generic `when: 'hmr`` still works as expected.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/util/ui/src/util.js
+++ b/modules/@apostrophecms/util/ui/src/util.js
@@ -128,7 +128,20 @@ export default () => {
   // THAT ONE WIDGET and NO OTHER. Don't worry about finding the
   // others, we will do that for you and we guarantee only one call per widget.
 
-  apos.util.widgetPlayers = {};
+  apos.util.widgetPlayersConfig = {
+    list: {},
+    initialized: false
+  };
+  apos.util.widgetPlayers = new Proxy(apos.util.widgetPlayersConfig.list, {
+    set(target, prop, value) {
+      target[prop] = value;
+      // run the player if we missed the initial run
+      if (apos.util.widgetPlayersConfig.initialized) {
+        apos.util.runPlayers(document, { [prop]: value });
+      }
+      return true;
+    }
+  });
 
   // Run the given function whenever the DOM has new changes that
   // may require attention. The passed function will be
@@ -164,6 +177,23 @@ export default () => {
   // Alias for onReadyAndRefresh, the recommended way to use and document this functionality
   apos.util.onReady = apos.util.onReadyAndRefresh.bind(apos.util.onReadyAndRefresh);
 
+  // Debounce a function execution. Only the last call within the
+  // given timeout will be executed.
+  // Example:
+  // const debounced = apos.util.debounce((msg) => console.log(msg), 1000);
+  // debounced('Hello');
+  // debounced('World');
+  // Only 'World' will be printed after 1 second.
+  apos.util.debounce = function (fn, timeout) {
+    let timer;
+    return (...args) => {
+      timer && clearTimeout(timer);
+      timer = setTimeout(() => {
+        fn(...args);
+      }, timeout);
+    };
+  };
+
   // Run all the players that haven't been run. Invoked for you at DOMready
   // time. You may also invoke it if you just AJAXed in some content and
   // have reason to suspect there could be widgets in there. You may pass:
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/util/ui/src/util.js
```diff
@@ -128,7 +128,20 @@ export default () => {
...
+  const widgetPlayersConfig = {
+    list: {},
+    initialized: false
+  };
+  apos.util.widgetPlayers = new Proxy(widgetPlayersConfig.list, {
+    set(target, prop, value) {
+      target[prop] = value;
+      // run the player if we missed the initial run
+      if (widgetPlayersConfig.initialized) {
+        apos.util.runPlayers(document, { [prop]: value });
```
boutell, line 140: runPlayers doesn't take a second argument. It sounds like you're suggesting one unofficially but I'd rather we not have this bite us later if we add one without looking at this. The lack of specificity won't hurt because elements already played won't get played again.

boutell (changes requested):

modules/@apostrophecms/util/ui/src/util.js
myovchev: This is a guard against attempting all players again. The internal player guard is not working well in many cases. We also need to look at the refresh calls in the Context Vue component someday, it causes some of the issues I observe @boutell

myovchev: See `apos.util.runPlayers = function (el, newPlayer) {`. We end up with players being called back to back and breaking a lot of stuff. Again, now we have a lot more room for racing conditions and it bites us (mostly in edit mode, when we have also ajax dom updates from what it seems and manipulate the dom while players are attempting to run).

modules/@apostrophecms/util/ui/src/util.js
```diff
@@ -184,24 +197,24 @@ export default () => {
   // DON'T try to find all the widgets. DO just enhance `el`.
   // This is a computer science principle known as "separation of concerns."
 
-  apos.util.runPlayers = function(el) {
-    const players = apos.util.widgetPlayers;
+  apos.util.runPlayers = function (el, newPlayer) {
```
boutell, line 200: Should be `newPlayers`

boutell: I still think we're fighting ghosts here, the real problem is players that replace the element completely, which we should document you should not do because it prevents tracking if players have been run (or your player must check that on its own if it insists on replacing the element)

But fine with this optimization if we name the parameter better

boutell (changes requested):

modules/@apostrophecms/util/ui/src/util.js
myovchev: I tested on players that are not replacing the element, but mounting (inserting child nodes). The player guard itself is mostly not working because the rapid event fire and DOM changes racing conditions. I spent a good amount of time debugging the whole. I think with this change, players are doing good on itself, eliminating the same tick executions during initialization. The next thing we might want to insect someday is the Vue context component `refresh` calls, that in some cases are spamming the refresh event - it should be probably better computed and/or slightly throttled.

myovchev: I came up with a better strategy I think. I like it more because it gives no "freedom" to pass any player outside of already registered, but still does improve the performance and guards against "player spamming".

modules/@apostrophecms/util/ui/src/util.js
```diff
@@ -183,25 +196,37 @@ export default () => {
   // Your player is guaranteed to run only once per widget. Hint:
   // DON'T try to find all the widgets. DO just enhance `el`.
   // This is a computer science principle known as "separation of concerns."
+  //
+  // The second argument is an options object. If the `init` option is true, only
```
boutell, line 200: Sure, but the name is quite confusing. I suggest that you call this option `newPlayersOnly` and have it filter when set to `true`. I also suggest this language:

"If this option is `true`, only widget player types whose players have not been invoked before will be invoked."

boutell (changes requested):

modules/@apostrophecms/util/ui/src/util.js
myovchev: Did some more testing and debugging... and reverted the whole new players thing. The performance gain is questionable, we need a more clear strategy for improvement here.

boutell (approved):

PR #4815, opened 2024-11-24
PRO-4431: Context menu dynamic focus trap [a11y]

myovchev:
## Summary

* Adds support for dynamic focus trap in Context menus (prop `dynamicFocus`). When set to `true`, the focusable elements are recalculated on each cycle step.
* Adds option to disable `tabindex` on `AposToggle` component. A new prop `noFocus` can be set to `false` to disable the focus on the toggle button. It's enabled by default.

Test on Document Versions modal context menu View Options. 

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/ui/ui/apos/components/AposToggle.vue
+++ b/modules/@apostrophecms/ui/ui/apos/components/AposToggle.vue
@@ -2,7 +2,7 @@
   <div class="apos-toggle__container">
     <div
       class="apos-toggle__slider"
-      tabindex="0"
+      :tabindex="noFocus ? null : '0'"
       :class="{'apos-toggle__slider--activated': !modelValue}"
       @click="$emit('toggle')"
       @keydown.stop.space="$emit('toggle')"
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/ui/ui/apos/components/AposToggle.vue
```diff
@@ -18,6 +18,10 @@ export default {
     modelValue: {
       type: Boolean,
       required: true
+    },
+    noFocus: {
```
stuartromanek, line 22: these props are usually named `focus` or `disableFocus`... does it make sense to rename this?

stuartromanek (commented):
Instead of re-querying on tab, should we be using a [mutation observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) on the element we want to confine focus to and querying then?

stuartromanek (approved):

PR #4816, opened 2024-11-26
Pro 6799 design default values invisible fields

ValJed:
[PRO-6528](https://linear.app/apostrophecms/issue/PRO-6528/find-out-whether-the-bug-in-select-is-a-project-level-bug-or-a-bug-in)

## Summary

* Adds control over when fields are ready and can be validated, by handling a new `fieldReady` property for each field.
Must work with async fields like dynamic choices, being able to set a field not ready during fetching and ready when operation is finished.
* Reworks schema level convert method. Check errors only when on root convert method to have access to the finale destination object to be able to check non visible fields, as well as non visible parents. No more order problem..
* Adds `aposPath` to fields during validation process.
* Adds tests.

[Cypress tests](https://github.com/apostrophecms/testbed/actions/runs/12549916545) 🟢 

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,9 @@
 
 ### Fixes
 
-<<<<<<< HEAD
 * Fixes a bug where images in Media manager are not selectable (click on an image does nothing) in both default and relationship mode.
-=======
 * In the schema convert method, we wait all sub convert to run, to have access to the final destination object. In order to check if 
 fields (or their parents) are visible and so related errors are discarded.
->>>>>>> e1fd5bd94 (changelog typo)
 
 ### Adds
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,12 @@
 
 ### Fixes
 
+<<<<<<< HEAD
 * Fixes a bug where images in Media manager are not selectable (click on an image does nothing) in both default and relationship mode.
+=======
+* In the schema convert method, we wait all sub convert to run, to have access to the final destination object. In order to check if 
+fields (or their parents) are visible and so related errors are discarded.
+>>>>>>> e1fd5bd94 (changelog typo)
 
 ## 4.11.1 (2024-12-18)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ did not actually use any noncompliant cookie names or values, so there was no vu
 rendered output.
 * Search bar will perform the search even if the bar is empty allowing to reset a search.
 * Fixes Color picker being hidden in an inline array schema field, also fixes rgba inputs going off the modal.
+* Simplifies the `convert` method from schema module.
 
 ### Adds
 
@@ -56,6 +57,7 @@ rendered output.
 * Adds `bundleMarkup` to the data sent to the external front-end, containing all markup for injecting Apostrophe UI in the front-end.
 * Warns users when two page types have the same field name, but a different field type. This may cause errors or other problems when an editor switches page types.
 * The piece and page `GET` REST APIs now support `?render-areas=inline`. When this parameter is used, an HTML rendering of each widget is added to that specific widget in each area's `items` array as a new `_rendered` property. The existing `?render-areas=1` parameter is still supported to render the entire area as a single `_rendered` property. Note that this older option also causes `items` to be omitted from the response.
+* Possibility to set a field not ready when performing async operations, when a field isn't ready, the validation and emit won't occur.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,6 @@
 * Adds option to disable `tabindex` on `AposToggle` component. A new prop `disableFocus` can be set to `false` to disable the focus on the toggle button. It's enabled by default.
 * Adds support for event on `addContextOperation`, an option `type` can now be passed and can be `modal` (default) or `event`, in this case it does not try to open a modal but emit a bus event using the action as name.
 
-
 ### Fixes
 
 * Focus properly Widget Editor modals when opened. Keep the previous active focus on the modal when closing the widget editor.
@@ -48,7 +47,6 @@ did not actually use any noncompliant cookie names or values, so there was no vu
 rendered output.
 * Search bar will perform the search even if the bar is empty allowing to reset a search.
 * Fixes Color picker being hidden in an inline array schema field, also fixes rgba inputs going off the modal.
-* Simplifies the `convert` method from schema module.
 
 ### Adds
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,11 @@
 ### Fixes
 
 * Fixes a bug where images in Media manager are not selectable (click on an image does nothing) in both default and relationship mode.
-* In the schema convert method, we wait all sub convert to run, to have access to the final destination object. In order to check if 
-fields (or their parents) are visible and so related errors are discarded.
+* Eliminated superfluous error messages. The convert method now waits for all recursive invocations to complete before attempting to determine if fields are visible.
 
 ### Adds
 
-* Possibility to set a field not ready when performing async operations, when a field isn't ready, the validation and emit won't occur.
+* Possibility to set a field not ready when performing async operations, when a field isn't ready, the validation and emit won't occur. 
 
 ## 4.11.1 (2024-12-18)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 fields (or their parents) are visible and so related errors are discarded.
 >>>>>>> e1fd5bd94 (changelog typo)
 
+### Adds
+
+* Possibility to set a field not ready when performing async operations, when a field isn't ready, the validation and emit won't occur.
+
 ## 4.11.1 (2024-12-18)
 
 ### Fixes
@@ -60,7 +64,6 @@ rendered output.
 * Adds `bundleMarkup` to the data sent to the external front-end, containing all markup for injecting Apostrophe UI in the front-end.
 * Warns users when two page types have the same field name, but a different field type. This may cause errors or other problems when an editor switches page types.
 * The piece and page `GET` REST APIs now support `?render-areas=inline`. When this parameter is used, an HTML rendering of each widget is added to that specific widget in each area's `items` array as a new `_rendered` property. The existing `?render-areas=1` parameter is still supported to render the entire area as a single `_rendered` property. Note that this older option also causes `items` to be omitted from the response.
-* Possibility to set a field not ready when performing async operations, when a field isn't ready, the validation and emit won't occur.
 
 ### Changes
 
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -682,6 +682,7 @@ module.exports = {
             const isVisible = await self.isVisible(req, schema, destination, field.name);
             if (!isVisible) {
               nonVisibleFields.add(curPath);
+              continue;
             }
             if (!field.schema) {
               continue;
--- a/modules/@apostrophecms/schema/index.js
+++ b/modules/@apostrophecms/schema/index.js
@@ -24,7 +24,6 @@ module.exports = {
     alias: 'schema'
   },
   init(self) {
-
     self.fieldTypes = {};
     self.fieldsById = {};
     self.arrayManagers = {};
@@ -585,20 +584,22 @@ module.exports = {
         {
           fetchRelationships = true,
           ancestors = [],
-          isParentVisible = true
+          rootConvert = true,
+          ancestorSchemas = {},
+          ancestorPath = ''
         } = {}
       ) {
         const options = {
           fetchRelationships,
           ancestors,
-          isParentVisible
+          ancestorSchemas,
+          ancestorPath
         };
         if (Array.isArray(req)) {
           throw new Error('convert invoked without a req, do you have one in your context?');
         }
 
-        const errors = [];
-
+        const convertErrors = [];
         for (const field of schema) {
           if (field.readOnly) {
             continue;
@@ -611,15 +612,11 @@ module.exports = {
           }
 
           const { convert } = self.fieldTypes[field.type];
-
           if (!convert) {
             continue;
           }
 
-          const isCurrentVisible = isParentVisible === false
-            ? false
-            : await self.isVisible(req, schema, data, field.name);
-
+          const fieldPath = ancestorPath ? `${ancestorPath}/${field.name}` : field.name;
           try {
             const isRequired = await self.isFieldRequired(req, field, destination);
             await convert(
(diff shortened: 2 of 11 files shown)


modules/@apostrophecms/schema/index.js
```diff
@@ -632,6 +632,15 @@ module.exports = {
                 }
               );
             } catch (error) {
+              const isVisible = isParentVisible === false
```
boutell, line 635: Thank you, I did wonder if there was a cleaner way.

boutell (commented):
I like where you're going with this.

modules/@apostrophecms/schema/index.js
```diff
@@ -612,90 +612,81 @@ module.exports = {
...
-            }
+          if (!convert) {
+            continue;
           }
-        }
 
-        const errorsList = [];
+          const isCurrentVisible = isParentVisible === false
+            ? false
+            : await self.isVisible(req, schema, data, field.name);
```
ValJed, line 621: Notable difference here (previous versions might be buggy). We pass data instead of destination because destination is not fully built at this point.
I think `isVisible` should compare data as they have been set by  the user and not the potential default values.
Tests are green with it but you might have some reasons to not allow this.
Let me know.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
```diff
@@ -52,6 +52,7 @@
         :server-error="fields[field.name].serverError"
         :doc-id="docId"
         :generation="generation"
+        @update:model-value="updateNextAndEmit"
```
ValJed, line 55: With this, no more need for a deep watcher.

modules/@apostrophecms/schema/ui/apos/logic/AposSchema.js
```diff
@@ -282,20 +276,23 @@ export default {
...
-          this.next.data[field.name] = this.modelValue.data[field.name];
-        }
-      });
+          ) {
+            changeFound = true;
+            this.next.data[field.name] = this.fieldState[field.name].data;
+          } else {
+            this.next.data[field.name] = this.modelValue.data[field.name];
+          }
+        });
```
ValJed, line 295: Only indentation in this block

modules/@apostrophecms/schema/index.js
boutell: Data can be unvalidated stuff of unexpected types etc., it could be anything. Trusting it is problematic. You might get away with that in `isVisible` because you're not saving the values, but then `isVisible` must be written carefully to not assume the values are at all well-formed, for instance it must not assume string fields are actually strings yet, etc. because nothing has been validated.

boutell (changes requested):
It is possible that changes are not necessary but see my note here about trusting `data`, `isVisible` must be written very defensively if it is going to see unvalidated user input.

ValJed:
@boutell @haroun Code could be cleaned but I would like your opinion on the general direction.

modules/@apostrophecms/schema/index.js
```diff
@@ -611,92 +620,204 @@ module.exports = {
...
+
+        async function handleConvertErrors({
+          req,
+          schema,
+          convertErrors,
+          nonVisibleFields,
+          destination,
+          destinationPath = '',
+          hiddenAncestors = false
+        }) {
```
ValJed, line 728: This method might be moved somewhere else

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 * Focus properly Widget Editor modals when opened. Keep the previous active focus on the modal when closing the widget editor.
 * a11y improvements for context menus.
 * Fixes broken widget preview URL when the image is overridden (module improve) and external build module is registered.
+* In th schema convert method, we wait all sub convert to run, to have access to the final destination object in order to check if
```
boutell, line 10: typo

modules/@apostrophecms/schema/index.js
```diff
@@ -585,20 +592,22 @@ module.exports = {
...
+          ancestorSchemas = {},
+          ancestorPath = ''
         } = {}
       ) {
         const options = {
           fetchRelationships,
           ancestors,
-          isParentVisible
+          ancestorSchemas,
+          ancestorPath
```
boutell, line 604: I think it is better to compute the `aposPath` of a field once, in the `validate` method of `schema` and its related functions, rather than pushing this boulder up the hill on each convert call 🪨

modules/@apostrophecms/schema/index.js
```diff
@@ -611,92 +620,204 @@ module.exports = {
...
+              // If invalid error has no sub error, this one can be removed
+              if (!subErrors.length) {
+                continue;
+              }
+
+              error.data.errors = subErrors;
+            }
+
+            if (typeof error !== 'string') {
+              self.apos.util.error(error.path + '\n' + error.stack);
```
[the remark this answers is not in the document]
boutell: I think it is better to log this detail later when logging the overall error. But I wouldn't die on that hill.

Will the error type ever be a string? I believe that went away in A3/A4...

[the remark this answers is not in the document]
boutell: I guess it's smart to check.

boutell (changes requested):
I like the general strategy. Still think it makes more sense to set `aposPath` on each `field` just once in `schema.validate` and its related functions.

modules/@apostrophecms/schema/index.js
ValJed: Yes good point. We agree that we are talking about the schema path and not the actual destination one.

modules/@apostrophecms/schema/index.js
```diff
@@ -611,92 +616,208 @@ module.exports = {
...
               }
-            }
+            );
+          } catch (err) {
+            const error = Array.isArray(err)
+              ? self.apos.error('invalid', { errors: err })
+              : err;
+
+            error.path = field.name;
+            error.schemaPath = field.aposPath;
```
ValJed, line 644: `aposPath` is now built for each schema in the `validate` method.

modules/@apostrophecms/schema/index.js
```diff
@@ -611,92 +616,207 @@ module.exports = {
...
+          req,
+          schema,
+          convertErrors,
+          destination,
+          nonVisibleFields
+        });
+
+        for (const error of validErrors) {
+          self.apos.util.error(error.stack);
+        }
```
ValJed, line 673: Moved log of errores out of `handleErrors` method.
Created a ticket to improve globally message printed from convert errors.

modules/@apostrophecms/schema/index.js
```diff
@@ -611,92 +616,207 @@ module.exports = {
...
+            : error.path;
+
+          // Case were this error field hasn't been treated
+          // Should check if path starts with, because parent can be invisible
+          const nonVisibleField = hiddenAncestors || nonVisibleFields.has(errorPath);
+
+          // We set default values only on final error fields
+          if (nonVisibleField && !error.data?.errors) {
+            const curSchema = self.getFieldLevelSchema(schema, error.schemaPath);
+            self.setDefaultToInvisibleField(curDestination, curSchema, error.path);
```
ValJed, line 749: I'm not sure this method is really useful.
I wrote a test for it and realized I saw it work only for a string that has a pattern.
In the case of in integer having a `min` or a `max` for example, the convert updates the value itself, it does not need this method.
Maybe there are some cases I didn't test.

modules/@apostrophecms/schema/index.js
```diff
@@ -611,92 +616,207 @@ module.exports = {
...
+        // is corrected
+        const field = schema.find(field => field.name === fieldName);
+        if (field) {
+          // To protect against security issues, an invalid value
+          // for a field that is not visible should be quietly discarded.
+          // We only worry about this if the value is not valid, as otherwise
+          // it's a kindness to save the work so the user can toggle back to it
+          destination[field.name] = klona((field.def !== undefined)
+            ? field.def
+            : self.fieldTypes[field.type]?.def);
```
ValJed, line 798: As descibed above, not sure where this method is useful except for strings have a `pattern` property.

modules/@apostrophecms/schema/index.js
```diff
@@ -705,7 +825,13 @@ module.exports = {
...
             if (field.if) {
               try {
-                const result = await self.evaluateCondition(req, field, field.if, object, conditionalFields);
+                const result = await self.evaluateCondition(
+                  req,
+                  field,
+                  field.if,
+                  destination,
+                  conditionalFields
+                );
```
ValJed, line 834: Just linting

modules/@apostrophecms/schema/index.js
```diff
@@ -1332,19 +1458,23 @@ module.exports = {
...
+            self.validateField(field, options, parent);
           }
         });
       },
 
       // Validates a single schema field. See `validate`.
       validateField(field, options, parent = null) {
+        field.aposPath = parent
+          ? `${parent.aposPath}/${field.name}`
+          : field.name;
```
ValJed, line 1476: Build `aposPath` by using parent `aposPath`

modules/@apostrophecms/schema/index.js
```diff
@@ -1332,19 +1458,23 @@ module.exports = {
...
 
-      validate(schema, options) {
+      validate(schema, options, parent = null) {
         schema.forEach(field => {
           // Infinite recursion prevention
           const key = `${options.type}:${options.subtype}.${field.name}`;
           if (!self.validatedSchemas[key]) {
             self.validatedSchemas[key] = true;
-            self.validateField(field, options);
+            self.validateField(field, options, parent);
```
ValJed, line 1467: `relationship` field uses `validate`u an not `validateField` like does `array` and `object` fields.
Maybe the three should use the validate method since we can now pass the parent?

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -1067,15 +1067,16 @@ module.exports = (self) => {
...
           const result = results.find(doc => (doc._id === item._id));
           if (result) {
             if (field.schema) {
+              const destItem = (Array.isArray(destination[field.name]) ? destination[field.name] : [])
+                .find((doc) => doc._id === item._id);
               result._fields = {
-                ...(destination[field.name]
-                  ?.find?.(doc => doc._id === item._id)
-                  ?._fields || {})
+                ...destItem?._fields || {}
```
ValJed, line 1077: Just make code more readable.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -1218,7 +1219,7 @@ module.exports = (self) => {
         self.validate(_field.schema, {
           type: 'relationship',
           subtype: _field.withType
-        });
+        }, _field);
```
ValJed, line 1222: Passing the parent to `validate`. Should we modify `object` and `array` to use the same `validate` method instead of `validateField`. (the validate method keeps tracks of already validated fields). I think it would make sense to use the same method for `relationship`, `object` and `array`.

modules/@apostrophecms/schema/index.js
ValJed: It now uses destination when eveything as been converted.

modules/@apostrophecms/schema/ui/apos/components/AposSchema.vue
ValJed: To keep the behavior consistent, `fieldState` was watched before, I just replaced the watch by this. `compareMetaState` wasn't.

modules/@apostrophecms/schema/index.js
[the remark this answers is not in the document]
ValJed: Idk, this was existing code, I removed it and moved the logs out of `handleErrors` function.

modules/@apostrophecms/schema/index.js
ValJed: Done, it's now computed during `validate`.

modules/@apostrophecms/schema/index.js
boutell: If the field is not visible and never gets touched in the UI, doesn't the frontend submit `def` anyway? It should.

boutell (changes requested):

modules/@apostrophecms/schema/index.js
ValJed: Yes I did not touch the existing line:
```
error.path = field.name;
```
It still happens

boutell: my bad on that one.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 ### Fixes
 
 * Fixes a bug where images in Media manager are not selectable (click on an image does nothing) in both default and relationship mode.
+* In the schema convert method, we wait all sub convert to run, to have access to the final destination object. In order to check if
```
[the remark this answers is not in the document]
ValJed: Done

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 ### Fixes
 
 * Fixes a bug where images in Media manager are not selectable (click on an image does nothing) in both default and relationship mode.
+* In the schema convert method, we wait all sub convert to run, to have access to the final destination object. In order to check if 
+fields (or their parents) are visible and so related errors are discarded.
+
+### Adds
+
+* Possibility to set a field not ready when performing async operations, when a field isn't ready, the validation and emit won't occur.
```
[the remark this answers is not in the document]
ValJed: Done

modules/@apostrophecms/schema/index.js
ValJed: I don't know, maybe it does.
My question was more about in what cases this is useful except for invisible string fields having a pattern that is not respected in the existing data.
I'm asking you because this function simply wraps existing code from you.
Also because you evoke min and max in the comments but I realized the convert already fixed min and max doe invisible fields.
```javascript
      setDefaultToInvisibleField(destination, schema, fieldPath) {
        // Field path might contain the ID of the object in which it is contained
        // We just want the field name here
        const [ _id, fieldName ] = fieldPath.includes('.')
          ? fieldPath.split('.')
          : [ null, fieldPath ];
        // It is not reasonable to enforce required,
        // min, max or anything else for fields
        // hidden via "if" as the user cannot correct it
        // and it will not be used. If the user changes
        // the conditional field later then they won't
        // be able to save until the erroneous field
        // is corrected
        const field = schema.find(field => field.name === fieldName);
        if (field) {
          // To protect against security issues, an invalid value
          // for a field that is not visible should be quietly discarded.
          // We only worry about this if the value is not valid, as otherwise
          // it's a kindness to save the work so the user can toggle back to it
          destination[field.name] = klona((field.def !== undefined)
            ? field.def
            : self.fieldTypes[field.type]?.def);
        }
      }
```
So I'm just wondering how useful is it, is it worth keeping it etc. Also how to properly test it.

boutell (approved):

haroun (approved):

PR #4819, opened 2024-12-11
Pro 6843 context menu additional items

ValJed:
[PRO-6843](https://linear.app/apostrophecms/issue/PRO-6843/as-a-user-i-want-to-be-able-to-move-palette-so-that-i-can-see-elements)

## Summary

I need to add in palette a new item in the context menu.
This item should not open any modal but just emit an event I could listen to know user clicked on it. So I cannot use `customOperations` here.

I just added the possibility to pass an array of extra items to a `AposDocContextMenu` component.
This one can be used to call an action like we already do, or to emit a simple event when closing to the parent, like I do with palette.

It allows me to do this in palette:
```javascript
          <AposDocContextMenu
            :doc="docFields.data"
            :show-edit="false"
            :additional-menu-items="[{
              label: 'aposPalette:resetPalettePosition',
              action: 'reset-position',
              emitEvent: true
            }]"
            @close="menuCloseHandler"
          />
       ...
    menuCloseHandler(_doc, action) {
      if (action === 'reset-position') {
        this.resetPosition();
      }
    }
```
If I pass `emitEvent` it will simply emit close with the `action` as parameter.

See above

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Adds support for supplying CSS variable names to a color field's `presetColors` array as selectable values.
 * Adds support for dynamic focus trap in Context menus (prop `dynamicFocus`). When set to `true`, the focusable elements are recalculated on each cycle step.
 * Adds option to disable `tabindex` on `AposToggle` component. A new prop `disableFocus` can be set to `false` to disable the focus on the toggle button. It's enabled by default.
+* Adds new prop `additionalMenuItems` to `AposDocContextMenu`. Allow to pass additional menu items that can trigger an action when clicked, or simply pass the action as the close event parameter.
 
 ## 4.10.0 (2024-11-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 * Adds support for supplying CSS variable names to a color field's `presetColors` array as selectable values.
 * Adds support for dynamic focus trap in Context menus (prop `dynamicFocus`). When set to `true`, the focusable elements are recalculated on each cycle step.
 * Adds option to disable `tabindex` on `AposToggle` component. A new prop `disableFocus` can be set to `false` to disable the focus on the toggle button. It's enabled by default.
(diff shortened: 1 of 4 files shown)


haroun (dismissed):

boutell (changes requested):
There is already a way to register additional context operations server side. The palette module can call:

```
self.apos.doc.addContextOperation({
  action: 'reset-palette-position',
  if: {
    type: self.__meta.type
  }  
});
```

This will make the operation available only for palette.

So there is no need to introduce a redundant way of registering custom context operations.

However I do think you may need a patch to allow them to trigger a bus event rather than opening a modal. This is the current logic that implements the "operation" property:

```
      const props = {
        moduleName: operation.moduleName || this.moduleName,
        // For backwards compatibility
        doc,
        ...docProps(doc),
        ...operation.props
      };
      await apos.modal.execute(operation.modal, props);
```

This will need an update to support emitting a bus event with an object containing the props as an alternative.

boutell:
(Excuse me, `self.__meta.name`, which is just a cute way of saying this module's name, you could also hardcode the string)

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Adds support for supplying CSS variable names to a color field's `presetColors` array as selectable values.
 * Adds support for dynamic focus trap in Context menus (prop `dynamicFocus`). When set to `true`, the focusable elements are recalculated on each cycle step.
 * Adds option to disable `tabindex` on `AposToggle` component. A new prop `disableFocus` can be set to `false` to disable the focus on the toggle button. It's enabled by default.
+* Adds new prop `additionalMenuItems` to `AposDocContextMenu`. Allow to pass additional menu items that can trigger an action when clicked, or simply pass the action as the close event parameter.
```
boutell, line 20: Change

boutell (approved):
Approving for time but please fix changelog

boutell (approved):

PR #4842, opened 2025-01-21
PRO-6856: batch localization

myovchev:
## Summary

**A change against a feature branch.**

Captures the PRO-6856 requirements but it's still missing the mocha tests.


boutell (commented):
General direction looks good.

PR #4857, opened 2025-02-06
PRO-6876: implicit properties for reverse relationship projectionsnship projections

boutell:
See changelog

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -1751,7 +1751,9 @@ module.exports = {
             const add = [];
             const remove = [];
 
-            // Add type in projection by default
+            // Add fundamentals by default to allow Apostrophe's core
+            // operations to find what they expect (determining document type,
+            // distinguishing documents from widgets)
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -1751,7 +1751,9 @@ module.exports = {
             const add = [];
             const remove = [];
 
-            // Add type in projection by default
+            // Add fundamentals by default to allow Apostrophe's core
```
boutell, line 1754: Unrelated update to clarify what happens here since it is more than just tpe

test/relationships.js
```diff
@@ -145,7 +148,8 @@ function getModules() {
             label: 'Rel page',
             type: 'relationship',
             withType: 'default-page',
-            withRelationships: [ '_articles._topics' ],
+            // Intentionally revisits the articles in order to test reverse relationships
```
boutell, line 151: Normally it would be silly to loop back like this but it's a simple way to test the reverse relationship.

PR #4871, opened 2025-02-24
Pro 7063 richtext csv

ValJed:
[PRO-7063](https://linear.app/apostrophecms/issue/PRO-7063/as-a-user-i-want-to-upload-a-csv-file-as-a-table-in-the-rich-text)

## Summary

Add support for importing csv files as tables in tiptap.

Try to add `imporTable` to your rich text _toolbar_ (supported since it supports tables too..) and _insert_. 
Both should work properly.

Cypress running [HERE](https://github.com/apostrophecms/testbed/actions/runs/13523219662) 🟢 

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


modules/@apostrophecms/rich-text-widget/index.js
```diff
@@ -298,6 +299,11 @@ module.exports = {
...
         label: 'apostrophe:richTextColor',
         command: 'setColor'
+      },
+      import: {
+        component: 'AposTiptapImportData',
+        icon: 'cloud-upload-icon',
+        label: 'Import CSV file' // TODO: i18n
       }
     },
     editorInsertMenu: {
```
boutell, line 309: Belongs here.

boutell (changes requested):
This should not be a button in the toolbar. That would replace the current selection, and be impossible to access if there is no selection. That's why options that insert things go on the insert menu instead.

It should only be a choice on the insert menu (accessed with the `/` key) if configured. e.g.:

```
// in some area
'@apostrophecms/rich-text-widget': {
  // note that regular 'table' is listed in both places because existing tables have
  // editing operations
  toolbar: [ 'bold', 'italic', 'table' ],
  insert: [ 'table', 'image', 'importTable' ]
}
```

See `editorInsertMenu` in `rich-text-widget/index.js`.

You can register a modal component to be displayed or a tiptap command to be immediately run. 

Let me know if you want to talk that through.

modules/@apostrophecms/notification/ui/apos/components/AposNotification.vue
```diff
@@ -247,10 +247,6 @@ function localize(s) {
   }
 }
 
-.apos-notification__close-icon {
-  height: 12px;
-}
```
ValJed, line 252: The close btn was badly displayed in notifications (should be more that 14px)

ValJed:
@boutell PR is still in draft so not ready for review.
I added support for insert menu. But don't know why it should not be available from toolbar like image and table? It should be available everywhere table are available.

modules/@apostrophecms/i18n/i18n/sk.json
```diff
@@ -570,6 +571,7 @@
   "updateAndReturn": "{{ saveLabel }} aktualizácie a vráťte sa do {{ typeLabel }} výpis.",
   "updateSingleton": "{{ saveLabel }} {{ typeLabel}}.",
   "updateTitle": "Aktualizovať názov",
+  "updateTooltip": "Publikovať aktualizácie",
   "updated": "Aktualizované",
   "uploadError": "Chyba nahrávania",
   "uploadForbidden": "Na nahranie súboru nemáte povolenie",
```
ValJed, line 577: Adding keys to all locale and sorting

boutell (approved):
I guess you're right.

After thinking it over, my new reasoning is:

* We already have horizontal rules enabled both ways...
* Even though there's nothing to "edit" about an existing horizontal rule...
* Because someone might want to select something and *replace* it with a horizontal rule, which is what an "insert" when there is already a selection does in tiptap and most editors.
* So it's fine to have "import table" work in both places too.

ValJed:
⚠️  Waiting for cypress tests to be green before to merge.

PR #4872, opened 2025-02-24
Implemented API endpoint to return current users info

sombitganguly:
Implemented GET /api/v1/@apostrophecms/login/whoami route such that it returns the details of the current user.

### DIFF
--- a/modules/@apostrophecms/login/index.js
+++ b/modules/@apostrophecms/login/index.js
@@ -67,7 +67,9 @@ module.exports = {
       allowedAttempts: 3,
       perMinutes: 1,
       lockoutMinutes: 1
-    }
+    },
+    minimumWhoamiFields: ['_id', 'username', 'title', 'email'],
+    whoamiFields: []
   },
   async init(self) {
     self.passport = new Passport();
@@ -346,6 +348,22 @@ module.exports = {
         // it should be accessed via POST because the result
         // may differ by individual user session and should not
         // be cached
+        async whoami (req) {
+          if(!req.user){
+            throw self.apos.error('notfound')
+          }
+
+          const fields = new Set([...self.options.minimumWhoamiFields, ...self.options.whoamiFields])
+          const user = {}
+
+          for (const field of fields){
+            if(req.user[field] !== undefined){
+              user[field] = req.user[field]
+            }
+          }
+
+          return user
+        },
         async context(req) {
           return self.getContext(req);
         },
@@ -361,7 +379,7 @@ module.exports = {
             self.apos.util.error(e);
             throw self.apos.error('invalid', req.t('apostrophe:loginResetInvalid'));
           }
-        }
+        },
       }
     };
   },


BoDonkey:
I haven't really checked out the code yet, but can you add two things to this PR:
1) Update the CHANGELOG to reflect the changes being added - this should be added to the `## UNRELEASED` section.
2) Add tests to show correct functionality.
Let me know if you need help with either of these and thanks again!

modules/@apostrophecms/login/index.js
```diff
@@ -346,6 +348,22 @@ module.exports = {
         // it should be accessed via POST because the result
         // may differ by individual user session and should not
         // be cached
+        async whoami (req) {
```
boutell, line 351: Please run `npx eslint . --fix` to correct minor whitespace and code style issues without pain 😄

boutell (changes requested):
Great stuff. Please run `npx eslint . --fix` to update the code style to match probably 100% of our guidelines without manual effort. Also please add it to the changelog.

sombitganguly:
> Great stuff. Please run `npx eslint . --fix` to update the code style to match probably 100% of our guidelines without manual effort. Also please add it to the changelog.

I have added it to the changelog as well as added test cases as instructed.

sombitganguly:
I have made all the requested changes, kindly review them. Thank you!

modules/@apostrophecms/login/index.js
```diff
@@ -346,6 +348,25 @@ module.exports = {
...
+          if (!req.user) {
+            throw self.apos.error('notfound');
+          }
+
+          // The query will contain comma separated additional field names
+          const requestedFields = req.query.additionalFields ? req.query.additionalFields.split(',') : [];
```
boutell, line 357: This is a security hole, so it should not be allowed, only configured fields should be returned by this API.

boutell (changes requested):
Looking really good otherwise, but allowing the API to fetch any properties it wants is a security hole, so that part needs to come out.

modules/@apostrophecms/login/index.js
sombitganguly: Security hole has been fixed

modules/@apostrophecms/login/index.js
```diff
@@ -353,8 +353,11 @@ module.exports = {
             throw self.apos.error('notfound');
           }
 
+          const allowedFields = new Set([ 'role' ]);
```
boutell, line 356: I apologize for the confusion, which started on our end because we didn't discuss the security implications enough internally first.

We don't need this `allowedFields` mechanism, and we don't need `req.query.additionalFields` either, because the only fields that should ever be returned are those in the `minimumWhoamiFields` and `whoamiFields` options. And those should always be returned.

We appreciate the contribution! The good news is you can simplify it now and remove the tests that relate to `req.query.additionalFields` because the feature is not needed.

boutell (changes requested):
Hi

boutell (changes requested):

boutell:
@sombitganguly are you interested in finishing your tests so this can be merged?

sombitganguly:
> @sombitganguly are you interested in finishing your tests so this can be merged?

absolutely, and i apologize for my sudden disappearance i got busy with some stuff.

boutell:
No worries, looking forward to it!

boutell (dismissed):
This is great; thanks so much!

PR #4883, opened 2025-03-13
PRO-7116: Unblock page manager UI while rearranging the tree

myovchev:
## Summary

Discrete and optimized update of the state when reorganizing the tree (without additional round trips)
- Queue the updates
- Use the response data to update the state (no `getAll`)
- Improve the `patch` endpoint to return meaningful data when `targetId` is provided
- Introduce a `expanded index` state to remember the expanded/collapsed nodes even when a full reload is performed (and when moving nodes around)
- Let asyncQueue report if there are ongoing/upcoming tasks
- Let `isModified` know when we have background tasks
- Allow scrolling while dragging a page in the Page Manager.

Drag-n-drop page tree nodes around without any UI blocking.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/page/index.js
```diff
@@ -1625,14 +1641,27 @@ database.`);
             parent = target._ancestors[0];
           }
           async function nudgeNewPeers() {
-            // Nudge down the pages that should now follow us
-            await self.apos.doc.db.updateMany({
+            const criteria = {
```
myovchev, line 1644: @boutell Isn't that a bug? The criteria doesn't involve any locale filter, which would eventually touch pages from a different locale?

boutell: I think you are right, I think there is a bug here.

I think it should work like the correctly written query on the `const modified =` line.

boutell (changes requested):
I see no bugs here, but I think it is worth fixing the bug you found.

If that will delay us we could push it back and open a ticket.

modules/@apostrophecms/page/index.js
myovchev: I only moved the locale criteria, which was the main issue. The moved document ID is mutated after nudging the peers and works correctly so I'll leave it be to not introduce regressions by an accident.

boutell (dismissed):

boutell (approved):
Golden

myovchev:
@boutell And another changelog conflict (see the last and only commit since the review)

PR #4893, opened 2025-03-28
PRO-7380 PRO-7483 Widget live preview

boutell:
See changelog. For the most interesting results, try the `preview-widgets` branch of `gt-samples` (the apostrophe project is in the `backend` subdirectory). In particular, in the "Box" tab try setting "width" to "custom percentage" and then dragging the slider.

Not available for rich text for now.


modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -197,6 +197,10 @@ export default {
     }
   },
   watch: {
+    // Note: please don't make this a deep watcher as that could cause
```
boutell, line 200: No change here, just futureproofing.

modules/@apostrophecms/area/views/area.html
```diff
@@ -1,6 +1,5 @@
 {# area needs its own copy of the widget options as #}
 {# JSON, for adding new widgets #}
-{%- set isSingleton = data.options.limit == 1 and data.options.type -%}
```
boutell, line 3: Unrelated: dead code removal.

modules/@apostrophecms/image-widget/views/widget.html
```diff
@@ -1,4 +1,6 @@
-{% if data.widget.aposPlaceholder and data.manager.options.placeholderUrl %}
+{% set attachment = apos.image.first(data.widget._image) %}
+
+{% if (data.widget.aposPlaceholder or not attachment) and data.manager.options.placeholderUrl %}
```
boutell, line 3: Another use case for the placeholder.

modules/@apostrophecms/video-widget/ui/src/index.js
```diff
@@ -1,4 +1,5 @@
 export default () => {
+  apos.oembedCache ||= {};
```
boutell, line 2: Cache oembed results while we're on the page. Video widget preview implemented in this way will never be perfect, but this reduces the OMG flicker factor.

modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -304,8 +308,8 @@ export default {
         ...this.next.slice(i + 2)
       ];
     },
-    async remove(i) {
-      if (this.docId === window.apos.adminBar.contextId) {
+    async remove(i, { autosave = true } = {}) {
```
boutell, line 311: the widget editor modal now calls back to these methods for preview effects and needs a way to say "don't autosave this, I'm going to clean it up later and revert if the user doesn't go through with it."

modules/@apostrophecms/video-widget/ui/src/index.js
```diff
@@ -49,33 +60,11 @@ export default () => {
         }
         inner.removeAttribute('width');
         inner.removeAttribute('height');
-        el.append(inner);
```
boutell, line 52: so much code removed by using aspect-ratio (standard since 2021)

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -201,11 +209,64 @@ export default {
...
+      }
+      const now = Date.now();
+      const body = () => {
+        this.lastPreview = now;
+        this.preview.area.update(this.getPreviewWidgetObject(), { autosave: false });
+      }
+      if (this.updatePreviewTimeout) {
+        clearTimeout(this.updatePreviewTimeout);
+      }
+      if (!this.lastPreview || (now - this.lastPreview > 250)) {
```
boutell, line 241: Simple debouncing is important because we're relying on server rendering.

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
```diff
@@ -62,7 +62,9 @@ export default {
         if (this.rendering && (isEqual(this.rendering.parameters, parameters))) {
           this.rendered = this.rendering.html;
         } else {
-          this.rendered = '...';
```
boutell, line 65: In retrospect this was never needed and it causes unnecessary flicker

boutell:
A draft until I write cypress tests

boutell:
Not a draft anymore, cypress tests might be deferred or written by someone else. Please do review.

boutell:
I pushed up fixes for the lint error. The lint warnings seem problematic and unrelated, I merged main and I have a ton of them for files that I didn't touch.

ValJed (changes requested):
Mocha tests are red.
Also is there a ticket associated with it? I'm not perfectly clear about steps to test this one.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -110,12 +111,16 @@ export default {
     parentFollowingValues: {
       type: Object,
       default: null
+    },
+    preview: {
+      required: false,
+      type: Object
+      // if present, has "area", "index" and "create" properties
```
ValJed, line 118: You need a default value to avoid eslint warning: `vue/require-default-prop`

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -190,11 +195,14 @@ export default {
...
-      return;
+    } else {
+      this.original = klona(defaults);
+      this.docFields.data = defaults;
     }
-
-    this.original = klona(defaults);
-    this.docFields.data = defaults;
+    if (!this.id) {
+      this.newId = createId();
```
ValJed, line 203: Good practice to instantiate all data inside `data` block even with a null value or whatever.
Something like that would be clear, and we would use only newId:
```javascript
return {
   ...
    newId: this.id || createId()
}
```

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -227,16 +287,39 @@ export default {
           });
           return;
         }
-        if (!widget.type) {
-          widget.type = this.type;
-        }
-        if (!this.id) {
-          widget._id = createId();
-        }
+        this.saving = true;
```
ValJed, line 290: Same, would be nice to instantiate this inside data.

ValJed (changes requested):
Tiny things + few questions:
- What about making it behavior by default?
- Would it be possible to not update the view if the schema isn't valid? Otherwise it triggers console errors + the `Unable to render widget` when an input isn't valid (like opening schema).
image

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
boutell: data is reactive, newId does not need to be reactive, so this would be unnecessary performance overhead.

boutell: (Because of the avoidable overhead, I don't agree it is a best practice to use data for things that are not reactive, although I would not reject a PR for doing it.)

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
boutell: Same issue, it doesn't need the reactivity.

modules/@apostrophecms/area/index.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
           const type = self.apos.launder.string(req.body.type);
           const _docId = self.apos.launder.id(req.body._docId);
           const field = self.apos.schema.getFieldById(areaFieldId);
-
+          const livePreview = self.apos.launder.boolean(req.body.livePreview);
```
boutell, line 40: So we can decide how to handle an error.

modules/@apostrophecms/area/index.js
```diff
@@ -51,7 +51,14 @@ module.exports = {
             self.warnMissingWidgetType(type);
             throw self.apos.error('invalid');
           }
-          widget = await sanitize(widget);
+          try {
+            widget = await sanitize(widget);
+          } catch (e) {
+            if (livePreview) {
+              return 'aposLivePreviewSchemaNotYetValid';
```
boutell, line 58: The result is HTML, not JSON, so we return a special string to avoid a status code that shows up as an "error" in the browser console. When previewing an error from sanitize is sometimes normal, so we return this code to indicate it's OK to do nothing.

modules/@apostrophecms/schema/index.js
```diff
@@ -684,10 +684,6 @@ module.exports = {
           nonVisibleFields
         });
 
-        for (const error of validErrors) {
```
boutell, line 687: See changelog — this has been the cause of a lot of verbose double-logging. We shouldn't log an error unless it is the ultimate result. At least not with util.error. At some point we might revisit the question of how to optionally enable a super-verbose mode and then this would be more appropriate.

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
```diff
@@ -52,21 +52,32 @@ export default {
...
-          this.rendered = await apos.http.post(`${apos.area.action}/render-widget?aposEdit=1&aposMode=${this.mode}`, {
+          // Don't use a placeholder here, it causes flickering in live preview mode.
+          // It is better to display the old until we display the new,
+          // we have "busy" for clarity
+          const result = await apos.http.post(`${apos.area.action}/render-widget?aposEdit=1&aposMode=${this.mode}`, {
             busy: true,
             body: parameters
           });
+          //
+          if (result !== 'aposLivePreviewSchemaNotYetValid') {
```
boutell, line 78: If we are previewing and the schema isn't valid at the moment according to the server, we have nothing to do here, leave the previous placeholder or preview alone.

modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
ValJed: I'm not sure this is the kind of performance overhead we should worry about.
I'm saying bad practice because I think it's a non document behavior so could lead to unexpected behavior, also think it's a bit less readable but I'm ok.

ValJed (approved):

PR #4910, opened 2025-04-10
PRO-7551 do not show editing UI of subwidgets when live previewing a widget

boutell:


modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -338,7 +338,9 @@ function guessOrigin(area) {
   // side of the screen will least obscure the widget
   const rect = area.$el.getBoundingClientRect();
   const cx = (rect.right - rect.left) / 2 + rect.left;
-  if (cx >= (window.innerWidth / 2)) {
+  // Favor the right hand side slightly because rich text
+  // subwidgets in centered areas are more intuitive that way
+  if (cx >= (window.innerWidth * 0.55)) {
```
boutell, line 343: Lindsay saw it as a defect that she couldn't see much when live previewing an area that is centered on the page. And when I think about it I realize that in LTR languages, favoring the right hand side a little for the widget editor will be best for most widgets. Rich text is just the most obvious.

At some point we may be asked to reconsider for RTL languages, but that's equally true for things like our default position for the regular widget editor modal and other things in our UI, so I don't consider it an issue for today.

stuartromanek:
@boutell i don't really understand what this does. Can you make a video or something?

stuartromanek (approved):
very nice 💯

PR #4915, opened 2025-04-14
Pro 7547 locale switcher

ValJed:
[PRO-7547](https://linear.app/apostrophecms/issue/PRO-7547/locale-switcher-does-not-work-for-pieces-if-the-piece-page-exists-but)

## Summary

See ticket:
Fix switch locale from piece to piece when the index page in the target locale is draft only.
⚠️ Most of the changes are linter only. I added two comments where the important code is located with some explanations.

* Create a piece in two locales
* Create an index page for these pieces in both locale
* In the target locale make the index page only draft
* Switch locale when on the piece page on the initial locale to the target locale
* It should find the target locale piece, even if its parent page is draft only

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/modules/@apostrophecms/i18n/index.js
+++ b/modules/@apostrophecms/i18n/index.js
@@ -415,14 +415,11 @@ module.exports = {
             locale: sanitizedLocale
           });
           if (_id) {
-            doc = await self.apos.doc.find(localeReq, {
-              aposDocId: _id.split(':')[0]
-            }).toObject();
-            if (!doc) {
-              const publishedLocaleReq = localeReq.clone({ mode: 'draft' });
-              doc = await self.apos.doc.find(publishedLocaleReq, {
-                aposDocId: _id.split(':')[0]
-              }).toObject();
+            const aposDocId = _id.split(':')[0];
+            doc = await self.apos.doc.find(localeReq, { aposDocId }).toObject();
+            if (!doc._url) {
+              const draftLocaleReq = localeReq.clone({ mode: 'draft' });
+              doc = await self.apos.doc.find(draftLocaleReq, { aposDocId }).toObject();
             }
           }
           if (!sanitizedLocale) {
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/piece-page-type/index.js
```diff
@@ -244,10 +244,16 @@ module.exports = {
...
+            req.aposParentPageCache[pieceName] = pages;
+          }
         }
         results.forEach(function (piece) {
-          const parentPage = self.chooseParentPage(req.aposParentPageCache[pieceName], piece);
+          const parentPage = self.chooseParentPage(
+            req.aposParentPageCache[pieceName] || [],
+            piece
+          );
+
```
ValJed, line 256: Real changes are here:
We don't set the cache if no data. This allow to retry to get the doc in draft after tryihng publish version but having no parent page (because this one is draft only).

modules/@apostrophecms/i18n/index.js
```diff
@@ -404,14 +415,11 @@ module.exports = {
...
-            if (!doc) {
-              const publishedLocaleReq = localeReq.clone({ mode: 'draft' });
-              doc = await self.apos.doc.find(publishedLocaleReq, {
-                aposDocId: _id.split(':')[0]
-              }).toObject();
+            const aposDocId = _id.split(':')[0];
+            doc = await self.apos.doc.find(localeReq, { aposDocId }).toObject();
+            if (!doc._url) {
+              const draftLocaleReq = localeReq.clone({ mode: 'draft' });
+              doc = await self.apos.doc.find(draftLocaleReq, { aposDocId }).toObject();
```
ValJed, line 422: real changes here:
We look for `doc._url` instead of just doc because published may have no parent page (draft only) so url could be set for a draft piece only.

boutell (changes requested):

boutell (approved):

PR #4920, opened 2025-04-16
fixes i18n locale route

ValJed:
Fixes i18N locale route


haroun (dismissed):

ETLaurent (changes requested):
changelog?

ETLaurent (approved):
discussed, fix on an unreleased feature, so no need for a changelog entry

PR #4930, opened 2025-04-25
Update getMangageApiProjection defaults

BoDonkey:
## Summary

This PR stemmed from trying to document `getManageApiProjection()` and finding that changing the default null projection would place a heavy burden on the developer. This PR adds the fields originally suggested in the code comments for the `options.manageApiProjection` to the core method.

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [X] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ A `secondaryLevel: true` option is available to add operations to the widget's c
 ### Fixes
 
 * Fixes all eslint warnings.
-* Fixes a RT bug where including `table` in `toolbar` but omitting an `insert` array crashed the rich text editor.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,11 @@ A `secondaryLevel: true` option is available to add operations to the widget's c
 ### Fixes
 
 * Fixes all eslint warnings.
-* Fixes a RT bug where including `table` in `toolbar` but omitting an `insert` array crashed the rich text editor. 
+* Fixes a RT bug where including `table` in `toolbar` but omitting an `insert` array crashed the rich text editor.
+
+### Changes
+
+* Updates the default fields for the `getMangageApiProjection()` to include a more sensible base configuration.
 
 ## 4.15.0 (2025-04-16)
 
--- a/modules/@apostrophecms/piece-type/index.js
+++ b/modules/@apostrophecms/piece-type/index.js
@@ -26,10 +26,8 @@ module.exports = {
     //   title: 1,
     //   _url: 1,
     // },
-    // By default the manager modal will get all the pieces fields below +
-    // all manager columns
-    // you can enable a projection using
-    // managerApiProjection: {
+    // By default the manager modal only fetches these fields:
+    // {
     //   _id: 1,
     //   _url: 1,
     //   aposDocId: 1,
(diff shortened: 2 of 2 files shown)


modules/@apostrophecms/piece-type/index.js
```diff
@@ -1156,19 +1157,29 @@ module.exports = {
         });
       },
       getManagerApiProjection(req) {
-        if (!self.options.managerApiProjection) {
-          return null;
-        }
-
-        const projection = { ...self.options.managerApiProjection };
+        // Start from the configured projection, or fall back
```
boutell, line 1160: If `managerApiProjection` is not configured at all, then we already bail out and don't project at all, the default behavior is and will remain "fetch everything" (see above).

So what I had in mind was different. We should *always* fetch these fields even if they are *not* included in the projection. In the same spirit as including the columns even if they are not part of the projection: we know you need them even if you don't.

We could reinforce this by accepting and documenting the value `true` ("please fetch less stuff according to your best guess, I have no fields to add") as an alternative to turning it on with an empty `{}`, although the latter should have the same effect ("please use your best guess, plus these fields... I guess I don't have any to suggest").

boutell (changes requested):

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ A `secondaryLevel: true` option is available to add operations to the widget's c
 ### Fixes
 
 * Fixes all eslint warnings.
+* Fixes a RT bug where including `table` in `toolbar` but omitting an `insert` array crashed the rich text editor.
```
boutell, line 16: This already shipped, see duplicate entry further down for 4.15.1

modules/@apostrophecms/piece-type/index.js
```diff
@@ -404,7 +405,7 @@ module.exports = {
           return self.apos.modules['@apostrophecms/job'].runBatch(
             req,
             req.body._ids,
-            async function(req, id) {
+            async function (req, id) {
```
boutell, line 408: I see this space before opening paren both appearing and disappearing in this PR.

We don't normally have this space, please remove.

boutell (changes requested):

modules/@apostrophecms/piece-type/index.js
BoDonkey: Removed - somehow my linter is behaving badly. I need to dig into why.

CHANGELOG.md
BoDonkey: Deleted.

modules/@apostrophecms/piece-type/index.js
```diff
@@ -430,7 +431,7 @@ module.exports = {
...
```
boutell, line 434: still one of these

modules/@apostrophecms/piece-type/index.js
```diff
@@ -457,7 +458,7 @@ module.exports = {
...
```
boutell, line 461: here's another

boutell (changes requested):

modules/@apostrophecms/piece-type/index.js
BoDonkey: Serched for and fixed additional instances.

boutell (approved):

PR #4933, opened 2025-04-28
PRO-7633: Hotfix 4.15.2

boutell:
hotfix for XSS issue.

The major version bump in dompurify is OK because all they did was remove IE support (not relevant in nodejs validation). I have verified that I can still upload SVG files to apostrophecms.


BoDonkey (approved):
Just updating the deps fixed this "security" hole?

boutell:
Yes, it was a dompurify security hole.

PR #4938, opened 2025-05-06
PRO-7613: Add widget width control

myovchev:
## Summary

Adds image widget width slider. Introduce widget options `defaultImageWidth` (percentage, 0 by default) and `imageResizeStep` to control the initial width of the image (0% means the setting is disabled) and the step of the width slider (default is 5) respectively.

The slider in the image widget editor controls the image width.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/image-widget/index.js
+++ b/modules/@apostrophecms/image-widget/index.js
@@ -7,18 +7,32 @@ module.exports = {
     dimensionAttrs: false,
     placeholder: true,
     placeholderClass: false,
-    placeholderImage: 'jpg'
+    placeholderImage: 'jpg',
+    // 0 means disabled width setting
+    defaultImageWidth: 0,
+    imageResizeStep: 5
   },
-  fields: {
-    add: {
-      _image: {
-        type: 'relationship',
-        label: 'apostrophe:image',
-        max: 1,
-        required: true,
-        withType: '@apostrophecms/image'
+  fields(self) {
+    return {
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/image-widget/index.js
```diff
@@ -7,18 +7,32 @@ module.exports = {
...
+          withType: '@apostrophecms/image'
+        },
+        width: {
+          type: 'range',
+          label: 'apostrophe:imageWidth',
+          help: 'apostrophe:imageWidthHelp',
+          min: 0,
+          max: 100,
+          step: self.options.imageResizeStep || 1,
+          def: self.options.defaultImageWidth ?? 100
```
boutell, line 32: This is going to work out to 0, not 100, because 0 is not nullish. So I think the `?? 100` does not do anything useful.

Now, assuming we just remove `?? 100`...

The user is going to see a slider set to 0, but a rendering that is probably 100% (whatever the CSS does by default).

But, this does result in backwards compatible behavior by default, and even provides a way to return to default behavior.

And "0%" itself is not a useful setting, therefore it's OK to customize the behavior.

So I guess I like it.

boutell (changes requested):

boutell (approved):

PR #4948, opened 2025-05-12
PRO-7756: fix the clipboard to always show the right content in menus previously toggled open and closed

boutell:


modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -118,31 +118,16 @@ export default function() {
   }
 
   function createWidgetClipboardApp() {
-    class Clipboard {
```
boutell, line 121: I simplified this class a lot but didn't change the interface. While I can't see any bug in the original logic for it, the "storage" event just wasn't firing for me when copying to the clipboard in a separate tab. This simplified logic always works because it always talks to the clipboard.

modules/@apostrophecms/area/ui/apos/components/AposAreaContextualMenu.vue
```diff
@@ -10,6 +10,8 @@
       :button="buttonOptions"
       :popover-modifiers="inContext ? ['z-index-in-context'] : []"
       :menu-id="menuId"
+      @open="buttonOpen = true"
```
boutell, line 13: This menu can get toggled open in two ways: via CSS classes injected by the parent, or via its own button (used when it's in an empty area). In both cases I need to trigger a fresh build of the menu so it includes the latest clipboard.

modules/@apostrophecms/area/ui/apos/components/AposAreaContextualMenu.vue
```diff
@@ -176,6 +182,20 @@ export default {
       return flag;
     },
     myMenu() {
+      // Ensures we take a fresh look at the clipboard when toggled open again.
```
boutell, line 185: All I really want to do is always touch "this.open" and "this.buttonOpen" in this `computed` function so that it responds to changes in their values. I did it this particular way to make sure neither developers nor compilers / optimizers ever decide it can be "optimized away."

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -92,6 +92,7 @@
           :tabbable="isHovered || isFocused"
           :menu-id="`${widget._id}-widget-menu-top`"
           :class="{[classes.open]: menuOpen === 'top'}"
+          :open="menuOpen === 'top'"
```
boutell, line 95: Making this explicitly available as a prop so the menu can react to it and regenerate itself.

stuartromanek (approved):

PR #4951, opened 2025-05-14
PRO-7773: missing functionality the import/export module needs now that we are enforcing our own rules better; also a bug fix for error reporting

boutell:


BoDonkey (approved):
I guess?

boutell:
The tests pass locally. We hit a quota in docker.

modules/@apostrophecms/schema/index.js
```diff
@@ -752,9 +752,8 @@ module.exports = {
           const [ destId, destPath ] = error.path.includes('.')
             ? error.path.split('.')
             : [ null, error.path ];
-
           const curDestination = destId
-            ? destination.find(({ _id }) => _id === destId)
+            ? (destination.items || destination || []).find(({ _id }) => _id === destId)
```
boutell, line 756: This logic does the right thing if destination is an area rather than an array.

BoDonkey (approved):

PR #4963, opened 2025-05-26
Pro 7746 batch tag images

haroun:
## Summary

Batch tagging for images

⚠️ once this is merged, be sure to change https://github.com/apostrophecms/testbed/blob/feature/batch-tagging/package.json#L59 with `"apostrophe": "github:apostrophecms/apostrophe#feature/batch-tagging",`

1. Select 1 or more images
2. Click on batch `Tag` operation
3. Select your tag
4. Verify that the left panel has been updated with your selected tag
5. Now select a single image
6. Add a tag or select an existing one
7. The right panel should have been updated
8. Select all the images
9. When you open batch `Tag` operation, you should see a checkbox with an "indeterminate" state
10. Click on it, it should remove the tag
11. Click on it again, it should add the tag (without the "indeterminate" state)
12. Reload the page just for fun, check the tags on the images you've updated so far

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


stuartromanek (changes requested):
These definitely need to be implemented
## UX stuff
- Clicking **Create New Tag** (As it is without a search query) doesn’t do anything. It should bring up the Create New Tag UI
image
This UI also offers a button that opens the Tag Manager as well
- If no tags exist when the dialogue is opened, display the empty state
image

These are some points I didn't detail or anticipate and would be interested in experimenting w them. Please create them as another branch from this one and not bound up with this PR
- Don’t re-sort tags during tagging. This is confusing when you click a tag and your images are tagged but the item you clicked gets shuffled to the top of the list and can essentially 'disappear' from view.. however
- Re-sort tags after tag search filtering. This applies to searching the tag list but also clearing the search. When I clear a search I should be returned to a list that is ordered like
- - Selected (A-Z)
- - Partially Selected (A-Z)
- - Unselected (A-Z)

PR #4985, opened 2025-06-19
PRO-7797: Ensure defaultNode has highest priority

myovchev:
## Summary

Unpin `prosemirror-model` and fix the underlying issue.

**The long story:**

We use custom extension named `defaultNode` in order to allow rich-text areas to have a "default style". A scenario - a configured by a developer rich text placed to produce `h1` on a page doesn't require the editor to switch the style manually to `h1`. 

The `defaultNode` can adapt and based on the configured style with `def: true`, become a simple extension of `paragraph`, `heading` or just a text node (`span`). The problem arising with this approach is how prosemirror handles parsing - it needs 1-to-1 mapping of a "nodeName" to a "nodeElement". When the DOM parser attempts to parse the tree and in our case - to validate a list, it matches the configured list item `content`, in our case `defaultNode`. However, when our `defaultNode` is just a `paragraph` extension (by default), the DOM parser matches the original `paragraph` node because it has a higher priority (it's registered first). The parser then don't recognize the node as the one configured for the list item content (it's `paragraph` and not `defaultNode`) and decides to "fix" the DOM by pulling it out of the list.

The current patch fixes that scenario with ensuring the right order - `defaultNode` is always registered first. This solves the problem at hand. We can collaborate on how to avoid prosemirror schema containing duplicate DOM representation in the future.

References:
- https://github.com/ProseMirror/prosemirror/issues/1534
- https://github.com/ueberdosis/tiptap/issues/6443

List should be properly parsed, UI tests should pass.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
+++ b/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
@@ -435,6 +435,8 @@ export default {
       .filter(Boolean)
       .concat(this.aposTiptapExtensions());
 
+    this.ensureExtensionsPriority(extensions);
+
     this.editor = new Editor({
       content: this.initialContent,
       autofocus: this.autofocus,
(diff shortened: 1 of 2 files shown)


boutell (approved):
Thank you!

PR #5003, opened 2025-07-15
PRO-8043: Opt-out from locale inferring when possible

myovchev:
## Summary

Improve the resolving logic of the relationship type `convert` handler:
- Ignore the request locale when it know it works with `_id`'s to avoid false positives
- This makes sense of the context of validating data, because the handler should care about the document and shouldn't be involved in any related to `aposMode` related logic. 
- When related items provide `_id` we can opt-out of auto-locale inferring, because the `_id` is unique.
- In all other cases the old logic applies (infer locale and mode)  

Providing a `published` request and `draft` document should successfully validate the draft document (and ignore `published` mode). 

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/schema/lib/addFieldTypes.js
+++ b/modules/@apostrophecms/schema/lib/addFieldTypes.js
@@ -1117,6 +1117,10 @@ module.exports = (self) => {
           }
         }
       }
+      // PRO-8043: When dealing with ID's only, request based locale inference
+      // is not needed. `undefined` - default auto locale inference,
+      // `null` - no locale inference.
+      const inferLocale = titlesOrIds.length > 0 ? undefined : null;
       const clauses = [];
       if (titlesOrIds.length) {
         clauses.push({
@@ -1138,6 +1142,7 @@ module.exports = (self) => {
         const results = await manager
           .find(req, { $or: clauses })
           .relationships(false)
+          .locale(inferLocale)
           .toArray();
         // Must maintain input order. Also discard things not actually found in
         // the db
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,15 @@
 
 ### Adds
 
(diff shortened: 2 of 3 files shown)


boutell:
"titlesOrIds" suggests there is more than one case in play here.

modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -1117,6 +1117,10 @@ module.exports = (self) => {
           }
         }
       }
+      // PRO-8043: When dealing with ID's only, request based locale inference
+      // is not needed. `false` - default auto locale inference,
+      // `null` - no locale inference.
+      const inferLocale = titlesOrIds.length > 0 ? false : null;
```
haroun, line 1123:
```suggestion
      const inferLocale = ids.length > 0 ? false : null;
```
If we have some titles, we will not be able to infer the locale or I'm reading this wrong?

haroun (changes requested):
I agree with the change. We should also update the code that triggers this issue to send the POST request as draft, otherwise we will have additional issues in some other places.

modules/@apostrophecms/schema/lib/addFieldTypes.js
myovchev: You got caught by the rather strange filter logic. See the comment above:
```js
      // PRO-8043: When dealing with ID's only, request based locale inference
      // is not needed. `false` - default auto locale inference,
      // `null` - no locale inference.
```

So, if we have ANY "possible" title, we pass the default value (false), otherwise we disable the filter (null).

myovchev: You can double check the behavior in the filter itself.

PR #5013, opened 2025-07-26
Fix #4979: Replace connect-multiparty with multer to resolve security vulnerability

### LINKED ISSUE #4979: connect-multiparty vulnerability and deprecation
Hi 👋

one of the direct dependencies [`connect-multiparty`](https://github.com/expressjs/connect-multiparty) seems to be no longer actively maintained.

Image

There is also already a vulnerability about it being reported: https://github.com/advisories/GHSA-w2xw-44r3-4v9g

Are there any plans to move away from this? 

Thank you in advance.

Cheers 
Andi

gingeekrishna:
Replace connect-multiparty with multer to resolve security vulnerability

- Remove connect-multiparty dependency (deprecated with CVE-2022-29623)
- Add multer v1.4.5-lts.1 as replacement for file upload handling
- Update attachment upload route to use multer.single('file')
- Update rich-text CSV upload to use multer middleware
- Migrate big-upload middleware to multer.any() for chunked uploads
- Update test suite to handle multer file format
- Maintain backward compatibility with file.name fallback
- Move form-data to devDependencies as only needed for testing
- I have validated via test cases with test file, please let me know this changes looks good

## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/http/lib/big-upload-middleware.js
```diff
@@ -163,7 +163,7 @@ module.exports = (self) => ({
         });
       }
       let n = 0;
-      req.files = {};
+      req.files = [];
```
boutell: The interface big-upload-middleware provides must not change to match multer's interface because we have already documented that it matches multiparty's interface. You can still use multer to implement it, but the `req.files` object exposed at the end must behave like the multiparty one to avoid a backwards compatibility break.

boutell (changes requested):
Overall this is excellent, thank you! However, there is a concern about big-upload-middleware. Its interface should not change — the `req.file` object it exposes at the end must continue to look like a `multiparty` one, because otherwise we'll have to hunt down other apostrophecms and third party uses of it and change them all.

To be clear, there's no problem with using multer to implement it, as long as the object delivered at the end behaves the old way, rather than the multer-style "array way."

Let me know if you can make that change.

Thanks!

modules/@apostrophecms/http/lib/big-upload-middleware.js
gingeekrishna: Hi. @boutell 

Thanks for looking into it, understand the concern and I have modified the changes now.

PR #5015, opened 2025-07-29
Resolved an issue affecting `withRelationships` with two or more steps

boutell:
This issue could cause a document to appear to be related to the same document more than once. If several documents in the first step of the `withRelationships` path are related to the same document in the second step, then widgets containing relationships in the second step might populate with multiple copies of the same image.

In particular: [ 'articles', 'authors' ] in apollo, where 'authors' have an area containing an image widget, and the same author was attached to several articles.

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
@@ -26,8 +26,8 @@
         :class="labelsClasses"
       >
         <ol
-          @click="isSuppressingWidgetControls = false"
           class="apos-area-widget__breadcrumbs"
+          @click="isSuppressingWidgetControls = false"
         >
           <li
             class="
--- a/modules/@apostrophecms/attachment/index.js
+++ b/modules/@apostrophecms/attachment/index.js
@@ -159,7 +159,6 @@ module.exports = {
       post: {
         upload: [
           self.canUpload,
-          //In the existing code, we are reading the zeroth element from the files array object, which results in processing only a single file. Therefore, I am currently reading just one file from the Multer package.
           require('multer')({ dest: require('os').tmpdir() }).single('file'),
           async function (req) {
             try {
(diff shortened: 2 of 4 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -26,8 +26,8 @@
         :class="labelsClasses"
       >
         <ol
-          @click="isSuppressingWidgetControls = false"
           class="apos-area-widget__breadcrumbs"
+          @click="isSuppressingWidgetControls = false"
```
boutell, line 30: unrelated, satisfies eslint

modules/@apostrophecms/attachment/index.js
```diff
@@ -159,7 +159,6 @@ module.exports = {
       post: {
         upload: [
           self.canUpload,
-          //In the existing code, we are reading the zeroth element from the files array object, which results in processing only a single file. Therefore, I am currently reading just one file from the Multer package.
```
boutell, line 162: unrelated, satisfies eslint, comment is not necessary anymore

modules/@apostrophecms/attachment/index.js
```diff
@@ -175,7 +174,6 @@ module.exports = {
 
               return attachment;
             } finally {
-              //Hence I am reading the single file from the upload and I am checking the condtion for the same
               if (req.file) {
```
boutell, line 177: unrelated, satisfies eslint, comment is not necessary anymore

modules/@apostrophecms/schema/lib/joinr.js
```diff
@@ -68,6 +68,14 @@ const joinr = module.exports = {
     getter,
     idMapper
   ) {
+    // This method never alters the items array itself, it alters
```
boutell, line 71: Note that reducing to unique elements in this way does not make new objects nor does it test actual object contents. This is intentional. We just want to eliminate duplicates *by reference*.

myovchev (approved):

PR #5029, opened 2025-08-07
PRO-8023: editor component

myovchev:
## Summary

- Allow control over area backend rendering via combination of schema and contextual (an object passed `with` custom tag keyword) options. This includes class name and style for the area wrapper, per widget item custom template, custom editor component. 
- Refactor Area Editor, extracting its logic
- Create a copy of Area Editor (Area Layout Editor) that reuses that logic and renders a development message to prove it's in use
- Use the custom Area Layout Editor with our new layout widget.

**The Result** 

The strategy "almost" works so far. This change allows the grid to be rendered in preview and public mode, but fails in edit mode. This is expected! In the next step we need the combination of the new area editor AND a custom widget component (probably inline) to handle the editing experience, while the rest should stay unchanged. 

Why not working? Because the Widget Editor also does its own thing (adding wrappers and number of dom enhancements) by default. But most importantly, it requests the widget contents directly, thus bypassing our new feature. This is fine! We'll handle this case manually in the next iteration.

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
+++ b/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
@@ -3,8 +3,6 @@ import createApp from 'Modules/@apostrophecms/ui/lib/vue';
 
 export default function() {
 
-  const component = apos.vueComponents.AposAreaEditor;
-
   let widgetsRendering = 0;
 
   apos.area.widgetOptions = [];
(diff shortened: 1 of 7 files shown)


modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -73,6 +71,14 @@ export default function() {
     const renderings = {};
     const _docId = data._docId;
 
+    let componentName = options.editorComponent;
+    if (!apos.vueComponents[componentName]) {
+      // eslint-disable-next-line no-console
+      console.error(`Area Editor component "${componentName}" not found. Switching to default.`);
```
boutell, line 78: This shouldn't be reported as an error, it is the typical case.

myovchev: It's a report worthy now - the initial value has the default value fallback. In this case, if it fails it's definitely a misconfiguration.

boutell (approved):

PR #5031, opened 2025-08-08
Feature layout widget

myovchev:
## Summary

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


stuartromanek:
This whips, Miro. A few final UX considerations:

- Can layout widget start out as ‘focused’ when added to the page? I think this would better communicate the intent of this special widget (here are areas (familiar) and special controls)
- if a column or widget inside a column is focused in Content mode, focus the parent layout as well. We need to keep the breadcrumb label and Content/Layout UI on the screen as much as makes sense.
- when column is drag-resizing or drag-moving keep the hover/active styles
    - make the cursor be `grabbing`
    - keep the four corner handles visible
    - keep the `.apos-layout__item-resize-handle-highlight-bar` visible (when resizing)
    - Keep the resize grip visible (when resizing)
- When trying to drag-resize a column in a direction that has no room to grow (immediate neighbor) the interaction feels ‘missed’ because there is no feedback about ‘this can’t work’..
    - Can we let the user expand the the ‘ghost purple’ box in the illegal direction but style it red/error? When they release in the illegal position the element snaps back to its original position

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
+<template>
+  <ol
+    v-if="widgetBreadcrumbActions.length > 0"
+    class="apos-area-widget__breadcrumbs apos-area-widget__breadcrumbs--action"
+    style="margin-left: 10px;"
```
ValJed: static style should go in style block.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -818,7 +917,8 @@ export default {
   .apos-area-widget__label {
     position: absolute;
     top: 0;
-    right: 0;
+    left: 0; // switch the root label position from right to left
+    // right: 0;
```
ValJed: Remove comments ?

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -440,8 +479,15 @@ export default {
   unmounted() {
     // Remove the focus parent listener when unmounted
     apos.bus.$off('widget-focus-parent', this.focusParent);
+    // Ensure click-outside listener is cleaned up
+    this.removeClickOutsideListener();
```
ValJed, line 478: It think the method name is clear enough no need comment

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -290,6 +325,7 @@ export default {
       isSuppressed: false,
       menuOpen: null,
       isSuppressingWidgetControls: false,
+      hasClickOutsideListener: false, // Track if click-outside listener is active
```
ValJed: I think the property name is clear enough, doesn't need a comment here

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -263,6 +284,20 @@ export default {
...
+    controlsDisabled: {
+      type: Boolean,
+      default: false
+    },
+    breadcrumbDisabled: {
+      type: Boolean,
+      default: false
+    },
+    // Whether or not this widget should be focused
+    shouldFocus: {
```
ValJed: name is clear enough, doesn't need comment, or rename it `shouldFocusWidget`

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -186,10 +203,14 @@
 
 <script>
 import AposIndicator from 'Modules/@apostrophecms/ui/components/AposIndicator.vue';
+import AposBreadcrumbSwitch from 'Modules/@apostrophecms/area/components/AposBreadcrumbSwitch.vue';
```
ValJed: You shouln't need to import components they are automatically registered.. (at least AposIndicator).

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
+<template>
+  <ol
```
ValJed, line 2: Is it numbered ordered list, could be `ul` I think ?

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+  >
+    <li class="apos-area-widget__breadcrumb">
+      <component
+        :is="operation.component"
+        v-for="operation in widgetBreadcrumbActions"
+        :key="operation.key"
+        v-slot="slotProps"
+        :data-apos-test-name="operation.name"
+        :data-apos-test-action="operation.action"
+        :data-apos-test-type="operation.type"
```
ValJed, line 17: 👍🏼

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+        v-bind="operation.props"
+        v-on="operation.listeners"
```
ValJed, line 18: I would put these two props at beginning, or at least before the `data-apos-test` props

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+import AposBreadcrumbSwitch from 'Modules/@apostrophecms/area/components/AposBreadcrumbSwitch.vue';
+
+export default {
+  name: 'AposBreadcrumbOperations',
+  components: {
+    AposIndicator,
+    AposBreadcrumbSwitch
+  },
+  props: {
+    i: {
```
ValJed, line 57: What is `i`, index ? Name could be clearer

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+    },
+    skipInfo: {
+      type: Boolean,
+      default: false
+    },
+    // Whether to teleport menu modals to the end of the document body.
+    teleportModals: {
+      type: Boolean,
+      default: false
+    }
```
ValJed, line 89: So context menu can use teleport or it can avoid to ? Why not using it all the time ?

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+        ...this.operationButtonDefault,
+        icon: operation.icon,
+        action: operation.action,
+        tooltip: operation.tooltip || null
+      };
+    },
+    getOperationListeners(operation) {
+      const listeners = {};
+      let setFocus = true;
+      let handleClick = true;
```
ValJed: I would find more readable something like:
```
const setFocus = ['switch', 'info'].includes(operation.type)
```
Same for `handleClick`, because I don't need to scan the entire function to guess their value.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+          this.getFocus(e, this.widget._id);
+        };
+      }
+      if (operation.action === 'remove') {
+        setFocus = false;
+      }
+
+      if (!handleClick) {
+        return listeners;
+      }
```
ValJed, line 234: By setting `setFocus` and `handleClick` as described above you could return earlier.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+
+      if (!listeners.click) {
```
ValJed: This if does nothing I think. I don't see any code above adding a click property to the listeners object ?

modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -0,0 +1,176 @@
...
+            }
+          }
+        },
+        content: {
+          type: 'area',
+          options: {
+            widgets: {
+              '@apostrophecms/rich-text': {},
+              '@apostrophecms/image': {},
+              '@apostrophecms/video': {}
```
ValJed, line 175: Users should be able to set these widgets, how it's done ?

modules/@apostrophecms/layout-meta-widget/index.js
```diff
@@ -0,0 +1,49 @@
...
+            auto: {
+              type: 'boolean',
+              def: true
+            }
+          }
+        }
+      }
+    }
+  }
+};
```
ValJed: Since it's not a real widget, shouldn't we get rid of these one and store these infos in the `layout-widget` itself ?

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -0,0 +1,209 @@
...
+        if (widgetTypes.length !== 2) {
+          throw new Error(
+            `The layout widget "columns" area must have exactly two widget types, it has ${widgetTypes.length}.`
+          );
+        }
+
+        const check = {
+          '@apostrophecms/layout-column': false,
+          '@apostrophecms/layout-meta': false
+        };
```
ValJed, line 165: widget names are set in variables in the init function, these variables should be used instead of hard coded strings in case we change their values. (metaWidgetName and columnWidgetName).
Also not sure what this `check` is about a more descriptive name could be helpful.

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -0,0 +1,209 @@
...
+          throw new Error(
+            'The layout widget "columns" area must have a widget type or a subtype of "@apostrophecms/layout-column"'
+          );
+        }
+
+        if (!check['@apostrophecms/layout-meta']) {
+          throw new Error(
+            'The layout widget "columns" area must have a widget type or a subtype of "@apostrophecms/layout-meta"'
+          );
+        }
```
ValJed, line 190: Same in these two if statements, we should use the variables.

modules/@apostrophecms/layout-widget/ui/apos/components/AposGridLayout.vue
```diff
@@ -0,0 +1,297 @@
...
+    <AposGridManager
+      v-if="isManageMode"
+      :grid-state="gridState"
+      :synthetic-items="syntheticItems"
+      :meta-id="meta._id"
+      :opstate="opstate"
+      @resize-start="isResizing = true"
+      @resize-end="$emit('resize-end', $event); isResizing = false"
+      @move-start="isMoving = true"
+      @move-end="$emit('move-end', $event); isMoving = false"
```
ValJed: I think logic should go in functions, especially when two instructions.

modules/@apostrophecms/widget-type/index.js
```diff
@@ -105,15 +105,67 @@
...
+// when the user selects that choice.
+// - `secondaryLevel`: If true, the operation appears in a secondary level of
+// the toolbar, accessed by clicking a "more" icon. This is only supported
+// for standard placement.
+// - `tooltip`: The tooltip to show on hover.
+// - `action`: A valid core action (e.g. remove, edit, etc.) to be
+// emitted when the operation is clicked. In effect only if no modal is specified.
+//
+// See also `skipOperations` option for a way to disable core operations
+// such as edit or remove.
```
ValJed, line 159: Would make sense to have it in documentation more than in comments maybe

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+        :key="`info-${operation.key}`"
+        :data-apos-test-name="operation.name"
+        v-bind="operation.props"
+        v-on="operation.listeners"
+      />
+    </li>
+  </ol>
+</template>
+
+<script>
```
ValJed, line 50: We stuck with options api forever ahah

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -0,0 +1,389 @@
...
+import AposIndicator from 'Modules/@apostrophecms/ui/components/AposIndicator.vue';
+import AposBreadcrumbSwitch from 'Modules/@apostrophecms/area/components/AposBreadcrumbSwitch.vue';
```
ValJed: Don't need to import components, at least `AposIndicator`

modules/@apostrophecms/layout-widget/ui/apos/components/AposAreaLayoutEditor.vue
```diff
@@ -0,0 +1,365 @@
...
+      <AposGridLayout
+        :options="gridModuleOptions"
+        :items="layoutColumnWidgets"
+        :meta="layoutMeta"
+        :layout-mode="layoutMode"
+        :device-mode="layoutDeviceMode"
+        :opstate="{
+          options: options,
+          disabled: field && field.readOnly,
+          operations: layoutBreadcrumbOperations || []
```
ValJed: No need ` || []` already done in computed.

ValJed (changes requested):
Few tiny things. A bit hard to modelize mentally between the widget themselves and the vue component. Would be very nice to be able to get read of the meta component, still don't really get what it does that the layout-widget couldn't.
Also, I see that main branch hasn't been merged since a while (1 or 2 months?).
Would make sense to do it before to go further (maybe some bugs have been fixed etc) and because it will become harder and harder.
Otherwise feature seems working well, just a bit a mess with the widget preview et cie, also hard to debug with devtools since layout editor and widgets are not in the same tree for example.
I created a branch to fix rich-text and focus issues. Also moved focus logic in the widget store, I know it's more code but the existing one was hard to read/track, and it's not a lot of code.
One last point, I see you used options API for all new components, would have been nice to use composition (easier to work with pinia, easier to share code, new idiomatic way of coding vue).

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: It's the choice of breadcrumbs original code, we are just following that standard and using the same styles

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev:
```
Attribute ":data-apos-test-name" should go before "v-on".eslint[vue/attributes-order](https://eslint.vuejs.org/rules/attributes-order.html)
```
Sadly, eslint is in my way. I did it originally as you suggest, eslint thinks otherwise.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: Same as AposAreaWidget, it's a pass-through.
```
    i: {
      type: Number,
      required: true
    },
```

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: Because it's too much side effects for the current work context. 
It can become gradually the default, but we don't know if we want and if we should (and if it works 100% of the time). 
It works for this use case and it's an option now (almost like a feature flag)

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: The example doesn't capture the exact logic, but I tried to make it more readable.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: You can't, you still need the listeners.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: Correct!

modules/@apostrophecms/layout-column-widget/index.js
myovchev: any of the following:
- Improve the layout-column-wdiget directly
- OR extend layout-column-widget, extend layout-widget to use your extended column, use your extended layout

modules/@apostrophecms/layout-meta-widget/index.js
myovchev: We can't access the data from the layout widget from within the custom area editor. 
However, we can remove the meta for this iteration and find a better way in the future because this one is TOO hacky. A separate ticket for this is created.

modules/@apostrophecms/widget-type/index.js
myovchev: This is the source that will help @BoDonkey update the documentation.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
myovchev: I think it's more appropriate for a lot of scenarios. Mainly - it doesn't need composables, no need of Composition API.  Options API is much more readable IMO (and doesn't look anything like React code which is always a plus!!!).

ValJed: It's the new idiomatic way to write vue. they kept options api for BC but nothing tells us they won't delete it in the future. When nicely written composition is very nice, still way better than react ahah

myovchev: I think they claimed that options API is here to stay. I'm pretty sure about it.

.npmrc
```diff
@@ -0,0 +1 @@
+public-hoist-pattern[]=*
```
ValJed, line 1: Do we want to keep that ?

PR #5033, opened 2025-08-13
stylelint mad about buttons

boutell:
Looks the same to me, but I did remove a qualifier (to satisfy the stylelint rule about no qualifiers on element types), so running this past you in order to get our tests passing

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
@@ -284,17 +284,18 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-button.apos-operation {
+.apos-operation {
   @include apos-transition();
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
```diff
@@ -284,17 +284,18 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-button.apos-operation {
+.apos-operation {
```
boutell, line 287: the only substantive change

modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
```diff
@@ -284,17 +284,18 @@ export default {
...
   @include apos-transition();
+
```
boutell, line 289: all whitespace and ordering changes intentional to please stylelint

PR #5036, opened 2025-08-14
Pro 8087 aspect ratio image widget

ValJed:
[PRO-8087](https://linear.app/apostrophecms/issue/PRO-8087/autocrop-doesnt-work-for-image-widgets)

## Summary

Fix autocrop in `image-widget`.

[Cypress tests](https://github.com/apostrophecms/testbed/actions/runs/17261371871) 🟢 

[Cypress PR](https://github.com/apostrophecms/testbed/pull/372).

Autocrop should work for:
* Quick image upload from media manager
* Drag and drop
* Quick image upload from your device
* Old school widget editor selection

Also it now works with live preview properly without slowing everything up.
It fixes a problem of updated data not being properly synced between the editor `docFields.data` and `AposSchema` `fieldState` / `next`. It's better that data updates come from inputs / schema and not from above to avoid bugs / confusion.

I moved the postprocess relationship logic into its own file, and in the context of the editors, it's made when a relationship field is upaded (inside the field itself when it changes). This way no need to create more complexity by comparing documents and stuff.
Works well, never triggered except if a relationship changed. for that specific relationship. So works nicely with widget preview.

- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
 * Importing a custom icon from an npm module using a `~` path per the admin UI now works per the documentation, as long as the Vue component used for the icon is structured like those found in `@apostrophecms/vue-material-design-icons`.
 * The `button: true` flag works again for piece module utility operations. Previously the button appeared but did not trigger the desired operation.
 * Fixes the widget data being cloned to be saved before the `postprocess` method being called, which leads to a loss of data in `AposWidgetEditor` (like the autocrop data).
+* Fixes autocropping not working when uploading or selecting images from the new quick image upload UI.
+* In editors like `AposWidgetEditor` relationships are now post processed after they are updated in `AposInputRelationship` only for the relationship that has been updated. 
+It allows live preview to work well with it, it also avoids complexity and fixes updated data not being properly synced between the editor and the `AposSchema`.
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Make conditional fields work in Image Editor.
 * Importing a custom icon from an npm module using a `~` path per the admin UI now works per the documentation, as long as the Vue component used for the icon is structured like those found in `@apostrophecms/vue-material-design-icons`.
 * The `button: true` flag works again for piece module utility operations. Previously the button appeared but did not trigger the desired operation.
+* Fixes the widget data being cloned to be saved before the `postprocess` method being called, which leads to a loss of data in `AposWidgetEditor` (like the autocrop data).
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * `apos.doc.edit` now accepts an optional `values` object as the final parameter, containing initial values for some or all fields. This is supported only when editing existing documents.
 * When specifying a modal name to be executed, developers may now register "transformers" to be invoked first, using pipe syntax. For example, the modal name `aposSectionTemplateLibraryWidgetToDoc|AposDocEditor` will invoke the transformer `aposSectionTemplateLibraryWidgetToDoc` with the original props, and pass the returned result to `AposDocEditor`. Note that transformers are awaited. Transformers are registered in frontend admin UI code by passing a name and a function to `apos.ui.addTransformer`.
 * Adds quick image upload UI to `@apostrophecms/image-widget`.
+* Makes autocropping work when uploading or selecting images from the new quick image upload UI.
 
 ### Fixes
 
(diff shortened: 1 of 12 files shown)


modules/@apostrophecms/schema/ui/apos/logic/AposSchema.js
```diff
@@ -296,22 +293,23 @@ export default {
...
+            // that's why it gets seen as different than next all the time
             this.next.data[field.name] = this.fieldState[field.name].data;
           } else {
             this.next.data[field.name] = this.modelValue.data[field.name];
           }
         });
-      if (
-        oldHasErrors !== this.next.hasErrors ||
-        oldFieldState !== newFieldState
-      ) {
```
ValJed: oldFieldState and newFieldState are always different because we instantiate a new object for each every time.
So the `update:model-value` event was always triggered, even when not needed.
Might be cases were this change could bring bugs.. Should be discussed and handled differently if needed

modules/@apostrophecms/ui/ui/apos/composables/AposTheme.js
```diff
@@ -1,11 +1,12 @@
...
+export function useAposTheme() {
+  const themeClass = computed(_themeClass);
   return { themeClass };
 };
+
+export function _themeClass() {
+  const classes = [];
+  classes.push(`apos-theme--primary-${window.apos.ui.theme.primary}`);
+  return classes;
+}
```
ValJed, line 12: Not related to this ticket, but I used the same sharing code logic for ApostTheme and AposThemeMixin. Was small enough to bring it here.

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
```diff
@@ -219,56 +219,6 @@ export default {
...
-            relationships = [
-              ...relationships,
-              findRelationships(field.schema, object[field.name] || {})
-            ];
-          }
-        }
-        return relationships;
-      }
-    }
+    async postprocess() {}
```
ValJed: Should we keep this function empty in case it's still called from users components?

modules/@apostrophecms/schema/ui/apos/logic/AposSchema.js
ValJed: edit: just few broken tests are linked to this change, some are legit one like no more confirmation modal when cancelling and editor without any change, so it's actually a fix.
Another a bit more annoying linked to hidden required fields, the error doesn't disappear directly when the fields is hidden.
We should find another solution to fix it.
(maybe I'll restaured this code and create a ticket to handle this).

ValJed: Edit: I rollbacked the previous code and will create a ticket for that.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 * Make conditional fields work in Image Editor.
 * Importing a custom icon from an npm module using a `~` path per the admin UI now works per the documentation, as long as the Vue component used for the icon is structured like those found in `@apostrophecms/vue-material-design-icons`.
 * The `button: true` flag works again for piece module utility operations. Previously the button appeared but did not trigger the desired operation.
+* Fixes the widget data being cloned to be saved before the `postprocess` method being called, which leads to a loss of data in `AposWidgetEditor` (like the autocrop data).
+* Fixes autocropping not working when uploading or selecting images from the new quick image upload UI.
```
boutell: As long as this fix ships when quick upload ships for the first time, we do not need to list it as a fix because there was no release with the issue.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 * Make conditional fields work in Image Editor.
 * Importing a custom icon from an npm module using a `~` path per the admin UI now works per the documentation, as long as the Vue component used for the icon is structured like those found in `@apostrophecms/vue-material-design-icons`.
 * The `button: true` flag works again for piece module utility operations. Previously the button appeared but did not trigger the desired operation.
+* Fixes the widget data being cloned to be saved before the `postprocess` method being called, which leads to a loss of data in `AposWidgetEditor` (like the autocrop data).
```
boutell, line 20: This item makes sense to include 👍

modules/@apostrophecms/modal/ui/apos/mixins/AposEditorMixin.js
boutell: Keep it, but log a warning that this function no longer does anything and in rare cases where relationship postprocessing is truly required, you should look to `AposImageWidget.vue` for a model.

CHANGELOG.md
ValJed: Indeed!

PR #5037, opened 2025-08-14
more merge stuff

boutell:
🙏 hoping it doesn't shift under my feet again after your next approvals

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
@@ -622,10 +622,10 @@ export default {
         });
       }
       if (!this.widgetHasInitialModal(name)) {
-        const widget = this.newWidget(name);
+        const newWidget = this.newWidget(name);
         return this.insert({
           widget: {
-            ...widget,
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -622,10 +622,10 @@ export default {
         });
       }
       if (!this.widgetHasInitialModal(name)) {
-        const widget = this.newWidget(name);
+        const newWidget = this.newWidget(name);
```
boutell, line 625: Following Val's new pattern without "else if" requires scoping a new variable here to avoid conflict with the parameter.

modules/@apostrophecms/area/ui/apos/components/AposAreaExpandedMenu.vue
```diff
@@ -285,8 +285,6 @@ export default {
 
 <style lang="scss" scoped>
 .apos-operation {
-  @include apos-transition();
```
boutell: addressed a sass warning by moving this down. doesn't appear to be a conflict

PR #5038, opened 2025-08-18
PRO-8187: render-areas should not crash when related docs with areas are present

boutell:
This is a blocker to get section template library working reliably for QA.

### DIFF
--- a/modules/@apostrophecms/area/index.js
+++ b/modules/@apostrophecms/area/index.js
@@ -383,8 +383,8 @@ module.exports = {
               return;
             }
             // We're only rendering areas on the document, not ancestor or
-            // child page documents.
-            const regex = /^_(ancestors|children)|\._(ancestors|children)/;
+            // child page or related documents.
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/area/index.js
```diff
@@ -383,8 +383,8 @@ module.exports = {
               return;
             }
             // We're only rendering areas on the document, not ancestor or
-            // child page documents.
-            const regex = /^_(ancestors|children)|\._(ancestors|children)/;
+            // child page or related documents.
+            const regex = /^_|\._/;
```
boutell, line 387: Even if we wanted the related documents to be rendered, and the existing exception for ancestor and children pages suggest we do not... it currently does not work and would be difficult to implement. So just blocking it makes sense for now at least. Those who really want the areas in related documents rendered can hit the piece REST APIs of those documents again with `?render-areas=1`.

BoDonkey (approved):

PR #5040, opened 2025-08-25
PRO-8218: fix the use of vue icons from npm modules

boutell:
It looks like this could never have worked as written, because the variable name associated with the import would have a `~` in it.

Fixed it by slugifying. Makes a long import name, but only the import file cares, in the end the developer gets the name they wanted.

I hand-verified that the changes in task.js and external-output-api.js are identical, and also tested both `testbed` and `starter-kit-essentials`.

### DIFF
--- a/modules/@apostrophecms/schema/lib/addFieldTypes.js
+++ b/modules/@apostrophecms/schema/lib/addFieldTypes.js
@@ -1183,9 +1183,9 @@ module.exports = (self) => {
         }
         destination[field.name] = actualDocs;
       }
-      // "min" and "required" are not enforced server-side for relationships because it is always possible
-      // for the related document to be removed independently at some point. This leads to too many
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/schema/lib/addFieldTypes.js
```diff
@@ -1183,9 +1183,9 @@ module.exports = (self) => {
         }
         destination[field.name] = actualDocs;
       }
-      // "min" and "required" are not enforced server-side for relationships because it is always possible
-      // for the related document to be removed independently at some point. This leads to too many
-      // edge cases and knock-on effects if enforced
+      // "min" and "required" are not enforced server-side for relationships because
+      // it is always possible for the related document to be removed independently
```
boutell, line 1187: eslint

PR #5041, opened 2025-08-25
PRO-8220: "button: true" works again for piece module utility operations

boutell:
Regression: this got broken when we expanded the amount of information these operations can pass, between April and July. Impacts the ai helper module, but also new work like section template library that needs a utility operation not hidden away in a menu.

### DIFF
--- a/modules/@apostrophecms/piece-type/ui/apos/components/AposUtilityOperations.vue
+++ b/modules/@apostrophecms/piece-type/ui/apos/components/AposUtilityOperations.vue
@@ -8,9 +8,9 @@
       :key="button.action"
       type="default"
       :icon-only="button.iconOnly"
-      :icon="button.icon || false"
+      :icon="button.icon || null"
       :label="button.label"
-      @click="handleUtilityOperation(button.action)"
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/piece-type/ui/apos/components/AposUtilityOperations.vue
```diff
@@ -8,9 +8,9 @@
       :key="button.action"
       type="default"
       :icon-only="button.iconOnly"
-      :icon="button.icon || false"
+      :icon="button.icon || null"
```
boutell, line 11: Unrelated: without this we get a type warning at runtime

modules/@apostrophecms/piece-type/ui/apos/components/AposUtilityOperations.vue
```diff
@@ -8,9 +8,9 @@
...
       :label="button.label"
-      @click="handleUtilityOperation(button.action)"
+      @click="handleUtilityOperation(button)"
```
boutell, line 13: Must pass the entire item, as we do in the kebab menu case

PR #5047, opened 2025-08-28
the moduleName passed to a doc type manager is the doc type, not the widget type you happen to be picking it for

boutell:
One liner fixes ftw


boutell:
Test failure is not this PR, npm failed in github actions, not worried about it

PR #5053, opened 2025-09-02
PRO-8223: be careful when assuming doc is an actual doc

boutell:
I tracked the issue to the render-widget route, which passes the widget to certain document loaders as if it were a document. Trouble occurs when the "areas" loader assumes the "doc" it has been given is a real doc and uses it to impute a value for docId.

This is problematic because docId is intentionally `null` in the editor dialog.

I corrected the issue by verifying the `metaType` of `doc` first.

I also added this check in a few other places.

Finally, one piece of frontend code assumed `docId` could not be null, so I had to fix that as well.

### DIFF
--- a/modules/@apostrophecms/doc-type/index.js
+++ b/modules/@apostrophecms/doc-type/index.js
@@ -2491,7 +2491,7 @@ module.exports = {
                     area._edit = true;
                   }
 
-                  area._docId = doc._id;
+                  area._docId = doc._docId || ((doc.metaType === 'doc') ? doc._id : null);
                   for (const item of area.items) {
                     if (area._edit) {
                       // Keep propagating ._edit so a widget can be passed
@@ -2499,7 +2499,7 @@ module.exports = {
                       // -Tom
                       item._edit = true;
                     }
-                    item._docId = doc._id;
+                    item._docId = area._docId;
                     if (!widgetsByType[item.type]) {
                       widgetsByType[item.type] = [];
                     }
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/doc-type/index.js
```diff
@@ -2491,15 +2491,15 @@ module.exports = {
...
+                  area._docId = doc._docId || ((doc.metaType === 'doc') ? doc._id : null);
                   for (const item of area.items) {
                     if (area._edit) {
                       // Keep propagating ._edit so a widget can be passed
                       // like a doc to aposArea if it contains nested areas.
                       // -Tom
                       item._edit = true;
                     }
-                    item._docId = doc._id;
+                    item._docId = area._docId;
```
boutell, line 2502: set in line 2494, no need to duplicate logic

modules/@apostrophecms/schema/ui/apos/lib/detectChange.js
```diff
@@ -64,7 +64,7 @@ export function detectFieldChange(field, v1, v2) {
       const newObject = {};
       for (const [ key, val ] of Object.entries(o)) {
         if (key === '_docId') {
-          newObject._docId = o._docId.replace(/:.*$/, '');
+          newObject._docId = (o._docId == null) ? null : o._docId.replace(/:.*$/, '');
```
boutell, line 67: don't crash on null, which is a thing in the modal

BoDonkey (approved):

PR #5057, opened 2025-09-03
PRO-8256: allow createWidgetOperations to be filtered based on a list of widget types

boutell:
... So we don't offer "insert section" if none of the widget types allowed in this area are permitted to be the basis for a section.

This does NOT mean you can never hit "insert section" and find an empty list. That would require an expensive API call, so it's not worth it.

### DIFF
--- a/modules/@apostrophecms/area/index.js
+++ b/modules/@apostrophecms/area/index.js
@@ -191,6 +191,7 @@ module.exports = {
       // interface regardless of whether `options.widgets` or `options.groups`
       // was used.
       getWidgets(options) {
+        // Keep in sync with client-side implementation
         let widgets = options.widgets || {};
 
         if (options.groups) {
--- a/modules/@apostrophecms/area/ui/apos/lib/get-widgets.js
+++ b/modules/@apostrophecms/area/ui/apos/lib/get-widgets.js
@@ -1,0 +1,14 @@
+// Requires an area field's .options property
+
+export default function getWidgets(options) {
+  let widgets = options.widgets || {};
+  if (options.groups) {
+    for (const group of Object.keys(options.groups)) {
+      widgets = {
+        ...widgets,
+        ...options.groups[group].widgets
+      };
+    }
+  }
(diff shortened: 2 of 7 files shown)


modules/@apostrophecms/area/index.js
```diff
@@ -191,6 +191,7 @@ module.exports = {
       // interface regardless of whether `options.widgets` or `options.groups`
       // was used.
       getWidgets(options) {
+        // Keep in sync with client-side implementation
```
boutell, line 194: Not a new issue, but I avoided further duplication within the client side.

Right now we have some ESM/commonjs issues that prevent us from sharing code as well as I'd like between front and back end.

modules/@apostrophecms/area/ui/apos/lib/get-widgets.js
```diff
@@ -0,0 +1,14 @@
+// Requires an area field's .options property
+
+export default function getWidgets(options) {
```
boutell, line 3: I didn't strictly have to factor this out today, but in the long run I'd like to have one implementation, so this is the right direction.

modules/@apostrophecms/widget-type/index.js
```diff
@@ -123,29 +123,6 @@ module.exports = {
     origin: null,
     preview: true
   },
-  handlers(self) {
```
boutell, line 126: Registering the widget operations late prevents the section template library from altering them cleanly if some modules are configured before and some after, which is likely because of our support for `modules.js` files with `nestedModuleSubdirs`. Yes this took me a while to figure out.

modules/@apostrophecms/widget-type/index.js
```diff
@@ -203,6 +182,24 @@ module.exports = {
         });
       },
 
+      composeWidgetOperations() {
```
boutell, line 185: no actual changes to the logic here.

myovchev (approved):

PR #5061, opened 2025-09-08
PRO-8268: do the URL encoding work before Astro.redirect sees the URL

boutell:
We do it with `encodeurl` because that is what Express does (I checked the source). That module drives defensively, it doesn't re-encode what is already encoded, so I don't want to do the naive thing and just call `encodeURI` instead.

### DIFF
--- a/modules/@apostrophecms/area/ui/apos/components/AposAreaContextualMenu.vue
+++ b/modules/@apostrophecms/area/ui/apos/components/AposAreaContextualMenu.vue
@@ -202,7 +202,10 @@ export default {
         return [];
       }
       const menu = [ ...this.contextMenuOptions.menu ];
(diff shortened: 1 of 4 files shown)


modules/@apostrophecms/area/ui/apos/components/AposAreaContextualMenu.vue
```diff
@@ -202,7 +202,10 @@ export default {
         return [];
       }
       const menu = [ ...this.contextMenuOptions.menu ];
-      const createWidgetOperations = filterCreateWidgetOperations(this.moduleOptions, this.options);
+      const createWidgetOperations = filterCreateWidgetOperations(
+        this.moduleOptions,
```
boutell, line 206: unrelated lint

PR #5064, opened 2025-09-09
PRO-8245 `widget` store, switch state control

myovchev:
## Summary

- A widgets central store was introduced (similar to modal).
- The breadrcrumb switch widget operation can now be controlled by any UI application via store options.


modules/@apostrophecms/ui/ui/apos/apps/AposWidgetStore.js
```diff
@@ -0,0 +1,14 @@
...
+  apos.widget = {
+    refs: widgetStore.refs,
+    toId: widgetStore.toId,
+    get: widgetStore.get,
+    set: widgetStore.set,
+    getOrSet: widgetStore.getOrSet,
+    update: widgetStore.update,
+    remove: widgetStore.remove
+  };
+};
```
ValJed, line 9: I think it would be preferable to call `useWidgetStore` wherever you need it, especially when it's in vue components.
Like we do for modals:
```javascript
import { useModalStore } from 'Modules/@apostrophecms/ui/stores/modal';

const store = useModalStore();
``` 
and use the store in your component the way you want, you'll be sure that reactivity is respected.

We did similar things (attaching store stuff to window.apos) because we needed to use these out of vue (stores can be used out of vue) but especially to avoid too much refactoring and time spent.
Also when doing that I'm not sure about the reactivity, like for refs. So When I did that I used  only functions that do stuff or return data, to be sure it would work even without reactivity.

Anyway I think it makes sense to just use this store where needed.

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbSwitch.vue
```diff
@@ -41,14 +45,15 @@ export default {
     'update'
   ],
   data() {
+    const name = `${this.name}:switch`;
     return {
-      next: this.value || null
+      next: this.value || null,
+      store: window.apos.widget.getOrSet(this.widgetId, name, this.value || null),
+      namespace: name
```
ValJed, line 56: Should get the store using:
```
const widgetStore = useModalStore();
return {
  widgetStore,
  store: widgetStore.getOrSet(this.widgetId, name, this.value || null),
}

```

modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbSwitch.vue
```diff
@@ -41,14 +45,15 @@ export default {
...
     };
   },
-  computed: {},
-  watch: {
-  },
-  mounted() {
+  unmounted() {
+    window.apos.widget.remove(this.widgetId, this.namespace);
```
ValJed: Then could just:
```
store.remove(this.widgetId, this.namespace);
```

modules/@apostrophecms/layout-widget/ui/apos/components/AposAreaLayoutEditor.vue
```diff
@@ -216,6 +216,7 @@ export default {
         index: 0
       });
       this.layoutMode = 'layout';
+      apos.widget.update(this.parentOptions?.widgetId, 'layout:switch', 'layout');
```
ValJed: Same here, could useWidgetStore in the setup function and then just use `store.update`, etc

PR #5066, opened 2025-09-09
PRO-8245 public build, tablet responsive fix

myovchev:
## Summary

Make sure that the last item, if it's an odd count, is marked for full width by the server. We need that because the visual order can be completely different than the natural order, which confuses any pure CSS solution.


BoDonkey (approved):
LGTM

PR #5071, opened 2025-09-11
Pro 8152 anchors

BoDonkey:
## Summary

This PR updates the `annotateAreaForExternalFront()` method to accept `annotateWidgetForExternalFront()` methods from any widget module. This will allow for individual widgets to pass options to an external frontend, like Astro.

*For example:*
1. In the `starter-kit-astro-essentials` backend add a project-level option to the image-widget.
2. In the frontend, modify the image widget to log the `options` from the frontmatter.
3. Add an image widget to any area and check the logs.

*(Check at least one)*

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 * Redirects to URLs containing accent marks and other non-ascii characters now behave as expected with Astro. Pre-encoding the URLs exactly the way `res.redirect` would before passing them to Astro prevents an error in Astro and allows the redirect to succeed.
 * Removes the non-functional `uniqueUsername` route from the `user` module
+* Modifies the `annotateAreaForExternalFront()` method of the `@apostrophecms/template` module to accept a per-module `annotateWidgetForExternalFront()` method. This allows widgets to send project-level options alongside the per-area options to external frontends.
 * Updated dependencies to address deprecation warnings.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 * Fixes a bug in the login `uponSubmit` filter where a user could login without meeting the requirement.
 
+### Fixes
+
+* Resolve inline image URLs correctly when in edit mode and not in the default locale.
+
 ### Changes
 
 * Redirects to URLs containing accent marks and other non-ascii characters now behave as expected with Astro. Pre-encoding the URLs exactly the way `res.redirect` would before passing them to Astro prevents an error in Astro and allows the redirect to succeed.
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -1319,7 +1319,7 @@ module.exports = {
             const module = self.apos.modules[moduleName];
 
             if (module && module.annotateWidgetForExternalFront) {
-              const widgetOptions = module.annotateWidgetForExternalFront(module);
+              const widgetOptions = module.annotateWidgetForExternalFront();
 
               // Add options directly to the widget item
               if (widgetOptions && Object.keys(widgetOptions).length > 0) {
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -1289,7 +1289,7 @@ module.exports = {
       // and `choices` properties, and guaranteeing that `items` exists,
       // at least as an empty array.
 
-      async annotateAreaForExternalFront(field, area) {
+      annotateAreaForExternalFront(field, area) {
         area.field = field;
         area.options = field.options;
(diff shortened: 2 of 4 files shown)


modules/@apostrophecms/template/index.js
```diff
@@ -1303,11 +1303,27 @@ module.exports = {
...
           }
+
+          // Annotate each individual widget with its options
+          // Each widget must elect into this by creating an
+          // `annotateWidgetForExternalFront() method.
+          const manager = self.apos.area.getManager?.(item.type) ||
+            self.apos.area.widgetManagers?.[item.type];
+
+          let widgetOptions = {};
+          if (manager && typeof manager.annotateWidgetForExternalFront === 'function') {
```
boutell: You don't need this if anymore. It's going to exist.

modules/@apostrophecms/widget-type/index.js
```diff
@@ -449,7 +449,12 @@ module.exports = {
           }
           return true;
         });
+      },
+
+      annotateWidgetForExternalFront() {
```
boutell, line 454: 👍

modules/@apostrophecms/template/index.js
boutell: By adding it to the base class we've made it part of the contract you can expect.

I guess you can still do if (manager).

BoDonkey: I thought I would have to error on the side of caution. Learning!

modules/@apostrophecms/template/index.js
```diff
@@ -1317,11 +1317,8 @@ module.exports = {
           const manager = self.apos.area.getManager?.(item.type) ||
             self.apos.area.widgetManagers?.[item.type];
 
-          let widgetOptions = {};
-          if (manager && typeof manager.annotateWidgetForExternalFront === 'function') {
-            widgetOptions = await manager.annotateWidgetForExternalFront() || {};
-          };
-          if (widgetOptions && Object.keys(widgetOptions).length > 0) {
+          const widgetOptions = await manager.annotateWidgetForExternalFront() || {};
+          if (Object.keys(widgetOptions).length > 0) {
```
boutell: Why not always provide it, so people on the astro side don't have to drive defensively with `?.`? An empty object isn't that expensive.

boutell (changes requested):

modules/@apostrophecms/template/index.js
BoDonkey: Updated

modules/@apostrophecms/template/index.js
```diff
@@ -1289,7 +1289,7 @@ module.exports = {
       // and `choices` properties, and guaranteeing that `items` exists,
       // at least as an empty array.
 
-      annotateAreaForExternalFront(field, area) {
+      async annotateAreaForExternalFront(field, area) {
```
boutell: Does this really need to be async? It's a major performance hit when multiplied by every area and then every widget, and there's no async work being done to look up options. Also I don't think you changed anything where it's called, so there's probably just a promise showing up there?

My starting assumption is that this annotation work should not be async. It's meant to be very lightweight to minimize the impact of using an external frontend.

modules/@apostrophecms/template/index.js
```diff
@@ -1303,11 +1303,22 @@ module.exports = {
...
             item._docId = area._docId;
           }
+
+          // Annotate each individual widget with its options
+          // Each widget must elect into this by creating an
+          // `annotateWidgetForExternalFront() method.
+          const manager = self.apos.area.getManager?.(item.type) ||
+            self.apos.area.widgetManagers?.[item.type];
+
+          const widgetOptions = await manager.annotateWidgetForExternalFront() || {};
```
boutell: Should not be async either.

modules/@apostrophecms/widget-type/index.js
boutell, line 454: Noticing this is already not async (and that's good), I think "async" in the area annotation method might have been a discarded early thought that hung around?

CHANGELOG.md
```diff
@@ -12,10 +12,15 @@
 
 * Fixes a bug in the login `uponSubmit` filter where a user could login without meeting the requirement.
 
+### Fixes
+
+* Resolve inline image URLs correctly when in edit mode and not in the default locale.
```
boutell, line 17: It looks like one of Miro's PRs got caught up in this PR somehow. But I'm going to ignore it because I know I approved it.

PR #5081, opened 2025-09-24
Process manually and inject CSS, fix manage layout CSS glitch

myovchev:
## Summary

- Load and preprocess the public styles on init
- Inject the appropriate style on runtime
- Remove the public sass
- Fix a glitch when manage UI is not properly shown when mobile preview is active

Closes PRO-8322

### DIFF
--- a/modules/@apostrophecms/layout-widget/index.js
+++ b/modules/@apostrophecms/layout-widget/index.js
@@ -228,20 +228,29 @@ module.exports = {
         try {
           const postcss = require('postcss');
           const cssnano = require('cssnano');
-          const mobilePreview = require('postcss-viewport-to-container-toggle');
+          const assetOptions = self.apos.asset.breakpointPreviewMode || {};
 
-          const resultApos = await postcss([
-            mobilePreview({
-              modifierAttr: 'data-breakpoint-preview-mode',
-              debug: self.apos.adminBar.breakpointPreviewMode?.debug === true,
-              transform: self.apos.adminBar.breakpointPreviewMode?.transform || null
-            }),
+          const resultPublic = await postcss([
             cssnano({
               preset: 'default'
             })
           ]).process(cssContent, { from: undefined });
 
-          const resultPublic = await postcss([
+          if (assetOptions.enable !== true) {
+            return {
+              apos: resultPublic.css,
+              public: resultPublic.css
+            };
+          }
+
+          const mobilePreview = require('postcss-viewport-to-container-toggle');
+
+          const resultApos = await postcss([
+            mobilePreview({
+              modifierAttr: 'data-breakpoint-preview-mode',
+              debug: assetOptions.debug === true,
+              transform: assetOptions.transform || null
+            }),
             cssnano({
               preset: 'default'
             })
--- a/modules/@apostrophecms/layout-widget/index.js
+++ b/modules/@apostrophecms/layout-widget/index.js
@@ -218,10 +218,12 @@ module.exports = {
         const mobileBreakpointPlus = mobileBreakpoint + 1;
         const tabletBreakpoint = self.options.tablet?.breakpoint || 1024;
         const tabletBreakpointPlus = tabletBreakpoint + 1;
-        cssContent = cssContent.replace(/\{\$mobile\}/g, mobileBreakpoint);
-        cssContent = cssContent.replace(/\{\$mobile-plus\}/g, mobileBreakpointPlus);
-        cssContent = cssContent.replace(/\{\$tablet\}/g, tabletBreakpoint);
-        cssContent = cssContent.replace(/\{\$tablet-plus\}/g, tabletBreakpointPlus);
+        cssContent = cssContent
+          .replace(/\{\$mobile\}/g, mobileBreakpoint)
+          .replace(/\{\$mobile-plus\}/g, mobileBreakpointPlus)
+          .replace(/\{\$tablet\}/g, tabletBreakpoint)
+          .replace(/\{\$tablet-plus\}/g, tabletBreakpointPlus);
+
         return self.processCss(cssContent);
       },
       async processCss(cssContent) {
@@ -250,18 +252,19 @@ module.exports = {
               modifierAttr: 'data-breakpoint-preview-mode',
               debug: assetOptions.debug === true,
               transform: assetOptions.transform || null
-            }),
-            cssnano({
-              preset: 'default'
             })
-          ]).process(cssContent, { from: undefined });
+          ]).process(resultPublic.css, { from: undefined });
 
           return {
             apos: resultApos.css,
             public: resultPublic.css
           };
         } catch (error) {
-          const errorTrace = error.stack ? error.stack.split('\n').slice(1).join('\n') : '';
+          const errorTrace = error.stack
+            ? error.stack
+              .split('\n')
+              .slice(1)
+            : [];
           self.logWarn('inline-css-minify-failed', error.message, {
             error: errorTrace
           });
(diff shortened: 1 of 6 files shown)


modules/@apostrophecms/layout-widget/index.js
```diff
@@ -179,6 +210,57 @@ module.exports = {
...
+          const postcss = require('postcss');
+          const cssnano = require('cssnano');
+          const mobilePreview = require('postcss-viewport-to-container-toggle');
+
+          const resultApos = await postcss([
+            mobilePreview({
+              modifierAttr: 'data-breakpoint-preview-mode',
+              debug: self.apos.adminBar.breakpointPreviewMode?.debug === true,
+              transform: self.apos.adminBar.breakpointPreviewMode?.transform || null
+            }),
```
ValJed: Should check that the option `breakpointPreviewMode` is enabled in `@apostrophecms/asset`. Otherwise the plugin shouldn't be used.

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -64,9 +68,17 @@ module.exports = {
...
     };
   },
-  init(self) {
+  async init(self) {
     self.metaWidgetName = '@apostrophecms/layout-meta';
     self.columnWidgetName = '@apostrophecms/layout-column';
+
+    if (
+      self.__meta.name === '@apostrophecms/layout-widget' &&
+      self.options.injectStyles !== false
```
ValJed, line 77: Not sure why you check the module's name, it should pass only once in this init function, unless this module is extended. I suppose you want this code to be ran only once at startup ?

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -179,6 +210,66 @@ module.exports = {
...
+        const cssFilePath = path.join(__dirname, 'ui', 'src', 'layout.css');
+        let cssContent = fs.readFileSync(cssFilePath, 'utf8');
+        const mobileBreakpoint = self.options.mobile?.breakpoint || 600;
+        const mobileBreakpointPlus = mobileBreakpoint + 1;
+        const tabletBreakpoint = self.options.tablet?.breakpoint || 1024;
+        const tabletBreakpointPlus = tabletBreakpoint + 1;
+        cssContent = cssContent.replace(/\{\$mobile\}/g, mobileBreakpoint);
+        cssContent = cssContent.replace(/\{\$mobile-plus\}/g, mobileBreakpointPlus);
+        cssContent = cssContent.replace(/\{\$tablet\}/g, tabletBreakpoint);
+        cssContent = cssContent.replace(/\{\$tablet-plus\}/g, tabletBreakpointPlus);
```
ValJed: What about 😁
```js
        const cssContent = fs.readFileSync(cssFilePath, 'utf8')
          .replace(/\{\$mobile\}/g, mobileBreakpoint)
          .replace(/\{\$mobile-plus\}/g, mobileBreakpointPlus)
          .replace(/\{\$tablet\}/g, tabletBreakpoint)
          .replace(/\{\$tablet-plus\}/g, tabletBreakpointPlus);
```

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -179,6 +210,66 @@ module.exports = {
...
+        return self.processCss(cssContent);
+      },
+      async processCss(cssContent) {
+        try {
+          const postcss = require('postcss');
+          const cssnano = require('cssnano');
+          const assetOptions = self.apos.asset.breakpointPreviewMode || {};
```
ValJed: they are more `breakpointPreviewModeOptions` than `assetOptions`

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -179,6 +210,66 @@ module.exports = {
...
+              preset: 'default'
+            })
+          ]).process(cssContent, { from: undefined });
+
+          return {
+            apos: resultApos.css,
+            public: resultPublic.css
+          };
+        } catch (error) {
+          const errorTrace = error.stack ? error.stack.split('\n').slice(1).join('\n') : '';
```
ValJed: multiline would help reading this.

modules/@apostrophecms/layout-widget/ui/src/layout.css
```diff
@@ -14,21 +6,21 @@
...
 
-  & > div {
-    grid-column: var(--colstart, auto) / span var(--colspan, 1);
-    grid-row: var(--rowstart, auto) / span var(--rowspan, 1);
-    justify-self: var(--justify);
-    /* stylelint-disable-next-line declaration-block-no-redundant-longhand-properties */
-    align-self: var(--align);
-    order: var(--order, 0);
-  }
+.layout-widget > div {
```
ValJed, line 11: layout-widget feels like a too random name, and not related to apos. I can imagine users having the same classes.. We should use a data attribute with apos in it.

modules/@apostrophecms/layout-widget/index.js
```diff
@@ -179,6 +210,66 @@ module.exports = {
...
+          const resultApos = await postcss([
+            mobilePreview({
+              modifierAttr: 'data-breakpoint-preview-mode',
+              debug: assetOptions.debug === true,
+              transform: assetOptions.transform || null
+            }),
+            cssnano({
+              preset: 'default'
+            })
+          ]).process(cssContent, { from: undefined });
```
ValJed: I think you should pass `resultPublic.css` and not `cssContent` otherwise the above processing is useless.

ValJed (changes requested):
Few things, mostly questions. I think a mistake in the `processCss` method.

modules/@apostrophecms/layout-widget/index.js
myovchev: You don't want to inject CSS more than once (it's a "one to rule 'em all" kind of situations).
The layout is always ON because it's registered by Apostrophe.

modules/@apostrophecms/layout-widget/ui/src/layout.css
myovchev: We have default `image-widget`. 
People are free to override stuff if they feel they need to.

modules/@apostrophecms/layout-widget/index.js
myovchev: ~~I don't understand this recommendation~~

modules/@apostrophecms/layout-widget/index.js
myovchev: I don't understand the problem, could you please clarify.

modules/@apostrophecms/layout-widget/index.js
myovchev: Done, it also don't need the final join.

modules/@apostrophecms/layout-widget/ui/src/layout.css
myovchev: The `apos-` convention is purely Admin UI related - the context where Apostorphe needs absolute control.
The public build is the opposite - the project has to be in absolute control. Thus `apos` prefix for public facing style makes no sense.

modules/@apostrophecms/layout-widget/index.js
myovchev: It's not useless, but I hear you. I assume minified content won't break the mobile preview pre-processor.

modules/@apostrophecms/layout-widget/index.js
ValJed: Just naming feedback

modules/@apostrophecms/layout-widget/ui/src/layout.css
ValJed: Yes good point.

PR #5086, opened 2025-10-01
Fix focus

ValJed:
## Summary

Fix layout widget focus not working when clicking on grid component.
Fix typing in rich text re rendering all widgets around and breaking focus
Put focus stuff in widget store because was a mess

Ref PRO-8326

Bugs are fixed

- [X] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other


stuartromanek (approved):
From a UX perspective this seems to do all the right things in terms of switching focus among nested widgets, focusing widgets appropriately during delete/paste.

modules/@apostrophecms/widget-type/ui/apos/mixins/AposWidgetMixin.js
```diff
@@ -12,8 +12,10 @@ export default {
   },
   watch: {
     modelValue: {
-      handler() {
-        this.renderContent();
+      handler(widget) {
+        if (widget.type !== '@apostrophecms/layout-column') {
+          this.renderContent();
+        }
```
ValJed, line 18: This fixes the issue of the rich text unfocusing because the entire area is re rendered through livre preview when typing (since before the rich text used it's own component without this mixin and now layout-widget and layout-column use this mixin with `AposWidget`).
I don't thing `layout-widget` and `layout-column-widget` should use `AposWidget` like regular widgets. I thing they should have their own basic componnent (like we did for quick image upload).
Let me know what you think.

myovchev: Both `layout-widget` and `layout-column-widget` should be regular apos widgets. The main driving power is the AposAreaLayoutEditor override. 
We need a more abstract, configurable way, to fix the problem. 

I don't think this is the right solution as well. We should discuss and find a better way.

ValJed: I would not say that they are regular widget, they're pretty special in my opinion, they wrap normal widgets and aren't content widget so reacting to their content doesn't make sense. For example they wrap rich text widget that has its own component and where changes should not trigger a re render.. Once wrapped in the layout this re renders,
I'm exploring the preview that is documented but this one doesn't work when set on iindividual widgets..

myovchev (approved):
Approving so that we can merge and QA. The marked problematic code will be handled in a separate ticket.

PR #5100, opened 2025-10-09
Pro 8285 login case insensitive

ValJed:
## Summary

Adds option to make login username / email case insensitive.

When option `caseInsensitve` is set to true a migration is run to make all users usernames / emails case insensitive. If a conflicts occurs we log an errors with the needed infos to fix the situation.

If option is disabled the migration is removed (will be re run if option is re set to true).

* usernames / emails are normalized when inserted in docs and safe collections.
* When verifying logins we normalize user input if option is true.

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Layout widget for dynamic grid layouts.
 * `widgetOperations` support for `placement: 'breadcrumb'` to add operations to the breadcrumb menu of widgets. Extend the widget operations configuration to support various features when in the breadcrumb menu.
 * Area template (Nunjucks) support for `aposStyle`, `aposClassName`, `aposParentOptions` and `aposAttrs` contextual named variables (`with {}` syntax).
-* New login option `caseInsensitive` to force login usernames and emails to be case insensitive. New task `login-case-insensitive` udpating all login names / email to lowercase, used by a new migration when switching ton `caseInsensitive`.
+* New login option `caseInsensitive` to force login usernames and emails to be case insensitive. New task `login-case-insensitive` updating all login names / email to lowercase, used by a new migration when switching to `caseInsensitive`.
 
 ### Changes
 


modules/@apostrophecms/login/index.js
```diff
@@ -57,6 +57,7 @@ module.exports = {
       username: 'apostrophe:enterUsername',
       password: 'apostrophe:enterPassword'
     },
+    caseInsensitive: false,
```
ValJed, line 60: Def value

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +999,87 @@ module.exports = {
...
+      },
+
+      async manageLoginCaseInsensitiveMigration() {
+        if (self.options.caseInsensitive) {
+          self.apos.migration.add('login-case-insensitive', async () => {
+            await self.caseInsensitiveTask();
+          });
+        } else {
+          self.apos.migration.cancel('login-case-insensitive');
+        }
```
ValJed, line 1011: If caseInsensitive is activated then deactivated, we remove the migration to re run it if the option is re activated. Makes sense to me to run it anytime you switch from sensitive to insensitive..
WDYT?

modules/@apostrophecms/login/index.js
```diff
@@ -781,10 +794,12 @@ module.exports = {
...
       // return an incomplete token.
       async initialLogin(req) {
-        const username = self.apos.launder.string(req.body.username);
+        const username = self.normalizeLoginName(
+          self.apos.launder.string(req.body.username)
+        );
         const password = self.apos.launder.string(req.body.password);
 
-        if (!(username && password)) {
+        if (!username || !password) {
```
ValJed, line 800: more readable

modules/@apostrophecms/login/index.js
```diff
@@ -808,7 +823,7 @@ module.exports = {
           await self.verifyRequirements(req, onTimeRequirements);
 
           // send log information
-          const user = await self.apos.login.verifyLogin(
+          const user = await self.verifyLogin(
```
ValJed, line 824: already in login module

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +996,87 @@ module.exports = {
...
+          });
+        } else {
+          self.apos.migration.cancel('login-case-insensitive');
+        }
+      },
+
+      async caseInsensitiveTask() {
+        const duplicatedUsernames = [];
+        await self.apos.migration.eachDoc({ type: '@apostrophecms/user' }, 1, async (user) => {
+          const usernameLower = user.username.toLowerCase();
```
boutell: Use the normalizeLoginName method here, otherwise as that method evolves this will get out of sync

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +996,87 @@ module.exports = {
...
+          const emailLower = user.email?.toLowerCase();
+          const shouldUpdateUsername = user.username !== usernameLower;
+          const shouldUpdateEmail = user.email && user.email !== emailLower;
+
+          if (!shouldUpdateUsername && !shouldUpdateEmail) {
+            return;
+          }
+          const existingUsername = await self.apos.doc.db.findOne({
```
boutell: Not just a username issue. The username, the email, or both could lead to duplication and therefore trigger a unique key error.

boutell: I suggest you actually do the update, catch errors, and test with `if (self.apos.doc.isUniqueErr(err)) { ... }`. This is how we solve it for slugs, it'll run a lot faster than fetching every user individually whether they have a conflict or not

modules/@apostrophecms/migration/index.js
```diff
@@ -67,16 +67,21 @@ module.exports = {
...
         self.migrations.push({
           name,
           options,
           fn: migrationFn
         });
       },
+
+      // Cancel a migration previously ran by removing it from the database
+      // Should not be used most of the time, except when some options change and
+      // that the migration should be run again.
```
boutell: I feel this is overkill. I think folks aren't going to migrate to lowercase and then go back to mixed case.

modules/@apostrophecms/user/index.js
```diff
@@ -194,12 +194,16 @@ module.exports = {
             doc.password = self.apos.util.generateId();
           }
         },
+        normalizeLoginNames(req, doc, options) {
```
boutell, line 197: This method might be unused.

boutell (changes requested):

modules/@apostrophecms/login/index.js
ValJed: Good point. but I thought email could be used as a username but still only the username would have been asked to be unique. I'll use this start is it's the way we do in other locations.

modules/@apostrophecms/migration/index.js
ValJed: Yes I just thought it would be very messy if they did, and this tiny hack could prevent it. But you're probably right.

modules/@apostrophecms/user/index.js
ValJed: This is a `beforeInsert` middleware that normalizes user names, not a method.

modules/@apostrophecms/login/index.js
ValJed: Problem is that running the task shouldn't depend on login option, it could end up with unexpected results.
Running the case insensitive task should make login info case insensitive.
I can add an option to override the default config for caseInsensitive.

modules/@apostrophecms/login/index.js
ValJed: Ok when implementing it I realized (thanks to my tests) that `username` has an index only in the safe collection not in the docs one..
So to be sure it breaks properly without updating the docs collection before the update must be done on the safe collection first.
Since there is an index in the safe collection and that username must be unique why not having an index in AposDocs user usernames?

modules/@apostrophecms/migration/index.js
ValJed: Removed this logic.

modules/@apostrophecms/login/index.js
boutell: There is no index on `username` or `email` in the main docs collection at all. It isn't necessary because we always insert or update in the safe first and that collection has the rule. Also it would be awkward for other piece and page types that might have `username` and `email` fields (more likely the latter) that are unrelated to login security.

modules/@apostrophecms/user/index.js
boutell: oops

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +994,77 @@ module.exports = {
...
+
+          const shouldUpdateUsername = user.username !== normalizedUsername;
+          const shouldUpdateEmail = user.email && user.email !== normalizedEmail;
+          if (!shouldUpdateUsername && !shouldUpdateEmail) {
+            return;
+          }
+
+          try {
+            if (shouldUpdateUsername) {
+              await self.apos.user.safe.updateOne(
```
boutell: The safe has both username and email properties you need to be worrying about, but for email you should only worry about it if `email` is truthy (this is important: the email index is "unique" but also "sparse," meaning you need to leave email undefined unless the user actually has one)

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +994,77 @@ module.exports = {
...
+                { _id: user._id },
+                {
+                  $set: {
+                    username: normalizedUsername
+                  }
+                }
+              );
+            }
+            await self.apos.doc.db.updateOne(
```
boutell: Why would you make this call at all if neither "should" flag is set?

I suggest that for both the safe and the main docs collection you build the same `$set` object with either one or two properties as needed and use it both calls to reduce code duplication.

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +994,77 @@ module.exports = {
...
+              { _id: user._id },
+              {
+                $set: {
+                  username: normalizedUsername,
+                  ...shouldUpdateEmail && { email: normalizedEmail }
+                }
+              }
+            );
+          } catch (err) {
+            if (self.apos.doc.isUniqueError(err)) {
```
boutell, line 1038: It might be the username, it might be the email. I think that's OK but we need to express that ambiguity to the user running the task.

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +994,77 @@ module.exports = {
...
+              return;
+            }
+            throw err;
+          }
+        });
+
+        if (duplicatedUsernames.length) {
+          self.logError(
+            'conflicting-usernames',
+            'Some usernames changed in lowercase already exist for other users, please fix it or they won\'t be able to log in anymore',
```
boutell: It might be the username, it might be the email. I think that's OK but we need to express that ambiguity to the user running the task.

test/login.js
```diff
@@ -999,3 +997,411 @@ describe('Login', function () {
...
+      const expected = {
+        loginError: false,
+        username: 'Virus',
+        email: 'Virus@gmail.com'
+      };
+
+      assert.deepEqual(actual, expected);
+    });
+
+    it('should migrate usernames and emails to lowercase when running the caseInsensitiveTask method', async function() {
```
boutell, line 1073: This test should also verify that the error logging communicates which accounts can't be updated due to a conflict in either username or email, and that two users with an *empty* email field are *not* treated as a conflict.

modules/@apostrophecms/login/index.js
ValJed: There is an early return if none should flag are set:
```js
          if (!shouldUpdateUsername && !shouldUpdateEmail) {
            return;
          }
```

I created a criteria, I missed the fact that safe users have email indeed.
So whenever I update the safe user I update the normal one too.

modules/@apostrophecms/login/index.js
ValJed: Right, updated the error to look like this:
```json
"failed": [
    {
      "user": {
        "_id": "uujr6k5k9n5ocqcwoyac98qc",
        "username": "Conflict",
        "email": "conflict@gmail.com"
      },
      "conflictingFields": {
        "username": "conflict"
      }
    }
```

modules/@apostrophecms/login/index.js
ValJed: updated to add emails.

modules/@apostrophecms/login/index.js
ValJed: added a test to check the error log

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Support for the `render-areas` query parameter in the REST API when using Astro as an external frontend, provided the Astro project has the corresponding route. This allows section template library previews to work in Astro projects. For ease of migration, if Astro cannot satisfy the request, ApostropheCMS will also attempt to render the widget natively.
 * Made `self.apos.externalFrontKey` available, simplifying API calls back to Astro.
+* New login option `caseInsensitive` to force login usernames and emails to be case insensitive.
```
boutell: Mention the task too. We will need documentation for this cc @BoDonkey

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +994,69 @@ module.exports = {
             required: true
           })
           );
+      },
+
+      normalizeLoginName(usernameOrEmail, {
+        caseInsensitive = self.options.caseInsensitive
```
boutell, line 1000: Ooh, never thought of accessing self.options here before

modules/@apostrophecms/login/index.js
boutell: This will add it yes and run it, one time ever, as a migration. For weird edge cases like turning it on and off again and on again, they should use the task.

modules/@apostrophecms/login/index.js
```diff
@@ -981,6 +994,69 @@ module.exports = {
...
+              return;
+            }
+            throw err;
+          }
+        });
+
+        if (duplicatedUsernames.length) {
+          self.logError(
+            'conflicting-usernames',
+            'Some usernames or emails changed to lowercase already exist for other users, please fix it or they won\'t be able to log in anymore',
```
boutell: Better wording:

"Accounts with certain usernames and/or emails would be in conflict with other accounts if changed to lowercase. Please review the following usernames and emails and address them manually."

boutell (changes requested):
Just minor stuff now.

CHANGELOG.md
ValJed: done

modules/@apostrophecms/login/index.js
ValJed: done

modules/@apostrophecms/login/index.js
ValJed: Pretty handy

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 * Support for the `render-areas` query parameter in the REST API when using Astro as an external frontend, provided the Astro project has the corresponding route. This allows section template library previews to work in Astro projects. For ease of migration, if Astro cannot satisfy the request, ApostropheCMS will also attempt to render the widget natively.
 * Made `self.apos.externalFrontKey` available, simplifying API calls back to Astro.
-* New login option `caseInsensitive` to force login usernames and emails to be case insensitive.
+* New login option `caseInsensitive` to force login usernames and emails to be case insensitive. New task `login-case-insensitive` udpating all login names / email to lowercase, used by a new migration when switching ton `caseInsensitive`.
```
boutell: udpating -> updating

ton -> on

(sorry)

ValJed: My bad

boutell (approved):

PR #5106, opened 2025-10-13
sticky controls

stuartromanek:
Custom sticky positioning for widget controls
- Keeps controls in view no matter how tall widget
- is aware of the admin bar and gets out of the way
- tolerates a certain amount of window resizing and 'does the right thing'
- 

https://github.com/user-attachments/assets/fe786800-f75e-4b8d-b389-69fc4102f1b3

### DIFF
--- a/modules/@apostrophecms/layout-widget/index.js
+++ b/modules/@apostrophecms/layout-widget/index.js
@@ -5,6 +5,7 @@ module.exports = {
   extend: '@apostrophecms/widget-type',
   options: {
     label: 'apostrophe:layout',
+    icon: 'view-column-icon',
     initialModal: false,
     skipOperations: [ 'edit' ],
     columns: 12,


modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -440,12 +459,109 @@ export default {
...
+          }
+
+          this.lastResizeTop = widgetRect.top;
+          this.resizeTicking = false;
+        });
+      }
+    },
+    stickyControlsScroll() {
+      if (!this.scrollTicking) {
+        window.requestAnimationFrame(() => {
```
ValJed: [detail]
Can get rid of `window` since you did in `stickyControlsResize`.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -440,12 +459,109 @@ export default {
...
+          }
+
+          // Widget is under admin bar
+          if (widgetRect.top <= this.adminBarHeight) {
+
+            // Widget bottom is approaching admin bar,
+            // position controls absolutely to the bottom
+            if (
+              visibleHeight <=
+                this.controlsHeight + this.adminBarHeight + (this.controlsMargin * 3)
```
ValJed: Could store the logic into variables to avoid the need for comments.
Would help follow the if else flow too/
Also this code `(this.controlsMargin * 3)` the 3 seems arbitrary? Maybe a variable for these random numbers for people to understand what it is for.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -440,12 +459,109 @@ export default {
...
+          ) {
+            this.stickyControlsStyles = newStyles;
+          }
+
+          this.scrollTicking = false;
+        });
+
+        this.scrollTicking = true;
+      }
+    },
```
ValJed, line 615: This one could be splitted in two.
this function in charge of the overall logic, and another function in charge of computing the `newStyles` and returning it.
Would make the code more readable, and remove hard to track reassignment.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
```diff
@@ -440,12 +459,109 @@ export default {
...
+          // Only update if styles changed
+          if (
+            Object.keys(newStyles).some(
+              key => this.stickyControlsStyles[key] !== newStyles[key]
+            )
+          ) {
+            this.stickyControlsStyles = newStyles;
+          }
+
+          this.scrollTicking = false;
```
ValJed, line 610: This is set to false on every frame, is it intentional? 
It's purpose is to know if some scroll logic is running right?
Maybe `this.scrollTicking = true;` should be at the start of the frame cb?

stuartromanek: it’s set to true right after scheduling (not inside) because requestAnimationFrame is async — setting it earlier prevents multiple frames from being queued before the first one runs. it’s reset to false at the end of the callback to allow the next frame.

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
stuartromanek: now computing this on mounted with comments, see `this.totalUiOffset`

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
stuartromanek: I did this by
- statically composing the `default` and `bottom` positions inside `data` as these don't change and aren't subject to the window
- Add a method computing the floating style, as it requires the latest `widgetRect` to compute the `right` prop
- Add a method for conditionally updating the styles that both handlers use

lmk what you think

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
stuartromanek: oh, misread your comment (about comments!) lmk if you want me to remove them (my comments, not your comments)

modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
ValJed: yes better !

ValJed (approved):

PR #5110, opened 2025-10-15
PRO-8435: log what widget type was missing

boutell:
* simple logging change
* plus some cleanup to avoid unnecessary logic

### DIFF
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -1314,11 +1314,14 @@ module.exports = {
           // Annotate each individual widget with its options
           // Each widget must elect into this by creating an
           // `annotateWidgetForExternalFront() method.
-          const manager = self.apos.area.getManager?.(item.type) ||
-            self.apos.area.widgetManagers?.[item.type];
-
-          const widgetOptions = manager.annotateWidgetForExternalFront() || {};
-          item._options = widgetOptions;
+          const manager = self.apos.area.getWidgetManager(item.type);
+          if (manager) {
+            const widgetOptions = manager.annotateWidgetForExternalFront();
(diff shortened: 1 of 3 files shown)


modules/@apostrophecms/template/index.js
```diff
@@ -1314,11 +1314,14 @@ module.exports = {
           // Annotate each individual widget with its options
           // Each widget must elect into this by creating an
           // `annotateWidgetForExternalFront() method.
-          const manager = self.apos.area.getManager?.(item.type) ||
```
boutell, line 1317: `getWidgetManager` always exists, the existing code was just calling the wrong method

modules/@apostrophecms/template/index.js
```diff
@@ -1314,11 +1314,14 @@ module.exports = {
...
-            self.apos.area.widgetManagers?.[item.type];
-
-          const widgetOptions = manager.annotateWidgetForExternalFront() || {};
```
boutell, line 1320: The default implementation returns `{}` and it's a new method so there should be no need for `||` here

modules/@apostrophecms/template/index.js
```diff
@@ -1314,11 +1314,14 @@ module.exports = {
...
-          item._options = widgetOptions;
+          const manager = self.apos.area.getWidgetManager(item.type);
+          if (manager) {
+            const widgetOptions = manager.annotateWidgetForExternalFront();
+            item._options = widgetOptions;
+          } else {
+            self.apos.area.warnMissingWidgetType(item.type);
```
boutell, line 1322: I'm following and upgrading an existing pattern to report this a bit more usefully before throwing the fatal error

BoDonkey (approved):

PR #5120, opened 2025-10-23
PRO-8530 autodetect ES bundles

boutell:
section-template-library is our first bundle delivered in ES module format.

Turns out, autodetection of ES bundles didn't work, and autodetection of improvements in ES bundles also didn't work.

We were sliding along just adding explicit `bundles` where it should not have been necessary and manually adding improvements to `modules` in `app.js` in places like testbed, which is particularly painful since that list can grow in a particular bundle, but not fully documenting these steps as requirements for section-template-library, probably because knew on some level they should not be required at all.

Turns out, there was a bit of autodetect code that never got modernized for ES modules. I've done that here following the same pattern as in `moog-require`.

This is the proper resolution to the problem iCIMS encountered installing section-template-library.

### DIFF
--- a/index.js
+++ b/index.js
@@ -440,7 +440,14 @@ async function apostrophe(options, telemetry, rootSpan) {
       const { default: apostropheModule } = await importFresh(npmPath);
       const bundle = apostropheModule.bundle;
       if (bundle) {
-        self.options.bundles =  [...new Set([...(self.options.bundles || []), apostropheModuleName])];
+        self.options.bundles = [
+          ...new Set(
+            [
+              ...(self.options.bundles || []),
+              apostropheModuleName
+            ]
+          )
+        ];
         const bundleModules = bundle.modules;
         for (const bundleModuleName of bundleModules) {
           if (!apostropheModules.includes(bundleModuleName)) {


index.js
```diff
@@ -14,6 +14,8 @@ const glob = require('./lib/glob.js');
 const moogRequire = require('./lib/moog-require');
 let defaults = require('./defaults.js');
 
+const importFresh = moduleName => import(`${moduleName}?${Date.now()}`);
```
BoDonkey, line 17: So bust cache by adding the date

index.js
```diff
@@ -435,16 +437,24 @@ async function apostrophe(options, telemetry, rootSpan) {
...
 
-      const apostropheModule = await self.root.import(npmPath);
-      if (apostropheModule.bundle) {
-        self.options.bundles = (self.options.bundles || []).concat(apostropheModuleName);
-        const bundleModules = apostropheModule.bundle.modules;
+      const { default: apostropheModule } = await importFresh(npmPath);
+      const bundle = apostropheModule.bundle;
+      if (bundle) {
+        self.options.bundles = [
+          ...new Set(
```
BoDonkey, line 444: Protect against `concat()` making duplicates by using a `Set()`, so just a new guardrail.

PR #5126, opened 2025-10-28
PRO-5472: file widget

boutell:
A very basic, functional treatment. You can set a class in the usual way. Fills a visible gap in the offerings.

### DIFF
--- a/modules/@apostrophecms/layout-column-widget/index.js
+++ b/modules/@apostrophecms/layout-column-widget/index.js
@@ -172,7 +172,8 @@ module.exports = {
             widgets: {
               '@apostrophecms/rich-text': {},
               '@apostrophecms/image': {},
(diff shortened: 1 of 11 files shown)


modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -172,7 +172,8 @@ module.exports = {
             widgets: {
               '@apostrophecms/rich-text': {},
               '@apostrophecms/image': {},
-              '@apostrophecms/video': {}
+              '@apostrophecms/video': {},
+              '@apostrophecms/file': {}
```
boutell, line 176: one of the straightforward core offerings, so I put it in column defaults.

stuartromanek (approved):

PR #5129, opened 2025-10-30
uses operation name for key and id (test purpose)

ValJed:
Fix layout broken tests


modules/@apostrophecms/area/ui/apos/components/AposBreadcrumbOperations.vue
```diff
@@ -137,7 +137,7 @@ export default {
           listeners: this.getOperationListeners(operation),
           name: operation.name || null,
           action: operation.action || null,
-          key: operation.action || operation.name,
+          key: operation.name,
```
ValJed, line 140: name was used before (because action and name were same), name is unique and defined thus should be used, also fixes tests

PR #5136, opened 2025-11-04
add get-apos-doc-id and set-apos-doc-id tasks

haroun:
## Summary

Adds `@apostrophecms/doc:get-apos-doc-id` and `@apostrophecms/doc:set-apos-doc-id` tasks

1. Tests must be green
2. You can try to manually run the tasks `@apostrophecms/doc:get-apos-doc-id` and `@apostrophecms/doc:set-apos-doc-id` and check the results

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/modules/@apostrophecms/doc/index.js
+++ b/modules/@apostrophecms/doc/index.js
@@ -1851,16 +1851,15 @@ module.exports = {
         if (!_id && !slug) {
           throw self.apos.error('invalid', 'Either _id or slug must be provided');
         }
-        if (!locale) {
+        if (!_id && !locale) {
           throw self.apos.error('invalid', 'Missing locale');
         }
 
-        const aposLocale = new RegExp(`^${self.apos.util.regExpQuote(locale)}:`);
         const criteria = _id
           ? { _id }
           : {
             slug,
-            aposLocale
+            aposLocale: new RegExp(`^${self.apos.util.regExpQuote(locale)}:`)
           };
 
         const doc = await self.apos.doc.db
@@ -1929,7 +1928,7 @@ module.exports = {
   tasks(self) {
     return {
       'get-apos-doc-id': {
-        usage: 'Retrieve the aposDocId from the _id or slug, and the locale. Usage: "node app.js @apostrophecms/doc:get-apos-doc-id --_id=pfh0haxfpzowht3oi213cqos:fr:draft --locale=fr" or "node app.js @apostrophecms/doc:get-apos-doc-id --slug=test --locale=fr"',
+        usage: 'Retrieve the aposDocId from the _id or slug, and the locale. Usage: "node app.js @apostrophecms/doc:get-apos-doc-id --_id=pfh0haxfpzowht3oi213cqos:fr:draft" or "node app.js @apostrophecms/doc:get-apos-doc-id --slug=test --locale=fr"',
         task: async (argv) => {
           const {
             _id, slug, locale
--- a/test/docs.js
+++ b/test/docs.js
@@ -1309,7 +1309,7 @@ describe('Docs: tasks', function () {
     await apos.lock.db.deleteMany({});
   });
 
-  it('should require the id or slug when calling @apostrophecms/doc:get-apos-doc-id task', async function() {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Adds
 
 * Translation strings added for the layout- and layout-column-widgets.
+* Adds `@apostrophecms/doc:get-apos-doc-id` and `@apostrophecms/doc-set-apos-doc-id` tasks.
```
boutell: Missing a : in the second one

modules/@apostrophecms/doc/index.js
```diff
@@ -1843,8 +1845,115 @@ module.exports = {
...
+        id, slug, locale
+      }) {
+        if (!id && !slug) {
+          throw self.apos.error('invalid', 'Either id or slug must be provided');
+        }
+        if (!locale) {
+          throw self.apos.error('invalid', 'Missing locale');
+        }
+
+        const req = self.apos.task.getReq({ mode: 'draft' });
```
boutell: You need to put locale in this req too. Otherwise if document A has slug /foo in "en" and unrelated document B has slug /foo in the intended locale you will get the wrong `aposDocId` back

modules/@apostrophecms/doc/index.js
```diff
@@ -1843,8 +1845,115 @@ module.exports = {
...
+
+        const criteria = id
+          ? { _id: self.apos.i18n.inferIdLocaleAndMode(req, id) }
+          : { slug };
```
boutell: Query by slug is going to do the wrong thing if req isn't for the right locale

modules/@apostrophecms/doc/index.js
```diff
@@ -1843,8 +1845,115 @@ module.exports = {
...
+        }
+
+        const originalId = (slug && !oldId)
+          ? await self.getAposDocId({
+            slug,
+            locale
+          })
+          : oldId;
+
+        const modes = [ 'draft', 'published' ];
```
boutell: actually draft, published and previous can all exist.

modules/@apostrophecms/doc/index.js
boutell: Consider a direct mongodb query, you have the ingredients for one and it gets `req` out of it and will be able to find edge cases like an archived document if the user wants to

boutell (changes requested):

modules/@apostrophecms/doc/index.js
```diff
@@ -1843,8 +1845,130 @@ module.exports = {
...
+        if (!_id && !slug) {
+          throw self.apos.error('invalid', 'Either _id or slug must be provided');
+        }
+        if (!locale) {
+          throw self.apos.error('invalid', 'Missing locale');
+        }
+
+        // const req = self.apos.task.getReq({ mode: 'draft', locale });
+        const aposLocale = new RegExp(`^${self.apos.util.regExpQuote(locale)}:`);
+        const criteria = _id
```
boutell, line 1858: One last thing, sorry... since the _id parameter has to be a full _id, which already contains the locale, there's no need to make locale mandatory or even part of the query when _id is present.

(Technically we don't need to make a query at all when _id is the argument, because _id already contains aposDocId, but I like the idea of making the query because it validates that the document actually does exist.)

boutell (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Add `@apostrophecms/migration:add-missing-schema-fields` task. This task does not run database migrations.
 * Translation strings added for the layout- and layout-column-widgets.
 * Adds `@apostrophecms/doc:get-apos-doc-id` and `@apostrophecms/doc:set-apos-doc-id` tasks.
+* New `box` schema field type
```
boutell, line 10: How did this creep into this PR? It's not hurting anything

PR #5137, opened 2025-11-04
 PRO-8154: Strip Latin accents

myovchev:
## Summary

Introduce a configurable option `stripUrlAccents` in the `@apostrophecms/i18n` module to control whether Latin accents are stripped from URLs (slugs) and attachment names.

- Default remains unchanged: accents are preserved in slugs (backward compatible).
- When `@apostrophecms/i18n.options.stripUrlAccents` is `true`:
  - New/updated document slugs (pages, pieces, slug fields) have accents removed.
  - Attachment `name` values are normalized without accents on upload.
  - The admin UI slug input reflects accent-stripping in real time.
- A new task `@apostrophecms/i18n:strip-slug-accents` is provided to update existing document slugs and attachment names when adopting this option.

1) Default behavior (accents preserved)
- Ensure `stripUrlAccents` is absent or set to `false` in `@apostrophecms/i18n` options.
- Create a page and a piece titled: `C'est déjà l'été`.
- Expected slugs:
  - Piece: `c-est-déjà-l-été`
  - Page: `/c-est-déjà-l-été`
- Upload an attachment named: `été-image.png`
  - Expected stored name in database (aposAttachments collection): `été-image`

2) Enable accent stripping for new/updated content
- Set `@apostrophecms/i18n.options.stripUrlAccents = true`.
- Create a new page and piece:
  - Title: `C'est déjà l'été`
  - Expected slugs:
    - Piece: `c-est-deja-l-ete`
    - Page: `/c-est-deja-l-ete`
- Upload an attachment named: `été-image.png`
  - Expected stored name in database (aposAttachments collection): `ete-image`
- In the admin UI, type an accented title and observe the slug field: it should show accent-free characters.

3) Migrate existing content (optional when adopting the option)
- With `stripUrlAccents: true`, run the task:
  - `node app @apostrophecms/i18n:strip-slug-accents`
- Expected:
  - Existing slugs like `/c-est-déjà-l-été` become `/c-est-deja-l-ete`.
  - Attachment names like `été-image` become `ete-image`.
  - Task logs counts of updated documents and attachments.

Notes
- The option only affects new/updated content unless the task is run.
- Non-Latin scripts (e.g., Cyrillic) continue to be handled by `sluggo`; the deburring only affects Latin accent marks.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


modules/@apostrophecms/doc/index.js
```diff
@@ -205,7 +205,10 @@ module.exports = {
       '@apostrophecms/doc-type:beforePublish': {
         testPermissions(req, info) {
           if (info.options.permissions !== false) {
-            if (!self.apos.permission.can(req, info.options.autopublishing ? 'edit' : 'publish', info.draft)) {
+            if (!self.apos.permission.can(
+              req,
+              info.options.autopublishing ? 'edit' : 'publish', info.draft
```
haroun:
```suggestion
              info.options.autopublishing ? 'edit' : 'publish',
              info.draft
```

modules/@apostrophecms/i18n/index.js
```diff
@@ -70,7 +70,9 @@ module.exports = {
     i18n: {
       ns: 'apostrophe',
       browser: true
-    }
+    },
+    // If true, slugifying will strip accents from Latin characters
+    stripUrlAccents: false
```
haroun, line 75: This and https://github.com/apostrophecms/apostrophe/pull/5137/files#diff-3520f0c3913f7e40c36f796cec657de7f9667bc7950f042a2490c236a89b9efbR54 will not allow someone to enable/disable this feature per-locale as stated in the ticket.

modules/@apostrophecms/i18n/index.js
```diff
@@ -677,7 +679,8 @@ module.exports = {
           debug: self.debug,
           show: self.show,
           action: self.action,
-          crossDomainClipboard: req.session && req.session.aposCrossDomainClipboard
+          crossDomainClipboard: req.session && req.session.aposCrossDomainClipboard,
+          stripUrlAccents: self.options.stripUrlAccents
```
haroun, line 683: this should be in the `locales` object instead

modules/@apostrophecms/i18n/index.js
```diff
@@ -757,7 +763,7 @@ module.exports = {
         }
         req.baseUrlWithPrefix = `${req.baseUrl}${self.apos.prefix}`;
         req.absoluteUrl = req.baseUrlWithPrefix + req.url;
-        req.prefix = `${req.baseUrlWithPrefix}${self.locales[req.locale].prefix || ''}`;
+        req.prefix = `${req.baseUrlWithPrefix}${self.locales[req.locale]?.prefix || ''}`;
```
haroun: Something is wrong if the request managed to reach this point with an incorrect `req.locale`. I think the i18n middleware already throws an invalid error. Why if this needed?

modules/@apostrophecms/i18n/index.js
```diff
@@ -1251,6 +1257,53 @@ module.exports = {
...
+      'strip-slug-accents': {
+        usage: 'Remove Latin accent characters from all document slugs and attachment names. Usage: node app @apostrophecms/i18n:strip-slug-accents',
+        async task() {
+          let docChanged = 0;
+          let attachmentChanged = 0;
+
+          await self.apos.migration.eachDoc({}, 5, async doc => {
+            const slug = doc.slug;
+            const req = self.apos.task.getAdminReq({
+              locale: doc.aposLocale?.split(':')[0] || 'en'
```
haroun: Forcing the `locale` to `en` seems wrong. We have `i18n` `defaultLocale` for that.
If you're using `doc.aposLocale` you can also pass it as `query: { aposLocale: doc.aposLocale }` and let the i18n middleware do its magic.

modules/@apostrophecms/i18n/index.js
```diff
@@ -1251,6 +1257,53 @@ module.exports = {
...
+            }
+
+            doc.slug = self.apos.util.slugify(doc.slug, {
+              stripAccents: true,
+              allow: '/'
+            });
+            if (slug !== doc.slug) {
+              await self.apos.doc.update(req, doc, { permissins: false });
+              docChanged++;
+              self.apos.util.log(`Updated doc [${req.locale}] "${slug}" -> "${doc.slug}"`);
```
haroun: During my testing I saw this

```
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-13"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-09"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-06"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-13"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-06"
Updated doc [en] "/@copy-of-copy-of-copy-of-template-02" -> "/@template-05"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-11"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-09"
Updated doc [en] "/@copy-of-copy-of-template-02" -> "/@template-04"
Updated doc [en] "/@copy-of-copy-of-copy-of-template-02" -> "/@template-05"
Updated doc [en] "/@copy-of-template-02" -> "/@template-03"
Updated doc [en] "/@copy-of-copy-of-template-02" -> "/@template-04"
Updated doc [en] "/@template-02" -> "/@template-02"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-11"
Updated doc [en] "/@template-02" -> "/@template-02"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-10"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-10"
Updated doc [en] "@template-article-03" -> "@template-article-03"
Updated doc [en] "/@copy-of-template-02" -> "/@template-03"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-12"
Updated doc [fr] "c-est-l-été-déjà" -> "c-est-l-ete-deja"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-12"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-07"
Updated doc [fr] "c-est-l-été-déjà" -> "c-est-l-ete-deja"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-07"
Updated doc [en] "@template-article-03" -> "@template-article-03"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-08"
Updated doc [en] "/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-08"
Updated 28 document slug(s) and 0 attachment name(s).
```

There are too many changes I think. Something seems off.

modules/@apostrophecms/i18n/index.js
```diff
@@ -1251,6 +1257,53 @@ module.exports = {
...
+            if (!self.shouldStripAccents(req)) {
+              return;
+            }
+
+            doc.slug = self.apos.util.slugify(doc.slug, {
+              stripAccents: true,
+              allow: '/'
+            });
+            if (slug !== doc.slug) {
+              await self.apos.doc.update(req, doc, { permissins: false });
```
haroun: There is a typo with `permissions`.

Updating using `self.apos.doc.update` means there are no redirects created. The old URLs give me 404s now.

modules/@apostrophecms/i18n/index.js
```diff
@@ -1251,6 +1257,53 @@ module.exports = {
...
+          });
+
+          const req = self.apos.task.getAdminReq();
+          await self.apos.attachment.each({}, 10, async (attachment) => {
+            if (!self.shouldStripAccents(req)) {
+              return;
+            }
+            const slug = self.apos.util.slugify(attachment.name, { stripAccents: true });
+            if (slug !== attachment.name) {
+              await self.apos.attachment.db.updateOne(
```
haroun: Using `self.apos.attachment.db.updateOne` means the file will not be updated in `uploadfs`. I'm not very familiar with all that logic, but we should think about it.

modules/@apostrophecms/page-type/index.js
```diff
@@ -468,14 +468,14 @@ module.exports = {
       },
       // If the page does not yet have a slug, add one based on the
       // title; throw an error if there is no title
-      ensureSlug(page) {
+      ensureSlug(page, { stripAccents } = {}) {
```
haroun: I don't think this scales well. We should work towards `ensureSlug` and `slugify` keeping the same signature somehow and fetching the `stripAccents` from the `i18n` `locales` option in my opinion. If we don't do that, it means that we have to change multiple external/pro modules to support the new feature, and apostrophe users will have to do the same. There will always be a piece of code somewhere where `stripAccents` is not passed and it will lead to inconsistent behavior that will be hard to track.

modules/@apostrophecms/util/index.js
```diff
@@ -291,8 +291,15 @@ module.exports = {
       // ONE punctuation character normally forbidden in slugs may
       // optionally be permitted by specifying it via options.allow.
       // The separator may be changed via options.separator.
+      // By default, accents are preserved. To strip accents,
+      // set options.stripAccents to true.
       slugify(s, options) {
-        return require('sluggo')(s, options);
+        const { stripAccents, ...opts } = options || {};
+        const slug = require('sluggo')(s, opts);
```
haroun, line 298: 1. We should put that `require` statement at the top.
2. Is the order of operations important here? slugify then strip accents vs strip accents then slugify.

modules/@apostrophecms/util/index.js
```diff
@@ -291,8 +291,15 @@ module.exports = {
       // ONE punctuation character normally forbidden in slugs may
       // optionally be permitted by specifying it via options.allow.
       // The separator may be changed via options.separator.
+      // By default, accents are preserved. To strip accents,
+      // set options.stripAccents to true.
       slugify(s, options) {
-        return require('sluggo')(s, options);
+        const { stripAccents, ...opts } = options || {};
```
haroun, line 297: This new option is a challenge because we need to update all the usages of the `slugify` method, in our modules but our clients needs to do the same. There is a default value, but this is some kind of a breaking change, mainly because it requires changes to the code for the behavior to remain consistent.

modules/@apostrophecms/schema/ui/apos/logic/AposInputSlug.js
```diff
@@ -48,6 +49,9 @@ export default {
     },
     localePrefix() {
       return this.field.page && apos.i18n.locales[apos.i18n.locale].prefix;
+    },
+    stripAccents() {
+      return apos.i18n.stripUrlAccents === true;
```
haroun, line 54: Having this here will make it difficult to have this feature enabled/disabled per locales.

modules/@apostrophecms/i18n/index.js
myovchev: This was a request from Tom, simple mvp, no per language control.

modules/@apostrophecms/i18n/index.js
myovchev: Not now.

modules/@apostrophecms/i18n/index.js
myovchev: It breaks tasks for no reason.

modules/@apostrophecms/i18n/index.js
myovchev: `aposLocale` can be undefined. I'll switch to default locale though.

modules/@apostrophecms/i18n/index.js
myovchev: This is exactly what I want.

modules/@apostrophecms/page-type/index.js
myovchev: ensureSlug is internal method. It's called internally via event. 
Utils should not fetch/know what i18n options are, so slugify is what it is - an util with an option. 
Having an option for ensureSlug is a clear contract. This is not a public API so no BC is required. 

I reverted option per locale because of many and bad showstoppers. I explained above that the there was an agreement to go with the simple "rule them all" option.

modules/@apostrophecms/util/index.js
myovchev: Moving the require is out of context for this PR. (Many other utils does it). 
The current order of operation makes sense. You don't want to change the behavior of sluggo with pre-filtering the input, but to filter the output of the sluggo.

modules/@apostrophecms/util/index.js
myovchev: The accent cleaning behavior can't break BC because it's a new feature. This code is fully BC.

modules/@apostrophecms/schema/ui/apos/logic/AposInputSlug.js
myovchev: This is one of the main showstoppers for per locale feature, which we don't implement with the current PR.

modules/@apostrophecms/i18n/index.js
myovchev: This is definitely off. This is very very off `"/@copy-of-copy-of-copy-of-copy-of-copy-of-copy-of-template-02" -> "/@template-08"` (completely different slugs). I'll be investigating

modules/@apostrophecms/i18n/index.js
myovchev: True, I'll switch to manager update.

modules/@apostrophecms/i18n/index.js
myovchev: Switched to deburr instead of slugify. Could you please retest with the DB that showed those results?

myovchev:
@haroun Added a patch to address the problems
- in the task, only accents are removed/compared, no double slugify is attempted; it should solve the problems
- in the task, switched to the default language
- in the task, we use manager to update the doc now, so that the URL redirects are recorded.

modules/@apostrophecms/i18n/index.js
haroun: We should investigate why, checking the code tells me this should not be possible. I'm afraid this is hiding a bug. By using optional chaining, we're might be delaying the moment we discover a bug that has been introduced recently.

test/utils.js
```diff
@@ -371,6 +372,49 @@ describe('Utils', async function() {
...
+      assert.equal(
+        apos.util.slugify('/C\'est déjà l\'été', {
+          stripAccents: true,
+          allow: '/'
+        }),
+        '/c-est-deja-l-ete',
+        'failed to slugify and strip accents with slug options'
+      );
+    });
+
```
haroun, line 434: we should also test with the `@apostrophecms/i18n` option set to `stripUrlAccents: true`

modules/@apostrophecms/i18n/index.js
```diff
@@ -1251,6 +1257,54 @@ module.exports = {
...
+            doc.slug = _.deburr(doc.slug);
+            if (slug !== doc.slug) {
+              const manager = self.apos.doc.getManager(doc.type);
+              if (!manager) {
+                return;
+              }
+              await manager.update(req, doc, { permissions: false });
+              docChanged++;
+              self.apos.util.log(`Updated doc [${req.locale}] "${slug}" -> "${doc.slug}"`);
+            }
```
haroun:
```suggestion
            if (slug === doc.slug) {
              return;
            }
            
            const manager = self.apos.doc.getManager(doc.type);
            if (!manager) {
              return;
            }
            await manager.update(req, doc, { permissions: false });
            docChanged++;
            self.apos.util.log(`Updated doc [${req.locale}] "${slug}" -> "${doc.slug}"`);
```

modules/@apostrophecms/i18n/index.js
haroun: I've checked the code in the attachment module + uploadfs, and I don't think this will work. Can we discuss about it?

modules/@apostrophecms/i18n/index.js
myovchev: Nvm, it's mostly test related and makes no sense to blow up. Reverting.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Adds
 
+* Adds `stripUrlAccents` option in `@apostrophecms/i18n` module to globally control whether accents are stripped from URLs. When set to `true`, all URLs (slugs) will have accents from Latin characters removed on document creation and updates. No existing documents are modified automatically; this only affects new or updated documents. A new task `node app @apostrophecms/i18n:strip-slug-accents` is provided to update existing document slugs and attachment `name` props in the database when needed.
```
boutell: Still mentions attachment name props, which are no longer part of this.

modules/@apostrophecms/i18n/index.js
```diff
@@ -1251,6 +1257,38 @@ module.exports = {
...
         }
+      },
+      'strip-slug-accents': {
+        usage: 'Remove Latin accent characters from all document slugs. Usage: node app @apostrophecms/i18n:strip-slug-accents',
+        async task() {
+          let docChanged = 0;
+
+          await self.apos.migration.eachDoc({}, 5, async doc => {
+            const slug = doc.slug;
+            const req = self.apos.task.getAdminReq({
```
boutell, line 1268: Feature is not per locale, as discussed. We can remove this req.

modules/@apostrophecms/schema/ui/apos/logic/AposInputSlug.js
boutell: As discussed the feature is not per locale.

PR #5140, opened 2025-11-05
PRO-8571: fix the astro section template library bug introduced in astro 4.9.0

boutell:


modules/@apostrophecms/area/index.js
```diff
@@ -376,7 +376,9 @@ module.exports = {
           const response = await fetch(`${self.apos.baseUrl}/api/apos-external-front/render-area`, {
             method: 'POST',
             headers: {
-              'apos-external-front-key': self.apos.externalFrontKey
+              'apos-external-front-key': self.apos.externalFrontKey,
+              // Without this Astro enforces CSRF protection starting in version 4.9.0
```
boutell, line 380: Of course it is always appropriate to specify a content type here so I'm not really blaming Astro but they started enforcing it recently

modules/@apostrophecms/area/index.js
```diff
@@ -376,7 +376,9 @@ module.exports = {
...
+              'content-type': 'application-json'
```
haroun: You mean this instead or I'm missing something?
```suggestion
              'content-type': 'application/json'
```

boutell:
Fixed that nasty typo, thank you!

haroun (approved):

PR #5153, opened 2025-11-11
Static build

boutell:
## Summary

Introduced static build support via external frontend.

### DIFF
--- a/packages/apostrophe-astro/lib/aposPageFetch.js
+++ b/packages/apostrophe-astro/lib/aposPageFetch.js
@@ -12,7 +12,7 @@ export default async function aposPageFetch(req) {
       });
     }
     const response = await aposResponse(request);
-    let headers =response.headers;
+    let headers = response.headers;
     aposData = await response.json();
 
     // Apostrophe's external-front middleware returns redirects as JSON
--- a/packages/apostrophe/modules/@apostrophecms/attachment/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/attachment/index.js
@@ -990,7 +990,8 @@ module.exports = {
       // schema recursively and collect `idsStorage` values from
       // relationship fields that point to types with attachment
       // fields.  Also handles the special case of rich-text
-      // widget inline images (`imageIds`).
+      // widget inline images (`imageIds`) and widgets whose own
+      // schema contains a direct `type: 'attachment'` field.
       //
       // Found IDs are added to the `target` Set
       // (aposDocId values, locale-agnostic).
@@ -1005,10 +1006,13 @@ module.exports = {
           },
           // Rich-text widgets store inline image references in
           // `imageIds` outside of any schema relationship field.
-          // Handled here via the `widget` handler so that a single
-          // schema-driven walk covers both relationships and
-          // rich-text image refs without a second full-doc walk.
+          // Widgets whose own schema has a direct `type: 'attachment'`
+          // field (e.g. a custom widget storing a file directly) also
+          // need to include the parent doc's ID.
           widget: (field, widget) => {
+            if (self.hasAttachmentFields(widget.type)) {
+              target.add(doc.aposDocId);
+            }
             if (
               widget.type === '@apostrophecms/rich-text' &&
               Array.isArray(widget.imageIds)
--- a/packages/apostrophe/test/static-build.js
+++ b/packages/apostrophe/test/static-build.js
@@ -749,6 +749,219 @@ describe('Static Build Support', function () {
     });
   });
 
+  describe('used scope with direct attachment fields', function () {
+    let apos;
+
+    before(async function () {
+      apos = await t.create({
+        root: module,
+        modules: {
+          '@apostrophecms/url': {
+            options: { static: true }
+          },
+          // Piece type with a direct attachment field in its own schema
+          'direct-attachment-piece': {
+            extend: '@apostrophecms/piece-type',
+            options: {
+              name: 'direct-attachment-piece',
+              label: 'Direct Attachment Piece',
+              alias: 'directAttachmentPiece'
+            },
+            fields: {
+              add: {
+                file: {
+                  type: 'attachment',
+                  label: 'File',
+                  group: 'office'
+                }
+              }
+            }
+          },
+          // Widget with a direct attachment field (not a relationship)
+          'attachment-widget': {
+            extend: '@apostrophecms/widget-type',
+            options: {
+              label: 'Attachment Widget'
+            },
+            fields: {
+              add: {
+                photo: {
+                  type: 'attachment',
+                  label: 'Photo',
+                  fileGroup: 'images'
+                }
+              }
+            }
+          },
+          // Page type with an area that allows the attachment widget
+          'test-page': {
+            extend: '@apostrophecms/page-type',
+            options: {
+              label: 'Test Page'
+            },
+            fields: {
+              add: {
+                body: {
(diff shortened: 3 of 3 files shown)


packages/apostrophe/test/static-build.js
```diff
@@ -0,0 +1,2015 @@
...
+      });
+
+      it('req.absoluteUrl uses staticBaseUrl during static builds', function () {
+        const req = apos.task.getAnonReq({
+          mode: 'published',
+          aposStaticBuild: true,
+          url: '/some-page'
+        });
+        assert(
+          req.absoluteUrl.startsWith('https://www.example.com'),
```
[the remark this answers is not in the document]
myovchev: Shut up. It's a test.

packages/apostrophe-astro/lib/aposPageFetch.js
```diff
@@ -11,8 +12,48 @@ export default async function aposPageFetch(req) {
       });
     }
     const response = await aposResponse(request);
+    let headers =response.headers;
```
boutell: formatting

packages/apostrophe-astro/lib/aposPageFetch.js
```diff
@@ -11,8 +12,48 @@ export default async function aposPageFetch(req) {
...
     aposData = await response.json();
-    aposData.aposResponseHeaders = response.headers;
+
+    // Apostrophe's external-front middleware returns redirects as JSON
+    // (e.g. { redirect: true, url: '/fr/', status: 302 }). When the
+    // redirect only adds or removes a trailing slash we should follow
+    // it internally rather than bouncing the browser — otherwise locale
+    // home pages like /fr/ cause an infinite redirect loop.
```
boutell, line 22: OK

packages/apostrophe-astro/lib/format.js
```diff
@@ -0,0 +1,34 @@
+// Build output formatting — matches Astro's native style (via piccolore)
```
boutell, line 1: !

packages/apostrophe-astro/index.js
```diff
@@ -1,14 +1,133 @@
...
+        // Resolve static build configuration.
+        // Environment variables take precedence over integration
+        // options so that CI/deployment pipelines can override
+        // without changing code.
+        const userStatic = options.staticBuild || {};
+        const staticBuild = {
+          attachments: process.env.APOS_SKIP_ATTACHMENTS
+            ? process.env.APOS_SKIP_ATTACHMENTS !== '1'
+            : (userStatic.attachments ?? true),
+          attachmentSizes: csvEnv('APOS_ATTACHMENT_SIZES')
```
boutell, line 114: Nice

packages/apostrophe-astro/index.js
```diff
@@ -1,14 +1,133 @@
...
+  // Resolved once and reused by the `astro:build:done` hook (which
+  // cannot access the Vite virtual module).
+  let resolvedAposHost;
+  let resolvedAposPrefix;
   return {
     name: 'apostrophe-integration',
     hooks: {
-      "astro:config:setup": ({ injectRoute, updateConfig, injectScript }) => {
+      "astro:config:setup": async ({ config, injectRoute, updateConfig, injectScript }) => {
+        isStaticBuild = config.output === 'static';
```
boutell, line 89: Oh, I guess we'll put the env var check for this in the starter kits. That makes sense.

packages/apostrophe-astro/index.js
```diff
@@ -1,14 +1,133 @@
...
         if (!options.widgetsMapping || !options.templatesMapping) {
           throw new Error('Missing required options')
         }
+
+        // Resolve aposHost and aposPrefix once.  Environment variables
+        // take precedence over integration options.  The resolved
+        // values are emitted into the virtual config module so that
+        // helpers and library code never need to re-check process.env.
+        resolvedAposHost = process.env.APOS_HOST || options.aposHost;
+        const aposPrefix = process.env.APOS_PREFIX
```
boutell, line 99: APOS_PREFIX is a VERY nice nice-to-have. I see it let you do github sites.

packages/apostrophe/modules/@apostrophecms/attachment/index.js
```diff
@@ -882,6 +885,260 @@ module.exports = {
...
+        }
+        for (const field of schema) {
+          if (field.type === 'attachment') {
+            return true;
+          }
+          if (
+            (field.type === 'array' || field.type === 'object') &&
+            field.schema
+          ) {
+            if (self.schemaHasAttachmentField(field.schema)) {
```
boutell, line 903: What about areas, widgets can own attachments directly too. Gotta recurse em all

boutell: What about rich text widgets, these often allow inline images, there is custom data for that

packages/apostrophe/modules/@apostrophecms/attachment/index.js
```diff
@@ -882,6 +885,260 @@ module.exports = {
...
+
+        self.collectDocAttachmentRefIds(doc, docIds);
+
+        // Convert aposDocId values to full _id format
+        return [ ...docIds ].map(
+          aposDocId => `${aposDocId}:${locale}:${mode}`
+        );
+      },
+
+      // Given a single document (raw MongoDB data), walk its
```
boutell, line 989: Oh this seems right, but I don't see where you recursed into widgets via areas to begin with?

packages/apostrophe/modules/@apostrophecms/piece-page-type/index.js
```diff
@@ -199,6 +199,26 @@ module.exports = {
 
       dispatchAll() {
         self.dispatch('/', self.indexPage);
+        if (self.apos.url.options.static) {
```
boutell, line 202: A question. This flag will be set when the dynamic editing site *intended for* static builds later is being edited right? It's best if URLs are built the same between the two.

boutell: But we could temporarily live without that if we must do so in order to release this on time.

myovchev: The build can't be initiated if this flag is not set. So basically - yes, it's indirectly enforced. There will be some differences how internal URLs will be built, but the major links - pagination, filters, etc should be the same.

packages/apostrophe-astro/index.js
myovchev: It's currently supported on a project level, I didn't change core to apply prefix based on env.

packages/apostrophe/modules/@apostrophecms/attachment/index.js
myovchev: Rich text is already recursively checked by `doc.walkByMetaType` - it is designed to walk recursively.
About direct attachment types - I think you are right, this should be done in the same spot that resolves the rich text references. A change in [the resolve handler](https://github.com/apostrophecms/apostrophe/pull/5153/changes/BASE..1751aabcd140b3eb15216d335a514ae2006be0d9#diff-be0ba871b263d3cebc6c462a8aa70a7b428d7aaf4c29b83275e10ff24dae5db9R1011) should fix that, looking at it now.

myovchev: Great catch and easy fix!

PR #5154, opened 2025-11-12
longPolling: false option

boutell:
Good testing is in place: we're using this already at the moment, via a branch dependency, in our public demo.

For user experience, long polling is good. For lighthouse scores when logged in, it is not so good, and some customers may find themselves under similar pressure. Now they have choices.

### DIFF
--- a/modules/@apostrophecms/notification/index.js
+++ b/modules/@apostrophecms/notification/index.js
@@ -81,9 +81,9 @@ module.exports = {
 
         async function attempt() {
           if (
-              self.options.longPolling &&
+            self.options.longPolling &&
               (Date.now() - start >= self.options.longPollingTimeout)
-            ) {
+          ) {
             return {
               notifications: [],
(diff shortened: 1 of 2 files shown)


modules/@apostrophecms/notification/index.js
```diff
@@ -64,7 +80,10 @@ module.exports = {
         return await attempt();
 
         async function attempt() {
-          if (Date.now() - start >= (self.options.longPollingTimeout || 10000)) {
+          if (
+            self.options.longPolling &&
+              (Date.now() - start >= self.options.longPollingTimeout)
```
BoDonkey, line 85: No fallback needed?

boutell: Correct, no fallback needed, because this is code that preempts making a database query at all if the long polling timer runs out, which isn't relevant in vanilla polling.

PR #5155, opened 2025-11-12
PRO-6669: Fix native windows

boutell:
Windows support. Validated this with NVM for Windows (not to be confused with regular nvm) and Git Bash. I never managed to figure out how to use npm at all in Powershell (not an ApostropheCMS issue).

Starter kits will need separate PRs to use `cross-env`, but you can test with `node app` in the meanwhile.

It's important for me to know whether (1) it npm installs and works successfully for someone else, and (2) accepts image uploads. If you get an error about `sharp` not being installed let me know. I found a workaround but it might only be needed in my preferred but wacky test environment (ARM Windows, definitely not normal)

### DIFF
--- a/modules/@apostrophecms/template/index.js
+++ b/modules/@apostrophecms/template/index.js
@@ -281,7 +281,7 @@ module.exports = {
         }
         if (process.platform === 'win32') {
           // CR/LF is a waste of space in HTML and breaks unit tests
-          result = result.replaceAll('\r', '');		  
+          result = result.replaceAll('\r', '');
         }
         return result;
       },


index.js
```diff
@@ -377,7 +377,7 @@ async function apostrophe(options, telemetry, rootSpan) {
     const reallyLocalPath = self.rootDir + localPath;
 
     if (fs.existsSync(reallyLocalPath)) {
-      local = await self.root.import(reallyLocalPath);
+      local = await self.root.import(pathToFileURL(reallyLocalPath));
```
boutell, line 380: "import" takes URLs, not filenames, although file:// is the default protocol. So without this, "C:" is treated as some newly invented protocol, and you are busted.

lib/moog.js
```diff
@@ -189,7 +189,7 @@ module.exports = function(options) {
         }
         // You can have access to options within a function, if you choose to
         // provide one
-        const properties = ((typeof step[cascade]) === 'function') ? step[cascade](that, options) : step[cascade];
+        const properties = ((typeof step[cascade]) === 'function') ? await step[cascade](that, options) : step[cascade];
```
boutell, line 192: New feature: allow cascades to do async work before returning their configuration. Allows project level palette configs to use async import.

modules/@apostrophecms/template/index.js
```diff
@@ -1151,16 +1151,16 @@ module.exports = {
...
-          if (node.comment) {
+          if (node.comment != null) {
             return `\n<!-- ${self.apos.util.escapeHtml(node.comment)} -->\n`;
           }
-          if (node.raw) {
+          if (node.raw != null) {
             return node.raw;
           }
-          if (node.name) {
+          if (node.name != null) {
```
boutell, line 1167: Unrelated bug. An empty text node is valid and should not print an error on the console.

PR #5163, opened 2025-11-19
Pro 8546 template page notfound

ValJed:
## Summary

⚠️ Should be tested with [this PR](https://github.com/apostrophecms/doc-template-library/pull/91)

Passes `targetId` from the post body to publish method when autopublish..

If all your pages are draft only and your create a page template, the publish failes because it didn't find the target..
By passing the `_home` we simply add it as a home child, like for the draft.
It still feel weird that we pass twice in these computations with different logic for draft and published. When the tree exist only for draft visually.
But feels like moving one piece of the page system would break things.

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/test/pages.js
+++ b/test/pages.js
@@ -2379,4 +2379,112 @@ describe('Pages', function() {
       assert.deepEqual(actual, expected);
     });
   });
+
+  describe.only('autopublish', function() {
+    before(async () => {
+      await t.destroy(apos);
+      apos = await t.create({
+        root: module,
+        modules: {
+          '@apostrophecms/express': {
+            options: {
+              apiKeys: {
+                [apiKey]: {
+                  role: 'admin'
+                }
+              }
+            }
+          },
+          '@apostrophecms/page': {
+            options: {
+              park: [],
+              types: [
+                {
+                  name: '@apostrophecms/home-page',
+                  label: 'Home'
+                },
+                {
+                  name: 'test-page',
+                  label: 'Test Page'
+                },
+                {
+                  name: 'autopublished-page',
+                  label: 'Autopublished Pages'
+                }
+              ],
+              publicApiProjection: {
+                title: 1,
+                _url: 1
+              }
+            }
+          },
+          'test-page': {
+            extend: '@apostrophecms/page-type'
+          },
+          'autopublished-page': {
+            extend: '@apostrophecms/page-type',
+            options: {
+              autopublish: true
+            }
+          }
+        }
+      });
+      await t.createAdmin(apos);
+    });
+
+    beforeEach(async function() {
+      await apos.doc.db.deleteMany({ type: { $in: [ 'test-page', 'autopublished-page' ] } });
+    });
+
+    it('should properly autopublish a page even if its page tree target is draft only', async function() {
+      const jar = await t.loginAs(apos, 'admin');
+      await apos.http.get('/', { jar });
+      const testItems = [
+        {
+          title: 'Page 1',
+          type: 'test-page',
+          slug: '/page1'
+        },
+        {
+          title: 'Page 2',
+          type: 'test-page',
+          slug: '/page2'
+        }
+      ];
+
+      for (const page of testItems) {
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/page/index.js
```diff
@@ -3241,16 +3241,20 @@ database.`);
           : parentAposDocId;
         const peerCriteria = {
           path: self.matchDescendants(parentPath),
-          level: doc.level
+          level: doc.level,
+          type: {
+            $nin: [ '@apostrophecms/archive-page', '@apostrophecms/search' ]
+          }
```
ValJed: I don't think we should be able to retrieve archive or search pages in peers when computing targets?
This change breaks mocha tests, keeping it because I want a feedback on it.

modules/@apostrophecms/page/index.js
```diff
@@ -3241,16 +3241,20 @@ database.`);
...
-          rank: 1
-        }).project({
-          _id: 1
-        }).toArray();
+        const peers = await self.apos.doc.db
+          .find(peerCriteria)
+          .sort({ rank: 1 })
+          .project({ _id: 1 })
+          .toArray();
+
```
ValJed, line 3254: readability

modules/@apostrophecms/doc-type/index.js
```diff
@@ -308,7 +308,9 @@ module.exports = {
           if (doc.aposLocale.includes(':draft')) {
             return self.publish(req, doc, {
               ...options,
-              autopublishing: true
+              autopublishing: true,
+              targetId: req.body?._targetId,
```
boutell: This is a method, not an API, so it really shouldn't manipulate `req.body`, new options should be passed to `autopublish` by the relevant route instead.

I may have other concerns, will keep reading.

modules/@apostrophecms/page-type/index.js
```diff
@@ -416,8 +416,12 @@ module.exports = {
           mode: 'published'
         });
         if (doc.level > 0) {
-          const { lastTargetId, lastPosition } = await self.apos.page
-            .inferLastTargetIdAndPosition(doc);
+          const { lastTargetId, lastPosition } = options.targetId
+            ? {
+              lastTargetId: options.targetId,
```
boutell, line 421: What does the browser know that we can't determine in `inferLastTargetIdAndPosition`?

modules/@apostrophecms/page/index.js
boutell: We should be able to do it, because pages can be moved around in the archive. The search page is an ordinary page that happens to exist, as a parked page, in some projects, it has no special status here. I do not think this change makes sense.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 * Specify the content type when calling back to Astro with JSON to render an area. This is required starting in Astro 4.9.0 and up, otherwise the request is blocked by CSRF protection.
 * Fixes `AposBreadcrumbSwitch` tooltip prop that is supposed to be an object, not a string. Object returned from the shared method `getOperationTooltip`.
 * Uses `modalData.locale` in `AposI18nLocalize` component. Fixes watcher on `relatedDocTypes` not being properly triggered (uses data and methods for more control instead).
+* Allows to pass `targetId` and `position` in options of the `publish` method. Passed by default during autopublish if it was existing in the body of the post request.
```
boutell, line 25: But you didn't actually pass it anywhere from the front end...?

boutell: Oh right, you pass it in the template module.

So here is my question. We could do it this way... but, is there really something the frontend knows that the backend doesn't? We can't infer the targetId and position in a beefed-up way when autopublishing a page (which normally never happens, but does happen with templates)?

boutell:
Closed after discussion, review and merge of my alternate implementation. Many thanks to Val for his deep insights & diagnosis here.

PR #5166, opened 2025-11-20
PRO-8546: template page not found, various issues with the page tree and published pages

boutell:


modules/@apostrophecms/page-type/index.js
```diff
@@ -173,7 +173,7 @@ module.exports = {
           });
           if (published && (doc.level > 0)) {
             const { lastTargetId, lastPosition } = await self.apos.page
-              .inferLastTargetIdAndPosition(doc);
+              .inferLastTargetIdAndPosition(doc, { publishedTargetsOnly: true });
```
boutell, line 176: Page tree editing occurs in draft, but we replay the movement afterwards in the published locale. When doing so we need to make sure we only look at published targets.

modules/@apostrophecms/page-type/index.js
```diff
@@ -417,13 +417,11 @@ module.exports = {
         });
         if (doc.level > 0) {
           const { lastTargetId, lastPosition } = await self.apos.page
-            .inferLastTargetIdAndPosition(doc);
+            .inferLastTargetIdAndPosition(doc, { publishedTargetsOnly: true });
```
boutell, line 420: Similarly, `insertPublishedOf` should only look at published targets.

modules/@apostrophecms/page/index.js
```diff
@@ -1361,8 +1361,8 @@ database.`);
         return query;
       },
       // Insert a page. `targetId` must be an existing page id, `_archive` or
-      // `_home`, and `position` may be `before`, `inside` or `after`.
-      // Alternatively `position` may be a zero-based offset for the new child
+      // `_home`, and `position` may be `before`, `after`, `firstChild` or `lastChild`.
```
boutell, line 1364: Unrelated correction to the documentation.

modules/@apostrophecms/page/index.js
```diff
@@ -3233,24 +3233,35 @@ database.`);
...
+      async inferLastTargetIdAndPosition(doc, { publishedTargetsOnly = false } = {}) {
         const parentPath = self.getParentPath(doc);
         const parentAposDocId = parentPath.split('/').pop();
         const parentId = doc.aposLocale
           ? `${parentAposDocId}:${doc.aposLocale}`
           : parentAposDocId;
+        if (publishedTargetsOnly) {
+          const parent = await self.apos.doc.db.findOne({ _id: parentId });
+          if (!parent?.lastPublishedAt) {
+            throw self.apos.error('notfound');
```
boutell, line 3245: We have a strict policy against publishing children of unpublished parents and the UI enforces that etc., so all we need to do is sanity check here. It works out for doc template library because doc templates are children of the home page, and parked pages are always published.

modules/@apostrophecms/page/index.js
```diff
@@ -3233,24 +3233,35 @@ database.`);
...
+        const allPeers = await self.apos.doc.db.find(peerCriteria).sort({
           rank: 1
         }).project({
-          _id: 1
+          _id: 1,
+          title: 1,
+          lastPublishedAt: 1
         }).toArray();
+        const peers = publishedTargetsOnly
+          ? allPeers.filter(peer => (peer._id === doc._id) || peer.lastPublishedAt)
```
boutell, line 3262: We're working with drafts, but consider only those that have a published equivalent (and the page in question itself, that information is used to determine relative ranks).

modules/@apostrophecms/page/index.js
```diff
@@ -3267,6 +3278,9 @@ database.`);
           targetId = peers[index - 1]._id;
           position = 'after';
         }
+        if (publishedTargetsOnly) {
+          targetId = targetId.replace(':draft', ':published');
```
boutell, line 3281: Return a published target id.

test/pages.js
```diff
@@ -1664,25 +1664,27 @@ describe('Pages', function() {
 
       // Publish and assert.
       // Obviously, should not throw an error.
+      // Publish them in reverse order, then verify the published ranks
```
boutell, line 1667: I updated these tests a lot for correctness and maintainability / readability.

test/pages.js
```diff
@@ -2112,7 +2114,7 @@ describe('Pages', function() {
           slug: '/level-1-page-2',
           path: `${home.aposDocId}/${level1Page2.aposDocId}`,
           level: 1,
-          rank: 4,
+          rank: 3,
```
boutell, line 2117: I went through this and determined that these new relative ranks are correct. It's just that the new code leaves fewer gaps between ranks.

test/pages.js
```diff
@@ -2379,4 +2381,127 @@ describe('Pages', function() {
       assert.deepEqual(actual, expected);
     });
   });
+
+  // We do not support autopublished page types per se, especially
```
boutell, line 2385: Here is my version of your new test, plus a comment for posterity to explain why we have it.

test/pages.js
```diff
@@ -2379,4 +2381,127 @@ describe('Pages', function() {
...
+        await apos.http.post('/api/v1/@apostrophecms/page?aposMode=draft&aposLocale=en', {
+          jar,
+          body
+        });
+      }
+    });
+  });
 });
+
+function getIndexes(all, mode, pages) {
```
boutell, line 2495: Utility to reduce boilerplate and return an object with named properties for more readable tests.

modules/@apostrophecms/page/index.js
ValJed: Yup makes sense.

modules/@apostrophecms/page/index.js
```diff
@@ -3233,24 +3233,35 @@ database.`);
...
       },
-      async inferLastTargetIdAndPosition(doc) {
+      async inferLastTargetIdAndPosition(doc, { publishedTargetsOnly = false } = {}) {
         const parentPath = self.getParentPath(doc);
         const parentAposDocId = parentPath.split('/').pop();
         const parentId = doc.aposLocale
           ? `${parentAposDocId}:${doc.aposLocale}`
           : parentAposDocId;
+        if (publishedTargetsOnly) {
+          const parent = await self.apos.doc.db.findOne({ _id: parentId });
```
ValJed, line 3243: Looks like `parentId` can also be an `aposDocId` if doc has no `aposLocale`, it might happen only on a non localized site? I still don't think this request would be valid.
Also, are we sure it will request the draft version of the parent? It's ok if we're sure doc is always draft.

modules/@apostrophecms/page/index.js
```diff
@@ -3233,24 +3233,35 @@ database.`);
...
           rank: 1
         }).project({
-          _id: 1
+          _id: 1,
+          title: 1,
+          lastPublishedAt: 1
         }).toArray();
+        const peers = publishedTargetsOnly
+          ? allPeers.filter(peer => (peer._id === doc._id) || peer.lastPublishedAt)
+          : allPeers;
```
ValJed, line 3263: Don't think you need more data in the projection, same for the filter  filter here:
```js
        const peerCriteria = {
          path: self.matchDescendants(parentPath),
          level: doc.level,
          ...doc.aposLocale && { aposLocale: doc.aposLocale },
          ...publishedTargetsOnly && { lastPublishedAt: { $ne: null } }
        };
        const peers = await self.apos.doc.db
          .find(peerCriteria)
          .sort({ rank: 1 })
          .project({ _id: 1 })
          .toArray();
```

modules/@apostrophecms/page/index.js
ValJed: Clearer, the `inferIdLocaleAndMode` also does that, it also mutates req which is hard to debug.
I prefer to do it here, not sure we can get rig of `inferIdLocaleAndMode` in `getTarget`.`

modules/@apostrophecms/page/index.js
boutell: aposDocId and _id are equal for a non localized page so this would not fail in that situation.

PR #5169, opened 2025-11-23
PRO-8743: windows fix 2b: back off on the file:/// URLs for vite

boutell:
those URLs actually breaks both windows and linux under some circumstances. Probably a vite bug, but vite will happily eat file paths starting with C:\, so.

### DIFF
--- a/modules/@apostrophecms/asset/lib/build/external-module-api.js
+++ b/modules/@apostrophecms/asset/lib/build/external-module-api.js
@@ -653,8 +653,8 @@ function invoke() {
             }
           }
         }
-        // You would think we should run pathToFileURL over realPath, but that actually breaks
-        // both Windows and Linux with Vite. -Tom
+        // You would think we should run pathToFileURL over realPath,
+        // but that actually breaks both Windows and Linux with Vite. -Tom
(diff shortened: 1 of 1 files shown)


modules/@apostrophecms/asset/lib/build/external-module-api.js
```diff
@@ -654,17 +653,17 @@ function invoke() {
             }
           }
         }
-        // We know realPath is a file path at this point
-        const importUrl = JSON.stringify(pathToFileURL(realPath));
+        // You would think we should run pathToFileURL over realPath, but that actually breaks both Windows and Linux with Vite. -Tom
+        const importPath = JSON.stringify(realPath);
```
myovchev, line 658: I'm confused. We should do relative paths and never absolute. I don't recall using explicitly `pathToFileURL()` during the implementation, this should be the refactored code and ONLY used in the default Webpack bundler. I don't have any other explanation (absolute paths can break our Vite integration and some features).

BoDonkey:
On my win11 box this is not spinning up correctly. It crashes trying to import the `import-all.scss` from the build folder.

PR #5179, opened 2025-11-28
PRO-8756: monorepo workflows

myovchev:
## Summary

Brings a tooling for detecting the change impact and provide a matrix for performant testing in GitHub Actions (only test what makes sense). 

### Detection

- Only packages that have `test` script in their `package.json` will participate
- Automatically detecting what packages are affected by the current changes
- If `[force ci]` (case insensitive) is found in a commit message, all packages will be tested
- A nightly run will test all packages, everyday. 
- A manual dispatch is possible. If _"Force tests for every package"_ is checked, all packages will be tested, otherwise only the auto detected packages will run. 

### Optimization & Configuration 

The following configuration in any `package/*/`package.json`
```json
  "apostropheTestConfig": {
    "requiresMongo": true,
    "requiresRedis": false
  }
```
will determine what services are needed during the CI test run for that package. The above showcases the default behavior if no configuration is found. 

`NODE_VERSIONS_JSON: "[20,22,24]"` and `MONGODB_VERSIONS_JSON: '["7","8"]'` environment variables in `monorepo.yml` control the version matrix of the respective runtime. `REDIS_VERSION: "7"` defines the Redis version (when in use). 

The services are pulled from the docker repository and cached for a month. No additional network requests will be made until cache miss.

### Technical details 

- Retire the legacy top-level “Tests” workflow in favor of two new monorepo-aware pipelines: a PR/branch workflow that only runs packages touched by a change and a nightly schedule that fans out across every package.
- Add `.github/workflows/scripts/detect-impacted-packages.mjs`, which diff-checks against the default branch, expands dependent packages, and feeds GitHub’s matrix jobs; mocha test ensures proper behavior. [Example output](https://github.com/apostrophecms/apostrophe/actions/runs/19895002627/job/57023016939#step:4:15).
- Add `.github/workflows/scripts/expand-runtime-matrix.mjs` to merge the detected packages matrix with the nodejs and mongodb version matrix. [Example output](https://github.com/apostrophecms/apostrophe/actions/runs/19895002627/job/57023016939#step:5:18)
- Allow commit messages containing `[force ci]` to force testing of all monorepo packages that have `test` package JSON script. 
- Expose a `run_all` workflow-dispatch flag plus nightly FORCE_ALL runs so we can still exercise the full workspace on demand.

1. Install dependencies (`pnpm install`) so the repo scripts/tests are available.
2. Run `npx mocha .github/workflows/scripts/test-impacted-packages.mjs` and confirm all detector tests pass.
3. (Optional sanity check) Modify a file inside one package, run `DEFAULT_BRANCH=main BASE_SHA=$(git merge-base HEAD origin/main) node .github/workflows/scripts/detect-impacted-packages.mjs`, and verify the JSON output only lists that package and its dependents.

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [x] Other

**Other information:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,17 @@
   "engines": {
     "node": ">=18"
   },
-
   "pnpm": {
-    "onlyBuiltDependencies": ["sharp", "vue-demi", "@parcel/watcher", "esbuild", "unrs-resolver"]
+    "onlyBuiltDependencies": [
+      "sharp",
+      "vue-demi",
+      "@parcel/watcher",
+      "esbuild",
+      "unrs-resolver"
+    ],
+    "ignoredBuiltDependencies": [
+      "protobufjs"
+    ]
   },
   "packageManager": "pnpm@10.18.0"
 }


.github/workflows/scripts/detect-impacted-packages.mjs
```diff
@@ -0,0 +1,291 @@
...
+      encoding: 'utf8',
+      stdio: ['ignore', 'pipe', 'ignore']
+    }).trim();
+    
+    if (!output) {
+      return false;
+    }
+    
+    const commitMessages = output.split('\n');
+    return commitMessages.some((msg) => /\[force ci\]/i.test(msg));
```
ValJed, line 183: So any commit on your PR containing this force ci will make them all run all the time (for this PR).
Something in the title of the PR would have been easier to track but not sure we can do that. I think it's ok but I don't see use case especially with the daily run.

.github/workflows/scripts/detect-impacted-packages.mjs
```diff
@@ -0,0 +1,291 @@
...
+
+function resolveBaseSha() {
+  try {
+    const mergeBase = execSync(`git merge-base HEAD origin/${defaultBranch}`, {
+      encoding: 'utf8',
+      stdio: ['ignore', 'pipe', 'ignore']
+    }).trim();
+    return mergeBase || '';
+  } catch (error) {
+    return '';
```
ValJed, line 163: [detail] Could be null for clarity since we check if it exist later.

.github/workflows/monorepo.yml
```diff
@@ -0,0 +1,277 @@
...
+
+      - name: Use Node.js 22
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.18.0
```
ValJed, line 95: the pnpm version could be an env too maybe?

.github/workflows/monorepo.yml
```diff
@@ -0,0 +1,277 @@
...
+
+      - name: Generate pnpm lockfile
+        run: pnpm install --lockfile-only --ignore-scripts
+
+      - name: Upload pnpm lockfile
+        uses: actions/upload-artifact@v4
+        with:
+          name: pnpm-lock
+          path: pnpm-lock.yaml
+          retention-days: 30
```
ValJed, line 111: Why 30 days? Is is shared between CI runs?

.github/workflows/scripts/detect-impacted-packages.mjs
```diff
@@ -0,0 +1,291 @@
...
+const repoRoot = path.resolve(__dirname, '..', '..', '..');
+const packagesRoot = path.join(repoRoot, 'packages');
+
+const defaultBranch = process.env.DEFAULT_BRANCH || 'main';
+const forceAll = process.env.FORCE_ALL === 'true';
+const baseShaEnv = (process.env.BASE_SHA || '').trim();
+
+const GLOBAL_IMPACT_PATHS = new Set([
+  'package.json',
+  // Keep it, although it's not under version control - be explicit.
```
ValJed, line 21: Yes but I see you use cache for the pnpm-lock file?

packages/boring/package.json
```diff
@@ -27,5 +27,8 @@
   },
   "devDependencies": {
     "mocha": "^6.2.0"
+  },
+  "apostropheTestConfig": {
+    "requiresMongo": false
   }
```
ValJed, line 33: I suppose that everytime there are mocha tests we need mongo? (Not  sure we might have unit tests).
If yes and if we follow a convention, any package without a script `mocha` could skip mongo. Your solution is clearer but I'm not sure people will think about updating this.

ValJed (approved):
Few questions, looks good!

packages/boring/package.json
myovchev: I think declaring explicitly what a package needs is a better way to go. For mongo it makes sense to explicitly opt-out, as most of the packages do require it.

.github/workflows/scripts/detect-impacted-packages.mjs
myovchev: It's generated on every run in the setup phase, reused in all other phases, including the test matrix jobs

.github/workflows/monorepo.yml
myovchev: It doesn't matter, it's overridden on every run and used down the pipeline.

.github/workflows/monorepo.yml
myovchev: That would be a great future improvement, yes. But we can leave it for when we bump pnpm

.github/workflows/scripts/detect-impacted-packages.mjs
myovchev: It's just a quick way to solve  "not forget to run all tests for my change for some strange reason" . This follows the convention for github owned `(skip ci)`

PR #5204, opened 2025-12-09
fix stale data after changing schema

ETLaurent:
Follow up of #5203, #5176

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
+++ b/packages/apostrophe/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
@@ -575,9 +575,9 @@ export default {
 
               if (
                 [ 'select', 'radio', 'checkboxes' ].includes(field.type) &&
-                  !object[field.name] &&
-                  !field.def &&
-                  !field.required
+                  !field.required &&
+                  field.def === undefined &&
+                  (object[field.name] === undefined || object[field.name] === null)
               ) {
                 continue;
               }


packages/apostrophe/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -552,14 +552,17 @@ export default {
           this.original = klona(docData);
           this.docFields.data = docData;
 
-          resetToSchema(
-            // Never allow changing the type of a page here
-            this.schema.filter(f => f.name !== 'type'),
-            this.docFields.data
-          );
```
ETLaurent, line 559: we actually want the pages to detect a change if a page type was removed from the Page manager list

packages/apostrophe/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -570,6 +573,15 @@ export default {
...
               }
 
+              if (
+                [ 'select', 'radio', 'checkboxes' ].includes(field.type) &&
+                  !object[field.name] &&
+                  !field.def &&
+                  !field.required
+              ) {
+                continue;
+              }
```
ETLaurent, line 583: if they don't have a default, nor they're required, leave 'em fields alone!

packages/apostrophe/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -570,6 +573,15 @@ export default {
                 resetToSchema(field.schema, object[field.name] || {});
               }
 
+              if (
+                [ 'select', 'radio', 'checkboxes' ].includes(field.type) &&
+                  !object[field.name] &&
+                  !field.def &&
```
myovchev: `field.def` and `object[field.name]` both can be `false` or `0`. The check should aim at  `null` and `undefined`. I'm not even sure about `null` in this specific test case. In theory I can set `def: null` in my schema.

boutell: For these field types `null` would never be distinct from no value.

I think you can use `!= null` which matches anything that is not null or undefined.

ETLaurent: that was bad, sorry

packages/apostrophe/modules/@apostrophecms/doc-type/ui/apos/components/AposDocEditor.vue
```diff
@@ -575,9 +575,9 @@ export default {
 
               if (
                 [ 'select', 'radio', 'checkboxes' ].includes(field.type) &&
-                  !object[field.name] &&
-                  !field.def &&
-                  !field.required
+                  !field.required &&
+                  field.def === undefined &&
+                  (object[field.name] === undefined || object[field.name] === null)
```
boutell, line 580: I like `== null` but this is fine

boutell (approved):

haroun (approved):

PR #5212, opened 2025-12-10
applying widget level styles in template

ETLaurent:
- ~fix box-type field to make it render valid css property~ --> separate PR: https://github.com/apostrophecms/apostrophe/pull/5227
  - ~we can now set a `border-%key%-width` property, for instance~ 
- render styles, using `self.apos.styles` methods

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/widget-type/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/widget-type/index.js
@@ -418,7 +418,9 @@ module.exports = {
           contextOptions: _with
         });
 
-        if (self.styles && self.options.stylesWrapper !== false) {
+        const hasStyles = Object.keys(self.styles || {}).length > 0;
+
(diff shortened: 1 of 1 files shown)


packages/apostrophe/modules/@apostrophecms/styles/ui/universal/customRules.mjs
```diff
@@ -37,12 +37,19 @@ export default {
...
+      if (property.includes('%key%')) {
+        parts.push(`${property.replace('%key%', side)}: ${val}${unit}${important ? ' !important' : ''}`);
+      } else {
         parts.push(`${property}-${side}: ${val}${unit}${important ? ' !important' : ''}`);
       }
-      rule = parts.join(';');
     }
 
+    rule = parts.join(';');
+
```
ETLaurent: that was set in the loop, no need

ValJed (approved):

PR #5217, opened 2025-12-12
PRO-8803: sourcemaps fixed all the ways

boutell:
* We now support `productionSourceMaps: true` in webpack (already existed in Vite, in buggy form)
* `productionSourceMaps: true` works properly in Vite
* We can use `devtool: 'source-map'` consistently now, removing complexity and the use of an inferior setting that was always a workaround for the deeper bug around combining multiple sourcemaps that I have fixed properly here
* As a result, webpack sourcemaps work a whole lot better in dev (they were already fine in Vite because Vite doesn't do bundling in dev, but some customers are happy with Webpack or need it for edge cases)
* New `productionSourceMapsDir` option for edge cases: GO wants to be able to generate production source maps but they will then push them to Datadog, not live to the public

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/asset/lib/build/internals.js
+++ b/packages/apostrophe/modules/@apostrophecms/asset/lib/build/internals.js
@@ -308,10 +308,9 @@ module.exports = (self) => {
             content = concat.content.toString('utf8') + `\n//# sourceMappingURL=${fileName}.map\n`;
             sourceMap = concat.sourceMap;
           } else {
-            content = files.map(f => fs.readFileSync(f, 'utf-8')
+            content = files.map(f => fs.readFileSync(f, 'utf-8'))
               .join('\n')
-              .trim()
-            );
+              .trim();
           }
           if (!content.trim().length) {
             delete bundles[target];


packages/apostrophe/modules/@apostrophecms/asset/lib/build/internals.js
```diff
@@ -207,6 +208,8 @@ module.exports = (self) => {
     // external build module build method (see `self.build()` and
     // `configureBuildModule()`).
     async computeBuildScenes(metadata, { write = true } = {}) {
+
+      const needSourceMap = self.options.productionSourceMaps;
```
boutell, line 212: Decision is simple because this code doesn't run for dev with vite etc.

packages/apostrophe/modules/@apostrophecms/asset/lib/build/internals.js
```diff
@@ -266,26 +269,66 @@ module.exports = (self) => {
...
+                  fs.readFileSync(map, 'utf8')
+                );
+              }
+            }
+            content = concat.content.toString('utf8') + `\n//# sourceMappingURL=${fileName}.map\n`;
+            sourceMap = concat.sourceMap;
+          } else {
+            content = files.map(f => fs.readFileSync(f, 'utf-8')
+              .join('\n')
+              .trim()
```
boutell: Keep it extremely lean, no more comments between files in this mode because people are beating us up over lighthouse scores

packages/apostrophe/modules/@apostrophecms/asset/lib/build/task.js
```diff
@@ -544,22 +545,52 @@ module.exports = (self) => ({
...
+        .map(([ name ]) => `${bundleDir}/${name}-build.${fileExt}`)
+        .filter(file => {
+          return (!checkForFile) || fs.existsSync(file);
+        });
+
+      const bundlePath = `${bundleDir}/${filePath}`;
+      const sourceMapDir = self.options.productionSourceMapsDir || bundleDir;
+      const sourceMapPath = `${sourceMapDir}/${filePath}.map`;
+
+      const needSourcemap = (process.env.NODE_ENV !== 'production') || self.options.productionSourceMaps;
```
boutell, line 559: For webpack this code is also relevant in dev, so this test has two parts

packages/apostrophe/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -25,14 +25,7 @@ module.exports = ({
   const mode = process.env.NODE_ENV || 'development';
   const pnpmModulePath = apos.isPnpm ? [ path.join(apos.selfDir, '../') ] : [];
   const dev = (mode === 'development');
-  // The default behavior of the security-headers module
```
boutell, line 28: I was able to get rid of all this blinking shit because the real underlying problem was always that we weren't concatenating sourcemaps correctly. In my tests devtool: 'source-map' works fine now. And it doesn't use eval so it should be CSP OK.

PR #5218, opened 2025-12-12
Add classes style prop support

myovchev:
## Summary

Support for the new class attribute of style fields


packages/apostrophe/modules/@apostrophecms/styles/ui/universal/render.mjs
```diff
@@ -470,3 +478,40 @@ function stringifyRules(styles, inline = false) {
...
+    for (const v of value) {
+      if (typeof v === 'string') {
+        storage.classes.add(v);
+      }
+    }
+    return value.length > 0;
+  } else if (typeof value === 'string') {
+    storage.classes.add(value);
+    return true;
+  }
```
ETLaurent, line 514:
```suggestion
  }
  if (typeof value === 'string') {
    storage.classes.add(value);
    return true;
  }
```

PR #5223, opened 2025-12-14
PRO-8613: Add widget render helpers

myovchev:
## Summary

Add helpers (including Nunjucks) to render style tag and widget wrapper attributes. 

Internally the widget output now should use:
```js
const styles = self.apos.styles.prepareWidgetStyles(widget);
const syleTag = self.apos.styles.getWidgetElements(styles);
const wrapperAttrs = self.apos.styles.getWidgetAttributes(styles);
```

This tasks DOES NOT implement the actual output change, nor the option for option-out of automatic rendering.

The helpers are sharing the same code. Example widget template (to be used when the automatic output is disabled)
```nunjucks
{%- set styles = apos.styles.render(data.widget) -%}
{{ apos.styles.elements(styles) }}
<article {{ apos.styles.attributes(styles, { class: 'fancy-article' }) }}>
  ...content ...
</article>
```

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/widget-type/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/widget-type/index.js
@@ -314,7 +314,7 @@ module.exports = {
           ...self.fields
         };
       },
-      // Return rednered styles object for a given widget instance.
+      // Return rendered styles object for a given widget instance.
       // This shouldn't be used directly, instead use the
       // `apos.styles.prepareWidgetStyles(widgetData)` helper method or
       // the corresponding Nunjucks helper `apos.styles.render(widget)`.


packages/apostrophe/modules/@apostrophecms/styles/lib/methods.js
```diff
@@ -152,6 +152,114 @@ module.exports = (self, options) => {
...
+      const classSet = new Set(classes || []);
+      if (additionalClasses) {
+        const extraClasses = Array.isArray(additionalClasses)
+          ? additionalClasses
+          : additionalClasses.split(/\s+/).filter(Boolean);
+        extraClasses.forEach(cls => classSet.add(cls));
+      }
+      if (classSet.size) {
+        attrs.push(`class="${[ ...classSet ].join(' ')}"`);
+      }
```
ETLaurent, line 238: 👍

packages/apostrophe/modules/@apostrophecms/widget-type/index.js
```diff
@@ -314,8 +314,20 @@ module.exports = {
           ...self.fields
         };
       },
-      getStylesheet(doc, styleId) {
-        return self.apos.styles.getWidgetStylesheet(self.schema, doc, {
+      // Return rednered styles object for a given widget instance.
```
ETLaurent:
```suggestion
      // Return rendered styles object for a given widget instance.
```

packages/apostrophe/test/modules/test-style-widget/views/widget.html
```diff
@@ -0,0 +1,5 @@
+{%- set styles = apos.styles.render(data.widget) -%}
+{{ apos.styles.elements(styles) }}
+<article {{ apos.styles.attributes(styles, { class: 'fancy-article' }) }}>
```
ETLaurent, line 3: Can we test `style` attributes and other extra ones?

myovchev: This is already done in the unit tests. If you look at that section of the tests, you'll see every possible variation is tested. 
This is more "integration" - validates that the helpers functions are properly wired.

PR #5225, opened 2025-12-15
Pro 8601 rich text styles

ValJed:
[PRO-8604](https://linear.app/apostrophecms/issue/PRO-8604/support-for-all-other-widget-types)
[PRO-8601](https://linear.app/apostrophecms/issue/PRO-8601/support-for-style-editing-of-rich-text-widgets)

## Summary

- Adds styles widget operations when widget type contains at least one style field.
- Adds Styles button to all widgets which opens the widget editor on the styles tab directly.
- Adds new `widgetStylesEditor` in components to manage the editor (default is `AposWidgetEditor`) but can be changed like we do for the rich text widget which has a special editor component.

When styles are configured  on a widget  you should be able to see a new styles button (palette icon) which opens the widget editor on the styles tab (normal widget and rich text).

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/rich-text-widget/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/rich-text-widget/index.js
@@ -11,14 +11,6 @@ const util = require('util');
 module.exports = {
   extend: '@apostrophecms/widget-type',
   cascades: [ 'linkFields' ],
-  styles(self, options) {
-    return {
-      add: {
-        myBorder: 'border',
-        alignment: 'alignment'
-      }
-    };
-  },
   linkFields(self, options) {
     const linkWithType = (Array.isArray(options.linkWithType)
       ? options.linkWithType


packages/apostrophe/modules/@apostrophecms/widget-type/index.js
```diff
@@ -320,14 +326,28 @@ module.exports = {
...
+                ...operation
+              }
+            ];
+          }, []);
+      },
+      disableWidgetOperation(opName, properties) {
+        if (opName === 'styles' && !Object.keys(self.styles).length) {
+          return true;
+        }
+        return false;
```
ValJed, line 350: We disable styles operation during composeWidgetOperations because styles aren't computed yet when defining widget operations object.

packages/apostrophe/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
```diff
@@ -71,6 +71,7 @@
         @update="update"
         @add="add"
         @paste="paste"
+        @edit-styles="editStyles"
```
myovchev, line 74: The same change is probably needed in the `AposAreaLayoutEditor.vue`

packages/apostrophe/modules/@apostrophecms/widget-type/index.js
myovchev: I think this might be generally useful information, precomputed during the schema composing, that I think comes first. It might be worth to save a flag during schema composition, that is later used by all interested parties, including widget operations and getBrowserData.

myovchev (changes requested):
The underlying widgets of Layout area editor need the `edit-styles` even as well.

packages/apostrophe/modules/@apostrophecms/area/ui/apos/components/AposAreaEditor.vue
ValJed: Yes there is another ticket for that: https://linear.app/apostrophecms/issue/PRO-8602/support-for-style-editing-of-layout-widgets (next task)

packages/apostrophe/modules/@apostrophecms/widget-type/index.js
ValJed: I'm not sure to understand, what might be a useful information?
I thought that, since we have the information backend, it makes sense to filter widget operations here (when doable), we do the same in other modules as well.
And yes `composeWidgetOperations` is run right after `composeSchema` and `addStylesFields`.

PR #5227, opened 2025-12-15
fix box field type to generate valid css properties

ETLaurent:
- **fix properties keys order to make valid css**
- **adapt tests**

## **Note for reviewers:** 

@stuartromanek I might have forgotten places where to adapt this change, or other cases where the changes in this PR wouldn't work.

## **Issue**

```js
            width: {
              label: 'apostrophe:styleWidth',
              type: 'box',
              def: {
                top: 1,
                right: 1,
                bottom: 1,
                left: 1
              },
              if: {
                active: true
              },
              unit: 'px',
              property: 'border-width'
            },
```

This was an issue because `top`, `right` etc. were added at the end of the property, which resulted in: `border-width-top`, `border-width-right`..., which do not exist.

The correct properties are `border-top-width`, `border-right-width` etc.

## **Proposed solution**

Be able to define where the key (`top`, `right` etc) is placed in the property:

```js
              property: 'border-%key%-width'
```

The `%key%` value template will then be replaced with the actual key.

If the property is written as usual, like:

```js
              property: 'margin'
```

Then the keys will be appended as before: `margin-top`, `margin-right`...


packages/apostrophe/modules/@apostrophecms/styles/ui/apos/universal/customRules.mjs
```diff
@@ -37,12 +41,21 @@ export default {
...
+        ? property
+          .replaceAll('-%key%', `-${side}`)
+          .replaceAll('%key%-', `${side}-`)
+        : `property-${side}`;
+
+      parts.push(`${normalizedProperty}: ${val}${unit}${important ? ' !important' : ''}`);
     }
 
+    rule = parts.join(';');
+
```
ETLaurent, line 58: ℹ️ `rule = parts.join(';');` moved out of the for loop

ETLaurent:
| before | after |
|--------|--------|
| image | image |

stuartromanek (approved):

PR #5229, opened 2025-12-15
context utility admin bar items support a label

stuartromanek:
## Summary

- context menu items can opt to display their label
- they can also omit an icon


stuartromanek:
needs documentation, will submit a PR after review

PR #5230, opened 2025-12-16
remove palette migrations

ETLaurent:
Should not be an automatic migration.
A task has been added to the palette module: https://github.com/apostrophecms/palette/pull/132

Also, remove the palette-related migrations, we should not need them.

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/styles/lib/handlers.js
+++ b/packages/apostrophe/modules/@apostrophecms/styles/lib/handlers.js
@@ -10,10 +10,9 @@ module.exports = self => {
             `
                        🎨
 ⚠️ @apostrophecms-pro/palette has been deprecated.
+Please install the latest version of @apostrophecms-pro/palette and run the following migration task:
 
-Please install the latest version of @apostrophecms-pro/palette and run the following migration task to move your existing palette configuration to
-@apostrophecms/styles: \`node app @apostrophecms-pro/palette:migrate-to-styles\`.
-Then remove @apostrophecms-pro/palette from your project to avoid conflicts.
+node app @apostrophecms-pro/palette:migrate-to-styles
                        🎨
 
 `


packages/apostrophe/modules/@apostrophecms/styles/lib/handlers.js
```diff
@@ -2,10 +2,22 @@ const { createId } = require('@paralleldrive/cuid2');
...
+                       🎨
+⚠️ @apostrophecms-pro/palette has been deprecated.
+
+Please install the latest version of @apostrophecms-pro/palette and run the following migration task to move your existing palette configuration to
+@apostrophecms/styles: \`node app @apostrophecms-pro/palette:migrate-to-styles\`.
+Then remove @apostrophecms-pro/palette from your project to avoid conflicts.
+                       🎨
+
+`
+          );
```
ETLaurent, line 19: Image

ETLaurent: this warning must catch one's eye

packages/apostrophe/modules/@apostrophecms/styles/lib/handlers.js
```diff
@@ -2,10 +2,22 @@ const { createId } = require('@paralleldrive/cuid2');
...
 module.exports = self => {
   return {
-    '@apostrophecms/doc:afterReplicate': {
-      async migrateLegacyGlobalPaletteFields() {
-        if (self.shouldMigrateLegacyGlobalPaletteFields) {
-          await self.migrateLegacyGlobalPaletteFields();
+    'apostrophe:modulesRegistered': {
+      warnDeprecatedPalette() {
+        const paletteModule = self.apos.modules['@apostrophecms-pro/palette'];
+        if (!paletteModule?.tasks?.['migrate-to-styles']) {
```
ETLaurent: would it be better to check the version of palette rather than checking if it contains the task? WDYT?

boutell: Feature checks are generally preferable to version checks.

packages/apostrophe/modules/@apostrophecms/styles/lib/handlers.js
```diff
@@ -2,10 +2,22 @@ const { createId } = require('@paralleldrive/cuid2');
...
+        const paletteModule = self.apos.modules['@apostrophecms-pro/palette'];
+        if (paletteModule && !paletteModule?.tasks?.['migrate-to-styles']) {
+          self.apos.util.warn(
+            `
+                       🎨
+⚠️ @apostrophecms-pro/palette has been deprecated.
+
+Please install the latest version of @apostrophecms-pro/palette and run the following migration task to move your existing palette configuration to
+@apostrophecms/styles: \`node app @apostrophecms-pro/palette:migrate-to-styles\`.
+Then remove @apostrophecms-pro/palette from your project to avoid conflicts.
```
myovchev: This is not the full story. You should copy the schema configuration to the core module. It might be a good idea to just put a URL to the docs there which @BoDonkey can probably provide?

myovchev: Also, I wander, why this should be done here? Isn't it much better to let palette output this information, if the core has `@apostrophecms/styles` available?

boutell: What if they haven't upgraded palette at the same time? In general we tell people they have to do that, but in this case, we can easily help them if they didn't.

ETLaurent: @myovchev I changed the message to keep it simple in this file, just saying "update palette and run the migration".

The rest of the steps are explained in the said migration: https://github.com/apostrophecms/palette/pull/132/changes

ETLaurent: > What if they haven't upgraded palette at the same time?

Exactly.

myovchev (approved):

PR #5234, opened 2025-12-17
provide widget styles for external front

ETLaurent:
- Use `annotateWidgetForExternalFront` to pass widget styles options to Astro.
- Add a `asObject` option to get styles not as a string but a plain object, for Astro

### DIFF
--- a/packages/apostrophe/test/areas.js
+++ b/packages/apostrophe/test/areas.js
@@ -360,35 +360,25 @@ describe('Areas', function() {
   let secondRendered;
 
   it('renders an area passed to the `renderArea` method', async function () {
-    const _generateId = apos.util.generateId;
     const req = apos.task.getReq();
-
-    apos.util.generateId = () => 'test-id';
     firstRendered = await apos.area.renderArea(req, rteArea, areaDocs[0]);
-    apos.util.generateId = _generateId;
-
-    const expected = `
-<div
+    assert(firstRendered);
+    assert.equal(firstRendered.trim(),
+`<div 
   class="apos-area"
-  ><div id="test-id" data-apos-widget-style-wrapper-for="ckjyvagy9000p2a67fbf8nwd3" data-apos-widget-style-classes="">
+  >
 <div data-rich-text>
   <p>Perhaps its fate that today is the 4th of July, and you will once again be fighting for our freedom, not from tyranny, oppression, or persecution -- but from annihilation.</p><p>We're fighting for our right to live, to exist.</p>
 </div>
-</div></div>
-`;
-
-    assert(firstRendered);
-    assert.equal(firstRendered.replaceAll(/\s/g, ''), expected.replaceAll(/\s/g, ''));
+</div>
+`.trim());
   });
 
   it('returns rendered HTML from the `renderArea` method for a mixed widget area', async function() {
-    const _generateId = apos.util.generateId;
     const req = apos.task.getReq();
     apos.area.prepForRender(mixedArea, areaDocs[1], 'main');
 
-    apos.util.generateId = () => 'test-id';
     secondRendered = await apos.area.renderArea(req, mixedArea, areaDocs[1]);
-    apos.util.generateId = _generateId;
 
     assert(secondRendered.includes(`<div data-rich-text>
   <p>Good morning.`));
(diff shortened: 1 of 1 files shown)


packages/apostrophe/modules/@apostrophecms/styles/lib/methods.js
```diff
@@ -206,7 +206,9 @@ module.exports = (self, options) => {
     // Optional second argument `additionalAttrs` is an object with
     // additional attributes to merge. `class` and `style` attributes
     // are merged with the styles values, keeping classes unique.
-    getWidgetAttributes(styles, additionalAttrs = {}) {
+    // Optional third argument `options`:
+    // - asObject: boolean - if true, returns attributes as an object
+    getWidgetAttributes(styles, additionalAttrs = {}, options = {}) {
```
myovchev, line 211: Yup, it will be (I think) useful for Val's PR as well. Please proxy the `options` from the corresponding nunjucks helper so it can be used in templates.

packages/apostrophe/modules/@apostrophecms/widget-type/index.js
```diff
@@ -712,8 +710,14 @@ module.exports = {
...
 
-      annotateWidgetForExternalFront() {
-        return {};
+      annotateWidgetForExternalFront(widget) {
+        const styles = self.apos.styles.prepareWidgetStyles(widget);
+
+        return {
+          aposStylesWrapper: self.options.stylesWrapper,
+          aposStylesElements: self.apos.styles.getWidgetElements(styles),
+          aposStylesAttributes: self.apos.styles.getWidgetAttributes(styles, {}, true)
```
myovchev: The last argument should be `{ asObject: true }` instead of boolean `true`.

packages/apostrophe/test/areas.js
```diff
@@ -360,39 +360,52 @@ describe('Areas', function() {
...
 
   it('renders an area passed to the `renderArea` method', async function () {
+    const _generateId = apos.util.generateId;
     const req = apos.task.getReq();
+
+    apos.util.generateId = () => 'test-id';
     firstRendered = await apos.area.renderArea(req, rteArea, areaDocs[0]);
-    assert(firstRendered);
-    assert.equal(firstRendered.trim(),
-`<div
```
ETLaurent, line 367: this contains a trailing space...

imagine the struggle when saving the file while having prettier removing it

packages/apostrophe/modules/@apostrophecms/widget-type/index.js
```diff
@@ -712,8 +710,16 @@ module.exports = {
         });
       },
 
-      annotateWidgetForExternalFront() {
-        return {};
+      annotateWidgetForExternalFront(widget) {
+        const styles = self.apos.styles.prepareWidgetStyles(widget);
```
myovchev, line 726: I think you should NOT parse and send anything beside the option if styles for this widgets are empty (see how getBrowserData does it), just a tiny performance improvement.

ETLaurent: well, if styles is empty, shouldn't `getWidgetElements(styles)` be empty? Same for `getWidgetAttributes`?

ETLaurent: Wait, `prepareWidgetStyles` doesn't check if the widget has a styles cascade defined?

If that's the case, you're right, and I should revert this: https://github.com/apostrophecms/apostrophe/pull/5234/changes#diff-43c79fcc1c463120e5592ba0126469ead0067fdb4444a6c8639087676cca57c2L423-L425

ETLaurent: Yeah, `getWidgetElements` and `getWidgetAttributes` are set to default values, but I added a check on whether widgets have styles configured or not nonetheless

ETLaurent: Developers will still be able to set their additional attributes, for widgets that don't have `styles` configured, unless we want to prevent them from doing that, but I don't see why we'd do that.

Image

Image

packages/apostrophe/modules/@apostrophecms/styles/lib/helpers.js
```diff
@@ -19,9 +19,9 @@ module.exports = self => {
         self.getWidgetElements(styles)
       );
     },
-    attributes(styles, additionalAttributes = {}) {
+    attributes(styles, additionalAttributes = {}, options = {}) {
```
ValJed, line 22: Would make sense to put additionalAttributes inside options no ? Since will be most of the time unset?

ETLaurent: Makes sense but I think it would be better to keep options (which can contain misc stuff) separate.

packages/apostrophe/modules/@apostrophecms/styles/lib/methods.js
ValJed: What about having additionalAttrs inside options? 😁

PR #5240, opened 2025-12-18
PRO-8836: getManagerOf should handle failure consistently

boutell:
MI requirement for release


packages/apostrophe/test/utils.js
```diff
@@ -876,6 +876,18 @@ describe('Utils', async function() {
       assert.deepEqual(results, [ 'task1', 'task2', 'task3' ]);
       assert.deepEqual(resolved, [ 'task1', 'task2', 'task3' ]);
     });
+
+    it('getManagerOf should log an error and politely return null if there is no metaType', function() {
```
boutell, line 880: These unit tests are our assurance that it doesn't break the world. After merge we will wait out the regression tests as well.

PR #5245, opened 2025-12-19
PRO-8828: Add core widget styles

myovchev:
## Summary

- add image and video widget styles
- document `stylesWrapper` widget option
- add missing tests for autowrapper opt-out

Closes PRO-8497, PRO-8501


packages/apostrophe/modules/@apostrophecms/styles/lib/presets.js
```diff
@@ -15,8 +15,12 @@ module.exports = (moduleOptions) => {
       label: 'apostrophe:styleAlignment',
       type: 'select',
       class: true,
-      def: 'apos-center',
+      def: '',
       choices: [
+        {
+          label: '',
+          value: ''
+        },
```
ETLaurent: remove `def` and the first empty choice and it'll take `null` by default?

PR #5251, opened 2025-12-24
PRO-8856: fix it so testing against Pro modules is optional (but always happens in our CI)

boutell:


CONTRIBUTING.md
```diff
@@ -98,6 +98,12 @@ were applicable. For Apostrophe core, that should be in the
 and for other modules, add it in their README files (unless the README directs
 you elsewhere).
 
+### Working with the monorepo
+
+All of our public npm modules are contained in this single monorepo. You can find the individual packages in the `packages/` subdirectory. This means you will not be using `npm link` in the way you may be used to.
+
+See MONOREPO.md for more information about how to work with the monorepo.
```
ETLaurent:
```suggestion
See [MONOREPO.md](./MONOREPO.md) for more information about how to work with the monorepo.
```

if that works

CONTRIBUTING.md
```diff
@@ -98,6 +98,12 @@ were applicable. For Apostrophe core, that should be in the
 and for other modules, add it in their README files (unless the README directs
 you elsewhere).
 
+### Working with the monorepo
```
ETLaurent, line 101: We should edit the "Contributing to Apostrophe Core" section below regarding the changelog

packages/import-export/test/index.js
```diff
@@ -1600,505 +1600,515 @@ describe('@apostrophecms/import-export', function() {
...
-          '@apostrophecms-pro/automatic-translation-deepl': {
-            options: {
-              apiSecret: 'test'
             },
-            extendMethods: (self) => ({
-              async requestTranslation(_super, req, text, source, target) {
-                if (text.some((t) => t.toLowerCase().includes('test error'))) {
-                  throw new Error('Translate test error');
+            ...(process.env.TEST_WITH_PRO
+              ? {
```
ETLaurent: already wrapped in `if (process.env.TEST_WITH_PRO)` at line 1603

ETLaurent: you probably want to keep this line and remove line 1603

MONOREPO.md
```diff
@@ -0,0 +1,182 @@
...
+
+Otherwise, pnpm installs dependencies of each dependencies and allow multiple modules to have the same dependency with a different version.
+
+We cannot do that because apostrophe expects to find everything at the root of  `node_modules`.
+
+That’s why we use this option. It still installs very fast because it uses symlinks.
+
+(In local development, pnpm has a global store where it stores dependencies and symlinks them in projects, that’s why it’s very fast.)
+
+- `pnpm-workspace.yaml` file. This file defines the structure of the monorepo to pnpm. It’s very simple for us, because we only have packages, bu we could have `apps` too for example with starter kits.
```
BoDonkey:
```suggestion
- `pnpm-workspace.yaml` file. This file defines the structure of the monorepo to pnpm. It’s very simple for us, because we only have packages, but we could have `apps` too, for example, with starter kits.
```

MONOREPO.md
```diff
@@ -0,0 +1,182 @@
...
+```jsx
+  "devDependencies": {
+    "eslint": "^9.39.1",
+    "eslint-config-apostrophe": "workspace:^"
+  }
+```
+
+`workspace:` allows us to install a package of the monorepo as a dependency of another package of the monorepo. It’s automatically resolved to the right versionwhen running `pnpm publish`.
+
+There is just onen issue with this, in `testbed` for example we install dependencies this way, directly from the monorepo.
```
BoDonkey:
```suggestion
There is just one issue with this, in `testbed`, for example, we install dependencies this way, directly from the monorepo.
```

PR #5254, opened 2025-12-26
Ai provider update

BoDonkey:
## Summary

This PR refactors the AI helper module to use a provider module structure, rather than being hardcoded for OpenAI. It also creates provider files for OpenAI, Anthropic, and Gemini. Finally, it adds a "skills" file that devs can use to speed their creation of custom provider files.
I will still need help before acceptance with the correct steps for the CHANGELOG in the new mono-repo

*For example:*
I can provide any API keys that are needed for testing purposes.
1. Install the base module and any desired providers.
2. Test text generation in rich text - all three providers.
3. Test image generation and variation - OpenAI and Gemini.
*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

- [X] Related documentation has been updated

**Other information:**

### DIFF
--- a/packages/ai-helper/skills/provider-creator.md
+++ b/packages/ai-helper/skills/provider-creator.md
@@ -1,602 +1,277 @@
 # AI Provider Creator (ApostropheCMS)
 
-**Purpose:** Define contract for custom AI providers integrating with `@apostrophecms/ai-helper`
+Create a **project-level** provider registration module + a provider **factory** for `@apostrophecms/ai-helper`.
 
-## Scope & Non-Goals
-
-This contract defines the **minimum required behavior** for AI providers
-integrating with `@apostrophecms/ai-helper`.
-
-> **Streaming responses are not currently supported.**
-> Providers must return complete results for text and image generation.
-> Partial or incremental responses will be ignored.
-
-Support for streaming may be added in a future major version.
-
-**Use for:** AWS Bedrock, Azure OpenAI, private LLM gateways, custom models
-
-**Reference implementations:** See bundled providers in `modules/@apostrophecms/ai-helper/providers/`
+This guide is optimized for an LLM to generate files.
 
 ---
 
-## Two Implementation Patterns
+## What you are generating
 
-Both patterns follow the same principle: **register a factory function**, not an instantiated provider. The AI Helper core calls your factory during activation.
+You will generate **two files**:
 
-### Module Pattern (Recommended)
-Self-contained, full lifecycle access, reusable across projects.
-```javascript
-// modules/@my-org/ai-helper-bedrock/index.js
+1. **Provider factory file** (ESM)
 
-/**
- * Bedrock provider factory
- * @param {Object} apos - Apostrophe instance
- * @param {Object} textOptions - Text generation options
- * @param {Object} imageOptions - Image generation options
- * @returns {Object} Provider object
- */
-const bedrockFactory = (apos, textOptions = {}, imageOptions = {}) => {
-  const apiKey = textOptions.apiKey || process.env.BEDROCK_API_KEY;
-  const textModel = textOptions.textModel || 'anthropic.claude-v2';
-  const textMaxTokens = textOptions.textMaxTokens || 1000;
-  const textRetries = textOptions.textRetries || 3;
+* Exports a function that returns a provider object (text/image methods + optional `validate`).
 
-  // Warn about invalid options
-  const validTextOptions = ['apiKey', 'textModel', 'textMaxTokens', 'textRetries'];
-  const invalidOpts = Object.keys(textOptions).filter(k => !validTextOptions.includes(k));
-  if (invalidOpts.length > 0) {
-    apos.util.warn(`Bedrock provider received invalid options: ${invalidOpts.join(', ')}`);
-  }
+2. **Provider registration module** (ESM Apostrophe module)
 
-  return {
-    name: 'bedrock',
-    label: 'AWS Bedrock',
-    capabilities: {
-      text: true,
-      image: false,
-      imageVariation: false
-    },
+* Extends `registerProviders()` and calls `self.registerProvider(...)` to register the factory.
 
-    validate() {
-      if (!apiKey && !process.env.APOS_AI_HELPER_MOCK) {
-        throw new Error(
-          'Bedrock provider requires an API key. ' +
-          'Set it via textProviderOptions.apiKey or export BEDROCK_API_KEY environment variable.'
-        );
-      }
-    },
-
-    async generateText(req, prompt, options = {}) {
-      const maxTokens = options.maxTokens || textMaxTokens;
-      const retries = options.textRetries || textRetries;
-      const model = options.model || textModel;
-
-      // Mock mode support
-      if (process.env.APOS_AI_HELPER_MOCK) {
-        return {
-          content: '# Sample Content\n\nThis is mock content for testing.',
-          metadata: {
-            model: 'mock-model',
-            usage: {
-              prompt_tokens: Math.ceil(prompt.length / 4),
-              completion_tokens: 250,
-              total_tokens: Math.ceil(prompt.length / 4) + 250
-            }
-          }
-        };
-      }
-
-      // Real implementation with retry logic
-      for (let attempt = 1; attempt <= retries; attempt++) {
-        try {
-          // PSEUDOCODE: AWS Bedrock requests require AWS SDK + SigV4 signing.
-          // This example shows response handling only, not a real endpoint.
-          const result = await apos.http.post('https://example-bedrock-endpoint.invalid', {
-            headers: {
-              'Authorization': `Bearer ${apiKey}`,
-              'Content-Type': 'application/json'
-            },
-            body: {
-              model,
-              max_tokens: maxTokens,
-              messages: [
-                {
-                  role: 'user',
-                  content: prompt
-                }
-              ],
-              system: options.systemPrompt || 'You are a helpful AI assistant.'
-            }
-          });
-
-          const content = result?.content?.[0]?.text;
-
-          if (!content) {
-            throw apos.error('error', 'No content returned from Bedrock');
-          }
-
-          return {
-            content,
-            metadata: {
-              usage: result.usage,
-              model: result.model,
-              ...(result.stop_reason && { stop_reason: result.stop_reason })
-            }
-          };
-
-        } catch (e) {
-          apos.util.error(`Bedrock request failed (attempt ${attempt}/${retries}):`, e.message);
-
-          // Don't retry on client errors
-          if (e.status === 400 || e.status === 401 || e.status === 403) {
-            throw e;
-          }
-
-          // Don't retry on last attempt
-          if (attempt === retries) {
-            throw e;
-          }
-
-          // Exponential backoff
-          await new Promise(resolve => setTimeout(resolve, attempt * 1000));
-          apos.util.info(`Retrying Bedrock request (attempt ${attempt + 1}/${retries})...`);
-        }
-      }
-    }
-  };
-};
-
-module.exports = {
-  async init(self) {
-    // Register the factory function with AI Helper
-    const aiHelper = self.apos.modules['@apostrophecms/ai-helper'];
-    
-    aiHelper.registerProvider('bedrock', {
-      factory: bedrockFactory,
-      label: 'AWS Bedrock',
-      capabilities: {
-        text: true,
-        image: false,
-        imageVariation: false
-      }
-    });
-  }
-};
-```
-
-**Enable in app.js:**
-```javascript
-import apostrophe from 'apostrophe';
-
-apostrophe({
-  root: import.meta,
-  modules: {
-    '@my-org/ai-helper-bedrock': {},
-    '@apostrophecms/ai-helper': {
-      options: {
-        textProvider: 'bedrock',
-        textProviderOptions: {
-          apiKey: process.env.BEDROCK_API_KEY,
-          textModel: 'anthropic.claude-v2',
-          textMaxTokens: 2000
-        }
-      }
-    }
-  }
-});
-```
-
-**How it works:**
-1. Your module registers the factory function during `init()`
-2. AI Helper stores the factory and metadata
-3. When `textProvider: 'bedrock'` is configured, AI Helper calls your factory with the options
-4. Your factory returns the instantiated provider
-5. AI Helper calls `validate()` and uses the provider for text generation
+> Note: `@apostrophecms/ai-helper` is CommonJS internally, but **project-level code examples here are ESM**.
 
 ---
 
-### Factory Pattern
-Lightweight, minimal overhead, follows bundled provider pattern. Good for simple integrations.
-```javascript
-// lib/providers/bedrock.js
+## Core behavior you must match
 
-/**
- * Bedrock provider factory
- * @param {Object} apos - Apostrophe instance
- * @param {Object} textOptions - Text generation options
- * @param {Object} imageOptions - Image generation options
- * @returns {Object} Provider object
- */
-module.exports = (apos, textOptions = {}, imageOptions = {}) => {
-  const apiKey = textOptions.apiKey || process.env.BEDROCK_API_KEY;
-  const textModel = textOptions.textModel || 'anthropic.claude-v2';
-  const textMaxTokens = textOptions.textMaxTokens || 1000;
-  const textRetries = textOptions.textRetries || 3;
+### Two-phase provider flow
 
-  return {
-    name: 'bedrock',
-    label: 'AWS Bedrock',
-    capabilities: {
-      text: true,
-      image: false,
-      imageVariation: false
-    },
+1. **Registration phase**
+   You register provider factories with metadata using:
 
-    validate() {
-      if (!apiKey && !process.env.APOS_AI_HELPER_MOCK) {
-        throw new Error('Bedrock provider requires an API key.');
-      }
-    },
+* `self.registerProvider(name, factory, capabilities)`
 
-    async generateText(req, prompt, options = {}) {
-      // Same implementation as module pattern
-      // ...
-    }
-  };
-};
-```
+2. **Activation phase**
+   `@apostrophecms/ai-helper` instantiates only the providers named in project options and (optionally) calls `provider.validate()`.
 
-**Register in your project-level module:**
-```javascript
-// modules/my-custom-module/index.js
-const bedrockFactory = require('../../lib/providers/bedrock');
+### Capabilities: source of truth
 
-module.exports = {
-  async init(self) {
-    const aiHelper = self.apos.modules['@apostrophecms/ai-helper'];
-    
-    // Register the factory function (not an instance!)
-    aiHelper.registerProvider('bedrock', {
-      factory: bedrockFactory,
-      label: 'AWS Bedrock',
-      capabilities: {
-        text: true,
-        image: false,
-        imageVariation: false
-      }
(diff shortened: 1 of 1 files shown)


packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+    // Default providers for text and image generation
+    textProvider: 'openai',
+    imageProvider: 'openai',
+    // Legacy option support
+    textMaxTokens: 1000,
+    // Usage tracking - can be set via option or
+    // APOS_AI_HELPER_LOG_USAGE env var
+    // When true, usage data is logged to the console
+    // for cost tracking and auditing
+    logUsage: process.env.APOS_AI_HELPER_LOG_USAGE === 'true' || false,
```
boutell: Just check it for truthiness IMHO, but if not we must support `1` in particular for consistency with common practice including our own.

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+  async init(self) {
+    // Storage for registered providers
+    self.providers = new Map();
+
+    // Initialize usage storage collection if enabled
+    if (self.options.storeUsage) {
+      self.aposAiHelperUsage = self.apos.db.collection('aposAiHelperUsage');
+      await self.aposAiHelperUsage.createIndex(
+        { createdAt: -1 },
+        { name: 'createdAt_-1' }
```
boutell: I'd remove this as it's not better than the default convention (I think it might be the default convention).

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+      );
+
+      // Index for per-user usage timelines
+      await self.aposAiHelperUsage.createIndex(
+        {
+          userId: 1,
+          createdAt: -1
+        },
+        { name: 'userId_1_createdAt_-1' }
```
boutell: Same re: not adding name if it's just the same as the default convention, which is fine

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
       browser: true
     }
   },
+
   bundle: {
     directory: 'modules',
     modules: getBundleModuleNames()
   },
-  options: {
-    textModel: 'gpt-5.1',
```
boutell, line 21: We have bc provisions I assume?

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+       * Register an AI provider module
+       * @param {Object} provider - The provider module (self)
+       * @param {Object} config - Provider configuration
+       * @param {string} config.name - Unique provider name (e.g., 'openai', 'anthropic')
+       * @param {string} config.label - Human-readable label
+       * @param {boolean} [config.text] - Supports text generation
+       * @param {boolean} [config.image] - Supports image generation
+       * @param {boolean} [config.imageVariation] - Supports image variations
+       */
+      registerProvider(provider, config) {
```
boutell: I actually think a full apostrophecms module is overkill for a provider. Not worth the startup overhead. I would go with passing an ordinary object. It's fine if it happens to be an apostrophe module.

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+          throw new Error('Provider must have a name');
+        }
+
+        if (self.providers.has(name)) {
+          throw new Error(`Provider "${name}" is already registered`);
+        }
+
+        // Validate that the provider implements required methods
+        if (text && typeof provider.generateText !== 'function') {
+          throw new Error(`Provider "${name}" claims text support but doesn't implement generateText()`);
```
boutell: Love the duck typing.

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+          image = false,
+          imageVariation = false
+        } = config;
+
+        if (!name) {
+          throw new Error('Provider must have a name');
+        }
+
+        if (self.providers.has(name)) {
+          throw new Error(`Provider "${name}" is already registered`);
```
boutell: We should allow this as an override mechanism for our core providers.

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+          module: provider,
+          label,
+          capabilities: {
+            text,
+            image,
+            imageVariation
+          }
+        });
+
+        self.apos.util.log(`AI provider registered: ${name} (${label})`);
```
boutell: Use the info log level.

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+       * @param {string} name - Provider name
+       * @returns {Object} Provider info
+       */
+      getProvider(name) {
+        const provider = self.providers.get(name);
+
+        if (!provider) {
+          const available = Array.from(self.providers.keys()).join(', ');
+          throw self.apos.error('notfound',
+            `AI provider "${name}" not found. Available providers: ${available || 'none'}`
```
boutell: Nice.

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+        if (!self.options.logUsage) {
+          return;
+        }
+
+        const username = req.user?.username || 'unknown';
+        const {
+          type, provider, prompt, metadata
+        } = data;
+
+        console.log(`\n[AI Usage] ${type} generation by ${username}:`);
```
boutell: What's with all the `console.log`? This is structured data, use our structured logging features.

boutell: Please add our standard linter to the module so we can detect things like `console.log` calls that should not have been made (I recognize this is a little beyond the scope and will make the PR harder to read on a one-time basis, but it helps catch stuff).

packages/ai-helper/index.js
```diff
@@ -2,34 +2,261 @@ const fs = require('fs');
...
+        console.log({
+          type,
+          provider,
+          ...metadata,
+          prompt: prompt.length > 100 ? prompt.substring(0, 100) + '...' : prompt
+        });
+      },
+
+      /**
+       * Store AI usage to MongoDB for permanent audit trail
```
boutell: An audit trail of apostrophecms actions in general is a desirable feature but shouldn't appear out of the blue in just one optional module, that will lead to many audit collections in many modules etc. Also, if you use structured logging it should be possible for customers who really want it to implement auditing. So I think you should remove the db collection and the logging functionality.

packages/ai-helper/index.js
boutell: I recognize people will want to override options of the individual providers, but they only need to do that for the ones they are actually using. So if you add `textProviderOptions` and `imageProviderOptions` to the main module that should address that concern.

packages/ai-helper/modules/@apostrophecms/ai-helper-anthropic/index.js
```diff
@@ -0,0 +1,120 @@
+module.exports = {
```
boutell: I think it should be an object and not a full apos module.

boutell: Developers can call `registerProvider` in project level `init` of the ai-helper module and pass their own provider objects. The main `init` would call it once for each of the standard providers.

packages/ai-helper/modules/@apostrophecms/ai-helper-anthropic/index.js
```diff
@@ -0,0 +1,120 @@
...
+              role: 'user',
+              content: prompt
+            }
+          ],
+          system: options.systemPrompt || 'You are a helpful AI assistant.'
+        };
+
+        // Retry logic for transient failures
+        let lastError;
+        for (let attempt = 1; attempt <= 3; attempt++) {
```
boutell: Make the number of attempts configurable.

packages/ai-helper/modules/@apostrophecms/ai-helper-anthropic/index.js
```diff
@@ -0,0 +1,120 @@
...
+                model: result.model,
+                ...(result.stop_reason && { stop_reason: result.stop_reason })
+              }
+            };
+
+          } catch (e) {
+            lastError = e;
+
+            // Log for debugging
+            console.error(`Anthropic request failed (attempt ${attempt}/3):`, e.message);
```
boutell: Providers should be registered as functions that accept a services object with more appropriate logging functions (e.g. access to structured logging).

packages/ai-helper/modules/@apostrophecms/ai-helper-gemini/index.js
```diff
@@ -0,0 +1,364 @@
...
+              aspectRatio
+            }
+          }
+        };
+
+        // Gemini generateContent endpoint
+        const url = `https://generativelanguage.googleapis.com/v1beta/models/${model}:generateContent`;
+
+        // Gemini returns one image per request,
+        // so make multiple requests for multiple images
```
boutell: Parallel requests?

packages/ai-helper/modules/@apostrophecms/ai-helper-openai/index.js
```diff
@@ -0,0 +1,258 @@
...
+module.exports = {
+  options: {
+    // API key - can be set via option or APOS_OPENAI_KEY env var
+    apiKey: process.env.APOS_OPENAI_KEY || null,
+    // Default models
+    textModel: 'gpt-5.1',
+    // Default text generation settings
+    textMaxTokens: 1000,
+    imageModel: 'gpt-image-1-mini',
+    imageSize: '1024x1024',
```
boutell: I didn't check, but make sure these are standardized across providers whenever it makes sense.

boutell (changes requested):
This is very cool stuff.

I'm pushing hard to get it right because I can see us pivoting this into core as the way we interface to models in general.

packages/ai-helper/index.js
BoDonkey: Changed.

packages/ai-helper/index.js
BoDonkey: Since we are removing the storage, this goes away.

packages/ai-helper/index.js
BoDonkey: `logUsage` eliminated.

packages/ai-helper/index.js
BoDonkey: Removed.

packages/ai-helper/index.js
BoDonkey: Provisions added.

packages/ai-helper/index.js
BoDonkey: Switched to factories.

packages/ai-helper/index.js
BoDonkey: Now allowed as last registered wins.

packages/ai-helper/index.js
BoDonkey: Changed

packages/ai-helper/index.js
BoDonkey: I think the correct linter - `eslint-config-apostrophe` is already in the project.

packages/ai-helper/index.js
BoDonkey: Both removed.

packages/ai-helper/modules/@apostrophecms/ai-helper-anthropic/index.js
BoDonkey: Changed.

packages/ai-helper/modules/@apostrophecms/ai-helper-anthropic/index.js
BoDonkey: Added and documented as an option.

packages/ai-helper/modules/@apostrophecms/ai-helper-anthropic/index.js
BoDonkey: Done.

packages/ai-helper/modules/@apostrophecms/ai-helper-gemini/index.js
BoDonkey: Refactored for parallel requests.

packages/ai-helper/modules/@apostrophecms/ai-helper-openai/index.js
BoDonkey: Option names were standardized where possible.

packages/ai-helper/modules/@apostrophecms/ai-helper-rich-text-widget/index.js
```diff
@@ -68,48 +69,32 @@ module.exports = {
...
+                  systemPrompt = `${systemPrompt} ${customPrompt}`;
                 } else {
-                  systemPrompt = self.options.customSystemPrompt;
+                  systemPrompt = customPrompt;
                 }
               }
 
-              const result = await provider.module.generateText(req, prompt, {
+              const result = await provider.generateText(req, prompt, {
                 maxTokens: aiHelper.options.textMaxTokens,
```
boutell, line 85: I thought you moved this into a `textProviderOptions` object.

packages/ai-helper/index.js
```diff
@@ -1,51 +1,114 @@
...
+        'Use "textProviderOptions.textMaxTokens" instead.'
       );
-      self.apos.util.log('AI Helper usage storage enabled');
+      textOpts.textMaxTokens = self.options.textMaxTokens;
     }
+    const textProvider = self.options.textProvider;
+    const imageProvider = self.options.imageProvider;
+
+    // Register bundled providers
+    self.registerProvider(openaiProvider(
```
boutell: I think we're trying too hard to provide for multiple text providers, multiple image providers. We don't have a proven need for that yet, so we shouldn't prematurely implement it.

Registering a provider shouldn't immediately instantiate it. It just puts it on the list of candidates that are valid for the `textProvider` and `imageProvider` options.

I think this flow makes sense:

* In `init`, call `self.registerProviders`, which calls `registerProvider` for the built-in choices, and is intended to be extended via `extendMethods` to register more providers
* Then call `activateProviders`, which activates only the needed providers (one for text and one for images)

packages/ai-helper/modules/@apostrophecms/ai-helper-rich-text-widget/index.js
```diff
@@ -35,6 +35,7 @@ module.exports = {
     return {
       post: {
         async aiHelper(req) {
+          console.log('req.body', req.body);
```
boutell, line 38: Cleanup

boutell (changes requested):

packages/ai-helper/index.js
myovchev: Looking at the providers, it's pretty cheap to instantiate. We basically store a simple object representing the meta AND the interface, it's quite smart. Splitting it to register and activate brings more chore IMO. 

I always prefer the idea of explicit registration call (`aiHelper.registerProvider(xxx)`) vs extending (things can go south much easily).

packages/ai-helper/modules/@apostrophecms/ai-helper-image/ui/apos/components/AposAiHelperImageEditor.vue
```diff
@@ -94,15 +125,21 @@ export default {
       this.$emit('modal-result', {
         action
       });
+    },
+    createVariant() {
+      if (!this.variantPrompt.trim()) return;
```
myovchev: Lint issue

packages/ai-helper/index.js
```diff
@@ -1,33 +1,244 @@
...
+
+        return providerInfo.provider;
+      },
+
+      /**
+       * List all registered providers
+       * @returns {Array} Array of provider info
+       */
+      listProviders() {
+        return Array.from(self.providers.entries()).map(([name, info]) => ({
```
myovchev: lint issue

myovchev (changes requested):
I'm giving a high level feedback - I like what I see. The provider contract is clean. It's flexible. I can see how even "non-active" providers (not pointed by the respective options) are still available and can be used in different scenarios programmatically (e.g. I can provide a runtime list in my custom project UI).

Few lint issue has to be fixed.

packages/ai-helper/modules/@apostrophecms/ai-helper-image/ui/apos/components/AposAiHelperImageEditor.vue
BoDonkey: Fixed. (Hate that rule though.)

packages/ai-helper/index.js
BoDonkey: Fixed.

packages/ai-helper/modules/@apostrophecms/ai-helper-rich-text-widget/index.js
```diff
@@ -65,7 +65,7 @@ module.exports = {
               };
               providerName = 'Mock Provider';
             } else {
-              const provider = aiHelper.getTextProvider();
+              const { provider } = aiHelper.getTextProvider();
```
boutell: Is there anything gained by calling getTextProvider, then calling generateText, as opposed to exposing a `generateText` method on `aiHelper` that calls `getTextProvider()` for you? Looking to minimize the effort for a developer wanting to generate some text using the model ApostropheCMS provides for that purpose. Same for images etc. Also helps enforce consistency in the APIs.

BoDonkey: I didn't understand this request at first. Then I thought about it in terms of core. Consolidating those two stepsand taking knowledge of the provider out of the loop then makes sense. Any module can easily call to generate text or images for their use and core handles all the setup. Correct?

BoDonkey: Fixed. Plus, I realized there were still `logUsage` and `storeUsage` calls that needed to be removed. I also added an icon to the generate variant button to fix a height problem.

packages/ai-helper/index.js
myovchev: I don't see it being allowed - not pushed yet?

packages/ai-helper/index.js
```diff
@@ -1,33 +1,321 @@
...
+
+      /**
+       * Get the configured text provider
+       * @returns {Object} Provider info and capabilities
+       */
+      getTextProvider() {
+        const providerName = self.options.textProvider;
+        const providerInfo = self.getProvider(providerName);
+
+        if (!providerInfo.capabilities.text) {
```
myovchev: This validation should be done on startup and not runtime, i.e. when registering/activating.

packages/ai-helper/index.js
```diff
@@ -1,33 +1,321 @@
...
+
+      /**
+       * Get the configured image provider
+       * @returns {Object} Provider info and capabilities
+       */
+      getImageProvider() {
+        const providerName = self.options.imageProvider;
+        const providerInfo = self.getProvider(providerName);
+
+        if (!providerInfo.capabilities.image) {
```
myovchev: Same here.

myovchev (changes requested):

myovchev (dismissed):

PR #5255, opened 2025-12-27
remove premature changelog entry

boutell:
there is a changeset file for this, and it wasn't part of the 4.52.0 release so putting it in the changelog file now was wrong all around


boutell:
I matched it to what was actually released in 4.25.0. I didn't remove any adds, changes or fixes that were actually released in 4.25.0.

PR #5259, opened 2025-12-31
PRO-8859: rich text widgets should not blow out styles when you edit the text

boutell:


packages/apostrophe/modules/@apostrophecms/area/ui/apos/logic/AposAreaEditor.js
```diff
@@ -572,7 +573,7 @@ export default {
       return null;
     },
     contextualWidgetDefaultData(type) {
-      return this.moduleOptions.contextualWidgetDefaultData[type];
+      return klona(this.moduleOptions.contextualWidgetDefaultData[type]);
```
boutell, line 576: Introducing this because with bigger structures it is more likely to be a bad idea to maybe accidentally modify the global defaults.

packages/apostrophe/modules/@apostrophecms/area/ui/apos/logic/AposAreaEditor.js
```diff
@@ -521,7 +522,7 @@ export default {
       if (this.widgetIsContextual(name)) {
         return this.insert({
           widget: {
-            type: name,
+            ...this.newWidget(name),
```
boutell, line 525: When we initialize a contextual widget we still want to initialize schema fields normally before applying the "contextual widget default data" for wacky weirdies (non-schema properties like `content`)

packages/apostrophe/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -334,6 +334,13 @@ export default {
   },
   watch: {
     modelValue(newVal, oldVal) {
+      // Accept any changes that were made to regular schema fields (like styles)
```
boutell, line 337: We were visualizing the styles changes but not recording them in `docFields.data`, so they were not becoming part of the next emitted `update` when typing in the editor, causing styles to disappear.

packages/apostrophe/modules/@apostrophecms/schema/lib/newInstance.js
```diff
@@ -99,7 +99,7 @@ function normalizeWidget(widgetInstance, self) {
     }
   }
 
-  // Contextual default data
+  // Contextual default data other than schema fields
```
boutell, line 102: This is just for clarity. This file ALREADY instantiates schema fields for contextual widgets, but this particular bit of code is about the wacky weirdies like `content`.

PR #5260, opened 2026-01-08
PRO-8899: browser side live preview of new widgets

boutell:


packages/apostrophe/modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -421,7 +418,7 @@ export default {
         ...props
       };
     },
-    getPreviewWidgetObject() {
+    getPreviewWidgetId() {
```
boutell, line 421: A little refactoring to expose just this

packages/apostrophe/modules/@apostrophecms/widget-type/ui/apos/components/AposWidgetEditor.vue
```diff
@@ -275,7 +275,7 @@ export default {
           checkIfConditionsFn: checkIfConditions,
           subset: this.moduleOptions.stylesFields
         });
-        this.applyPreviewStyles(styles, value.data);
+        this.applyPreviewStyles(styles);
```
boutell, line 278: value.data was only passed to determine the id and we have a better way

PR #5275, opened 2026-01-16
PRO: Direction locale option

myovchev:
## Summary

Add direction locale option,  i18n template data, and direction for localized documents.


packages/apostrophe/modules/@apostrophecms/i18n/index.js
```diff
@@ -175,6 +175,12 @@ module.exports = {
         // The array is provided in the order in which locales are configured.
         // The current locale is included and has the property `current: true`.
         async addLocalizations(req) {
+          const locale = req.locale || self.defaultLocale;
```
myovchev, line 178: The changes here will also land in the external front-ends.

BoDonkey (approved):

PR #5279, opened 2026-01-16
Feature RTL

myovchev:
## Summary

Add support for RTL locales.

### DIFF
--- a/.changeset/lucky-towns-check.md
+++ b/.changeset/lucky-towns-check.md
@@ -2,4 +2,4 @@
 "apostrophe": minor
 ---
 
-In `@apostrophecms/i18n` add `direction` property to locale configuration to support RTL languages (e.g., `he`, `ar`), `slugDirection` option to control default direction of slug fields. Add `direction` property in the schema field definitions to override and validate text direction (supports `ltr` and `rtl`) of input fields.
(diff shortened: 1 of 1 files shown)


.changeset/lucky-towns-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"apostrophe": minor
+---
+
+In `@apostrophecms/i18n` add `direction` property to locale configuration to support RTL languages (e.g., `he`, `ar`), `slugDirection` option to control default direction of slug fields. Add `direction` property in the schema field definitions to override and validate text direction (supports `ltr` and `rtl`) of input fields.
```
boutell: Please clarify that admin UI layout and labels overall are still LTR only for now, but these changes accommodate editing RTL locale content within that.

For best results the feature should be combined with `adminLocales` and `defaultAdminLocale`, e.g. the admin UI itself should remain in English or another LTR language for now.

packages/apostrophe/modules/@apostrophecms/box-field/index.js
```diff
@@ -64,6 +64,10 @@ module.exports = {
             const defProps = [ 'top', 'right', 'bottom', 'left' ];
             let defMin = 0;
 
+            if (field.direction && ![ 'ltr', 'rtl' ].includes(field.direction)) {
```
boutell, line 67: Numbers are still displayed LTR in the major RTL languages, but this is handled automatically by Unicode when RTL direction is otherwise in effect. So I guess supporting this for "box" doesn't actually hurt anything and might be good somehow.

packages/apostrophe/modules/@apostrophecms/i18n/index.js
```diff
@@ -32,6 +32,11 @@
 // ### `encoding`
 //
 // Defaults to `'utf-8'`. You almost certainly do not want to change this.
+//
+// ### `slugDirection`
+//
+// Controls the default `direction` value of slug schema. Can be `ltr`, `rtl` or
```
boutell, line 38: We already gave Michelin a way to force simplified LTR slugs anyway, so I doubt this will get used much.

PR #5282, opened 2026-01-19
Fix rich text styles computation performance

myovchev:
## Summary

A simple patch that prevents style recomputation on every rich text keystroke.
Additionally, fix missing default values from the styles schema.


packages/apostrophe/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposRichTextWidgetEditor.vue
```diff
@@ -334,6 +336,9 @@ export default {
   },
   watch: {
     modelValue(newVal, oldVal) {
+      if (newVal.content !== oldVal.content) {
```
myovchev, line 339: Performance, no style related change.

PR #5283, opened 2026-01-20
PRO-8875-direction-validation

myovchev:
## Summary

- Validate direction schema for the relevant core fields
- Revert `range` field direction support
- Add direction support for `box` field

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
+++ b/packages/apostrophe/modules/@apostrophecms/area/ui/apos/components/AposAreaWidget.vue
@@ -149,6 +149,7 @@
         :is="widgetEditorComponent(widget.type)"
         v-if="isContextual && !foreign"
         :key="generation"
+        :class="adminContentDirectionClass"
         :options="widgetOptions"
         :type="widget.type"
         :model-value="widget"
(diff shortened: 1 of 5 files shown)


packages/apostrophe/modules/@apostrophecms/schema/ui/apos/logic/AposInputRange.js
```diff
@@ -1,9 +1,8 @@
 import AposInputMixin from 'Modules/@apostrophecms/schema/mixins/AposInputMixin.js';
-import AposFieldDirectionMixin from 'Modules/@apostrophecms/schema/mixins/AposFieldDirection.js';
```
myovchev, line 2: revert range - it's technically an input but it's broken with RTL direction.

packages/apostrophe/modules/@apostrophecms/box-field/ui/apos/logic/AposInputBox.js
```diff
@@ -38,11 +39,18 @@ export default {
     tabindex () {
       return this.field.disableFocus ? '-1' : '0';
     },
-    classes() {
```
myovchev, line 41: `classes` is not referenced in the Vue component, orphan computed prop.

PR #5298, opened 2026-01-30
Test coverage for the @ syntax of PATCH when applied to area ids

boutell:
Turns out it works fine, but the test is healthy to have. Created with claude code.

### DIFF
--- a/packages/apostrophe/test/patch-image-widget.js
+++ b/packages/apostrophe/test/patch-image-widget.js
@@ -23,7 +23,16 @@ describe('PATCH Image Widget to Area', function() {
                 type: 'area',
                 options: {
                   widgets: {
-                    '@apostrophecms/image': {}
+                    '@apostrophecms/image': {},
+                    '@apostrophecms/rich-text': {}
+                  }
+                }
+              },
+              secondary: {
+                type: 'area',
+                options: {
+                  widgets: {
+                    '@apostrophecms/rich-text': {}
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


packages/apostrophe/test/patch-image-widget.js
```diff
@@ -0,0 +1,171 @@
...
+    const user = apos.user.newInstance();
+    user.title = 'admin';
+    user.username = 'admin';
+    user.password = 'admin';
+    user.email = 'ad@min.com';
+    user.role = 'admin';
+    await apos.user.insert(apos.task.getReq(), user);
+  });
+
+  it('should be able to login as admin', async function() {
```
myovchev, line 60: This should be part of the `before` hook. The rest of the tests shouldn't depend on any other test. `before` in this case is fine because there is no `reset` between tests (otherwise `beforeEach` would be more appropriate)

packages/apostrophe/test/patch-image-widget.js
```diff
@@ -0,0 +1,171 @@
...
+      body: {
+        username: 'admin',
+        password: 'admin',
+        session: true
+      },
+      jar
+    });
+  });
+
+  it('should be able to upload an image and create an image piece', async function() {
```
myovchev, line 77: Same here. This is just state used by other tests, in this case `before` is the right place.

myovchev (changes requested):
Just a little refactoring for consistency with the "non-dependent" test standards (raised some time ago). There are also 2 eslint problems to be fixed.

PR #5299, opened 2026-02-02
Pro 8395/widget editor window

stuartromanek:
## Scope of changes
- Adds a 'window' modal type
- Abstracts the window dragging mechanism used by Palette (global styles) and implements it in the new window modal type
- widget module's `width` option expanded to accept `window`, which is now the module default. Can still set `one-third`, `two-thirds`, and `half`, as before.
- The editor can toggle the modal type with an icon UI. Changing the position will save this preference locally **for that widget type**
- Updates the strategy for placing widget editor modal tabs, which used to be hardcoded as a tab length but are now have their rendered sizes calculated to determine whether or not the additional menu is needed. The menu no longer appears if unneeded.

### Additional changes
- Updates modal windows to ensure they are scrolled to top when they are opened and that tabs scroll to the top when changed
- Updates Box field type to switch between Individual and Shorthand views instead of disabling Shorthand when Individual is active
- In Object field type add `options.flat` and `options.hideLabel` to alter rendered field. `flat` flattens the visualization of subprops (no nesting or left border), `hideLabel` will remove the parent label. This is implemented in the `boxShadow` and `border` presets
- Updates various inputs to pass modifier props, print proper BEM-style class names, micro styles

## Manual testing
Test against https://github.com/apostrophecms/public-demo/tree/pro-8395/widget-editor-window-demo. In it are lots of juicy widget schema examples. 
For the biggest test add a 'Dummy Widget' to a page (very bottom of Add Content menu on home page, no preview). It will add an H1 to the page and the schema will not impact the display of the widget. The schema of Dummy is taken from Testbed's `modules/topic/index.js`schema, which is meant to test most all field types. give it a looksie.

## Related PRs
Testbed: https://github.com/apostrophecms/testbed/pull/417
Cypress-tools: https://github.com/apostrophecms/cypress-tools/pull/90

### DIFF
--- a/.changeset/nice-vans-boil.md
+++ b/.changeset/nice-vans-boil.md
@@ -1,0 +1,5 @@
+---
+"@apostrophecms/cli": patch
+---
+
+- Validates the `shortName` argument in the `create` command to only allow letters, numbers, hyphens, and underscores, preventing potential command injection.
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -26,6 +26,11 @@ module.exports = function (program) {
       'Add a connection string to connect to a hosted MongoDB instance.'
     )
     .action(async function (shortName, options) {
+      if (!/^[\w-]+$/.test(shortName)) {
+        await util.error('create', 'The shortname must only contain letters, numbers, hyphens, and underscores.');
+        return false;
+      }
+
       // If options.starter is undefined, use config.BOILERPLATE
       const input = options.starter ? options.starter : config.BOILERPLATE;
 


packages/apostrophe/modules/@apostrophecms/schema/ui/apos/logic/AposInputObject.js
```diff
@@ -39,6 +39,16 @@ export default {
     };
   },
   computed: {
+    options() {
+      return this.field?.options || {};
+    },
+    finalizedField() {
+      const field = this.field;
+      if (this.options.hideLabel) {
+        delete this.field.label;
```
myovchev: This mutation can do harm (prop mutation). Clone (spread) the object before deleting the prop. 
```js
const field = { ...this.field };
```
should be enough.

packages/apostrophe/modules/@apostrophecms/ui/ui/apos/composables/useDraggableWindow.js
```diff
@@ -0,0 +1,391 @@
...
+    if (storageKey) {
+      try {
+        localStorage.setItem(storageKey, JSON.stringify(position.value));
+      } catch (err) {
+        // Ignore storage errors (e.g. private mode)
+      }
+    }
+
+    window.removeEventListener('mousemove', drag);
+    window.removeEventListener('mouseup', stopDragging);
```
myovchev: It looks to me that removing the listener should be the first thing to happen here (to avoid racing conditions when more "dragging" happens while it computes stuff)

stuartromanek: moved this

packages/apostrophe/modules/@apostrophecms/schema/ui/apos/logic/AposInputObject.js
stuartromanek: geeze didn't finish this. fixed.

packages/apostrophe/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
```diff
@@ -262,18 +264,22 @@ const {
...
+  cleanup: cleanupDraggable
 } = useDraggableWindow({
   size,
   getDefaultPosition: () => getDefaultPosition(props.modal.origin),
-  minWidth: props.modal.minWidth !== undefined ? props.modal.minWidth : 280,
-  maxWidth: props.modal.maxWidth !== undefined ? props.modal.maxWidth : null,
-  minHeight: props.modal.minHeight !== undefined ? props.modal.minHeight : 200,
+  minWidth: props.modal.minWidth !== undefined ? props.modal.minWidth : 300,
+  maxWidth: props.modal.maxWidth !== undefined ? props.modal.maxWidth : 800,
+  minHeight: props.modal.minHeight !== undefined ? props.modal.minHeight : 300,
```
myovchev: More readable version of these:
```
minWidth: props.modal.minWidth || 300
```
This would fallback to 300 when the value is one of `false | null | udnefined | 0 | '(empty string)'`

packages/apostrophe/modules/@apostrophecms/ui/ui/apos/composables/useDraggableWindow.js
```diff
@@ -374,6 +503,26 @@ export function useDraggableWindow({
...
+    window.removeEventListener('mousemove', doResize);
+    window.removeEventListener('mouseup', stopResizing);
+    dragging.value = false;
+    resizing.value = false;
+    dragClassApplied.value = false;
+    resizeClassApplied.value = false;
+    document.getSelection().removeAllRanges();
+  }
```
myovchev, line 529: Double check the cleanup, it duplicates the cleanup code from stopDragging and stopResizing (and there are some related changes I think). It would be nice that this uses a shared with the stop handlers logic (single source of truth) so we don't introduce bugs/memory leaks in the future. What I mean here is `cleanupDragging` and `cleanupResizing` handlers that combined do what we see here. Then here we call them and they can be used in stopDragging/stopResizing respectively.

packages/apostrophe/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
stuartromanek: applied

packages/apostrophe/modules/@apostrophecms/ui/ui/apos/composables/useDraggableWindow.js
stuartromanek: consolidated everything into its respective `cleanup` handler and cleanups are now called within respective `stop` handlers. @myovchev does this seem right?

PR #5326, opened 2026-02-13
PRO-9175: fix regression that broke on-page, in-context area editing for certain areas

boutell:
To verify this fix, you can edit your page template so that it has at least two separate area fields, and one of the `{% area... %}` calls is nested in a div, while the other is not.

Without the fix, the one nested in a div is not editable on-page.

With the fix, they are both editable on-page.

The fix has also been verified on the Wissahickon Charter School site.

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
+++ b/packages/apostrophe/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
@@ -49,11 +49,17 @@ export default function() {
     els.filter(el => depth(el) === lowest).forEach(el => createAreaApp(el));
   }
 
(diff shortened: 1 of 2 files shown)


packages/apostrophe/modules/@apostrophecms/area/ui/apos/apps/AposAreas.js
```diff
@@ -49,11 +49,17 @@ export default function() {
     els.filter(el => depth(el) === lowest).forEach(el => createAreaApp(el));
   }
 
+  // Determine how deeply it is nested in other areas. We don't care about
```
boutell, line 52: Without this, projects that happen to nest "root" areas in non-area-related markup to a mix of depths have problems: the more deeply nested root areas and their descendants never become editable on-page.

PR #5351, opened 2026-03-10
PRO-9225: Combine Styles and Column configuration

myovchev:
## Summary

Combine Styles and Column configuration in a single Styles Editor experience

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
@@ -198,7 +198,7 @@ module.exports = {
       },
 
       migrateColumnWidget(widget, dotPath) {
-        if (typeof widget.showTablet === 'boolean') {
+        if (!widget.desktop) {
           return null;
         }
         const d = widget.desktop || {};
--- a/packages/apostrophe/test/layout-widget-migration.js
+++ b/packages/apostrophe/test/layout-widget-migration.js
@@ -58,32 +58,6 @@ describe('Layout Widget Migration', function () {
       });
     });
 
-    it('returns null for already-migrated widget (idempotency)', function () {
-      const widget = {
-        _id: 'w1',
-        type: '@apostrophecms/layout-column',
-        metaType: 'widget',
-        desktop: {
-          colstart: 1,
-          colspan: 6
-        },
-        tablet: {},
-        mobile: {},
-        colstart: 1,
-        colspan: 6,
-        rowstart: 1,
-        rowspan: 1,
-        order: 0,
-        justify: 'center',
-        align: 'start',
-        showTablet: true,
-        showMobile: true
-      };
-
-      const result = migrate(widget, 'main.items.0');
-      assert.equal(result, null);
-    });
-
     it('applies defaults when optional fields are missing', function () {
       const widget = {
         _id: 'w1',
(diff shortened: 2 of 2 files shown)


packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -103,128 +50,177 @@ module.exports = {
...
+    group: {
+      utility: {
+        fields: [ 'colstart', 'colspan', 'rowstart', 'rowspan', 'order', 'content' ]
+      }
+    }
+  },
+  styles(self, options) {
     const breakpoints = options?.labelBreakpoints ||
-      options?.breakpoints || // BC, derpecated
+      options?.breakpoints || // BC, deprecated
```
BoDonkey, line 94: Golden typo!!

packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -103,128 +50,177 @@ module.exports = {
...
         }
       }
     };
+  },
+  init(self) {
+    self.addColumnWidgetMigration();
+  },
+  methods(self) {
+    return {
+      addColumnWidgetMigration() {
```
myovchev, line 179: @boutell The column migration

packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -103,128 +50,177 @@ module.exports = {
...
+            });
+          }
+        );
+      },
+
+      migrateColumnWidget(doc) {
+        let changed = false;
+        self.apos.area.walk(doc, (area) => {
+          for (const item of (area.items || [])) {
+            if (item.type !== self.name) {
```
myovchev: The `name` should be the right thing to compare to

packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -103,128 +50,177 @@ module.exports = {
...
+              continue;
+            }
+            // Idempotency check
```
boutell: Technically unnecessary, but I'm good with defense in depth!

packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -103,128 +50,177 @@ module.exports = {
...
+    self.addColumnWidgetMigration();
+  },
+  methods(self) {
+    return {
+      addColumnWidgetMigration() {
+        self.apos.migration.add(
+          `${self.__meta.name}:flatten-column-schema`,
+          async () => {
+            await self.apos.migration.eachDoc({}, 5, async (doc) => {
+              if (await self.migrateColumnWidget(doc)) {
```
boutell: Use `eachWidget` and `$set`. `eachWidget` provides the dot path. You still have to check the widget type.

myovchev:
> Your code is correct. But. From long experience:
> 
> Just don't rip out the old properties at all. If you don't remove it, a bug can't bite you leaving no path forward.
> 
> You can do your idempotency check based on the existence of the new properties only.

And it solves millions of things that could go wrong - this was the missing part I needed :) It will be up in a minute.

packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
myovchev: Actually this created problems. I'm glad I sanity checked. I need to investigate why.

packages/apostrophe/modules/@apostrophecms/layout-column-widget/index.js
```diff
@@ -198,30 +198,22 @@ module.exports = {
       },
 
       migrateColumnWidget(widget, dotPath) {
-        // Idempotency check
-        if (!widget.desktop && typeof widget.colstart === 'number') {
+        if (typeof widget.showTablet === 'boolean') {
```
boutell: Unfortunately... the `addMissingSchemaFields` logic may prevent this from working properly. In fact, it *should* prevent it from working properly, according to my reading of the migration module source. Even if something is currently preventing that from happening, I wouldn't trust it.

So we have two options:

* We can trust apostrophecms to track completed migrations. It does do that.
* We can *rename* rather than *remove* the old properties, so that your original idempotency check can be used.

boutell (changes requested):
Sorry! LOL

PR #5352, opened 2026-03-10
PRO-6706: optional pretty URLs for PDFs (files, not images)

boutell:
A simple and effective implementation. Optional because there is a performance tradeoff but this is a reasonably performant solution. I don't think we should implement it for images, because it would hurt performance more significantly, and RA agrees that's not important. People care *much* more for PDFs because they are pages, of a sort, and people see that URL.

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/file/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/file/index.js
@@ -109,7 +109,7 @@ module.exports = {
         for (const file of files) {
           if (self.options.prettyUrls) {
             const { extension } = file.attachment;
-            const baseUrl = self.apos.url.getBaseUrl(req, { prefix: true })
+            const baseUrl = self.apos.url.getBaseUrl(req, { prefix: true });
             file._url = `${baseUrl}${self.options.prettyUrlDir}/${file.slug.replace(self.options.slugPrefix || '', '')}.${extension}`;
             file.attachment._prettyUrl = file._url;
           } else {


packages/apostrophe/lib/stream-proxy.js
```diff
@@ -0,0 +1,49 @@
...
+    // Can't make a self-request without an absolute URL
+    url = `${req.baseUrl}${url}`;
+  }
+  let response;
+  try {
+    response = await fetch(url);
+  } catch (e) {
+    return send502(e);
+  }
+  for (const header of [ 'content-type', 'etag', 'last-modified', 'content-disposition', 'cache-control' ]) {
```
boutell, line 19: This is the set of useful headers I can proxy through without causing any problems if the original is gzip encoded.

packages/apostrophe/modules/@apostrophecms/doc/index.js
boutell, line 57: I think eslint --fix did this.

packages/apostrophe/modules/@apostrophecms/file/index.js
```diff
@@ -93,9 +95,58 @@ module.exports = {
   methods(self) {
     return {
       addUrls(req, files) {
-        _.each(files, function (file) {
-          file._url = self.apos.attachment.url(file.attachment);
-        });
+        for (const file of files) {
+          if (self.options.prettyUrls) {
+            const { extension } = file.attachment;
+            file._url = `${req.baseUrl}${self.options.prettyUrlDir}/${file.slug.replace(self.options.slugPrefix || '', '')}.${extension}`;
```
boutell: The slugPrefix isn't "pretty" (/files/file-name is not as nice as /files/name)
Slug without prefix is still unique among pieces of this type.

packages/apostrophe/modules/@apostrophecms/file/index.js
```diff
@@ -93,9 +95,58 @@ module.exports = {
...
+            file.attachment._prettyUrl = file._url;
```
boutell, line 114: A hint to .url() for later

packages/apostrophe/modules/@apostrophecms/file/index.js
```diff
@@ -93,9 +95,58 @@ module.exports = {
...
+      }
+    };
+  },
+  routes(self) {
+    if (!self.options.prettyUrls) {
+      return;
+    }
+    return {
+      get: {
+        async [`${self.options.prettyUrlDir}/*`](req, res) {
```
boutell, line 128: Implements /files/pretty-name-here.pdf

packages/apostrophe/modules/@apostrophecms/file/index.js
myovchev: This will blow up static builds. 
```js
const baseUrl = self.apos.url.getBaseUrl(req, { prefix: true })
```
The Base URL should respect:
- prefix
- static base URL if set
- etc - all rules applied by the single source of truth now above

if prefix shouldn't be respected, `prefix: false` can be passed.
Even with that, static build should be also validated, because 
- we expect relative URLs when getting metadata
- and single base URL for the attachments 
- I doubt this will be included in the attachments metadata at all
- We ignore the files and images modules metadata when processing "used" attachments, but `files` can register these URLs as "literal content" I guess, which can save the day
- If we do that, static build will have duplicated (different paths) files, but this is expected and the right thing to do

packages/apostrophe/modules/@apostrophecms/file/index.js
```diff
@@ -104,9 +106,59 @@ module.exports = {
...
+          try {
+            const matches = (req.params[0] || '').match(/^([^.]+)\.\w+$/);
```
BoDonkey, line 130: This regex expects that the file name will have a single `.`, so a name like `my-report.v2.pdf` would not match. Slugify shouldn't let this through, so this case should never occur, but do we need to modify this to recognize the last `.` in the string, not the first?

packages/apostrophe/modules/@apostrophecms/file/index.js
```diff
@@ -104,9 +106,59 @@ module.exports = {
...
+            if (!matches) {
```
BoDonkey, line 131: You look at the file name, but don't save the extension and do any checks. Do we need to further confirm that the file extension is `pdf`?

packages/apostrophe/modules/@apostrophecms/file/index.js
boutell: Indeed slugify will not let it through, so we won't find a match. It can't be used maliciously, which is what matters here. If dots occurred in our slugs it would need more work.

packages/apostrophe/modules/@apostrophecms/file/index.js
boutell: We do not because we know ApostropheCMS guarantees a unique slug for each file piece, regardless of extension.

PR #5356, opened 2026-03-16
PRO-9314: Add support for files pretty URLs in static builds

myovchev:
## Summary

Added support for pretty URL file attachments in the static build metadata pipeline. When `@apostrophecms/file` has `options.prettyUrls` enabled, the getAllUrlMetadata API now annotates affected attachment properly. The backend streaming proxy route was also fixed to correctly resolve relative uploadfs URLs during static builds.

The `writeAttachments` step now supports per-entry base URL resolution, correctly downloading and writing pretty URL files to the appropriate output directory (e.g. `dist/files/`). A new `attachmentFilter` option ('all' or 'prettyOnly') lets you skip regular uploadfs attachments when those are served by a CDN, while still including backend-served pretty URL files in the static output. Configurable via the `staticBuild.attachmentFilter` integration option or the `APOS_ATTACHMENT_FILTER` environment variable.

Files should be properly copied and referenced in static builds when pretty URL option for files is enabeld.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


.changeset/wicked-ducks-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apostrophecms/apostrophe-astro": minor
+---
+
+The `writeAttachments` step now supports per-entry base URL resolution, correctly downloading and writing pretty URL files to the appropriate output directory (e.g. `dist/files/`). A new `attachmentFilter` option ('all' or 'prettyOnly') lets you skip regular uploadfs attachments when those are served by a CDN, while still including backend-served pretty URL files in the static output. Configurable via the `staticBuild.attachmentFilter` integration option or the `APOS_ATTACHMENT_FILTER` environment variable.
```
boutell, line 5: I think the prettyOnly distinction does not make sense because you are 100% guaranteed of failure if you do not copy the "pretty URL" files. It's still true or false, but pretty URL files should still be copied with "false".

packages/apostrophe/modules/@apostrophecms/file/index.js
```diff
@@ -105,19 +105,109 @@ module.exports = {
...
+
+        for (let i = 0; i < attachmentMeta.results.length; i += batchSize) {
+          const batch = attachmentMeta.results.slice(i, i + batchSize);
+          const batchIds = batch.map(a => a._id);
+
+          // Lightweight direct query — no builders, no relationship
+          // resolution.  Only needs slug and attachment ID/extension.
+          const fileDocs = await self.apos.doc.db.find({
+            type: self.options.name,
+            'attachment._id': { $in: batchIds },
```
boutell, line 178: We need to add an index then. Could be painful otherwise. Full collection scan.

.changeset/wicked-ducks-greet.md
myovchev: This is not the purpose of the `prettyOnly` filter. You have 2 options "All" or "prettyOnly". You can't exclude them but you can copy only them (and leave the CDN files alone).

packages/apostrophe/modules/@apostrophecms/file/index.js
myovchev: Even when it's only used for static-build and issued once? Isn't that overkill for the production node backends?

boutell: Static build performance is likely to be a concern customers express down the road, but approving and merging because this is good enough and I know you're out of hours. Thanks

PR #5403, opened 2026-05-08
clarifications to the changesets pointing out that certain fixes are security related, although not severe

boutell:


.changeset/wild-forks-fetch.md
```diff
@@ -1,5 +1,5 @@
 ---
-"apostrophe": minor
+"apostrophe": patch
```
boutell, line 2: We didn't remove anything that worked, and it is security related although none of the old services have been allowed to expire and been pwn3d... yet.

BoDonkey (approved):

PR #5404, opened 2026-05-08
Latest merge prerelease 2026 05

boutell:
I'll follow a similar process on Wednesday after we land the advisories in main. Then `latest` (NOT `main`) will be ready for our normal release PR procedures. Then after releasing it we can just merge `latest` back to `stable` because there is no divergence yet. We won't make releases from `stable` this month for the same reason, that starts next month.


BoDonkey (approved):

boutell:
Christ, I got the base wrong, gotta watch that

PR #5409, opened 2026-05-13
mergeback

boutell:


boutell:
This PR is bigger than I expected. It looks like some work went to `latest` originally and never `main` until now. Well, here it is, finding its way to `main`.

BoDonkey (approved):

PR #5411, opened 2026-05-13
fix: skip ACL params when bucketObjectsACL is false

### LINKED ISSUE #5366: S3: uploads fail with AccessControlListNotSupported on buckets with BucketOwnerEnforced
## Problem

Since April 2023, AWS S3 creates all new buckets with `ObjectOwnership: BucketOwnerEnforced` 
by default, which disables ACLs entirely. uploadfs always sends `ACL: 'public-read'` 
(or whatever `bucketObjectsACL` is set to) in every `copyIn`, `enable`, and `disable` call, 
causing all uploads to fail with:
```
AccessControlListNotSupported: The bucket does not allow ACLs
```

## Root Cause

In `lib/storage/s3.js`, the `ACL` param is always included in requests:
```js
// copyIn
const params = {
  Bucket: bucket,
  ACL: bucketObjectsACL, // always sent, no way to disable
  ...
};

// enable / disable also always call PutObjectAclCommand
```

## Expected Behavior

It should be possible to pass `bucketObjectsACL: false` (and `disabledBucketObjectsACL: false`) 
to skip sending ACL headers entirely, for use with modern S3 buckets or S3-compatible 
storage that does not support ACLs.

## Suggested Fix
```js
bucketObjectsACL = options.bucketObjectsACL === false
  ? false
  : (options.bucketObjectsACL || 'public-read');

disabledBucketObjectsACL = options.disabledBucketObjectsACL === false
  ? false
  : (options.disabledBucketObjectsACL || 'private');
```

Then in `copyIn`, `enable`, and `disable` — skip ACL if value is `false`.

This is fully backward compatible — existing users who don't set these options 
get the same `public-read` default as before.

## Related

- AWS announcement: https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/

ghost:
## Summary

Fixes uploads failing with `AccessControlListNotSupported` on S3 buckets 
with `BucketOwnerEnforced` policy. Since April 2023, AWS S3 creates all new 
buckets with ACLs disabled by default, but `lib/storage/s3.js` always sent 
ACL headers in every request with no way to disable them.

When `bucketObjectsACL: false` is passed, ACL headers are now skipped 
entirely in `copyIn`, `enable`, and `disable` calls.

Closes apostrophecms/apostrophe#5366

1. Create an AWS S3 bucket with `ObjectOwnership: BucketOwnerEnforced` (default since April 2023)
2. Configure uploadfs with `bucketObjectsACL: false` and `disabledBucketObjectsACL: false`
3. Try uploading a file — it should succeed without ACL errors
4. Verify existing users without these options still get `public-read` default behavior

- [x] Bug fix

- [x] It includes the existing issue ID being resolved (apostrophecms/apostrophe#5366)

**Other information:**

Note: I have not updated the changelog or tests yet — 
happy to do so if the maintainers approve this approach first.

### DIFF
--- a/packages/uploadfs/lib/storage/s3.js
+++ b/packages/uploadfs/lib/storage/s3.js
@@ -8,6 +8,7 @@ const {
   S3Client,
   GetObjectCommand,
   DeleteObjectCommand,
+  CopyObjectCommand,
   PutObjectAclCommand
 } = require('@aws-sdk/client-s3');
 const { Upload } = require('@aws-sdk/lib-storage');
@@ -48,7 +49,7 @@ module.exports = function() {
       }
 
       bucket = options.bucket;
-      bucketObjectsACL = options.bucketObjectsACL || 'public-read';
+      bucketObjectsACL = (options.bucketObjectsACL === false) ? false : (options.bucketObjectsACL || 'public-read');
       disabledBucketObjectsACL = options.disabledBucketObjectsACL || 'private';
       noGzipContentTypes = options.noGzipContentTypes || require('./noGzipContentTypes');
       addNoGzipContentTypes = options.addNoGzipContentTypes || [];
@@ -116,11 +117,14 @@ module.exports = function() {
 
       const params = {
         Bucket: bucket,
-        ACL: bucketObjectsACL,
         Key: utils.removeLeadingSlash(self.options, path),
         Body: inputStream,
         ContentType: contentType
       };
+
+      if (bucketObjectsACL !== false) {
+        params.ACL = bucketObjectsACL;
+      }
 
       if (gzipAppropriate(contentType)) {
         params.ContentEncoding = 'gzip';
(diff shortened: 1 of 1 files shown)


boutell:
This eliminates the ability to enable or disable files entirely, or makes it only pretend to do those things, which would not meet the need.

A better approach would be to look at what's done in the local and azure storage backends, both of which handle disabling access by hashing the filename in a secure way.

Ideally that approach should be made available globally with reusable logic.

boutell (changes requested):
Thanks for contributing!

Unfortunately this eliminates the ability to enable or disable files entirely, or makes it only pretend to do those things, which would not meet the need.

A better approach would be to look at what's done in the local and azure storage backends, both of which handle disabling access by hashing the filename in a secure way.

Ideally that approach should be made available globally with reusable logic.

boutell:
(Submissions must also pass the test suite.)

ghost:
Hi @boutell — I've refactored the approach based on your feedback:

disable() and enable() now use filename hashing via utils.getDisabledPath() when disabledFileKey is set — same pattern as the local and azure backends
Old ACL-based path kept as fallback for backward compatibility
copyIn now also skips ACL param when bucketObjectsACL: false

Please let me know if this approach looks good and I'll update the changelog and tests too!

boutell:
Sounds promising. We'll check it out in a week and a half or so, I am on
vacation.

Thank you!


THOMAS BOUTELL | CHIEF EXECUTIVE OFFICER
APOSTROPHECMS | apostrophecms.com | he/him/his

ghost:
Hi @boutell — I've updated the PR with the filename hashing approach using utils.getDisabledPath() for disable/enable, same as the local and azure backends, with ACL fallback for backward compatibility. 
Changelog and tests will be added once you confirm this looks good. 
Enjoy your vacation!

packages/uploadfs/lib/storage/s3.js
```diff
@@ -230,24 +234,36 @@ module.exports = function() {
     },
 
     enable: function(path, callback) {
+      if (self.options.disabledFileKey) {
+        const dPath = utils.getDisabledPath(path, self.options.disabledFileKey);
+        return self._copyObject(dPath, path, function(err) {
+          if (err) return callback(err);
```
boutell, line 240: Please follow the convention (no single-line statements in if, use {})

boutell (changes requested):
We'll need test coverage showing the tests work with or without this option. Also a minor nitpick about code formatting.

We're also going to need to think about a migration path, but this is fine for new projects as written, so I won't hold you up for that.

PR #5417, opened 2026-05-14
Fix/new area fields rendering

stuartromanek:


BoDonkey (dismissed):
Two thoughts:
First, `ensureMissingSchemaAreasForEditableDocTree` now runs on every call to `annotateDocForExternalFront`, walking the full doc tree and calling `getManagerOf` + schema inspection on every editable object, even after all stubs are materialized. For pages with large widget counts or many joined pieces, this could add up. I'm slightly nervous about the performance hit on large pages with nested areas.
Second, `_edit` is currently checked per object, but the entire doc needs to be walked to find each object. Could we get an early return if the doc itself isn't editable?

BoDonkey (commented):
I'm really worried about the performance hit here. I asked Miro to take a look and he wants to discuss further.

PR #5425, opened 2026-05-19
Feature create-apostrophe

myovchev:
## Summary

The modern `create-apostrophe` CLI experience.


BoDonkey (dismissed):

packages/create-apostrophe/src/core/steps/sample-data.js
```diff
@@ -0,0 +1,491 @@
...
+          fetchImpl, assets.uploads, files.uploads, (f) => t.progress?.(f)
+        )
+      );
+    }
+
+    await runTask(task('Importing sample content'), async () => {
+      if (assets.db) {
+        files.db = path.join(tmpDir, 'database.zip');
+        await downloadAndVerify(fetchImpl, assets.db, files.db);
+        const uri = resolveDbUri({
```
BoDonkey, line 132: Claude flagged this as potentially throwing an error that bypasses StageError wrapping. It is essentially an error that would likely never get hit unless a dev was passing all the parameters as a CLI command and forgot the dbUri.

BoDonkey (dismissed):
Testing feels solid. Claude picked up one edge case that I commented on. There is also an edge case where if user:add hits a duplicate name and user:changePassword also fails, the error report will say that user:add failed.

myovchev:
> Claude picked up one edge case that I commented on.

This is unrealistic, as command arguments are validated before the flow is invoked. Interactive prompts are validating the dbUri on the fly.

 > There is also an edge case where if user:add hits a duplicate name and user:changePassword also fails, the error report will say that user:add failed.

It's a legit reporting issue, fixed.

BoDonkey (approved):

PR #5432, opened 2026-05-23
Fix raw-text sanitization bypass vulnerability and add regression tests

Dipanshusinghh:
"Hi @BoDonkey,

I have audited the codebase & successfully addressed the underlying vulnerability & parsing state risk associated with raw-text elements (CVE-2026-44990).

The Fix
I updated the raw-text handler conditions in packages/sanitize-html/index.js to check tagAllowed(tag). this ensures that if the raw-text element (like xmp, textarea, script, or style) is disallowed and discarded, its inner content is safely escaped rather than being appended as raw HTML.

I hve also added 6 new test cases to the test suite covering these regression paths & edge cases, & all tests are passing successfully.

Proof of Verification:-
Here is a screenshot of the PoC script execution and Mocha test results running on my local environment:
image

### DIFF
--- a/packages/sanitize-html/index.js
+++ b/packages/sanitize-html/index.js
@@ -566,13 +566,13 @@ function sanitizeHtml(html, options, _recursing) {
 
       if (options.disallowedTagsMode === 'completelyDiscard' && !tagAllowed(tag)) {
         text = '';
-      } else if ((options.disallowedTagsMode === 'discard' || options.disallowedTagsMode === 'completelyDiscard') && ((tag === 'script') || (tag === 'style'))) {
+      } else if (tag && tagAllowed(tag) && (options.disallowedTagsMode === 'discard' || options.disallowedTagsMode === 'completelyDiscard') && ((tag === 'script') || (tag === 'style'))) {
         // htmlparser2 gives us these as-is. Escaping them ruins the content. Allowing
         // script tags is, by definition, game over for XSS protection, so if that's
         // your concern, don't allow them. The same is essentially true for style tags
         // which have their own collection of XSS vectors.
         result += text;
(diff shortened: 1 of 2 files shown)


boutell (approved):
Thank you for reporting and fixing this!

Next time, please reach out to security@apostrophecms.com or report the vulnerability via:

https://github.com/apostrophecms/apostrophe/security/advisories

Rather than opening a public PR or issue.

This avoids premature public disclosure, which is the situation we are in now.

We will push out a hotfix to resolve this. In the future it would be better if we were able to do that in our normal release cycle. But, of course we appreciate the contribution!

boutell:
Looking at this more carefully, it appears there is no risk unless nonTextTags has been overridden to remove well-known tags of that kind, which is an unusual choice.

If that's the case then I think there is no urgency to release a hotfix. removing script from nonTextTags, for instance, would not be a sensible developer decision.

Let me know if you agree with my assessment.

Dipanshusinghh:
Thanks for the follow-up, @boutell 
You're right — in the default configuration, the practical risk is limited. Exploiting this would require nonTextTags to be explicitly overridden in an atypical way, which is not a common or recommended usage.

That said, the underlying branching logic does leave the raw-text passthrough active even when a tag has been discarded via tagAllowed(tag) returning false. My fix tightens that condition to ensure consistent, predictable behavior across all configurations — not just the default one.

I'm comfortable treating this as a robustness and consistency improvement rather than a critical hotfix, and the regression tests now cover parsing paths that weren't previously validated.

Happy to update the PR description to better reflect this scope if that would help.

boutell:
Thanks! I will add an acknowledgement to the changelog for our Wednesday release when this will come out.

PR #5433, opened 2026-05-25
PRO-9542: fix the bug that breaks sitemaps for RA

boutell:
Tested against AKA EWCS.

### DIFF
--- a/.changeset/sparkly-experts-walk.md
+++ b/.changeset/sparkly-experts-walk.md
@@ -1,0 +1,5 @@
+---
+"apostrophe": patch
+---
+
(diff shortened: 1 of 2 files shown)


myovchev (approved):
It's strange to me how `getUrlMetadata` can affect non-static build process. I'm probably missing something.

boutell:
The site map module uses it. If it belongs in a static site, it's
enumerable so it also belongs in a site map. Single source of truth.

The bug would also have prevented them from using our static site feature.


THOMAS BOUTELL | CHIEF EXECUTIVE OFFICER
APOSTROPHECMS | apostrophecms.com | he/him/his

PR #5441, opened 2026-06-04
Pro 9551 apos astro v2

BoDonkey:
*Check one*

- [X] main
- [X] latest
- [ ] stable

## Summary

This PR introduces a major version update to the apostrophe-astro package to make it compatible with Astro v6. It maintains BC with v5. It is also a significant refactoring of helpers to provide a more viable architecture for moving forward. It has been tested for SSR and static build generation with the latest versions of both v5 and v6. Closes PRO-9551.

*For example:*

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**


haroun:
@BoDonkey Any plans to update undici? Asking because undici 6 is supported for Node.js 18, 20, 21, 22 only. No support for 24 nor 26. It works, but Node.js 24 ships with undici 7.x by default.

Undici Version | Bundled in Node.js | Node.js Versions Supported | End of Life
-- | -- | -- | --
5.x | 18.x | ≥14.0 (tested: 14, 16, 18) | 2024-04-30
6.x | 20.x, 22.x | ≥18.17 (tested: 18, 20, 21, 22) | 2027-04-30
7.x | 24.x | ≥20.18.1 (tested: 20, 22, 24) | 2028-04-30
8.x | 26.x | ≥22.19.0 (tested: 22, 24, 26) | 2029-04-30

source: https://undici.nodejs.org/#/?id=long-term-support

> Astro 6 also now requires Node 22 or later, dropping support for Node 18 and Node 20, which have reached or are approaching end-of-life. Node 22 is faster, more secure, and lets us drop polyfills for older Node versions — resulting in a smaller, more maintainable package and better performance in Astro across the board.

source: https://astro.build/blog/astro-6/

> Vite now requires Node.js 20.19+, 22.12+. We have dropped Node.js 18, now that it has reached its [EOL](https://endoflife.date/nodejs) at the end of April 2025.

source: https://vite.dev/blog/announcing-vite7#node-js-support

BoDonkey:
> @BoDonkey Any plans to update undici? Asking because undici 6 is supported for Node.js 18, 20, 21, 22 only. No support for 24 nor 26. It works, but Node.js 24 ships with undici 7.x by default.
> 
> Undici Version	Bundled in Node.js	Node.js Versions Supported	End of Life
> 5.x	18.x	≥14.0 (tested: 14, 16, 18)	2024-04-30
> 6.x	20.x, 22.x	≥18.17 (tested: 18, 20, 21, 22)	2027-04-30
> 7.x	24.x	≥20.18.1 (tested: 20, 22, 24)	2028-04-30
> 8.x	26.x	≥22.19.0 (tested: 22, 24, 26)	2029-04-30
> source: https://undici.nodejs.org/#/?id=long-term-support
> 
> > Astro 6 also now requires Node 22 or later, dropping support for Node 18 and Node 20, which have reached or are approaching end-of-life. Node 22 is faster, more secure, and lets us drop polyfills for older Node versions — resulting in a smaller, more maintainable package and better performance in Astro across the board.
> 
> source: https://astro.build/blog/astro-6/
> 
> > Vite now requires Node.js 20.19+, 22.12+. We have dropped Node.js 18, now that it has reached its [EOL](https://endoflife.date/nodejs) at the end of April 2025.
> 
> source: https://vite.dev/blog/announcing-vite7#node-js-support

Thanks for this! We are also handling another issue between 24 and 25, so I will update this before the release.

packages/apostrophe-astro/helpers/server/fetch.js
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * @example
+ * ```astro
+ * ---
+ * import { aposPageFetch } from '@apostrophecms/apostrophe-astro/helpers/server';
+ * const aposData = await aposPageFetch(Astro.request);
+ * ---
+ * ```
+ */
+export async function aposPageFetch(req) {
```
myovchev: This should not be part of public documentation and should have explicit "for internal use only" here. 
I understand why it's refactored and why here - the slug.astro in projects probably uses it. However, it's meant to be used only there, in our entrypoint. `aposFetch` is the tool that can be exposed for developers. 

I'd argue that this refactoring makes no sense - we can only re-export it from `lib/` (it is indeed internal thing) - similar with what we do for `static`.

packages/apostrophe-astro/package.json
```diff
@@ -10,11 +10,47 @@
...
+    },
+    "./helpers/server": {
+      "types": "./types/helpers/server/index.d.ts",
+      "default": "./helpers/server/index.js"
+    },
+    "./helpers/universal": {
+      "types": "./types/helpers/universal/index.d.ts",
+      "default": "./helpers/universal/index.js"
+    },
+    "./helpers/client": {
```
myovchev: We probably don't need that (and the empty index.js and types). I think we can document that it's reserved for client side only (browser) helpers.

packages/apostrophe-astro/lib/static.js
```diff
@@ -2,12 +2,28 @@ import { writeFile, mkdir, readFile, readdir, rm } from 'node:fs/promises';
 import { dirname, join } from 'node:path';
 import { bgGreen, black, blue, dim, green, yellow, red, getTimeStat, timestamp } from './format.js';
 
-const CACHE_DIR = join(process.cwd(), 'node_modules', '.apostrophe-astro');
-const CONFIG_CACHE = join(CACHE_DIR, '_config.json');
-const ATTACHMENTS_CACHE = join(CACHE_DIR, '_attachments.json');
+let cacheDir = join(process.cwd(), 'node_modules', '.apostrophe-astro-static');
```
myovchev: A caution about process.cwd() here - this can break things if process points to the wrong path (not sure if possible though).

myovchev (changes requested):
There are some minor issues. Also I think the big library page request handler refactoring is unnecessary - it's not a helper, it's internal lib that we need to reach in our starter kit (or manual integrating projects would need as a bootstrap). Probably no need of risky refactoring when we can only re-export it.

We should rethink the major release bump. Currently we re-export a bunch of old exports for "BC reasons".
We could probably do that for the rest (marking them again as deprecated) so that we match the exact shape of the things before the helpers refactoring, assuming there are no conflicts. Then we can remove the export map in package.json and just publish it as a major bump - the other changes are internal implementation details.

packages/apostrophe-astro/helpers/server/fetch.js
```diff
@@ -1,12 +1,12 @@
-import config from 'virtual:apostrophe-config';
+import config from 'apostrophe-astro-config/config';
 import { getAposHost } from './url.js';
```
myovchev, line 2: `/helpers/server/static.js` re-exports the lib so that the project can import from `/helpers/server` , however the `helpers/server/fetch.js` now is reverted (as requested) but doesn't re-export the lib. This sends mixed signals - the project will import static from helpers but the standard page loading from `/lib`? I think `aposPageFetch()` should be also re-exported here.

packages/apostrophe-astro/helpers/index.js
```diff
@@ -1,3 +1,14 @@
...
+ * // Server-side helpers (Astro frontmatter, endpoints, prerendering)
+ * import { aposFetch, getAposHost, isStaticBuild, getAllStaticPaths } from '@apostrophecms/apostrophe-astro/helpers/server';
+ *
+ * // Universal helpers (server + client)
+ * import { slugify, getAttachmentUrl, aposSetQueryParameter } from '@apostrophecms/apostrophe-astro/helpers/universal';
+ * ```
+ */
+
+export { aposFetch, getAposHost, isStaticBuild, getAllStaticPaths, getAllUrlMetadata, getLocales } from './server/index.js';
+export { buildPageUrl, getFilterBaseUrl, aposSetQueryParameter, slugify, stylesElements, stylesAttributes, getFocalPoint, getAttachmentUrl, getAttachmentSrcset, getWidth, getHeight } from './universal/index.js';
```
myovchev, line 14: This now re-exports everything. I think it should only re-export what was exported previously.

myovchev (changes requested):
We still have some export inconsistencies (see the code comments). Additionally:
- For every legacy export, it should be made clear that this is not the right import path - with dev JSDoc and documentation, that also point to the proper import path. For example helpers/index.js should be never used. Same for /lib/xxx - we allow it for BC reasons. I'm not sure if this can be done with JSDoc and @deprecated (because it's just re-export) but at least the file can have a clear comment that this should never be directly imported from that location. 
- Our starter kits should always showcase the right import paths.

packages/apostrophe-astro/MIGRATION.md
```diff
@@ -0,0 +1,62 @@
...
+
+// After
+import { getAllStaticPaths } from '@apostrophecms/apostrophe-astro/helpers/server';
+```
+
+---
+
+## Removed: Vite virtual modules
+
+`virtual:apostrophe-config` and `virtual:apostrophe-doctypes` were private implementation details and are no longer available. If you were importing either of these directly, remove those imports — there is no public replacement, as they were never part of the supported API.
```
myovchev: Replace them with imports and example? Or this is internal detail that should be not exposed?

packages/apostrophe-astro/vite/vite-plugin-apostrophe-generated-config.js
```diff
@@ -0,0 +1,226 @@
...
+  // Serialise the resolved integration config as a static ES module export.
+  // All keys are always present so consumer code can rely on their shape.
+  const configObj = {
+    aposHost: integrationConfig.aposHost ?? null,
+    aposPrefix: integrationConfig.aposPrefix ?? '',
+    includeResponseHeaders: integrationConfig.includeResponseHeaders ?? null,
+    excludeRequestHeaders: integrationConfig.excludeRequestHeaders ?? null,
+    viewTransitionWorkaround: integrationConfig.viewTransitionWorkaround ?? false,
+    staticBuild: integrationConfig.staticBuild ?? null
+  };
```
myovchev: What bothers me here is we are constructing the "integration config" twice in two different places. The other place is apostrophe-astro/index.js on line 165. This means, when modifing the configuration (e.g. adding new prop or removing existing) we need to touch 2 files, doing the same change. This is usually a sign of bad architecture. I'd say, trust the input here, ensure that the integrationConfig object is properly sent (same default values logic). Or if absolutely needed (some logic I don't see ATM) - refactor through shared function or something like that. The goal is to have a single source of truth - modifying in one place.

packages/apostrophe-astro/package.json
```diff
@@ -10,11 +10,24 @@
...
+  },
   "author": "Apostrophe Technologies",
   "license": "MIT",
   "dependencies": {
     "lodash.deburr": "^4.1.0",
     "sluggo": "^1.0.0",
     "undici": "^8.5.0"
+  },
+  "peerDependencies": {
+    "astro": ">=5.0.0 <8.0.0"
```
myovchev, line 26: Should probably allow 8?

myovchev (dismissed):
A small change that is more code quallity than anything else. Some questions as well. 
I love the dotpath and "fake node module" trick that fixes the problem "but Vite is not in play here". 
A very clean rewrite.

packages/apostrophe-astro/package.json
BoDonkey: We have no idea what breaking changes 8 will bring. We have tested on 5, 6, and 7.

packages/apostrophe-astro/MIGRATION.md
BoDonkey: They are internal and not exposed, but I added alternatives based on how I'm guessing that a dev might be trying to use them.

packages/apostrophe-astro/vite/vite-plugin-apostrophe-generated-config.js
BoDonkey: Good catch. The `??` defaults in the plugin were defensive coding left over from an earlier iteration — index.js is the only call site and already guarantees every key has a concrete value, so the re-defaulting in the plugin is dead code. Simplifying to pass integrationConfig through directly and removing the redundant configObj construction.

myovchev (approved):

PR #5447, opened 2026-06-07
fix: treat col as a self-closing tag

### LINKED ISSUE #5443: sanitize-html: col element is self-closing
PLEASE NOTE: make sure the bug exists in the latest patch level of the project. For instance, if you are running a 2.x version of Apostrophe, you should use the latest in that major version to confirm the bug.

## To Reproduce

1. Go to https://codesandbox.io/p/devbox/l6hczn
2. Click the "Start" item in the lower right corner
3. Note the outputs: sanitize-html turns the valid self-closing `<col>` into `<col></col>`, causing the sanitized HTML to fail validation.

## Expected behavior

sanitize-html creates valid HTML under its default settings.

## Describe the bug

sanitize-html accepts `<col>` tags in its default configuration, but it fails to mark `<col>` as void elements (self-closing). As a result, it turns them into `<col></col>`, which is not valid HTML.

## Details

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/col

**Version of Node.js:**
*PLEASE NOTE: Only stable LTS versions (20.x, 22.x, 24.x as of 2026-02-24) are fully supported but we will do our best with newer versions. Avoid odd-numbered versions, which are not considered stable.*

**Server Operating System:**
*The server (which might be your dev laptop) on which Apostrophe is running. Linux? MacOS X? Windows? Is Docker involved?*

**Additional context:**
> Add any other context about the problem here. If the problem is specific to a browser, OS or mobile device, specify which.

**Screenshots**
*If applicable, add screenshots to help explain your problem.*

vansh1011:
## Summary

Fixes #5443 by adding `col` to the `selfClosing` tag list.

`<col>` is a void HTML element and should not be serialized with a closing tag. Previously, sanitize-html could output invalid HTML by converting a `<col>` element into `<col></col>`.

### Changes

* Added `col` to the default `selfClosing` tag list.
* Added a regression test covering this behavior.

### Testing

* Added a test that reproduces the issue.
* Verified the test passes after the fix.
* Ran the sanitize-html test suite.

### DIFF
--- a/.changeset/sixty-hats-kneel.md
+++ b/.changeset/sixty-hats-kneel.md
@@ -1,0 +1,5 @@
+---
+"apostrophe": patch
+---
+
+Fix invalid HTML output for <col> elements in sanitize-html (treat void elements correctly)
--- a/packages/sanitize-html/index.js
+++ b/packages/sanitize-html/index.js
@@ -953,7 +953,7 @@ sanitizeHtml.defaults = {
     'alt'
   ],
   // Lots of these won't come up by default because we don't allow them
-  selfClosing: [ 'img', 'br', 'hr', 'area', 'base', 'basefont', 'input', 'link', 'meta' ],
+  selfClosing: [ 'img', 'br', 'hr', 'area', 'base', 'basefont', 'input', 'link', 'meta', 'col' ],
   // URL schemes we permit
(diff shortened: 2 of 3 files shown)


BoDonkey (approved):
Looks good to me. I didn't realize that sanitize added a trailing slash to self-closing tags, so it caused me to look at the test twice.

PR #5456, opened 2026-06-09
remove absent options

boutell:


BoDonkey (approved):
Miro pointed out that there are some new flags coming from the create-apostrophe work, but I don't think they are crucial right now. Especially if we plan to bring the --starter flag back.

boutell:
I agree.

We don't have to bring `--starter` back exactly as-is, but we should address making `apos create` useful to experienced devs. If they need to put some kind of JSON file in their repo to more robustly answer questions about what it is and what it can do, we can define that.

PR #5475, opened 2026-06-12
Add explicit test for inline configuration collapsing

myovchev:
Explicit tests related with PRO-9564.

### DIFF
--- a/packages/apostrophe/test/config-cascade-merge.js
+++ b/packages/apostrophe/test/config-cascade-merge.js
@@ -1,0 +1,57 @@
+const t = require('../test-lib/test.js');
+const assert = require('assert');
+
+// PRO-9564
+// Documents established moog-require behavior (see lib/moog-require.js): when a
+// module is configured BOTH via the app.js config object AND a project-level
+// index.js, app.js wins for any section declared in both. Only `options` gets a
+// further merge so index.js can fill gaps; cascade sections like `fields` do not
+// merge — the app.js declaration replaces the project-level one.
+//
+// This is intentional and long-standing; changing it is a major BC break. It is
+// the reason multisite enhances @apostrophecms/user through the module chain (a
+// bundled improvement) rather than inline config, so it composes with a
+// project's dashboard/modules/@apostrophecms/user/index.js instead of clobbering
+// it. This test guards the contract so it is not "fixed" by accident.
+//
+// The project-level fixture lives in test/modules/collision-piece/index.js.
+
+describe('config-object vs project-level cascade merge (moog-require contract)', function() {
+  this.timeout(t.timeout);
+  let apos;
+  after(async () => {
+    await t.destroy(apos);
+  });
+
+  it('app.js `fields` wins over the project-level file when both declare it', async function() {
+    apos = await t.create({
+      root: module,
+      modules: {
+        'collision-piece': {
+          extend: '@apostrophecms/piece-type',
+          options: { alias: 'collisionPiece' },
+          fields: {
+            add: {
+              configField: {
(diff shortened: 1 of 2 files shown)


BoDonkey (dismissed):

packages/apostrophe/test/config-cascade-merge.js
```diff
@@ -0,0 +1,54 @@
+const t = require('../test-lib/test.js');
+const assert = require('assert');
+
+// PRO-9564
+// Documents established moog-require behavior (see lib/moog-require.js): when a
+// module is configured BOTH via the app.js config object AND a project-level
+// index.js, app.js wins for any section declared in both. Only `options` gets a
+// further merge so index.js can fill gaps; cascade sections like `fields` do not
```
BoDonkey: I think I've had this logic backward in my head. I also thought fields would merge like options.

BoDonkey (dismissed):

boutell (changes requested):
Bob, you're not crazy!

Miro I don't know what you mean by a "further merge." Cascade sections of course do merge. Options also merge. However keep in mind that if you override an option that's an object or string there is no "deep merge," you're replacing that entire option... and exactly the same thing happens if you specify an field twice. It is completely overridden.

This is why, in some cases, we have hacks on a per-option basis like overriding `minimumAllowlist` versus `allowlist`. There have been various proposals for a way to opt into deep merge, etc., but none have ever been implemented.

So I think this comment needs clarification before it can be included in the code.

boutell:
Updated comment, sorry:

Bob, you're not crazy!

Miro I don't know what you mean by a "further merge." Cascade sections of course do merge. Options also merge. However keep in mind that if you override an option that's an object or string there is no "deep merge," you're replacing that entire option... and exactly the same thing happens if you specify a field twice. It is completely overridden.

This is why, in some cases, we have hacks on a per-option basis like overriding `minimumAllowlist` versus `allowlist`. There have been various proposals for a way to opt into deep merge, etc., but none have ever been implemented.

So I think this comment needs clarification before it can be included in the code.

myovchev:
@BoDonkey @boutell I'm rephrasing it to make it more clear, but I don't think we are on the same page here. In fact I think you both are missing the point so I'll try to explain why I need this set in stone as explicit contract (test) and what exactly it represents. The perfect way to give you the context is to use the Tom's words, written 5 years ago, showcasing the designed intent (that is now tested so this never changes - I'm doing it because I had a thought about changing it, which could lead to devistating results)

```js
    _.defaults(definition, projectLevelDefinition);
    // app.js wins if a section is declared twice. For most sections
    // there is no further merge algorithm between app.js and index.js.
    // However if an option exists in index.js and not in app.js, that
    // should be honored.
    if (definition.options && projectLevelDefinition.options) {
      _.defaults(definition.options, projectLevelDefinition.options);
    }
```

- We are talking about "configuration collision" - outside of the merge chain of cascades/options. The above is part of the moog's `define` logic. 
- The above happens BEFORE we enter the extend/improve chain. It resolves configuration from 2 different sources - inline (app.js) and the owner index.js
- The logic is simple - fill the gaps of the ENTIRE configuration object, second pass - fill the gaps of only config.options (that's why different named options survive the configuration conflict)
- The `fields` in app.js "wins" unconditionally and the index.js fields is ignored forever - it never appears in the cascade chain.
- everything in the inheritance chain works as expected after this point - i.e. fields are merged (at a top level only, shallow merge)

I hope this explains the case better (and it's better worded as a dev comment now)

boutell (approved):
Makes sense now. Thanks Miro.

PR #5476, opened 2026-06-15
Literal content automatic registration

myovchev:
*Check one*

- [x] main
- [x] latest
- [ ] stable

## Summary

Allow literal content routes like `/robots.txt` to be exposed by their respective owning modules and auto-registered and handled in Apostrophe Astro. 

URLs like `/sitemap.xml`, `/robots.txt` are properly resolved (when their respective modules are enabled in the backend) without any project level configuration. 

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/packages/apostrophe-astro/lib/aposLiteralContentMiddleware.js
+++ b/packages/apostrophe-astro/lib/aposLiteralContentMiddleware.js
@@ -1,0 +1,100 @@
+// Astro SSR middleware that routes module-declared "literal content" URLs
+// (robots.txt, sitemap.xml, llms.txt, …) straight to Apostrophe via the raw
+// proxy, bypassing the catch-all page renderer - which would `JSON.parse` the
+// non-HTML body and 500.
+//
+// This is the SSR counterpart to the static build's literal-content handling.
+// It is additive: the config-time `proxyRoutes` option is unaffected and keeps
+// working alongside this. Because the decision is made per request, the route
+// list can come from the running backend (unknown at config/startup time).
+//
+// The pattern list is the manifest served by the backend at
+// `/api/v1/@apostrophecms/url/literal-routes` (the
+// `@apostrophecms/url:getLiteralContentRoutes` event). It is fetched lazily on
+// the first request and cached with a TTL, so a newly declared route appears
+// without a frontend restart.
+import { defineMiddleware } from 'astro:middleware';
+import config from 'virtual:apostrophe-config';
+import aposResponse from './aposResponse.js';
+
+// Cache the compiled manifest for this long before refetching.
+// TODO: expose as a config option.
+const TTL_MS = 5 * 60 * 1000;
+// Shorter retry window when a fetch fails, so a briefly-unavailable backend
+// is not hammered on every request.
+const ERROR_TTL_MS = 15 * 1000;
+
+const EXTERNAL_FRONT_KEY = process.env.APOS_EXTERNAL_FRONT_KEY;
+
+let cache = null; // { matchers, expires }
+let inflight = null;
+
+// Compile a prefix-free path pattern to a RegExp.
+// `*` matches within a path segment; `**` matches across segments.
+// A trailing slash is tolerated.
+function toRegExp(pattern) {
+  const source = pattern
+    .replace(/[.+?^${}()|[\]\\]/g, '\\$&') // escape regex metachars
+    .replace(/\*\*/g, ' ') // placeholder for **
+    .replace(/\*/g, '[^/]*') // * stays within a segment
+    .replace(/ /g, '.*'); // ** spans segments
+  return new RegExp(`^${source}/?$`);
+}
+
+async function fetchManifest() {
+  const url = new URL(
+    (config.aposPrefix || '') + '/api/v1/@apostrophecms/url/literal-routes',
+    config.aposHost
+  );
+  const res = await fetch(url, {
+    headers: {
+      'x-requested-with': 'AposExternalFront',
+      'apos-external-front-key': EXTERNAL_FRONT_KEY
+    }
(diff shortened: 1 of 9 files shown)


packages/sitemap/index.js
```diff
@@ -115,6 +115,12 @@ module.exports = {
             });
           }
         }
+      },
+      '@apostrophecms/url:getLiteralContentRoutes': {
+        addSitemapRoutes(req, patterns) {
+          patterns.push('/sitemap.xml');
+          patterns.push('/sitemaps/*');
```
BoDonkey, line 122: Should this be guardrailed so it is only exposed when `perLocale` is `true`?

BoDonkey (dismissed):

packages/apostrophe-astro/lib/aposLiteralContentMiddleware.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+  if (!res.ok) {
+    throw new Error(`literal-routes manifest fetch failed (${res.status})`);
+  }
+  const { patterns } = await res.json();
+  return (patterns || []).map(toRegExp);
+}
+
+function getMatchers() {
+  if (cache && cache.expires > Date.now()) {
```
boutell: We don't have a cache for anything else and sitemaps etc. could potentially change faster than you think. I do not think we should introduce a cache for this one thing (or in general). It is better to fetch the manifest once at startup. Best practice is to deploy the apps together if anything changes and our own hosting in fact runs them in the same container.

packages/apostrophe/modules/@apostrophecms/url/index.js
```diff
@@ -94,6 +113,41 @@ module.exports = {
         return !!req.aposExternalFront;
       },
 
+      // Returns the list of "literal content" route patterns declared by
```
boutell, line 116: I think we could greatly reduce the amount of new code needed if we instead supported calling `GET /api/v1/@apostrophecms/url?literal=1`, which would return only metadata entries with both `sitemap: false` and `contentType` properties. We already have, or should have, the plumbing for this for all of the same URLs.

boutell (changes requested):
I understand the reasoning now. I agree it should be declarative. However I think there is a way to eliminate the double maintenance burden, and I also think the cache is not necessary. See comments.

packages/apostrophe-astro/lib/aposLiteralContentMiddleware.js
boutell: I am not 1000% sure this cache is merely for the metadata (in which case, we don't need it because it only changes on deployments and the two should be deployed together), or also for individual literal URL requests (in which case, we shouldn't have it because they can change and while people change them infrequently, they expect results when they do change them).

boutell:
There is a problem with my suggestion to prevent double maintenance: the full list of static URLs would still get generated, expensively, even if only some are returned with the query parameter. And some of the routes do involve wildcards, and the list of sitemaps for instance can indeed change without a redeploy.

So I think you are right to introduce this, and you are also right to require that we separately handle it for dynamic and static situations. Fortunately there aren't all that many of these URLs.

But I still don't think we need the caching. If the caching is for the list of literal routes itself, that should only change on a redeploy, and best practice is to deploy the apps together, and having to wait 4 hours for it to "fix itself" would still feel broken anyway.

If it is not just for the list of literal routes itself then it is also unnecessary because these files are not end-user-oriented and they will still be served quite fast if proxied, also people do expect to be able to review these files manually right after they make a change.

packages/apostrophe/modules/@apostrophecms/url/index.js
myovchev: I went through that idea, not worth. They are different format and most importantly - different data. Literal content we need in this case is "possible routes, wildcards supported", while originally during static build we want ALL URLs, not routes. 
I'd say there is major incompatibility and friction if they go through one and the same route. The new code delivers unconditional routes. 
Again, I went through this direction already and it's a dead end.

myovchev:
@boutell The cache invalidation is gone. However, I went a mile further about backing off in case (somehow) the project doesn't have the recent enough apostrophe version - this is one of these rare cases when module is tied to a specific core feature.

packages/apostrophe-astro/lib/aposLiteralContentMiddleware.js
```diff
@@ -60,21 +68,32 @@ async function fetchManifest() {
...
 
 function getMatchers() {
-  if (cache && cache.expires > Date.now()) {
-    return Promise.resolve(cache.matchers);
+  if (cached) {
+    return Promise.resolve(cached);
   }
   if (!inflight) {
     inflight = fetchManifest()
       .then((matchers) => {
```
boutell, line 76: Why is this not written as an async function, why the explicit chaining?

boutell (changes requested):
I think it's the right logic now, one concern about maintainability: why not an async function?

packages/apostrophe-astro/lib/aposLiteralContentMiddleware.js
myovchev: It's the more clean way of doing it. The whole reason is the inflight guard. If you look closely you'll see that the middleware will receive one and the same "inflight" promise which is exactly ONE request to our backend. The cached value will be set exactly once no matter how many times the function is executed (same promise). getMatcher is the sync control of this guard.

I tried to recreate this with same sync getMatcher but async anonymous function (self executing) for inflight and it's much more messy code.

boutell (approved):

PR #5478, opened 2026-06-15
Fix AposArea crash when area prop is undefined

### LINKED ISSUE #5477: apostrophe-astro: AposArea crashes when area prop is undefined
## Summary

`AposArea.astro` throws when the `area` prop is `undefined`, even though other parts of the same component already treat a missing area as empty.

## Package

`packages/apostrophe-astro` (`@apostrophecms/apostrophe-astro`)

## Error

```
TypeError: Cannot read properties of undefined (reading 'options')
    at AposArea.astro:46:45
```

## Root cause

In `components/AposArea.astro`, the component partially guards against a missing `area`:

```js
const widgets = area?.items || [];
```

…but later unconditionally reads `area.options`:

```js
const widgetOptions = getWidgetOptions(area.options);
```

When an optional CMS area field is unset, Apostrophe often omits it from the API response entirely (`undefined`), which is expected backend behavior. Passing that through to `<AposArea area={piece.footerArea} />` then crashes during SSR.

## Reproduction

1. Define an optional `area` field on a piece type (not required, may be empty/unset).
2. Create or use a piece/document where that area was never populated.
3. Render `<AposArea area={doc.optionalArea} />` in an Astro template.
4. Request the page in public (non-edit) mode.

## Expected behavior

`AposArea` should no-op gracefully when `area` is missing in public mode (same as when `area.items` is empty): render nothing and do not throw.

## Suggested fix

Use the same defensive pattern already used for `items`:

```js
const widgetOptions = getWidgetOptions(area?.options ?? {});
```

Optionally short-circuit entirely when `!area && !isEdit`, if edit mode should remain strict.

## Context

We hit this on an extension detail page where `footerArea` is optional and unset for some extensions. The application-level fix is to conditionally render the footer section, but the component inconsistency still looks like a bug in `AposArea` itself.

## Environment

- `@apostrophecms/apostrophe-astro`: 1.12.0 (also present in ^1.4.0 dependency range)
- Astro: 5.x
- Node: current LTS

Dipanshusinghh:
Closes #5477
AposArea.astro already handles a missing area when accessing items, but later accesses area.options without a guard. This can cause a runtime error when an optional area field is undefined.
This change adds a safe fallback when reading area options so that undefined areas are handled consistently.
What are the specific steps to test this change?
Use an optional area field that has not been populated.
Render <AposArea area={doc.optionalArea} />.
Verify that the page renders without throwing an error.
Verify that populated areas continue to render as expected.
What kind of change does this PR introduce?
Bug fix
Make sure the PR fulfills these requirements:

It includes a clear description of the bug it resolves

The changelog is updated

Related documentation has been updated

Related tests have been updated


packages/apostrophe-astro/components/AposArea.astro
```diff
@@ -56,64 +43,39 @@ if (isEdit) {
   };
 }
 const Wrapper = isEdit || forceWrapper ? "div" : Fragment;
-const widgetOptions = getWidgetOptions(area?.options);
+const widgetOptions = getWidgetOptions(area?.options ?? {});
```
myovchev: No need of ` ?? {}` - `getWidgetOptions` already handles that. 
Also this is already fixed in `main`.

myovchev:
I'm confused. This PR is trying harden the Astro area options to avoid missing area crashing the Astro app but in the same time it fixes "something" in the sanitize HTML land?

myovchev (changes requested):
The problem is already fixed in `main` (stale fork probably). 
Please revert the Astro area related change, rework the PR body/title to give us an idea what is really fixed in the sanitize HTML package.

packages/apostrophe-astro/components/AposArea.astro
Dipanshusinghh: Thanks for the review, @myovchev!
You're right — I will revert the AposArea.astro change since it's already fixed in main (my fork was stale).
The actual fix in this PR is for a raw-text sanitization bypass vulnerability in the sanitize HTML layer. I'll update the PR title and description to reflect that clearly, and keep only the relevant commits.
Will push the updated changes shortly.

myovchev:
Nothing to review after syncing with main.

PR #5482, opened 2026-06-18
fix: remove potential XSS vector in anchors widget

MatrixNeoKozak:
The anchors widget output method was directly interpolating `widget.anchorId` into an HTML attribute without sanitization. If a user was able to control the `anchorId` value, they could inject arbitrary HTML attributes, leading to a cross-site scripting (XSS) vulnerability. The fix escapes the `widget.anchorId` value before inserting it into the `div` element, ensuring it is treated as a plain string. Additionally, the attribute name is now checked against a whitelist to prevent injection into the attribute name itself.

### DIFF
--- a/packages/anchors/modules/@apostrophecms/anchors-widget-type/index.js
+++ b/packages/anchors/modules/@apostrophecms/anchors-widget-type/index.js
@@ -24,8 +24,15 @@ module.exports = {
 
         if (widget.anchorId && self.options.anchors !== false) {
           // Apply a wrapper div with the anchor attribute.
-          const attr = self.options.anchorAttribute || 'id';
-          const opener = `<div ${attr}=${widget.anchorId}>`;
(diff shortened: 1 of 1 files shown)


boutell:
Module options are set by developers. They are not user input. Therefore no vulnerability and no reason to restrict this.

But if you think I'm mistaken please clarify how this can come from user input.

boutell:
Also, if reporting a publicly undisclosed vulnerability please follow responsible disclosure by using github's advisory disclosure form:

https://github.com/apostrophecms/apostrophe/security/advisories

Or writing to security@apostrophecms.com.

Thanks!

boutell:
I missed that you are also escaping anchorId.

However, anchorId is of `type: 'slug'` which is already sanitized through Apostrophe's slugification mechanisms. These don't allow any dangerous markup.

Again, if you find and demonstrate a specific counterexample, please feel free to reopen.

PR #5496, opened 2026-06-29
allow titlefield to be set to false

stuartromanek:
Arrays default back to using `title` as the default `titleField` with no way to opt out. My `title` field was an object and using it as a string was obviously breaking UI

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/schema/ui/apos/logic/AposInputArray.js
+++ b/packages/apostrophe/modules/@apostrophecms/schema/ui/apos/logic/AposInputArray.js
@@ -399,7 +399,9 @@ export default {
       return newInstance(this.schema);
     },
     getLabel(id, index) {
-      const titleField = this.field.titleField || 'title';
+      const titleField = this.field.titleField === false
+        ? null
+        : (this.field.titleField || 'title');
       const item = this.items.find(item => item._id === id);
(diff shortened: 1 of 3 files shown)


BoDonkey (commented):
Never mind - edited because I read the note up top.

BoDonkey:
Does the numbered index get automatically added to the UI?

boutell (changes requested):
I'd classify it as "minor" but no problem with adding the feature. Needs to make it to documentation as well

packages/apostrophe/modules/@apostrophecms/schema/ui/apos/logic/AposInputArray.js
```diff
@@ -399,7 +399,9 @@ export default {
       return newInstance(this.schema);
     },
     getLabel(id, index) {
-      const titleField = this.field.titleField || 'title';
+      const titleField = this.field.titleField === false
+        ? null
+        : (this.field.titleField || 'title');
       const item = this.items.find(item => item._id === id);
       return get(item.schemaInput.data, titleField) || `Item ${index + 1}`;
```
myovchev, line 406: Be sure that `get(data, null)` doesn't panic.

myovchev (approved):
A nice tiny feature. I agree it's a minor release (patch is a bug fix usually).

BoDonkey (approved):

PR #5503, opened 2026-07-08
PRO-9688: prevent crash due to double defineProperty, but I have deja vu

boutell:
This seems to be why those asset builds fail in the deploy, Claude sorted it out quickly from the error

But I recognized it

Didn't we do this before?


myovchev:
I remember that. I caught test issues and had the choice to patch the module or ensure single instance of the package. I chose the latter because I didn't have enough context on the former.

myovchev (approved):

PR #5506, opened 2026-07-09
fix(schema): launder email field value before validating and storing

spokodev:
## Bug

The `email` schema field's `convert` launders the submitted value into `destination[field.name]`, but then validates the **raw** `data[field.name]` and overwrites the destination with that raw value, discarding the laundered result.

Two consequences:

1. A valid address with surrounding whitespace is rejected. The validation regex is anchored (`^…$`) and excludes whitespace, so `"  test@example.com  "` fails to match and throws `invalid` — even though the first line already computed the trimmed value that was meant to be stored.
2. Non-string input (e.g. a number sent through the REST API) throws an uncaught `TypeError` (`data[field.name].match is not a function`) instead of failing validation gracefully. Every other string-based field type launders first and so coerces such input.

## Repro

```js
const schema = apos.schema.compose({
  addFields: [ { type: 'email', name: 'authorEmail', label: 'Author Email' } ]
});
const result = {};
await apos.schema.convert(apos.task.getReq(), schema, { authorEmail: '  test@example.com  ' }, result);
// expected: result.authorEmail === 'test@example.com'
// actual:   throws [ 'invalid: invalid' ]
```

## Fix

Validate and store `destination[field.name]` (the laundered value) and drop the redundant final overwrite. This trims surrounding whitespace before validating, matches the behavior of the other string-based field types, and safely coerces non-string input via `launder.string`.

Adds a regression test in `test/schemas.js` and a changeset.

### DIFF
--- a/.changeset/email-field-trim-whitespace.md
+++ b/.changeset/email-field-trim-whitespace.md
@@ -1,0 +1,5 @@
+---
+"apostrophe": patch
+---
+
+The `email` schema field now validates and stores the laundered value, so surrounding whitespace is trimmed rather than causing a valid address to be rejected as invalid. Non-string input (for example a number sent through the REST API) is now coerced by the launder step instead of throwing an uncaught error.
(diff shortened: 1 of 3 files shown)


boutell:
Makes good sense. Thank you!

boutell (approved):

PR #5507, opened 2026-07-10
fix(rich-text-widget): match most specific style when tags overlap

### LINKED ISSUE #3622: Rich text editor style selection bug
basically if you have multiple style types on the same HTML tag, like

```
{
  tag: 'p',
  label: 'Paragraph (P)',
  def: true,
},
{
  tag: 'p',
  label: 'Small (P)',
  class: 'small',
},
```
When you select Small (P) in the style picker dropdown it'll change the style, but the picker will mark Paragraph (P) as selected, which means you can't undo it. You need to select a different tag entirely, like

```
{
  tag: 'h2',
  label: 'Heading 2 (H2)',
},
```
and THEN you can select the original Paragraph (P) to revert the style selection

**Posted this to the discord and Alex Bea requested a github issue

mmustafasenoglu:
## What does this PR do?

Fixes #3622 — the rich text style picker marks the wrong style as active when multiple styles share the same tag (e.g. a plain `<p>` and a `<p class="small">`).

## Why

The `active` computed in `AposTiptapStyles.vue` used `nodes.findIndex(...)`, which returns the **first** match in array order. So a `<p class="small">` matched the generic `<p>` style first, leaving "Paragraph" highlighted instead of "Small" — making the style impossible to undo without first selecting a different tag.

Following maintainer guidance in the issue, styles are now matched in **descending order of specificity** (number of classes). The most specific style wins the first match, then it is resolved back to its original index so the dropdown (which keeps the original order) highlights the correct option.

## How to test

Configure a rich text widget with overlapping tags, e.g.:

```js
{
  tag: 'p',
  label: 'Paragraph (P)',
  def: true,
},
{
  tag: 'p',
  label: 'Small (P)',
  class: 'small',
},
```

Apply "Small (P)" — the picker now correctly shows "Small (P)" as the active style, and selecting "Paragraph (P)" reverts it as expected.

## Checklist

- [x] I have read the `CONTRIBUTING` document.
- [x] I have added a changeset (`.changeset/rich-text-style-selection-bug.md`).
- [x] My commit messages and PR title follow the repo conventions.

### DIFF
--- a/.changeset/rich-text-style-selection-bug.md
+++ b/.changeset/rich-text-style-selection-bug.md
@@ -1,0 +1,5 @@
+---
+"apostrophe": patch
+---
+
+Fixed the rich text style picker marking the wrong style as active when multiple styles share the same tag. Styles are now matched in descending order of specificity (number of classes), so a `<p class="small">` correctly shows "Small" as the active style instead of the plain "Paragraph".
--- a/packages/apostrophe/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposTiptapStyles.vue
+++ b/packages/apostrophe/modules/@apostrophecms/rich-text-widget/ui/apos/components/AposTiptapStyles.vue
@@ -118,11 +118,24 @@ export default {
           if (activeEls[0].name === 'defaultNode') {
             return 1;
           } else {
-            const match = nodes.findIndex(node =>
+            // Match the most specific style first. When several styles
+            // share the same tag (e.g. a plain `<p>` and a `<p class="small">`)
+            // a `<p class="small">` would otherwise match the generic `<p>`
+            // style, marking the wrong option as active. We sort a copy of
+            // the nodes by descending class count and pick the first match,
+            // then resolve it back to its original index so the dropdown
+            // (which keeps the original order) highlights the correct option.
+            const sortedNodes = [ ...nodes ].sort((a, b) => {
+              const aCount = a.class ? a.class.trim().split(/\s+/).length : 0;
+              const bCount = b.class ? b.class.trim().split(/\s+/).length : 0;
+              return bCount - aCount;
+            });
+            const matchedNode = sortedNodes.find(node =>
               node.class === activeEls[0].class &&
               node.type === activeEls[0].name &&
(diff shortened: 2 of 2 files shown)


BoDonkey:
I don't think I'm able to reproduce the issue with the current version of Apostrophe. I think it was fixed in 4.20.0, but the related issue wasn't closed. I might be testing incorrectly, so if you can provide a test scenario where the bug replicates, I'd be happy to review your PR.

mmustafasenoglu:
Hi @BoDonkey, thanks for looking into this. The bug only manifests when you have **multiple styles sharing the same HTML tag** with different classes. Here is a minimal reproduction:

**1. Configure the rich text widget with overlapping styles:**

```js
// In your module that registers the rich-text-widget options:
styles: [
  { tag: 'p', label: 'Paragraph (P)', def: true },
  { tag: 'p', label: 'Small (P)', class: 'small' },
  { tag: 'h2', label: 'Heading 2 (H2)' },
]
```

**2. Steps to reproduce:**
- Open the rich text editor
- Type some text in a paragraph
- Open the style dropdown and select **"Small (P)"**
- The text gets the `small` class correctly
- **But the dropdown now shows "Paragraph (P)" as active** instead of "Small (P)"

**3. Why it happens:**
The `active` computed in `AposTiptapStyles.vue` uses `nodes.findIndex(...)`, which returns the first match in array order. When checking `class === "small"`, it hits the plain `<p>` (class: null) first which does not match, but the matching logic iterates in array order (least specific first). With `def: true` styles, the default paragraph appears first in the array and wins the race.

**The fix** (already in this PR) sorts styles by specificity (number of classes) so the most specific style matches first, then maps back to the original dropdown index. This follows boutell's recommended approach in #3622.

**To test in the Apostrophe admin UI:**
1. Go to an admin template
2. Open the template editor (or use the rich-text-widget playground)
3. Configure the styles as above
4. Apply "Small (P)" and check which style is highlighted in the dropdown

Let me know if you need any additional information!

mmustafasenoglu:
Closing due to inactivity. Feel free to reopen if needed.

PR #5518, opened 2026-07-11
fix(apostrophe-astro): fix a bug where redirect responses drop headers

eri-vi-maddyie:

*Check one*
- [x] main
- [x] latest
- [x] stable

## Summary

When ApostropheCMS is deployed as a sidecar with no external network exposure (i.e. `APOS_HOST` is pointed at Astro such that all traffic to ApostropheCMS is routed through `aposProxy.js`), some logins (e.g. Google OAuth SSO with `@apostrophecms/passport-bridge`) fail to persist session cookies after the callback redirect.

This is caused by `aposProxy.js` rebuilding redirect responses using Astro's `redirect()` function which is hard-coded to drop all headers except for `Location` - e.g. dropping headers like `Set-Cookie` from the auth callback's redirect response.

**For example (Google auth):**

```mermaid
sequenceDiagram
    autonumber
    participant Browser
    participant Astro as Astro proxy (aposProxy.js)
    participant Apos as Apostrophe (passport-bridge)
    participant Google

    Browser->>Astro: GET /auth/google/login
    Astro->>Apos: proxy request
    Apos-->>Astro: 302 Location: accounts.google.com/...
    Astro-->>Browser: 302 (forwarded)

    Browser->>Google: GET /o/oauth2/auth
    Google-->>Browser: 302 Location: /auth/google/callback?code=...

    Browser->>Astro: GET /auth/google/callback?code=...
    Astro->>Apos: proxy request
    Apos-->>Astro: 302 Location: /<br/>Set-Cookie: apos.sid=...

    rect rgb(255, 220, 220)
    Note over Astro: BUG: aposProxy.js calls Astro's redirect()<br/>which drops the Set-Cookie header
    end
    Astro-->>Browser: 302 Location: / (no Set-Cookie)

```

### Fix

The `aposResponse()` function already builds complete working responses with all backend headers intact.

```diff
-export async function ALL({ params, request, redirect }) {
+export async function ALL({ request }) {
   try {
-    const response = await aposResponse(request);
-    if ([301, 302, 307, 308].includes(response.status)) {
-      return redirect(response.headers.get('location'), response.status);
-    }
-    return response;
+    return await aposResponse(request);
   } catch (e) {
     return new Response(e.message, { status: 500 });
   }
 };
```

This is consistent with existing behaviour: `aposResponse` forwards headers on backend responses.

### Cause

Traced back to [apostrophecms/apostrophe-astro@4cf1797](https://github.com/apostrophecms/apostrophe-astro/commit/4cf1797b1f26ffdf6fb08c145600ef0586337fa1).

Before that commit `aposProxy.js` always returned `aposResponse()` (no special case for redirects).

The bug this commit fixed was in `aposResponse.js` was real and fixed properly, but it also added a special case in `aposProxy.js` which excluded redirects from the fix.

It effectively bundled two changes:
1. Correctly building headers and passing statusCode, which fixed the described bug, and
2. An unnecessary branch in `aposProxy.js` rerouting redirects through Astro's `redirect()` causing the headers to drop.

NB: I have checked that my change does not regress the issue fixed by this change - i.e. visiting `/login` while already logged in.

## Follow up

When forwarding the response with `aposResponse.js` the response body may be dropped (on specific status, e.g. redirects) but the entity headers describing the body (e.g. content-length, content-encoding, transfer-encoding) were not removed .
A stale content-length may make strict clients (reverse proxies, non-browser HTTP clients) hang waiting for body bytes that will never arrive.

Fixed by removing the stale headers alongside removing the response body.

- Added `test/endpoints/aposProxy.test.js`. Redirect responses forward `Set-Cookie` and custom headers, non-redirect responses pass through unchanged.
-  Added a case to `test/lib/aposResponse.test.js`. For each of the impacted statuses 204/304/301/302/307/308, content-length, content-encoding, and transfer-encoding are stripped while set-cookie survives.
- Package test suite passes (`pnpm --filter @apostrophecms/apostrophe-astro run test`).
- Manual test: Google OAuth login with passport-bridge against a sidecar deployment (e.g. `APOS_HOST` set to the Astro origin).

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/.changeset/late-comics-chew.md
+++ b/.changeset/late-comics-chew.md
@@ -3,3 +3,4 @@
 ---
 
 Fixed a bug where redirect responses proxied through `aposProxy.js` (e.g. OAuth login callbacks) lost every header except `Location`. This affected setups where Apostrophe is only reachable through the Astro proxy (e.g. a sidecar deployment), causing the session `Set-Cookie` header from an OAuth callback redirect to be silently dropped and the login to appear to fail.
+Included fix to `aposResponse.js` to remove entity headers describing the body when dropping the body (e.g. redirects) which may otherwise cause strict clients to hang waiting for bytes.
--- a/.changeset/late-comics-chew.md
+++ b/.changeset/late-comics-chew.md
@@ -1,0 +1,5 @@
+---
+"@apostrophecms/apostrophe-astro": patch
+---
+
+Fixed a bug where redirect responses proxied through `aposProxy.js` (e.g. OAuth login callbacks) lost every header except `Location`. This affected setups where Apostrophe is only reachable through the Astro proxy (e.g. a sidecar deployment), causing the session `Set-Cookie` header from an OAuth callback redirect to be silently dropped and the login to appear to fail.
--- a/packages/apostrophe-astro/endpoints/aposProxy.js
+++ b/packages/apostrophe-astro/endpoints/aposProxy.js
@@ -1,12 +1,8 @@
-import aposResponse from "../lib/aposResponse";
+import aposResponse from "../lib/aposResponse.js";
 
-export async function ALL({ params, request, redirect }) {
+export async function ALL({ request }) {
   try {
-    const response = await aposResponse(request);
-    if ([301, 302, 307, 308].includes(response.status)) {
-      return redirect(response.headers.get('location'), response.status);
-    }
-    return response;
+    return await aposResponse(request);
   } catch (e) {
     return new Response(e.message, { status: 500 });
   }
--- a/packages/apostrophe-astro/lib/aposResponse.js
+++ b/packages/apostrophe-astro/lib/aposResponse.js
@@ -83,12 +83,16 @@ export default async function aposResponse(req) {
 
     const { headers, statusCode, ...rest } = res;
 
-    // Statuses whose body we never send to the client: 204/304 carry none,
-    // and redirects (301/302/307/308) become a fresh Astro redirect built
-    // from the Location header in aposProxy. Dump the undici body so its
-    // socket returns to the pool instead of being held open until GC.
+    // Statuses we forward with no body: 204/304 carry none, and for
+    // redirects (301/302/307/308) Location + Set-Cookie are all the client
+    // needs. Dump the undici body so its socket returns to the pool, and
+    // strip the entity headers that described it - a stale Content-Length
+    // would make the client wait for bytes that never arrive.
     if ([204, 304, 301, 302, 307, 308].includes(statusCode)) {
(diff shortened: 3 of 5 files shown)


myovchev:
Thanks for the detailed report. The code in question is not touched since the dawn of time (around the initial version of the Astro plugin). This proxy handles the "unknown" (not mapped to pages/API) routes. I can't agree more that we need a single source of truth for redirects - aposResponse does correct header manipulation. I need to manually validate and test the change first to ensure we are not missing something. 

The test failure is related with the CI configuration, not related with your changes. I'll look into that in a separate ticket.

myovchev (changes requested):
The change is correct and fixes an actual bug exactly as described. However, I found a side effect that may cause the redirect to hang in some cases - especially with clients that are stricter than browsers (e.g., reverse proxies). The reason for that is now all headers are passed through, including these describing the body (`content-length` and friends). 

To fight that, a change in `lib/aposResponse.js` is required. On line 86, you'll find: 
```js
    // Statuses whose body we never send to the client: 204/304 carry none,
    // and redirects (301/302/307/308) become a fresh Astro redirect built
    // from the Location header in aposProxy. Dump the undici body so its
    // socket returns to the pool instead of being held open until GC.
    if ([204, 304, 301, 302, 307, 308].includes(statusCode)) {
      await res.body.dump().catch(() => {});
      return new Response(null, { ...rest, status: statusCode, headers: responseHeaders });
    }
```

This block should become: 
```js
    // Statuses we forward with no body: 204/304 carry none, and for
    // redirects (301/302/307/308) Location + Set-Cookie are all the client
    // needs. Dump the undici body so its socket returns to the pool, and
    // strip the entity headers that described it - a stale Content-Length
    // would make the client wait for bytes that never arrive.
    if ([204, 304, 301, 302, 307, 308].includes(statusCode)) {
      await res.body.dump().catch(() => {});
      responseHeaders.delete('content-length');
      responseHeaders.delete('content-encoding');
      responseHeaders.delete('transfer-encoding');
      return new Response(null, { ...rest, status: statusCode, headers: responseHeaders });
    }
```

This also handles a stale dev comment related with the already removed code in  aposProxy.js. I single test, as addition to the already provided, should showcase that the headers in question are dropped.

eri-vi-maddyie:
Thanks! Fixed, added tests, and updated the changeset and PR description.

myovchev (approved):

myovchev:
Thanks for the contribution!

PR #5528, opened 2026-07-17
Remove stricting the engine for sanitize-html

alexander-schranz:

*Check one*
- [x] main
- [ ] latest
- [ ] stable

## Summary

Not all projects are `node` based / server side rendering and so forcing the node version should not be done for a package which also can be used in the browser. Or atleast only be done via a new Major release, else existing pipelines, fail out of the box.

```
npm ERR! engine Not compatible with your version of node/npm: sanitize-html@2.17.6
npm ERR! notsup Not compatible with your version of node/npm: sanitize-html@2.17.6
npm ERR! notsup Required: {"node":">=22.12.0"}
npm ERR! notsup Actual:   {"npm":"9.9.4","node":"v20.19.2
```

npm install via node 20, npm 8, ... with strict engine flag active.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/packages/sanitize-html/package.json
+++ b/packages/sanitize-html/package.json
@@ -25,9 +25,6 @@
   ],
   "author": "Apostrophe Technologies, Inc.",
   "license": "MIT",
-  "engines": {
-    "node": ">=22.12.0"
-  },
   "dependencies": {
     "deepmerge": "^4.2.2",
     "escape-string-regexp": "^4.0.0",


boutell:
You are using an unsupported, insecure version of Node.js that is past its end-of-life. You should not do that. We support only supported versions of Node.

boutell:
See: https://nodejs.org/en/about/eol

Note the list of vulnerabilities for each end-of-life version.

alexander-schranz:
Hello @boutell,

> You are using an unsupported, insecure version of Node.js that is past its end-of-life. You should not do that.

I understand what you want todo but keep in mind not all what using the package are `Node` applications. We are using just NPM to download the packages, our application and others is not node based and never will be so the Node version doesn't care here as not run in prod system.

The problem is more that you did add engines restriction during a minor or patch release which break several CIs using older Node / NPM version. NPM will not fallback to a older version for engines restrictions instead it will just start failing the install. So any application who can not yet update would require to update there `package.json` to older version.

As said best would be to add engines restricition in the NPM world only for Major releases this avoids that other CIs can just start crashing.

boutell:
I appreciate the inconvenience. But packages we depend on have moved on to ESM, because in properly updated releases, node 22 and up can import ESM with "require" (as long as there is no top level await in the ESM). And versions of node earlier than the one we're requiring are not supported anyway, so it is not a breaking change if following best practices.

Everyone needs to move on from unsupported versions of node, and that's the position we are sticking to.

PR #5530, opened 2026-07-20
AI generate

myovchev:
Implement `apos.ai.generate(req, promptOrOptions, options?)` for plain text and
multi-turn chat, end to end against one real provider (Anthropic), together
with all the cross-cutting plumbing the core owns once: retries, error
normalization, failure logging, prompt-cache policy, usage metadata, events,
and mock mode.


BoDonkey (approved):
Not sure about the exact models and token amounts chosen, but otherwise valid.

boutell:
Nice. I assume you're going to write one mocha test suite and iterate it across all of the adapters.

PR #5536, opened 2026-07-23
Image generation

myovchev:
Implement `apos.ai.generateImage(req, prompt, options)` on the two
image-capable standard providers (OpenAI, Google), covering both modes:
text → image, and image(s) + text → image (editing/variations, when `images`
sources are passed).

**Note:** code refactoring and full doc (comments) rewrite upcoming with the next change.

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/ai-adapter-anthropic/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/ai-adapter-anthropic/index.js
@@ -7,6 +7,15 @@
 // The transport is `apos.http`, no SDK. Projects can adjust the dialect
 // by extending this module and overriding its methods.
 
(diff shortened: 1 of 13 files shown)


BoDonkey (approved):
So, I'm not positive if ALL openai-compatible tools are going to have image capability - maybe that should be false by default?

boutell:
Yeah they won't all have it. But, we shouldn't make the default behavior of the most common thing (actual OpenAI) suffer as a result. More like we need the  ability to configure the adapter to not expose image capabilities.

myovchev:
@BoDonkey @boutell The `image: true` can be overridden by the adapter that a project configures (for another provider or just in-place override of that adapter). The logic is exactly what Tom assumes - on by default (yes, Open AI compat supports image), configuration can opt-out.

PR #5537, opened 2026-07-24
AI non-blocking generate (job)

myovchev:
`apos.ai.generateJob` brings **non-blocking generation**: it accepts everything `generate` accepts, returns `{ jobId, cancel }` immediately, and stores the exact blocking result on an **expiring job record**, with **`onMessage`/`onEnd` hooks** for callers. It rides **additive `@apostrophecms/job` extensions** (job owner with scoped routes, cancel flag and route, `cancelled` status, `error` payload, `startedAt`/`endedAt`, opt-in per-document TTL), supports **in-process and cross-process cancellation** that waits out the in-flight step and preserves partial results (`finishReason: 'cancel'`), publishes **progress notifications** (started / per-turn / ended, correlated by `jobId`, opt-out via `notify: false`), and introduces a first-class **notification `bus: true` flag** — a never-rendered message-bus carrier emitted in exactly one tab — retiring the hidden-notification hack.

> Note: The job tests change do not imply a BC break - it fixes old race that was exposed by the additional DB roundtrip that we made to support job canceling. It's internal behavior change bringing enough latency to expose the race more often.

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/ai/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/ai/index.js
@@ -30,7 +30,15 @@ module.exports = {
     retryBaseDelay: 1000,
     // Elapsed-time budget in milliseconds for one call including its
     // retry waits; a delay that would land past it stops the call
-    retryMaxElapsed: 60000
+    retryMaxElapsed: 60000,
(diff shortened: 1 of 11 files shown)


BoDonkey (approved):

boutell (approved):
Didn't read every line, but the direction feels good

PR #5538, opened 2026-07-26
Self-debouncing widget renders. The widget component now handles this on its own

boutell:
Additional work could be done to get rid of the nonfunctional async plumbing elsewhere that never really worked, but it makes sense to get this to Go7 as a beta.

It doesn't solve the underlying problem of a widget being very slow to render (I haven't reproduced their results locally yet) but it should greatly mitigate the pain and this is what we always intended so it's considered a bug fix.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.1.2/schema.json",
+  "changelog": "@changesets/cli/changelog",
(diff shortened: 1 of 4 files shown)


BoDonkey (dismissed):

boutell:
My manual testbed CI run passed fully except for one platform combo that cancelled itself, something at the github actions level went wrong there, nothing to do with the feature.

BoDonkey (approved):

PR #5541, opened 2026-07-28
Regenerate session when invalid

myovchev:

*Check one*
- [x] main
- [x] latest
- [x] stable

## Summary

Fixed a 500 error when logging in on a session that had just been invalidated by a password change or by a disabled account. Such sessions are now regenerated rather than destroyed, so the request keeps a usable session and the invalidated one is removed from the store before the response is sent.

Added a regression test that reproduces the 500 error without the fix.

This is a very old code (5+ years) and not a recent regression. The race comes on request basis (second request at the right time while the session is being deleted). The actual cause is consistently reproducible. 

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/packages/apostrophe/modules/@apostrophecms/login/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/login/index.js
@@ -1180,8 +1180,12 @@ module.exports = {
             req.user.loginInvalidBefore &&
             (!req.session.loginAt || (req.session.loginAt < req.user.loginInvalidBefore))
           ) {
-            req.session.destroy();
             delete req.user;
+            // Regenerate rather than destroy. `destroy` removes `req.session`
+            // right away and does not wait for the store, so later middleware
+            // and routes, such as a fresh login attempt, are left without a
+            // session, and the old one can still be loaded by the next request.
+            return req.session.regenerate(() => next());
           }
           return next();
         }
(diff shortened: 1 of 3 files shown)


BoDonkey (approved):

packages/apostrophe/modules/@apostrophecms/login/index.js
```diff
@@ -1180,8 +1180,12 @@ module.exports = {
             req.user.loginInvalidBefore &&
             (!req.session.loginAt || (req.session.loginAt < req.user.loginInvalidBefore))
           ) {
-            req.session.destroy();
```
boutell, line 1183: `req.session.destroy` also supports a callback. Are you sure you can't just use that to invoke `next` here? It is most commonly associated with logout, which is what we're doing here.

boutell (changes requested):

packages/apostrophe/modules/@apostrophecms/login/index.js
myovchev: What I saw in the original code is that `regenerate` does "delete" + "create". The main culprit here is 500 error (with right request timing) as a result of a session-less request. I'll research more to find the answer of this question.

boutell (approved):
Approving. I've verified regenerate makes a new session, and `req.session` straight up not existing is not really well supported or expected by apostrophe. An empty one is fine.

PR #5543, opened 2026-08-03
CON-47: field ids should not depend on md5, makes deployments too sensitive

boutell:
To replace md5 we need a sufficiently unique path through the schema of the document. This is technically a bc break for people who overrode these mechanisms, but the only people likely to be doing that are the direct beneficiaries of the fix, and it is a necessary bug fix based on the recent experience of customers. Therefore no major version bump.

### DIFF
--- a/.changeset/stable-schema-field-ids.md
+++ b/.changeset/stable-schema-field-ids.md
@@ -1,0 +1,9 @@
+---
+"apostrophe": minor
+---
+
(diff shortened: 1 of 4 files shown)


BoDonkey (approved):
Somehow feels hinky that you can substitute something so unique as an md5 hash with a simple path.name, but...

boutell:
The path is what we really wanted all along, it uniquely describes it within apostrophe. Using its actual contents to describe it was a hack to get around getting the full path right, but in the end it cost us because customers with rolling deployments are adversely affected if anything changes.

PR #5553, opened 2026-08-13
fix(rich-text-widget): match most specific style when tags overlap

### LINKED ISSUE #3622: Rich text editor style selection bug

mmustafasenoglu:
Fixes #3622

Supersedes #5507. The matching logic now lives in a small helper (`findActiveStyleIndex`) with unit tests covering the overlapping-tags scenario discussed there: when several configured styles share the same tag, the most specific one (most classes) wins, so a `<p class="small">` shows "Small" instead of the plain "Paragraph".

Unit tests: packages/apostrophe/test/rich-text-style-selection.js
Changeset included.

### DIFF
--- a/.changeset/rich-text-style-selection-bug.md
+++ b/.changeset/rich-text-style-selection-bug.md
@@ -1,0 +1,5 @@
+---
+"apostrophe": patch
+---
+
+Fixed the rich text style picker marking the wrong style as active when multiple styles share the same tag. Styles are now matched in descending order of specificity (number of classes), so a `<p class="small">` correctly shows "Small" as the active style instead of the plain "Paragraph".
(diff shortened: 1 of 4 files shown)


boutell (changes requested):
If I read the old code right, it already would match only if `class` is a perfect match, whether in its absence or having exactly the same list of classes.

So sorting by length of match doesn't alter anything there. Only one could be an exact match anyway.

What am I missing?

PR #5566, opened 2026-08-27
Send external front redirect payloads with a 200 status

boutell:
## The bug

Visiting a page at a slug it used to live at returns a **500** instead of redirecting, when the site is served through an external front end (Astro). Reproduced on apostrophecms.com: `/extensions/template-library` (moved to `/extensions/document-template-library`) 500s, while `/extensions/document-template-library` is fine.

## Why

Apostrophe can't issue a redirect itself when an external front end holds the browser's connection, so `@apostrophecms/express` replaces `res.redirect` with one that *describes* the redirect as JSON for the front end to act on:

```js
res.redirect = function(...args) {
  // ...
  return res.send({ redirect: true, url, status });
};
```

That response deliberately carries no `Location` header — the external front is not the party being redirected. But it was inheriting whatever status the request had already set. `@apostrophecms/soft-redirect` sets `req.statusCode = 302`, and `@apostrophecms/page`'s `serveDeliver` copies that onto the response *before* calling `res.redirect`:

```js
if (req.statusCode) {
  req.res.statusCode = req.statusCode;   // page/index.js
}
if (req.redirect) {
  return req.res.redirect(status, req.redirect);
}
```

So the front end received a **302 with a body and no `Location`** — and `@apostrophecms/apostrophe-astro` reasonably treats 3xx as bodiless, dumping the body before parsing it. `aposPageFetch` then throws `Unexpected end of JSON input`, which surfaces as the 500.

Redirects created with `@apostrophecms/redirect` were unaffected: those are emitted from middleware that runs before page serving and never sets a status first, so they were already going out as 200 and worked correctly. That asymmetry is what made the failure look intermittent — identical payloads got different envelopes purely by internal code path.

## The fix

Set the status explicitly:

```js
return res.status(200).send({ redirect: true, url, status });
```

A 3xx is an instruction aimed at the browser, and the browser is not on the Astro→Apostrophe hop; its `Location` contract can't be honored there. The status the browser should ultimately see already travels in the body, where the front end reads it — that redundant `status` field, and the original `res.send` (rather than `res.status(302).send`), are both evidence the 3xx envelope was never intended.

This also fixes the live site with no adapter upgrade: the currently released `apostrophe-astro` already handles the 200-plus-JSON shape correctly, as `@apostrophecms/redirect` redirects demonstrate today.

## Test

`test/soft-redirects.js` gains a third describe block covering the external front. It fails on `main` (`302 !== 200`) and passes here. The assertion also pins the body, so the intended 302 must still reach the front end.

## Verification

- `test/soft-redirects.js`, `external-front.js`, `express.js`, `i18n.js`, `pages.js`, `static-build.js`, `urls.js` — 300 passing, 0 failing.
- `eslint` clean on both changed files.
- No consumer depends on the 3xx envelope: `apostrophe-astro` dispatches on the body's `redirect` flag, and its own `aposPageFetch` tests already model these responses as status 200.

### DIFF
--- a/.changeset/external-front-redirect-envelope.md
+++ b/.changeset/external-front-redirect-envelope.md
@@ -1,0 +1,9 @@
+---
+"apostrophe": patch
+---
+
+Redirects reported to an external front end are now sent with an HTTP status of 200, fixing a 500 error in Astro when a soft redirect was followed.
+
+Apostrophe cannot issue a redirect itself when an external front end holds the browser's connection, so `@apostrophecms/express` replaces `res.redirect` with one that describes the redirect as JSON (`{ redirect: true, url, status }`) for the front end to act on. That response carries no `Location` header, because the external front end is not the party being redirected. It was, however, inheriting whatever status the request had already set, and `@apostrophecms/soft-redirect` sets `req.statusCode` to 302 before the redirect is emitted. The result was a 302 with a body and no `Location` — which `@apostrophecms/apostrophe-astro` reasonably treated as bodiless, discarding the very payload it needed and failing with `Unexpected end of JSON input`.
+
+Visiting a page at a slug it used to live at therefore produced a 500 rather than a redirect. Redirects created with `@apostrophecms/redirect` were unaffected, as those are emitted from middleware that never sets a status first.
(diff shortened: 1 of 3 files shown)


BoDonkey (approved):

boutell:
Asset module tests failing a bunch in CI but passing locally, and not relevant to the change. Merging.

PR #5573, opened 2026-08-30
fix(sanitize-html): preserve incomplete tag prefix dropped by htmlparser2

NitinSemwal2605:
Fixes an issue where sanitize-html incorrectly strips the `<` and following characters of an incomplete tag ending in a space (e.g. `a <i `) when htmlparser2 drops them. Also properly escapes incomplete tags in escape/discard mode.

### DIFF
--- a/packages/sanitize-html/index.js
+++ b/packages/sanitize-html/index.js
@@ -598,6 +598,11 @@ function sanitizeHtml(html, options, _recursing) {
       frame.openingTagLength = result.length - frame.tagPosition;
     },
     ontext: function(text) {
+      if (html[parser.startIndex] === '<' && text[0] !== '<') {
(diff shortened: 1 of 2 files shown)


boutell (commented):
The current behavior matches what Chrome does if you give it your test cases.

Given that, I am inclined to say sanitize-html is already doing the right thing. I don't see a compelling use case for preserving this. I'm open to discussion however.

NitinSemwal2605:
That makes complete sense. If Chrome handles it by dropping the incomplete tag, then keeping sanitize-html's behavior aligned with browser parsing is definitely the right move for consistency and security. I don't have a strong use case for preserving it other than trying to prevent data loss on malformed input, but browser parity is more important.

boutell:
No worries, thanks for your understanding.

PR #5577, opened 2026-09-03
PRO-9948 fix our tests by accounting for recent webpack changes

boutell:
webpack moved the goalposts in a minor release, one of their fave things to do

this will make our mocha tests pass, possibly cypress also we'll see

all those package.json files were touched by `pnpm update -r`, it's just reflecting latest minor and patchlevel, exactly as npm update of a project would in practice, sorry for the noise


packages/apostrophe/modules/@apostrophecms/asset/lib/webpack/apos/webpack.config.js
```diff
@@ -33,7 +33,14 @@ module.exports = ({
     context: __dirname,
     mode,
     optimization: {
-      minimize: process.env.NODE_ENV === 'production'
+      minimize: process.env.NODE_ENV === 'production',
+      // Webpack 5.110+ auto-minifies CSS assets (experiments.css now
```
boutell, line 37: This optimization is not worth it to us for the bc break to tests and the fact that webpack is legacy anyway with vite available

BoDonkey (approved):

PR #5587, opened 2026-09-09
PRO-9904 Fix apos.boxField.toCss invalid CSS for non-uniform values and harden…

stuartromanek:
… missing value/property (PRO-9904)

*Check one*
- [X] main
- [ ] latest
- [ ] stable

## Summary
- Non-uniform box values no longer emit a single space-joined declaration that browsers drop. Per-side rules are joined with ;, and a plain property name like 'padding' becomes padding-top instead of property-top.
- Missing value or property no longer throws: undefined values return '', and a missing property warns once in development then returns ''.
- Adds helper tests for uniform/non-uniform output and the missing-argument cases.

*For example:*
1. Render a widget with a uniform box ({ top: 24, right: 24, bottom: 24, left: 24 }) via apos.boxField.toCss(..., 'padding-%key%') and confirm padding: 24px;
2. Change one side (e.g. top/bottom 32, left/right 24) and confirm per-side declarations with semicolons are applied in the browser
3. Call toCss(value, 'padding') without %key% and confirm padding-top / padding-right / etc., not property-top
4. Call toCss with undefined value or no property and confirm the page renders; in development, missing property logs box-field-toCss-property once
5. `npx mocha -t 10000 test/box-field.js in packages/apostrophe`

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/.changeset/big-regions-buy.md
+++ b/.changeset/big-regions-buy.md
@@ -1,0 +1,5 @@
+---
+"apostrophe": patch
+---
+
+Fix apos.boxField.toCss invalid CSS for non-uniform values and harden missing value/property (PRO-9904)
--- a/packages/apostrophe/modules/@apostrophecms/box-field/index.js
+++ b/packages/apostrophe/modules/@apostrophecms/box-field/index.js
@@ -124,17 +124,25 @@ module.exports = {
       }
     };
   },
-  helpers() {
+  helpers(self) {
     return {
       toCss(value, property, unit = 'px') {
+        if (typeof property !== 'string' || !property) {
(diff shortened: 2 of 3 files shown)


stuartromanek:
@myovchev do we want the included test?

myovchev (approved):

myovchev:
> @myovchev do we want the included test?

Sure. Be sure Cypress tests don't break though, there is a good chance we assert some related stuff.

PR #5592, opened 2026-09-15
Document versions I (wip)

myovchev:
Closes PRO-9336, PRO-9971, PRO-9973, PRO-9974, PRO-9977

## Phases

- **01 Port into core** (PRO-9336): the versions module ships in core as `@apostrophecms/document-versions` (alias `docVersions`), with its UI, i18n strings and tests.
- **02 Compressed records** (PRO-9971): version content is stored gzipped; a migration converts existing records.
- **03 Draft capture** (PRO-9973, PRO-9974, PRO-9977): draft saves are recorded under the capture rules, Save Draft always creates a version, each locale keeps its own history.
- **04 AI attribution** (PRO-9975, PRO-9976): versions record AI involvement and the user who started it; automatic translation marks its saves (together with the automatic-translation PR). Both tickets close with the AI indicator in phase 07.

## Outstanding changes

- Version `docId` is now `aposDocId`, with `mode` and `locale` next to it, because draft and published share one timeline per locale.
- `@apostrophecms/db-connect` exports `EJSON`, reached as `apos.modules['@apostrophecms/db'].EJSON`; db-connect must be released with core (changeset).
- Core refuses to start next to a pro Document Versions module that still records versions; only its method-less deprecation release may stay installed.
- `convert-legacy-versions` is a migration and also a task, for legacy records written by old instances during a rolling deploy.
- New public `apos.doc.setSaveFlags(req, input)`: virtual `_explicitSave`, `_restoreVersion` and `_ai` on REST writes; `localize` applies it after `beforeLocalize`.
- A publish always records a version, even without changes, so Unpublish always has its publication point.
- import-export: translated imports are marked as AI; a page override now moves before it patches, which keeps relationships to documents not yet in the target locale (changeset).
- CI: `APOS_AUTOMATIC_TRANSLATION_REF` (workflow env, or the `automatic-translation-ref` manual input) picks the automatic-translation ref for import-export's Pro tests; temporarily `document-versions`, default branch when empty.
- i18n key `notFound` renamed `versionsNotFound` (core already had `notFound`); the locale rename task is `@apostrophecms/document-versions:rename-locale`.


packages/apostrophe/modules/@apostrophecms/document-versions/index.js
```diff
@@ -0,0 +1,982 @@
...
+          localized
+        } = moduleOptions;
+
+        if (versions === false) {
+          return false;
+        }
+
+        if (versions !== true) {
+          if (autopublish === true || localized === false) {
+            return false;
```
boutell, line 325: ... Excluding autopublish from versions is... an existing decision we WILL NOT touch right now! Maybe much later we'll talk about it 😄

boutell (approved):
I'm good with the code. I had a little trouble starting it symlinked in testbed, advice? Not sure how I could both have support for AI providers at all *and* not have the openai one right now:

```
13:22:15 [ERROR] [apostrophe] startup-error: Unknown AI provider: openai  outcome=exiting
    invalid: Unknown AI provider: openai
```

PR #5595, opened 2026-09-15
Do not seize undo/redo shortcuts while a modal is open

stuartromanek:
Fixes [PRO-9821](https://linear.app/apostrophecms/issue/PRO-9821).

## Problem

In-context undo and redo are page-level operations: they replay or roll back patches against the document being edited on the page. The command menu registered `Ctrl/Cmd+Z` and `Ctrl/Cmd+Shift+Z` globally, so those keystrokes were seized even while a modal was open. Pressing Cmd+Z in a dialog rolled back page changes the editor could not see, and the browser's native text undo never ran.

## Approach

Commands can now declare `skipInModal`, and the admin bar's `undo` and `redo` entries set it. `TheAposCommandMenu` carries the flag through to the shortcut map and consults it in `keyboardShortcutListener` before calling `preventDefault()`, so when the modal stack is non-empty the event is left alone for the browser (or a more specific listener) to handle.

The decision itself lives in a small `should-handle-shortcut.mjs` helper rather than inline in the component, so it can be unit tested without mounting the admin UI.

Note the check runs *after* the existing `requireWidgetFocus`, `skipOnTextSelection`, and `trigger === 'native'` guards, so it only changes behavior for shortcuts that would otherwise have been seized.

## Testing

`test/command-menu.js` gains two cases: one asserting the admin bar registers `undo` and `redo` with `skipInModal`, and one covering the helper across an open and closed modal stack.

End-to-end coverage lives in a companion testbed PR, which opens a widget editor, presses both shortcuts, and asserts the admin bar's Redo button stays disabled — an undo would have moved a patch onto the redo stack and enabled it.

## Testbed run
https://github.com/apostrophecms/testbed/actions/runs/35027048682

Made with [Cursor](https://cursor.com)


boutell (approved):
Seems pretty elegant.

myovchev (approved):

PR #5596, opened 2026-09-16
Document Versions in core phases 05–06

myovchev:
Closes PRO-9978, PRO-9972, PRO-9997, PRO-9995, PRO-10017, PRO-10018

## Phases

- **05 Unpublish and attachments** (PRO-9978, PRO-9972, PRO-9997): Unpublish reads the previous publication point from the versions store and the `previous` mode is gone; attachments held only by versions survive a reference recount; versions follow `aposDocId` changes.
- **06 Diff engine** (PRO-9995): server-side change rows at full schema depth, text representations, WYSIWYG annotation and consolidation, exposed through `GET :versionId/changes`, `GET changes?ids=` and `annotate=1` on the single version route.

## Outstanding changes

- `aposMode: 'previous'` is gone: `publish` writes no `:previous` copy; migrations `seed-publication-points` and `remove-previous-mode-docs` on the versions module give existing documents their publication points and delete the old copies.
- A second Unpublish is `invalid` until the next publish, detected by `restoredFrom` on the newest published record.
- `versions: false` types record at most two published versions per document and locale, no drafts, so Unpublish keeps working; no versions UI, REST routes `notfound`.
- New request flag `aposSkipVersion`; core sets it on the page tree replay after Unpublish, which otherwise recorded a second published version.
- New public `apos.attachment.addDocSource(name, fn)`; `recomputeAllDocReferences` runs every registered source after `aposDocs`; versions register one.
- New event `@apostrophecms/doc:afterChangeDocIds(pairs, { keep, skipReplace })`; a renamed `aposDocId` is replaced in the content of every stored version (page paths, relationship ids).
- Restoring a document from the archive records no version.

## Bugs fixed

- PRO-10017: moving a page to the last child of the home page saved it twice, the first time without a rank; move results carried an empty `changed` entry instead of the descendants.
- PRO-10018: `changeDocIds` lost relationship field values and rewrote page paths by substring; new public `apos.doc.replaceDocIdReferences(doc, { oldId, newId })`.
- `test/assets.js`: the load-dependent "faster with cache" assertion is disabled, the case renamed `should build with cache`.


packages/apostrophe/modules/@apostrophecms/document-versions/index.js
```diff
@@ -10,10 +10,17 @@ const zlib = require('node:zlib');
 const { createId } = require('@paralleldrive/cuid2');
 const _ = require('lodash');
 const { stripIndent } = require('common-tags');
+const diff = require('./lib/diff.js');
+const text = require('./lib/text.js');
 
 const gzip = promisify(zlib.gzip);
 const gunzip = promisify(zlib.gunzip);
 
+// A type that records published versions only keeps this many per document
```
boutell: This is very unclear and does not explain the purpose of the feature, which is to honor a request that versions NOT be kept at all for a particular type as much as possible while still allowing the "unpublish" button to function.

packages/apostrophe/modules/@apostrophecms/document-versions/index.js
```diff
@@ -48,6 +55,7 @@ module.exports = {
     await self.enableCollection();
     await self.createIndexes();
     self.addMigrations();
+    self.apos.attachment.addDocSource(self.__meta.name, self.eachVersionDoc);
```
boutell: I would go with `docSource` for the method name. `eachVersionDoc` doesn't say much.

boutell (changes requested):
In general this looks good. One concern about a comment.

packages/apostrophe/modules/@apostrophecms/document-versions/index.js
myovchev: Corrected.

PR #5599, opened 2026-09-17
Document Versions in core, phases 07–08

myovchev:
Closes PRO-9994, PRO-9980, PRO-9975, PRO-9976, PRO-9981, PRO-9983, PRO-9987, PRO-10036, PRO-10037

## Phases

- **07 Modal shell and version list**: the modal follows the prototype layout, with collapsible tabs, a sliding versions panel and list entries with status, time, editor, edit count and the AI mark.
- **08 Change list, expand, filter**: View Changes opens the version's grouped change list with breadcrumbs, word diffs, the type and AI filter and a change navigator.
- **03 correction**: a publish takes over the draft record it saved first.

## Outstanding changes

- The versions modal is rewritten on the composition API with the ported race conditions removed; new `useAdvisoryLock` composable in `@apostrophecms/ui`, next to `AposAdvisoryLockMixin`.
- Compare removed: route `GET compare/:vid1/:vid2`, methods `getCompareSchema`, `getCompareData`, `removeIfFrom`, and the i18n keys `compare`, `compareVersions`, `onlyShowDifferences`, `viewOptions`, `publishedEdit`.
- `AposModalRail` gains a `collapsible` prop; `AposModal` narrows the rail column to the new `$modal-rail-collapsed-w`.
- The modal focus trap skips elements inside `[inert]`, in every modal.
- A publish promotes the newest draft record (same author, same AI flag, not a restore, same content) into the publication point instead of inserting a second record.
- The `dateTimeFormatOptions` option defaults to `null`, a compact localized "day · time" in the admin UI locale.
- New dependency `diff` (jsdiff) `^9`; `changes` route rows carry `diff: [ { text, change } ]`.
- A change navigator from the prototype steps through the change groups and highlights the field or widget in the body.
- Expanded changes show the whole value; the design's first-line clip is dropped.

## Bugs fixed

- PRO-10036: `AposInputOembed` threw when a form reloaded in place onto an empty oembed value, and a late oembed answer wrote onto the newer value.
- PRO-10037: `useInfiniteScroll` could stop after the first page when started inside modal content that was still hidden; it re-checks when the scroll root resizes (also the recently edited manager).


packages/apostrophe/modules/@apostrophecms/modal/ui/apos/components/AposModal.vue
```diff
@@ -515,7 +515,7 @@ function onKeydownTab(event) {
   // which puts the content under display:none and makes every candidate's
   // offsetParent null.
   const elements = getFocusableElements(modalEl.value)
-    .filter(el => el.offsetParent !== null);
+    .filter(el => el.offsetParent !== null && !el.closest('[inert]'));
```
boutell, line 518: nice

packages/apostrophe/modules/@apostrophecms/oembed-field/ui/apos/components/AposInputOembed.vue
```diff
@@ -10,7 +10,7 @@
       <div class="apos-input-wrapper">
         <input
           :id="uid"
-          v-model="next.url"
+          v-model="url"
```
boutell, line 13: Why was this input touched at all... it's only ever used for video widgets. I trust that some bug was fixed...?

packages/apostrophe/modules/@apostrophecms/ui/ui/apos/composables/useAdvisoryLock.js
```diff
@@ -0,0 +1,188 @@
+import { onUnmounted, ref } from 'vue';
+
+const REFRESH_INTERVAL = 10000;
+
+/**
+ * Advisory lock on a document, the composition counterpart of
+ * `AposAdvisoryLockMixin`. Obtains the lock with `lock(url)`, refreshes
```
boutell, line 7: Shouldn't the mixin be using this internally to avoid doubled implementation?

boutell (changes requested):
Is it possible that detectChange.js could be reused rather than having a separate change detection algorithm, or is this just a vastly more sophisticated use case?

packages/apostrophe/modules/@apostrophecms/oembed-field/ui/apos/components/AposInputOembed.vue
myovchev: It's in the PR description. A bug indeed, nothing too bad. It has dedicated ticket.

packages/apostrophe/modules/@apostrophecms/ui/ui/apos/composables/useAdvisoryLock.js
myovchev: I moved it to composition API. The final result should be mostly neutral when it comes to bundle size.

myovchev:
> Is it possible that detectChange.js could be reused rather than having a separate change detection algorithm, or is this just a vastly more sophisticated use case?

Your last assumption is correct. detectChanges.js is the browser version of schema owned getChanges. We use the latter where possible (when boolean answer is enough). Keep in mind our diff computations are entirely server side driven. Our additional diff code creates the changes list, consumed by the UI change list - very different data requirements. It also drives the doc annotate logic, that should give us the in-context change view (modified). 

Let me know if I didn't get your question right.

PR #5602, opened 2026-09-18
Document Versions in core, phases 09–10

myovchev:
Closes PRO-9982, PRO-9988, PRO-10019, PRO-9984, PRO-9985, PRO-9986, PRO-9979

## Phases

- **09 Consolidation and restore**: consolidated versions served by the server with a cursor pager, Restore Version on the entry, one publication per save for autopublished types, and a change of order as its own row.
- **10 WYSIWYG markers and rich text**: changed widgets and fields are marked where they render (Added, Modified, Deleted, Moved, AI), rich text shows inline marks, a marker opens its rows in the change list, and widgets can opt in to their previous data.

## Outstanding changes

- The versions list pages by cursor (`before` in, `{ results, next }` out); `getRestPager` is replaced by `getTimelinePage`, and page numbers are gone from the response.
- `?consolidate=1` on the list, `:versionId/changes` and `:versionId`; `changes?ids=` removed; new `lib/consolidation.js`, methods `getSequenceFrom`, `getVersionPairs`, `getChangeCount`, property `sequenceBatchLimit`.
- `changeCount` counts change rows, not top-level fields; a restore stores `0`.
- Capture rules: a draft save after a restore starts a new version; the draft of an autopublish type never records (`isAutopublishedDraft`).
- Core `ensureDraftExists` makes the draft copy of a published-mode insert without recording a version, for every type.
- Undo Publish falls back to the version's `createdAt` when the recorded document has no `lastPublishedAt`.
- Change rows gain an order row for arrays and areas; the diff context takes an optional `t`.
- Core `AposSchema` gains a `beforeField` slot; `AposAreaWidget` shows version markers from the new `docVersionMarkers` store and emits `doc-version-marker` on `apos.bus`.
- The versions modal owns its widget graph key, so its areas no longer register the version's widgets in the page's graph.
- `@apostrophecms/rich-text-widget` sets `renderVersions: true`; `annotate` marks changed rich text widgets and `richText` fields with `del`/`ins`, and gives up past 2000 differing tokens.
- `render-widget` accepts `_olderVersion` for a type that sets `renderVersions`; `AposRichTextEditor` renders static markup inside the versions modal.
- `AposDocVersionsList` exposes `focus(versionId)`, `AposDocVersionsChanges` exposes `reveal()`; new component `AposDocVersionAiBadge`.

## Bugs fixed

- `useInfiniteScroll` stopped after the first page in Firefox, which reports a stale and a live entry in one callback; it reads the newest (versions list, recently edited manager).


packages/apostrophe/modules/@apostrophecms/document-versions/lib/rich-text-diff.js
```diff
@@ -0,0 +1,397 @@
...
+  }, []);
+  joined.forEach((region, index) => {
+    if (region.removed?.length && !region.added.length) {
+      settle(region, joined[index - 1], joined[index + 1]);
+    }
+  });
+  return joined;
+}
+
+// Where a removal starts is open to choice when the tokens before it repeat
```
boutell, line 319: How about:

"When there is more than one way to describe a removal, favor a description involving complete HTML elements."

... If that's what it even means 😄

boutell (changes requested):
I like the shape of it. It seems a pity we have to "own" text diff and can't reuse a good implementation, but then again, that would be one more external dependency going bad on us at some point.

I did request that one comment be clarified. You can do that and merge.

packages/apostrophe/modules/@apostrophecms/document-versions/lib/rich-text-diff.js
myovchev: True story :)
Better explained now.

myovchev:
> I like the shape of it. It seems a pity we have to "own" text diff and can't reuse a good implementation, but then again, that would be one more external dependency going bad on us at some point.
> 
> I did request that one comment be clarified. You can do that and merge.

Our rich text diff is 400 lines of HTML aware tokenizing and rendering built on top of 2 libraries - jsdiff and cheerio. The former is also the backbone of all our text change diff. No tool exists that can handle everything we need. So I'd say it's a good balance of custom code and well written and maintained libraries. 

One thing that I plan to do in the finishing cleanup is to look for exposing high level diff utils for other modules and widgets.

PR #5603, opened 2026-09-18
customizable focus method

boutell:
This is needed so that really interesting custom wysiwyg inline fields become compatible with clicking on the breadcrumb to focus them.


boutell:
(Realized the need when trying to generate a docs example that wasn't just string fields all over again.)

BoDonkey (approved):

